PR #60, opened 2020-04-15
Snap to grid

JShep1:
In order to keep consistency of the snapping behavior to the visual grid, we'll need to make an update to the grid so that an odd cell count won't shift the grid to be misaligned with the true world coordinate frame.  That issue is documented [here](https://github.com/ignitionrobotics/ign-rendering/issues/70)

Ready for review


chapulina:
@osrf-jenkins run tests

JShep1:
Manually set DCO to pass, I ran into a lot of headache trying to get the signoffs on past commits and not completely sure what's going on.  I just didn't want to muddy the commit history anymore with all the rebasing.

JShep1:
snaptogrid

src/gui/plugins/transform_control/CMakeLists.txt
```diff
@@ -1,6 +1,8 @@
 gz_add_gui_plugin(TransformControl
   SOURCES TransformControl.cc
   QT_HEADERS TransformControl.hh
+  PUBLIC_LINK_LIBS
```
iche033, line 4: I think this can be `PRIVATE`?

src/gui/plugins/transform_control/TransformControl.cc
```diff
@@ -24,6 +24,14 @@
 #include <ignition/plugin/Register.hh>
 #include <ignition/transport/Node.hh>
 #include <ignition/transport/Publisher.hh>
+#include <ignition/rendering/Visual.hh>
+#include <ignition/rendering/Geometry.hh>
+#include <ignition/rendering/Grid.hh>
+#include <ignition/rendering/Material.hh>
```
iche033, line 30: Material.hh not needed

src/gui/plugins/transform_control/TransformControl.cc
```diff
@@ -98,6 +131,162 @@ void TransformControl::OnMode(const QString &_mode)
   this->dataPtr->node.Request(this->dataPtr->service, req, cb);
 }
 
+/////////////////////////////////////////////////
+void TransformControl::OnSnapToGrid()
+{
+  this->dataPtr->snapToGrid = true;
```
iche033, line 137: I notice that the tool button is not checkable . Is it possible to toggle snap to grid?

JShep1: I didn't think this functionality should be a toggle.  Clicking the button will simply update the snap values to the grid cell length.  Introducing a toggle snap to grid may add extra complications to the custom snap dialog - it would need to be set to read only to the user while the button is toggled on, I didn't necessarily want to introduce more logic that might in the end confuse the user (ie, a novel user might wonder why they can't update the custom snap values if they have the snap to grid button pressed or forget they have it pressed), so I thought that a simple click to set values to the visual grid would be the most intuitive approach.

iche033: ah I see! I didn't realize clicking on the button also updated the custom snap values. Not sure what the best way is to give feedback

a different issue: while testing the snapping tool, I noticed that if the model is selected with the translation tool active, the snapping behavior does not work the first time but works the second time, e.g.

1. click model
2. select translation tool button -> arrows now displayed over the model
3. click snap to grid button
4. Hold ctrl and drag arrow -> no snapping
5. Release ctrl key and mouse button
6. Repeat 4. -> snapping works.

JShep1: > I didn't realize clicking on the button also updated the custom snap values

@chapulina and I decided that it would likely be best to keep the Custom Snap Value dialog consistent with the current snapping values so that when the snap the grid button is clicked, the internal snap values are updated to the grid cell length and then reflected in the dialog.

> I noticed that if the model is selected with the translation tool active, the snapping behavior does not work the first time but works the second time, e.g. 

That's actually been a recurring issue that isn't necessarily tied to this PR, this behavior occurs in other cases and has been around since keybindings were added, essentially when the "scene" window is not prioritized and the `CTRL` (or any other button, in fact) is pressed, `Scene3D` does not register the button press, which is where all the logic for those functionalities is held.  This is an artifact of QML's event hierarchy and likely needs a separate issue, Louise and I did some researching into this and couldn't quite find a good solution.

iche033: Got it, that makes sense now. The render window only gets focus only when the mouse is pressed: 
https://github.com/ignitionrobotics/ign-gazebo/blob/master/src/gui/plugins/scene3d/Scene3D.cc#L2244

So I think we don't get the any key events before dragging the mouse. Maybe we have to force focus to be active when the mouse hovers over the render window. Anyways that's a separate issue to tackle.

PR #62, opened 2020-04-16
Optional physics features

### LINKED ISSUE #58: Optional physics features
**Original report ([archived issue](https://osrf-migration.github.io/ignition-gh-pages/#!/ignitionrobotics/ign-gazebo/issues/58)) by Louise Poubel (Bitbucket: [chapulina](https://bitbucket.org/%7B5cfa2075-477b-4ded-bdb9-8d2479544ec4%7D/), GitHub: [chapulina](https://github.com/chapulina)).**

----------------------------------------
Currently, all features used by the physics system are required to load a physics engine. In the interest of supporting multiple physics engines with different sets of features, the physics system can be refactored so that some features are optional. Optional features are just skipped if not supported by a physics engine, while required features prevent the engine from being loaded.

[Pull request #573](https://osrf-migration.github.io/ignition-gh-pages/#!/ignitionrobotics/ign-gazebo/pull-requests/573) started work on this but couldn't be merged in time for GitHub migration. We'll reopen the PR there.

chapulina:
Reopening [this BitBucket PR](https://osrf-migration.github.io/ignition-gh-pages/#!/ignitionrobotics/ign-gazebo/pull-requests/573/page/1), resolves #58.

---

# Overview

Reduce `MinimumFeatureList` so that physics engines are not required to implement all features, and some of them are treated as optional.

# Bullet

This enables running Bullet with Ignition. You must use `ign-physics`' [bullet_chapulina](https://github.com/ignitionrobotics/ign-physics/compare/ign-physics1...bullet_chapulina) branch. :warning:  As of that branch, Bullet is not fully functional yet though. But it loads :slightly_smiling_face: 

# Implementation

This PR makes the following features optional:

* Features for detachable joints:
    * `ignition::physics::AttachFixedJointFeature`
    * `ignition::physics::DetachJointFeature`
    * `ignition::physics::SetJointTransformFromParentFeature`

* Features for bounding box:
    * `ignition::physics::GetModelBoundingBox`

* Features for joint velocity commands:
    * `ignition::physics::SetJointVelocityCommandFeature`

* Features for 3D meshes:
    * `ignition::physics::mesh::AttachMeshShapeFeature`

* For each group of optional features:
    * Create a feature list, i.e. `DetachableJointFeatureList`
    * Create entity types which use that list, i.e. `JointDetachableJointPtrType`
    * Create an entity map that holds the specialized entity type, i.e. `entityJointDetachableJointMap`
    * Use `entityCast` to cast from the original entity, created with `MinimumFeatureList`, to the specialized entity. `entityCast` will also populate the specialized map the first time it's called and get the entity from that map on subsequent calls. This means that `RequestFeature` is only called once for each combination of entity + feature list.

Any engine implementing all the minimum features will be loaded. If, at runtime, there's an attempt to use an unsupported feature, a warning is printed. That is:

* If a `component::DetachableJoint` is created, a warning is printed and the joint isn't attached or detached.
* If a `components::AxisAlignedBox` is created, a warning is printed and the box is not populated.
* If a `components::JointVelocityCmd` is created, a warning is printed and the command has no effect.
* If a collision has a mesh geometry, a warning is printed and the geometry isn't created.

I'll remove features not needed by TPE on a subsequent PR, that's currently on the `chapulina/tpe` branch.

### DIFF
--- a/src/systems/physics/Physics.cc
+++ b/src/systems/physics/Physics.cc
@@ -120,11 +120,11 @@ namespace components = ignition::gazebo::components;
 // Private data class.
 class ignition::gazebo::systems::PhysicsPrivate
 {
+  /// \brief This is the minimum set of features that any physics engine must
+  /// implement to be supported by this system.
+  /// New features can't be added to this list in minor / patch releases, in
+  /// order to maintain backwards compatibility with downstream physics plugins.
   public: using MinimumFeatureList = ignition::physics::FeatureList<
-          ignition::physics::AttachFixedJointFeature,
-          ignition::physics::DetachJointFeature,
-          ignition::physics::SetJointTransformFromParentFeature,
-          // FreeGroup
           ignition::physics::FindFreeGroupFeature,
           ignition::physics::SetFreeGroupWorldPose,
           ignition::physics::FreeGroupFrameSemantics,
@@ -134,12 +134,9 @@ class ignition::gazebo::systems::PhysicsPrivate
           ignition::physics::GetEntities,
           ignition::physics::GetContactsFromLastStepFeature,
           ignition::physics::RemoveEntities,
-          ignition::physics::mesh::AttachMeshShapeFeature,
           ignition::physics::GetBasicJointProperties,
           ignition::physics::GetBasicJointState,
           ignition::physics::SetBasicJointState,
-          ignition::physics::SetJointVelocityCommandFeature,
-          ignition::physics::GetModelBoundingBox,
           ignition::physics::sdf::ConstructSdfCollision,
           ignition::physics::sdf::ConstructSdfJoint,
           ignition::physics::sdf::ConstructSdfLink,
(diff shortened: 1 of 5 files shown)


chapulina:
@osrf-jenkins run tests

azeey:
@chapulina, should we conclude the test failures from [thils thread](https://bitbucket.org/ignitionrobotics/ign-gazebo/pull-requests/573/optional-physics-features/diff?w=1#comment-145092050) were flaky tests?

chapulina:
> should we conclude the test failures from thils thread were flaky tests?

No, I think they're introduced by this PR. I'll take a look. Also, I'd like to refactor this a bit to prevent some `Each` calls for unsupported features. Sorry I only remembered all this after reopening the PR.

chapulina:
> >    should we conclude the test failures from thils thread were flaky tests?
>
> No, I think they're introduced by this PR.

@azeey , I just realized that the test passed on the latest OSX build. :thinking: 

> I'd like to refactor this a bit to prevent some Each calls for unsupported features.

Done on db37fb6. I had thought of caching an `engine` pointer for each set of features and then just checking that pointer before some `Each` calls, but I thought that may be overkill? Went instead with breaking the `Each` loop. Let me know what you think.

src/systems/physics/Physics.cc
```diff
@@ -722,25 +825,27 @@ void PhysicsPrivate::CreatePhysicsEntities(const EntityComponentManager &_ecm)
...
         {
-          ignwarn << "DetachableJoint's child link entity ["
-                  << _jointInfo->Data().childLink << "] not found in link map."
-                  << std::endl;
-          return true;
+          ignwarn << "Can't process DetachableJoint component, physics engine "
+                  << "missing AttachFixedJointFeature" << std::endl;
+
+          // Break Each call since no DetachableJoints can be processed
+          return false;
```
azeey, line 837: By returning early, we are saying that the only way `entityCast` returns a `nullptr` is if the requested feature is missing, but looking at the code in `entityCast`, it can also return a `nullptr` if the entity is not in the `minimumMap`. I'm not sure, at least in this case, if the code will get to calling `entityCast` with the entity not being in `minimumMap`, but the check is in there nonetheless. 

This got me thinking...do we need to pass in the `minimumMap` at all? Can we not just pass the `EntityPtr` found from the the `minimumMap`? Looking at most places where `entityCast` is used, the code surrounding it already does a lookup in the `minimumMap`, so the extra check in `entityCast`  can be avoided by passing in the `EntityPtr`.

chapulina: Yeah you're right, most of the time the lookup has already happened. I think the first couple that I implemented didn't need the external lookup and that's why I moved it into the cast. Let me move it out.

chapulina: I guess it was a tie, 3 needed map lookup while 3 didn't: 1ce537a

PR #68, opened 2020-04-20
[master] Fix README hyperlinks

tykurtz:
Fixes the slightly annoying problem of the markdown links not working in `README.md`.

Reproduce this by clicking hyperlinks in [ignitionrobotics/ign-gazebo/README.md](https://github.com/ignitionrobotics/ign-gazebo)

Verify this fixes it by clicking the hyperlinks in [tykurtz/ign-gazebo/README.md](https://github.com/tykurtz/ign-gazebo/tree/fix-readme-links)

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,7 +96,7 @@ The binary install method will use pre-built packages which are typically
 available through a package management utility such as [Apt](https://wiki.debian.org/Apt).
 This approach eliminates the need to download and compile source code, and dependencies
 are handled for you. The downside of a binary install is that you won't be able to modify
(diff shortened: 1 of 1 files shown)


chapulina:
Thanks for the PR. I believe the links used that pattern so that they worked both on BitBucket. And then we adapted our [website](https://ignitionrobotics.org/libs/gazebo)* so that the links work there as well. We should make sure that this pattern also works there.

CC @nkoenig 

\* which hasn't been updated to point to GitHub yet

PR #79, opened 2020-04-20
Refactor Gui.hh so that the Gazebo GUI can be ran from other packages

azeey:
With this PR the `GazeboGui` plugin in ign-launch can run the Gazebo GUI much less code duplication. By doing so, this adds support for saving world files from ign-launch (with the corresponding PR TBD).


azeey:
Looks like the build is failing on homebrew. I'm looking into it.

azeey:
Fixed by putting the function definitions for `runGui` and `createGui` under the version inline namespace.

PR #88, opened 2020-04-23
Exit early if deployments are not allowed

nkoenig:
This prevents a topic from creation, and also limits some computation.


src/systems/breadcrumbs/Breadcrumbs.cc
```diff
@@ -45,6 +45,16 @@ void Breadcrumbs::Configure(const Entity &_entity,
     EntityComponentManager &_ecm,
     EventManager &_eventMgr)
 {
+  this->maxDeployments =
+      _sdf->Get<int>("max_deployments", this->maxDeployments).first;
+
+  // Exit early if breadcrumbs deployments are not possible.
+  if (this->maxDeployment <= 0)
```
iche033, line 52:
```suggestion
  if (this->maxDeployments <= 0)
```

src/systems/breadcrumbs/Breadcrumbs.cc
```diff
@@ -45,6 +45,16 @@ void Breadcrumbs::Configure(const Entity &_entity,
...
 {
+  this->maxDeployments =
+      _sdf->Get<int>("max_deployments", this->maxDeployments).first;
+
+  // Exit early if breadcrumbs deployments are not possible.
+  if (this->maxDeployments <= 0)
+  {
+    ignmsg << "Breadcrumbs max deployment is <= 0. Breadcrumbs are disabled."
+      << std::endl;
+  }
```
azeey, line 56: Should there be a return here?
```suggestion
    return;
  }
```

nkoenig: Added: 154fdfdd7a46ef4d5a05f37ee0a99725616aa37d

chapulina:
Does the doxygen need to be updated?

https://github.com/ignitionrobotics/ign-gazebo/blob/536a08d403e1bb608e349063b11250301c53e74d/src/systems/breadcrumbs/Breadcrumbs.hh#L60-L63

nkoenig:
Documentation has been updated.

PR #95, opened 2020-04-25
[Blueprint] Add force focus when mouse enters render window

JShep1:
Solves issue mentioned [here](https://github.com/ignitionrobotics/ign-gazebo/pull/60#discussion_r414903153).

This PR focuses the render window if the mouse enters it.  Previously, the render window was only focused if the user clicked it, introducing a slightly annoying behavior where the user had to click twice in order to use the snapping behavior given the `CTRL` press wouldn't be registered until after the window was focused.


iche033:
update changelog please

PR #97, opened 2020-04-25
[Dome] Add force focus when mouse enters render window

JShep1:
Solves issue mentioned [here](https://github.com/ignitionrobotics/ign-gazebo/pull/60#discussion_r414903153).

This PR focuses the render window if the mouse enters it.  Previously, the render window was only focused if the user clicked it, introducing a slightly annoying behavior where the user had to click twice in order to use the snapping behavior given the `CTRL` press wouldn't be registered until after the window was focused.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,9 +2,6 @@
 
 ### Ignition Gazebo 4.X.X
 
-1. Add window focus upon mouse entering the render window
-    * [Github pull request 97](https://github.com/ignitionrobotics/ign-gazebo/pull/97)
-
 ### Ignition Gazebo 4.0.0 (20XX-XX-XX)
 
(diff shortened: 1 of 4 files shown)


chapulina:
> Solves issue mentioned here.

That was merged into `ign-gazebo3`. Do we need to backport this fix to that branch?

Changelog.md
```diff
@@ -2,6 +2,9 @@
 
 ### Ignition Gazebo 4.X.X
 
+1. Add window focus upon mouse entering the render window
+    * [Github pull request 97](https://github.com/ignitionrobotics/ign-gazebo/pull/97)
```
iche033, line 6: dome has not been release yet :) can you move this entry down to the 4.0.0 section?

PR #108, opened 2020-04-28
Fix crash when running the Gazebo GUI

azeey:
The crash (introduced in PR #79) was caused by an access to an invalid address within Qt when
trying to access the command line arguments to the application. These
are passed as argc and argv to QApplication, but for argc, we were
passing a reference to a local.


JShep1:
Builds and runs fine for me now!

PR #127, opened 2020-05-07
Port support for computing model bounding box in physics system

iche033:
Porting changes from this bitbucket PR that was merged into ign-gazebo2:
https://osrf-migration.github.io/ignition-gh-pages/#!/ignitionrobotics/ign-gazebo/pull-requests/546/page/1

see original changes in this commit: 4cdef0c1ea957439b6d87ebdf3650f3d4b967827


src/systems/physics/Physics.cc
```diff
@@ -129,6 +132,7 @@ class ignition::gazebo::systems::PhysicsPrivate
           ignition::physics::GetBasicJointState,
           ignition::physics::SetBasicJointState,
           ignition::physics::SetJointVelocityCommandFeature,
+          ignition::physics::GetModelBoundingBox,
```
chapulina, line 135: I started work to make this feature optional on the `ign-gazebo3` branch here:

https://github.com/ignitionrobotics/ign-gazebo/blob/2259f027cd3a244f2c841b62c968088eace1b338/src/systems/physics/Physics.cc#L370-L387

What do you think of doing this from this PR? It will reduce my PR later. You'd need to forward-port just the [entityCast](https://github.com/ignitionrobotics/ign-gazebo/pull/62/files#diff-2b54a7efa7163b4546a1568fdd51cbbb) function from #58 .

If you need this urgently, the optional feature can wait.

iche033: ok d18df79 should do it. I notice the one of the `entityCast` args changed between blueprint and your branch. Not sure which one I should use so I picked the blueprint version but should be easy to change.

nkoenig:
Is an ign-physics release needed to fix the builds?

chapulina:
> Is an ign-physics release needed to fix the builds?

src/systems/physics/Physics.cc
chapulina: Thanks! Yeah the Blueprint function is the correct one.

PR #132, opened 2020-05-11
DetachableJoint system: Add option to suppress warning about missing child model

azeey:
This is useful for avoiding warning messages when parent and child models are spawned with a delay in between.


src/systems/detachable_joint/DetachableJoint.hh
```diff
@@ -49,6 +49,10 @@ namespace systems
   /// creating a fixed joint with a link in the parent model.
   ///
   /// <topic> (optional): Topic name to be used for detaching connections
+  ///
+  /// <suppress_child_warning> (optional): If true, the system will wait for the
```
nkoenig, line 53:
```suggestion
  /// <suppress_child_warning> (optional): If true, the system
```

PR #136, opened 2020-05-14
2 ➡️ 3

chapulina:
Merge `ign-gazebo2` forward to `ign-gazebo3`.

@azeey, I had to make some changes to `SdfGenerator_TEST` on 9ecee15 so that the worlds would load with SDFormat 9. Mind checking if that's the best way to fix those tests?

@iche033 , I have 2 tests failing locally, `INTEGRATION_gpu_lidar` and `INTEGRATION_sensors_system`, both at the destruction of `GpuLidarSensor` with the backtrace below. Have you seen this before and know why it would be failing on this branch? If not, I can dig into it.

<details><summary>Backtrace</summary>

```
Thread 1 "INTEGRATION_sen" received signal SIGSEGV, Segmentation fault.
0x00007fffc1c8d969 in ?? () from /usr/lib/x86_64-linux-gnu/libGLdispatch.so.0
(gdb) bt
#0  0x00007fffc1c8d969 in ?? () from /usr/lib/x86_64-linux-gnu/libGLdispatch.so.0
#1  0x00007fffb616f4c8 in Ogre::GL3PlusFrameBufferObject::~GL3PlusFrameBufferObject() ()
   from /usr/lib/x86_64-linux-gnu/OGRE-2.1/OGRE/RenderSystem_GL3Plus.so
#2  0x00007fffb616eaf2 in ?? () from /usr/lib/x86_64-linux-gnu/OGRE-2.1/OGRE/RenderSystem_GL3Plus.so
#3  0x00007fffb619527a in Ogre::v1::GL3PlusTextureBuffer::~GL3PlusTextureBuffer() ()
   from /usr/lib/x86_64-linux-gnu/OGRE-2.1/OGRE/RenderSystem_GL3Plus.so
#4  0x00007fffb61952b9 in Ogre::v1::GL3PlusTextureBuffer::~GL3PlusTextureBuffer() ()
   from /usr/lib/x86_64-linux-gnu/OGRE-2.1/OGRE/RenderSystem_GL3Plus.so
#5  0x00007fffb61921eb in Ogre::GL3PlusTexture::freeInternalResourcesImpl() ()
   from /usr/lib/x86_64-linux-gnu/OGRE-2.1/OGRE/RenderSystem_GL3Plus.so
#6  0x00007fffc2c0e57d in Ogre::Texture::freeInternalResources() () from /usr/lib/x86_64-linux-gnu/libOgreMain.so.2.1.0
[...]

PR #140, opened 2020-05-14
Support TPE (more optional physics features)

### LINKED ISSUE #58: Optional physics features

chapulina:
Builds on top of #136, closes #58.

As a follow up to https://github.com/ignitionrobotics/ign-gazebo/pull/62, this pull request makes more physics features optional so that we can load the latest awesome engine everyone is talking about, the _Trivial Physics Engine_.

To use it with TPE, do for example:

`ign gazebo -v 2 -r shapes.sdf --physics-engine ignition-physics-tpe-plugin`

This should lower the bar for other physics engines besides TPE as well. Features being made optional:

* Joints
* Collisions / contacts (even though TPE supports them now)
* Applying force

---

@azeey , I'd like your opinion on making `UpdateSim` and `UpdateCollisions` not `const`. That's necessary so we can populate the new maps.


azeey:
>  I'd like your opinion on making UpdateSim and UpdateCollisions not const. That's necessary so we can populate the new maps.

Yeah, I think that's fine.

src/systems/physics/Physics.cc
```diff
@@ -727,7 +781,16 @@ void PhysicsPrivate::CreatePhysicsEntities(const EntityComponentManager &_ecm)
...
         else
         {
-          collisionPtrPhys = linkPtrPhys->ConstructCollision(collision);
+          auto linkCollisionFeature = entityCast(_parent->Data(), linkPtrPhys,
+              this->entityLinkCollisionMap);
+          if (!linkCollisionFeature)
+          {
+            ignwarn << "Can't process Collision entity, physics engine "
+                    << "missing ConstructSdfCollision feature" << std::endl;
+            return true;
```
claireyywang, line 790: should we still warn about it if we made this feature optional? same with other optional features' warnings.

src/systems/physics/Physics.cc
```diff
@@ -125,24 +124,12 @@ class ignition::gazebo::systems::PhysicsPrivate
...
-          ignition::physics::DetachJointFeature,
-          ignition::physics::SetJointTransformFromParentFeature,
-          // FreeGroup
           ignition::physics::FindFreeGroupFeature,
           ignition::physics::SetFreeGroupWorldPose,
           ignition::physics::FreeGroupFrameSemantics,
           ignition::physics::LinkFrameSemantics,
-          ignition::physics::AddLinkExternalForceTorque,
           ignition::physics::ForwardStep,
-          ignition::physics::GetEntities,
```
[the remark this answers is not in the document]
chapulina: The reason I removed it here is that this plugin itself is not using any of the functions contained in the `GetEntities` feature list. It's ok if a physics engine implements and uses these features, but they're not required to load the engine into `ign-gazebo` right now.

src/systems/physics/Physics.cc
chapulina: The idea is that the user will only see these warnings if they're trying to use an unsupported feature. If they loaded a world that needs a certain feature, but that feature is not supported, then maybe the user is doing something they don't mean to. For example, a user loads a model that contains joints, but the engine is going to ignore joints. Then we warn them that the joints are being ignored.

But it's true that many times the user may be aware that the feature is not supported and they're fine ignoring it. In this case the warnings will be spammy. For example, the user wants to load the same model into TPE and DART at different times to test different things. They know that the model's joints will be ignored by TPE and they're fine with it. And they don't want to maintain 2 different models, one for TPE and another for DART.

For this situation, how about offering the user a way to silence the warnings? This could be done within the SDF, for example:

```
    <plugin
      filename="libignition-gazebo-physics-system.so"
      name="ignition::gazebo::systems::Physics">
      <engine>
        <filename>ignition-physics-tpesim-plugin</filename>
        <!-- Don't emit warnings if the following unsupported features are used -->
        <suppress_warnings>GetBasicJointState</suppress_warnings>
      </engine>
    </plugin>
```

src/systems/physics/Physics.cc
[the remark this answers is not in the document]
claireyywang: I see. That makes sense.

src/systems/physics/Physics.cc
claireyywang: I guess the warnings were helpful when I was debugging TPE. Giving option to suppress it sounds like a good idea, but I think listing feature  in sdf file requires a bit much work and knowledge about features used by each world, which isn't super obvious. How about using `igndbg` instead of `ignwarn`? `igndbg` seems to only show up at a more verbose level if I remember correctly, so that users can still access this info if they notice something wrong.

chapulina: > How about using igndbg instead of ignwarn?

My immediate impulse was to treat it as a warning because I see it as a ["potentially harmful situation"](https://stackify.com/logging-levels-101/) when it isn't straight up an error ("problem that needs to be dealt with"). But given the conversation above and the fact that this could become spammy, I took your suggestion, and made it so each message is only printed once: fa29326 .

PR #144, opened 2020-05-20
3 ➡️ 4

### LINKED ISSUE #131: ign-gazebo regressions test failing INTEGRATION_rgbd_camera.test_ran
On linux: https://build.osrfoundation.org/job/ignition_gazebo-ci-master-bionic-amd64/
And MacOS: https://build.osrfoundation.org/job/ignition_gazebo-ci-master-homebrew-amd64/

Test `INTEGRATION_rgbd_camera.test_ran` is failing. 

`Expected: (depthBuffer) != (nullptr), actual: NULL vs 8-byte object <00-00 00-00 00-00 00-00>`

From: https://build.osrfoundation.org/job/ignition_gazebo-ci-master-bionic-amd64/lastCompletedBuild/consoleText

<details>

```
109: /var/lib/jenkins/workspace/ignition_gazebo-ci-master-bionic-amd64/ign-gazebo/test/integration/rgbd_camera.cc:95: Failure
109: Expected: (depthBuffer) != (nullptr), actual: NULL vs 8-byte object <00-00 00-00 00-00 00-00>
109/116 Test #109: INTEGRATION_rgbd_camera ..............................***Exception: SegFault  1.90 sec
```
</details>

chapulina:
@mabelzhang , I'm having a hard time getting the logging tests to pass. I'm opening this PR for visibility but I'm still looking into these tests. Let me know if you already have all the answers :wink: 

It's a bit confusing because we deprecated the `<path>`, but that's still the way we inject the path into the plugin, right? I think we could change this approach to simplify things a bit. For example, we could put the log record configuration into a component. We're doing something similar to propagate the physics engine from the command line to the physics plugin. If we take this route, I recommend we do the minimum we need to get the tests to pass here, and change the approach on a follow up PR.

fixes #131


chapulina:
> I recommend we do the minimum we need to get the tests to pass here, and change the approach on a follow up PR.

Done on b47cb81 and d693bb6. The tests passed for me locally, :crossed_fingers: for CI.

@mabelzhang , I'd appreciate your :heavy_check_mark: on these changes.

src/ign.cc
```diff
@@ -192,7 +192,6 @@ extern "C" IGNITION_GAZEBO_VISIBLE int runServer(const char *_sdfString,
       //   In Blueprint and Citadel, LogRecord will record to <path> if no
       //   --record-path is specified on command line.
       serverConfig.SetLogRecordPath(recordPathMod);
```
mabelzhang, line 194: This line can be moved outside of and after the if-else statement, and the TODO deleted.

src/ign.cc
```diff
@@ -192,7 +192,6 @@ extern "C" IGNITION_GAZEBO_VISIBLE int runServer(const char *_sdfString,
...
-      serverConfig.SetLogIgnoreSdfPath(true);
```
mabelzhang: The `logIgnoreSdfPath` in `ServerConfig.cc` has a comment that it should be deprecated in public APIs in Ignition-D (and removed in Ignition-E). Probably its corresponding functions `LogIgnoreSdfPath()` and `SetLogIgnoreSdfPath()` too. I'm not sure how to do that.

test/integration/log_system.cc
```diff
@@ -445,12 +445,14 @@ TEST_F(LogSystemTest, LogPaths)
...
   // Check state.tlog is no longer stored to path specified in SDF
-  EXPECT_FALSE(common::exists(common::joinPaths(this->logDir,
-      "state.tlog")));
+  auto stateFile = common::joinPaths(this->logDir, "state.tlog");
+  EXPECT_FALSE(common::exists(stateFile)) << stateFile;
 #ifndef __APPLE__
   EXPECT_EQ(0, entryCount(this->logDir));
 #endif
 
+  return;
```
mabelzhang: Is this `return` supposed to be here?

src/ServerPrivate.cc
```diff
@@ -296,19 +241,11 @@ void ServerPrivate::AddRecordPlugin(const ServerConfig &_config)
...
               cPathElem = pluginElem->GetElement("compress_path");
               cPathElem->AddValue("string", "", false, "");
-              cPathElem->Set<std::string>(cmpPath);
+              cPathElem->Set<std::string>(_config.LogRecordCompressPath());
             }
 
             return;
           }
-
-          // If playback plugin also specified, do not add a record plugin
```
mabelzhang: I think this block is still needed, to make sure they don't specify both playback and record at the same time.

src/ServerPrivate.cc
```diff
@@ -207,67 +207,12 @@ void ServerPrivate::AddRecordPlugin(const ServerConfig &_config)
           //   SDF.)
           if (pluginName->GetAsString() == LoggingPlugin::RecordPluginName())
           {
-            std::string recordPath = _config.LogRecordPath();
-            std::string cmpPath = _config.LogRecordCompressPath();
-
-            // Set record path
-            if (!_config.LogRecordPath().empty())
```
mabelzhang: I think the `if (!_config.LogRecordPath().empty())` check is still needed, around the new lines 210-215, in case the string is empty.

test/integration/log_system.cc
chapulina: I had pushed that by mistake, it was removed in d693bb636e09754ec16c782ff22d9dc6b93282ad

src/ServerPrivate.cc
chapulina: Oh sorry you're right, I lost track of the ifs and thought this was inside `if (pluginName->GetAsString() == LoggingPlugin::RecordPluginName())`

src/ign.cc
chapulina: The deprecation warnings are already there:

https://github.com/ignitionrobotics/ign-gazebo/blob/6332a61de6728b3aeb080fd587a04b569788f2a5/include/ignition/gazebo/ServerConfig.hh#L246-L254

src/ServerPrivate.cc
chapulina: I removed it because it was making a test fail. But I made other changes to the test since then, so I can try adding it again.

chapulina: This is one of the tests that are broken if I check for an empty string:

https://github.com/ignitionrobotics/ign-gazebo/blob/d693bb636e09754ec16c782ff22d9dc6b93282ad/test/integration/log_system.cc#L331-L352

I'm confused about the expectations though, because recording does take place, just not to `<path>`.

mabelzhang: That `Recording will not take place` output looks right. Basically that's a case that would only happen if someone uses the C++ API to record and forgot to specify a path. In normal usage, the path is specified in `ign.cc`, so if they don't go through the command line, then the plugin doesn't know where to record to - and we decided not to initialize the ignLogDirectory inside the plugin.

Maybe some tests need to be changed since we remove the <path> in Dome, I will take a look. I have a library failure at runtime after I reinstalled using colcon and haven't been able to run the tests, ha... See if I can fix it.

src/ign.cc
chapulina: @mabelzhang , while you're looking at the empty `LogRecordPath`, mind moving this too? I'm not sure exactly what `if / else` it's referring to, and since all tests pass at the current state, I don't know if we should add a test to cover what will be changed by this move, or change an existing test's expectation.

src/ServerPrivate.cc
chapulina: eec3984

chapulina:
The homebrew CI is broken on master: https://build.osrfoundation.org/job/ignition_gazebo-ci-master-homebrew-amd64/

So I think we can merge this and deal with OSX problems once that's back up.

PR #150, opened 2020-05-21
Added origin axes config plugin

ahcorde:
I have been playing a little bit around ignition-gazebo. I have created a config plugin to add and configure the origin axes. I was trying to do something similar as the grid. This is created in ignition-rendering using the method `showGrid()`. I was planning to add a new method called `showOriginAxes()`.  but arrows are build based on two visuals: a cone and a cylinder. I was not able to get the arrows in the config plugins based only in the ID. is there any way to get the arrows based on the name?

Origin_axes_plugin_config


src/gui/plugins/origin_axes/OriginAxesConfig.cc
```diff
@@ -0,0 +1,213 @@
...
+  if (this->dataPtr->x_axis || this->dataPtr->y_axis || this->dataPtr->z_axis)
+    return;
+
+  // Create grid
+  ignwarn << "Creating origin axes" << std::endl;
+
+  auto root = scene->RootVisual();
+  this->dataPtr->x_axis = scene->CreateArrowVisual();
+  this->dataPtr->y_axis = scene->CreateArrowVisual();
+  this->dataPtr->z_axis = scene->CreateArrowVisual();
```
chapulina, line 166: On Gazebo-classic the origin axes were implemented using dynamic lines instead of arrows. Most of the time users need the origin to be long and without arrows, see [this comment](https://osrf-migration.github.io/gazebo-gh-pages/#!/osrf/gazebo/pull-requests/1700/page/1#comment-7000228). @iche033 can correct me if I'm wrong, but I believe the lines should be less expensive to render, since we don't have the arrow head and the extra triangles from the cylinders. I'm not advocating for a change, I just think it should be pointed out.

---

Having said all that, I think it's totally valid to add support for custom axes like [RViz's axes displays](http://wiki.ros.org/rviz/DisplayTypes/Axes), which ties into the work @Sarath18 will be doing :smile: To that end, here are some suggestions to make the current plugin more flexible:

* Add an option to set the pose
* Add an option to set the radius
* Add an option to change the arrows into lines

This is definitely feature creep for this PR, but I'll just mention it as an idea for the future:

* Let the user attach an axis to any entity that has a pose component

src/gui/plugins/origin_axes/OriginAxesConfig.qml
```diff
@@ -0,0 +1,87 @@
...
+    Layout.columnSpan: 1
+    Layout.rowSpan: 1
+    Layout.fillWidth: true
+  }
+
+  CheckBox {
+    Layout.alignment: Qt.AlignHCenter
+    id: showgrid
+    Layout.columnSpan: 2
+    text: qsTr("Show/Hide Grid")
```
chapulina, line 43:
```suggestion
    text: qsTr("Show/Hide Axis")
```

src/gui/plugins/origin_axes/OriginAxesConfig.qml
```diff
@@ -0,0 +1,87 @@
...
+  // Left spacer
+  Item {
+    Layout.columnSpan: 1
+    Layout.rowSpan: 1
+    Layout.fillWidth: true
+  }
+
+  CheckBox {
+    Layout.alignment: Qt.AlignHCenter
+    id: showgrid
```
chapulina, line 41: This `id` can be removed since it's not being used anywhere

src/gui/plugins/origin_axes/OriginAxesConfig.qml
```diff
@@ -0,0 +1,87 @@
...
+    Layout.columnSpan: 1
+    id: vercelltext
+    color: "dimgrey"
+    text: "Size(meters):"
+    Layout.fillWidth: true
+  }
+
+  IgnSpinBox {
+    Layout.columnSpan: 1
+    id: verticalCellCount
```
chapulina, line 74: This id doesn't make sense for the axes. Maybe `length`?

ahcorde:
Thank you for the review @chapulina. 

I forgot to mention that this PR builds on top of this other PR https://github.com/ignitionrobotics/ign-rendering/pull/87.

I created this other PR https://github.com/ignitionrobotics/ign-rendering/pull/88 to scale the `BaseAxisVisual` properly.

I can also modify the current implementation of `BaseAxisVisual` to allow to choose between: arrows or lines

src/gui/plugins/origin_axes/OriginAxesConfig.cc
iche033: yes lines are less expensive to render. I think a generic axes plugin would be useful, e.g. for visualizing frames. That could be done by changing the name of this plugin to just AxesPlugin and adding a pose widget.

ahcorde:
- Select the origin of the different entities
 - Add an option to set the pose
 - Add an option to change the arrows into lines

axes

src/gui/plugins/axes/AxesConfig.cc
```diff
@@ -0,0 +1,285 @@
...
+  }
+}
+
+void AxesConfig::onCurrentIndexChanged(int _index)
+{
+  std::string name_entity = m_comboList[_index].toStdString();
+  LoadAxesbyName(name_entity);
+}
+
+void AxesConfig::EntitiesInScene()
```
ahcorde, line 232: @chapulina, Is there other way to get the entitie names from a plugin?

src/gui/plugins/CMakeLists.txt
```diff
@@ -84,6 +84,7 @@ add_subdirectory(modules)
 add_subdirectory(align_tool)
 add_subdirectory(entity_tree)
 add_subdirectory(grid_config)
+add_subdirectory(axes)
```
chapulina, line 87: Nit: alphabetize

src/gui/plugins/axes/AxesConfig.cc
chapulina: If you change the plugin to inherit from `gazebo::GuiSystem`, then you could use an `Each` call to get for example the names of all visuals, links, collisions, models, etc. The downside is that this plugin would become Gazebo-specific so we can't port it to `ign-gui` later.

src/gui/plugins/axes/AxesConfig.cc
```diff
@@ -0,0 +1,285 @@
...
+    if (tokens.size() == 1)
+    {
+      set_entities.insert(tokens[0]);
+    }
+  }
+
+  QStringList local_combolist;
+  for (auto s : set_entities)
+  {
+    local_combolist << s.c_str();
```
chapulina, line 252: Remember that the Ignition style is `camelCase` :slightly_smiling_face:

src/gui/plugins/axes/AxesConfig.cc
```diff
@@ -0,0 +1,285 @@
...
+
+    auto axes = this->dataPtr->scene->CreateAxisVisual(this->dataPtr->name_axes + "Axes");
+    root->AddChild(axes);
+  }
+}
+
+void AxesConfig::onCurrentIndexChanged(int _index)
+{
+  std::string name_entity = itemComboList[_index].toStdString();
+  LoadAxesbyName(name_entity);
```
chapulina, line 229: This needs to be done in the rendering thread, otherwise we get occasional crashes like this one I just got:

```
Invalid parameter at line 374 (@foreach). 'hlms_lights_directional' is not a number nor a variable
terminate called after throwing an instance of 'Ogre::RenderingAPIException'
  what():  OGRE EXCEPTION(3:RenderingAPIException): Fragment Program 537100288PixelShader_ps failed to compile. See compile log above for details. in GLSLShader::compile at /var/lib/jenkins/workspace/ogre-2.1-debbuilder/repo/RenderSystems/GL3Plus/src/GLSL/OgreGLSLShader.cpp (line 308)
```

To use the rendering thread, listen to events like this:

https://github.com/ignitionrobotics/ign-gazebo/blob/aea1accff6d8daee0358ae9025e2a9797d292844/src/gui/plugins/transform_control/TransformControl.cc#L94-L95

And look for the `Render` event:

https://github.com/ignitionrobotics/ign-gazebo/blob/aea1accff6d8daee0358ae9025e2a9797d292844/src/gui/plugins/transform_control/TransformControl.cc#L220-L222

src/rendering/RenderUtil.cc
```diff
@@ -1281,6 +1282,15 @@ void RenderUtil::ShowGrid()
   root->AddChild(visual);
 }
 
+/////////////////////////////////////////////////
+void RenderUtil::ShowOriginAxes()
+{
+  rendering::VisualPtr root = this->dataPtr->scene->RootVisual();
+
+  auto worldOriginAxes = this->dataPtr->scene->CreateAxisVisual("worldOriginAxes");
```
chapulina, line 1290: Now that the plugin displays entity axes. I think words like "origin" and "world" can be misleading.

I'd actually suggest removing this logic from `RenderUtil` and creating the axes directly from `AxesConfig`, within the rendering thread.

src/rendering/RenderUtil.cc
```diff
@@ -320,6 +320,7 @@ void RenderUtil::Update()
     this->dataPtr->scene->SetBackgroundColor(scene.Background());
     if (scene.Grid() && !this->dataPtr->enableSensors)
       this->ShowGrid();
+    this->ShowOriginAxes();
```
chapulina, line 323: Now that the plugin displays axes for entities, I think it doesn't make sense to initialize it here.

If we want a plugin that displays the world origin axes visual, here we should check `if (scene.OriginVisual() && !this->dataPtr->enableSensors)`.

src/gui/plugins/axes/AxesConfig.qml
```diff
@@ -0,0 +1,255 @@
...
+      AxesConfig.onCurrentIndexChanged(index)
+      axesLength.value = AxesConfig.length
+      x.value = AxesConfig.axesX
+      y.value = AxesConfig.axesY
+      z.value = AxesConfig.axesZ
+      roll.value = AxesConfig.axesRoll
+      pitch.value = AxesConfig.axesPitch
+      yaw.value = AxesConfig.axesYaw
+    }
+    onAccepted: {
```
chapulina, line 102: Each time I choose a new model, a new visual is created. Is that intentional? I end up with many of them

image

src/gui/plugins/axes/AxesConfig.qml
```diff
@@ -0,0 +1,255 @@
...
+  Item {
+    Layout.columnSpan: 1
+    Layout.rowSpan: 1
+    Layout.fillWidth: true
+  }
+
+  CheckBox {
+    Layout.alignment: Qt.AlignHCenter
+    Layout.columnSpan: 2
+    text: qsTr("Lines/Arrow Axes")
```
chapulina, line 66: Reading this I wasn't sure what to expect. Maybe just "Show arrows"?

src/gui/plugins/axes/AxesConfig.qml
```diff
@@ -0,0 +1,255 @@
...
+  Item {
+    Layout.columnSpan: 1
+    Layout.rowSpan: 1
+    Layout.fillWidth: true
+  }
+
+  CheckBox {
+    Layout.alignment: Qt.AlignHCenter
+    Layout.columnSpan: 2
+    text: qsTr("Show/Hide Axes")
```
chapulina, line 42: I would just say "Show axes" - so when the box is checked, it's showing, and when unchecked, it's not showing

src/gui/plugins/axes/AxesConfig.cc
```diff
@@ -0,0 +1,373 @@
...
+/////////////////////////////////////////////////
+double AxesConfig::axesYaw() const
+{
+  return this->dataPtr->pose.Rot().Euler().Z();
+}
+
+/////////////////////////////////////////////////
+void AxesConfig::EntitiesInScene()
+{
+  std::set<std::string> set_entities;
```
chapulina, line 318: Style :wink: 
```suggestion
  std::set<std::string> setEntities;
```

ahcorde:
This plugin will create an axes for each entity in the world. I can extend the plugin to accept other types of objects in the scene.

chapulina:
Removing the beta label, this can go in a minor release.

src/gui/plugins/axes/AxesConfig.qml
ahcorde: @chapulina, yes, this is intentional. The GUI allows you to check or uncheck the `Show Axes` option that enables the axes for the selected entity. I did it like this because you may want to visualize more than one axes on the scene.

Do you expect a different behaviour ?

chapulina: I see. The usage is still a bit not intuitive to me. For example, if multiple models have axes showing, and I hide the axes for one of them. When I change to another entity, that entity's axes disappear:

axes

I'd expect that when I select the "sphere" from the dropdown, the sphere keeps its axes showing, and the "Show Axes" box checks itself. That is, the widget updates itself to match the currently model on the dropdown.

src/gui/plugins/axes/AxesConfig.qml
```diff
@@ -0,0 +1,255 @@
...
+  }
+
+  // Left spacer
+  Item {
+    Layout.columnSpan: 1
+    Layout.rowSpan: 1
+    Layout.fillWidth: true
+  }
+
+  ComboBox {
```
chapulina, line 87: All the information on the plugin applies to the entity that's selected on the combo box. So I think it makes sense to put the dropdown on the top of the plugin, above the checkboxes.

chapulina (commented):
Since this PR doesn't need anything from `ign-gazebo`, what do you think of moving it to `ign-gui`, so other applications can also use it? Now that `ign-gazebo` can use `ignition::gui::events::Render`, this plugin doesn't need to be here.

ahcorde:
Closing this PR in favour of this other PR in ign-gui https://github.com/ignitionrobotics/ign-gui/pull/182

Issue #43, opened 2019-12-10
Add file extension to file name in double pendulum video record demo


osrf-migration:
**Original report ([archived issue](https://osrf-migration.github.io/ignition-gh-pages/#!/ignitionrobotics/ign-gazebo/issues/43)) by John Shepherd (Bitbucket: [John Shepherd](https://bitbucket.org/%7Bc4803212-effe-4921-9323-02cd01b8bc9c%7D/), GitHub: [jshep1](https://github.com/jshep1)).**

----------------------------------------
# Summary

Currently the file extension isn't appended to the end of the typed file name for the `video_record_dbl_pendulum.sdf` example.

# Motivation

This would make the example more user friendly.

mabelzhang:
@mcres will take a stab at this.

mcres:
Yup, let's see if I can solve this one!

PR #153, opened 2020-05-22
Fix for issue 43

mcres:
Issue #43 

This will filter the record menu and write the format to the file according to which button the user pushed (mp4 or ogv). 

This is my first PR, I apologize if there are any obvious errors!


mcres:
You're right, I didn't notice that. Thanks for the suggestions! I'll work on that.

I think we should respect the typed name as well, independently of the button pushed in the record menu. 
However, what happens when the user tries to save the file as something other than `*.mp4` or `*.ogv`, for instance `*.avi`? Maybe we should respect that, even if the format doesn't exist at all? If I save a video as `ignition.myOwnFormat`, VLC is still able to open and play it.

From a user's point of view, that should probably be the behaviour. But I've realized that there could be a discrepancy between `VideoRecorder::OnStart` and `VideoRecorder::OnSave` as we could be recording in one format and saving in another one? It doesn't seem to be a problem, but I put the question here just in case.

mcres:
With this commit, I save the file with the format specified in the drop-down list (which in turn is initially set depending on the selected button in the record menu, but can be changed). 
If the user types any format, I respect that one. 

This is the most intuitive behavior for me, but any comments are appreciated!

mabelzhang:
> discrepancy between `VideoRecorder::OnStart` and `VideoRecorder::OnSave`
Ohhh, I tried it out and I see what you mean. That is a problem. In that case, we cannot respect the user's input of extension (sorry for suggesting that), because the format we started with must be the file we save.

Is there a way to pre-populate the file name box with the extension `OnStart` was called with? I notice that it's usually populated with the name in a previous save. That would be a better cue to the user that they are required to save it in that format.

Then we can revert to the original behavior of appending the extension to the end, regardless of what the user puts in. Except if they already put the correct extension in the file name box, then we don't add the extension again.

> If I save a video as `ignition.myOwnFormat`, VLC is still able to open and play it.
That doesn't mean it's a valid format. In Ubuntu, if you right-click on the file and go to Properties, then under the Video tab > General > Container, you will see the actual format of the file, regardless of what the extension is.

For example, I was able to save a .mp4 file whose actual Container says Ogg, and an .ogv file with Video > Codec H.264, which is a mp4 encoding. If I select mp4 and save a .avi file, it still also says Codec H.264. Any media player is able to look at the actual format, not the extension, to determine how to play it.

mcres:
> Then we can revert to the original behavior of appending the extension to the end of the user input, regardless of what the user puts in. Except if the user already put the correct extension in the file name box, then we don't add the extension again.

OK, let's do that. If the user writes `fileName.avi`, should that be converted to `fileName.avi.mp4` or `fileName.mp4`? I guess the second one, right?

> Is there a way to pre-populate the file name box with the extension `OnStart` was called with? 

I looked at the [FileDialog documentation](https://doc.qt.io/qt-5/qml-qt-labs-platform-filedialog.html), but I'm not sure how I can find out what's the exact Qt version Ignition is using to check the corresponding documentation. 
For example, `defaultSuffix` property could be used but seems to belong to a more recent Qt version.
Another possibility would be using `fileUrl`, but it turns out it is a read-only property.

I didn't found much on Google, but I'll keep searching and see if something comes up.

mabelzhang:
> If the user writes `fileName.avi`, should that be converted to `fileName.avi.mp4` or `fileName.mp4`? I guess the second one, right?

Let's do the first one, because the user might want to put a period somewhere in the file name, e.g. `2020.05.07`, and we wouldn't want to deal with figuring out what they meant. In that example, we can't just strip everything after the `.` and append `mp4`, as it'd result in `2020.05.mp4`.

> what's the exact Qt version Ignition is using

The CMakeLists.txt shows it's Qt5 https://github.com/ignitionrobotics/ign-gazebo/blob/master/CMakeLists.txt#L80

mcres:
It seems that currently there's no way to preset the name field in `FileDialog`, or at least I couldn't find one... Maybe [this property](https://doc.qt.io/qt-5/qml-qtquick-dialogs-filedialog.html#defaultSuffix-prop) can be helpful if Qt libraries are updated someday? 

Tight now, there's no problem when saving multiple times in the same format. However, if the first time I save as `.mp4` and then I decide to save as `.ogv`, the preset name field of the file dialog will be something like `fileName.mp4`. In that case, I could change from `fileName.mp4` to `fileName.ogv` for the user, but it doesn't seem intuitive either.

mabelzhang:
@iche033 @JShep1 Do you guys have input on the user interface for saving a recorded video? This PR adds the file extension to the file name, but the current behavior is a bit clunky. Maybe you know of a better way to access things in Qt. I would also like another pair of eyes on this PR before merging as I haven't done stuff with GUI.

iche033:
we're using the Qt version on ubuntu bionic, which is 5.9.5, and I don't think that has support for the `defaultSuffix` property.

If the user tries to save in a different file extension than recorded, I think we can still respect the user's choice of filename but the encoding will not match, i.e. it'll be whatever format that was selected originally.

If we want to go one step further, we could pop up a confirmation dialog saying that the file extension in the user specified name does not match the encoding format. If user confirms, we save and close file dialog, else return to the file dialog and let user specify a different name.

mcres:
@iche033  Thanks for the feedback!

> If we want to go one step further, we could pop up a confirmation dialog saying that the file extension in the user specified name does not match the encoding format. If user confirms, we save and close file dialog, else return to the file dialog and let user specify a different name.

I think this is a good workaround, as we can't use the `defaultSuffix` property.
I'll work on that.

mcres:
I've implemented the string logic in the .qml, so the other files remain the same as previous to the PR. 
(Didn't know I could do that and I thought it was a good idea).

There's also the confirmation dialog suggested by @iche033.

src/gui/plugins/video_recorder/VideoRecorder.qml
```diff
@@ -37,16 +37,67 @@ ToolBar {
...
+    Dialog {
+      id: mismatchDialog 
+      title: "Enconding and filename mismatch"
+      modal: true
+      focus: false
+      width: 700
+      height: 200
+      parent: ApplicationWindow.overlay
+      x: (parent.width - width) / 2
+      y: (parent.height - height) / 2
```
mcres, line 78: @iche033 I'd really appreciate if you could tell me if there's anything wrong with this. I made research and it works for me, but I've got the feeling that on a screen other than my laptop's weirds things could happen. 

Thank you!

src/gui/plugins/video_recorder/VideoRecorder.qml
```diff
@@ -37,16 +37,67 @@ ToolBar {
...
+      standardButtons: Dialog.Save | Dialog.Abort
+
+      GridLayout {
+        id: grid
+        rows: 2
+        flow: GridLayout.TopToBottom
+        anchors.fill: parent
+        Label { text: "File extension doesn't correspond with video encoding."; font.pixelSize: 20 }
```
mabelzhang, line 86: Not sure about our style guide for QML, but I have a feeling the things in the braces `{}` should be on a new line, and then one line per field like the other existing items.

mcres: Thanks! Yes, I think so.

src/gui/plugins/video_recorder/VideoRecorder.qml
```diff
@@ -73,13 +132,17 @@ ToolBar {
       MenuItem {
         text: "mp4"
         onTriggered: {
+          fileDialog.nameFilters = [ "*.mp4" ]
```
mabelzhang, line 135: Last nit, remove spaces within the brackets.

src/gui/plugins/video_recorder/VideoRecorder.qml
```diff
@@ -73,13 +132,17 @@ ToolBar {
...
+          fileDialog.selectedFormat = "mp4"
           VideoRecorder.OnStart("mp4")
           animation.start()
         }
       }
       MenuItem {
         text: "ogv"
         onTriggered: {
+          fileDialog.nameFilters = [ "*.ogv" ]
```
mabelzhang, line 144: Same as above

PR #155, opened 2020-05-26
Allow renaming breadcrumb models if there is a name conflict

azeey:
Adds a new parameter `<allow_renaming>` that, when set to true, will rename the spawned breadcrumb model by adding a numeric suffix if there is already a model with the same name in the world. This allows using multiple instances of the same Breadcrumb system and configuration.


src/systems/breadcrumbs/Breadcrumbs.hh
```diff
@@ -128,6 +132,10 @@ namespace systems
     /// \brief Whether the deployed models will be performers
     private: bool isPerformer{false};
 
+    /// \brief Whether the deployed model should be renamed if a model with the
+    /// same name already exists
+    private: bool allowRenaming{false};
```
nkoenig, line 137: A default value of `true` seems like a better option. It is different from the released behavior, but I think a typical user would want renaming.

chapulina: Defaulting to false matches the behaviour of the `UserCommands` system. In that case, I think that renaming without the user explicitly asking for it can have unintended consequences, like changing topic names, or even spawning the same model twice by mistake.

https://github.com/ignitionrobotics/ign-gazebo/blob/44951e3ddfd238f24182d4d80b1376f0d426bd43/src/systems/user_commands/UserCommands.cc#L456

I haven't been using the breadcrumbs plugin to have a notion of how the breadcrumbs themselves are being used. But if it's common to use their names in topics, or we want to safeguard against deploying the same model twice, I think false could be a good default here too. It also has the advantage of being backwards compatible.

azeey: I'd agree with @chapulina. Renaming should not occur in typical usage of Breadcrumbs. It came up for dynamic obstacles because we wanted all obstacles to have the same name (hence, topic). Even before this PR, breadcrumbs are assigned a numeric suffix equal to the number of breadcrumbs deployed so far, so there should not be any conflict among models spawned by a single Breadcrumb system. The conflict is with multiple Breadcrumb systems or with other models that already exist in the system.

nkoenig: I see, thanks for the explanation.

Issue #91, opened 2020-04-25
Dynamically update actor trajectories


chapulina:
Here is an idea for how to expose skeleton control to systems. Currently, actors can only be controlled from the SDF `<script>`. With these components, systems should have the flexibility to control actors in various different ways. The only thing missing is a way for a system to set waypoints that are automatically interpolated, but I'm not sure how useful that would be.

:arrow_right: `components::Pose`
* Implemented on [this PR](https://osrf-migration.github.io/ignition-gh-pages/#!/ignitionrobotics/ign-gazebo/pull-requests/542/page/1)
* Defines the reference point for all trajectory waypoints (i.e. origin).
* Initialized to <actor><pose>.
* Already merged into Dome.
* Works whether ScriptFromSdf or not.

:arrow_right: `components::ScriptFromSdf`
* Defaults to true
* If false, the entire <script> from SDF is ignored.
* If set back to true, SDF script starts working again.

:arrow_right: `components::TrajectoryPose`
* Added on [this PR](https://osrf-migration.github.io/ignition-gh-pages/#!/ignitionrobotics/ign-gazebo/pull-requests/579/page/1) (and [this](https://osrf-migration.github.io/ignition-gh-pages/#!/ignitionrobotics/ign-gazebo/pull-requests/549/page/1) declined one)
* Defines the pose of the root bone with respect to components::Pose.
* If ScriptFromSdf, it will be set according to the current interpolation between `<waypoint><pose>`s.
* If not ScriptFromSdf, unlike <waypoint><pose>, it's not interpolated from the previous pose and performs teleportation.

:arrow_right: `components::AnimationName`
* Name of animation being currently played.
* If ScriptFromSdf, it will be set according to the current <trajectory type> (i.e. <animation name>).
* It must correspond to one of the `<animation>`s loaded with the SDF.

:arrow_right: `components::AnimationTime`
* Time within animation being currently played.
* If ScriptFromSdf, it will be set according to the animation being currently played.
* If not ScriptFromSdf, BoneTransforms will be ignored and the render engine is responsible for animating them.

:arrow_right: `components::UseBoneTransforms`
* Defaults to false
* If true, bone transformations will be calculated by Ignition Common and set through the BoneTransforms component.
* If false, the BoneTransforms component is ignored and the render engine is responsible for calculating the transforms based on AnimationName and AnimationTime.

:arrow_right: `components::BoneTransforms`
[...]

PR #157, opened 2020-05-27
More actor components and follow plugin

chapulina:
Addresses part of #91.

actor_follow_4

Add a couple of components:

* `AnimationName`
* `AnimationTime`

Also added the `FollowActor` plugin as a useful*, which is a port of [servicesim::FollowActorPlugin](https://github.com/osrf/servicesim/blob/master/servicesim_competition/src/FollowActorPlugin.hh), with a subset of functionalities:

* An entity to follow can be set on the SDF (`<target>`)
* A minimum distance from the target's origin can be set
* If the target is further than the maximum distance, the actor stops following (this was useful in the ServiceSim context, maybe not so much outside of it)
* In case the actor has many animations, a specific animation one can be chosen in the SDF
* The actor's velocity is set on the SDF

\* Actually useful, unlike the [examples/plugin/command_actor](https://github.com/ignitionrobotics/ign-gazebo/tree/master/examples/plugin/command_actor) I had been adding.

---

:warning: One inconvenient part is having to manually set the "animation X velocity", which is different for each animation file.

I looked into calculating it automatically inside rendering like we do when using `interpolate_x`. That approach involves calculating the average speed of the animation based on the start and end positions of the root node. That's only an approximation though. Plugin developers have more flexibility if they can directly set `AnimationTime` at each time step. That's more flexible but also more work. I took that route here mainly because that interface was exposed in Gazebo-classic's [Actor::SetScriptTime](https://github.com/osrf/gazebo/blob/d33e6429e3aa63181c3cef6d292370e6dc3f4320/gazebo/physics/Actor.hh#L132-L136), which I saw used on at least one other plugin besides ServiceSim's.

In the future, we could look into exposing the skeleton information to all systems as a component. We need to be careful when the `SceneManager` is only on the client though.


src/systems/follow_actor/FollowActor.cc
```diff
@@ -0,0 +1,277 @@
...
+
+  if (_info.paused)
+    return;
+
+  // TODO(louise) Throttle this system
+
+  // Time delta
+  std::chrono::duration<double> dtDuration = _info.simTime -
+      this->dataPtr->lastUpdate;
+  double dt = dtDuration.count();
```
luca-della-vedova, line 201: More a curiosity here, can we just use _info.dt or is there any benefit in calculating the dt manually?

chapulina: Yeah good point. I was planning to throttle this system so it doesn't run at 1000 / 500 Hz. I'll go ahead and do that.

src/systems/follow_actor/FollowActor.cc
```diff
@@ -0,0 +1,277 @@
...
+  if (nullptr == animationNameComp)
+  {
+    _ecm.CreateComponent(_entity, components::AnimationName(animationName));
+  }
+  else
+  {
+    *animationNameComp = components::AnimationName(animationName);
+  }
+  _ecm.SetChanged(_entity,
+      components::AnimationName::typeId, ComponentState::OneTimeChange);
```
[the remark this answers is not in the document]
chapulina: Yeah we definitely need to document those. It actually took me a while to debug what was missing. Added some comments here: 71466eead2ffdc77d7a26515965b520724907568

chapulina:
> one behavior of the follow actor system I noticed is that if the box moves of out range and back within range again, the actor would no longer follow the box.

That was actually a feature on ServiceSim. If the robot moved too fast, the human would get lost and stop following. It wouldn't restart following until the robot asked to be followed again.

Since this port doesn't have a transport interface, there's no way to get the actor to follow again. I realize this isn't very useful. So I changed the implementation so that the actor always follows if within range: e9588c2

PR #160, opened 2020-05-29
Use collision bitmask flag for meshes

luca-della-vedova:
Preliminary support for collision bitmask was introduce in ign-physics with [this PR](https://osrf-migration.github.io/ignition-gh-pages/#!/ignitionrobotics/ign-physics/pull-requests/116/page/1).

That introduces filling the bitmask and introducing the filtering in the [ConstructSdfCollision](https://github.com/ignitionrobotics/ign-physics/blob/ign-physics2/dartsim/src/SDFFeatures.cc#L499) function. The issue is that the function is called in an [else statement](https://github.com/ignitionrobotics/ign-gazebo/blob/master/src/systems/physics/Physics.cc#L730) only if the object is not a mesh.
This means that before this PR meshes didn't implement the collide_bitmask flag, now they do. This effectively makes the [change in the ConstructSdfCollision](https://github.com/ignitionrobotics/ign-physics/blob/ign-physics2/dartsim/src/SDFFeatures.cc#L574-L579) obsolete and it can be removed.

While this is just a few lines and it fixes the problem, it introduces an additional feature to the MinimumFeatureList which is probably not desirable. Also, I guess there should be a discussion about whether more features from the ConstructSdfCollision functions should be ported to make them work with meshes as well, from what I can see in the code there is a whole chunk of SDF [parameters related to friction](https://github.com/ignitionrobotics/ign-physics/blob/ign-physics2/dartsim/src/SDFFeatures.cc#L539-L565) that will be applied to primitives but not to meshes, which doesn't sound correct.

Open to ideas on whether this fix is acceptable (and whether we should find a way to port the above) or if I should look at alternative fixes.

### DIFF
--- a/src/systems/physics/Physics.cc
+++ b/src/systems/physics/Physics.cc
@@ -745,7 +745,7 @@ void PhysicsPrivate::CreatePhysicsEntities(const EntityComponentManager &_ecm)
         sdf::Collision collision = _collElement->Data();
         collision.SetRawPose(_pose->Data());
         collision.SetPoseRelativeTo("");
-        uint16_t collideBitmask = collision.Surface()->Contact()->CollideBitmask();
+        auto collideBitmask = collision.Surface()->Contact()->CollideBitmask();
 
         ShapePtrType collisionPtrPhys;
         if (_geom->Data().Type() == sdf::GeometryType::MESH)
--- a/src/systems/physics/Physics.cc
+++ b/src/systems/physics/Physics.cc
@@ -835,6 +835,17 @@ void PhysicsPrivate::CreatePhysicsEntities(const EntityComponentManager &_ecm)
         {
           filterMaskFeature->SetCollisionFilterMask(collideBitmask);
         }
+        else
+        {
+          static bool informed{false};
+          if (!informed)
+          {
+            igndbg << "Attempting to set collision bitmasks, but the physics "
+                   << "engine doesn't support feature [CollisionFilterMask]. "
+                   << "Collision bitmasks will be ignored." << std::endl;
+            informed = true;
+          }
+        }
 
         this->entityCollisionMap.insert(
             std::make_pair(_entity, collisionPtrPhys));
--- a/src/systems/physics/Physics.cc
+++ b/src/systems/physics/Physics.cc
@@ -311,8 +311,7 @@ class ignition::gazebo::systems::PhysicsPrivate
   public: using CollisionFeatureList = ignition::physics::FeatureList<
             MinimumFeatureList,
             ignition::physics::GetContactsFromLastStepFeature,
-            ignition::physics::sdf::ConstructSdfCollision,
-            ignition::physics::CollisionFilterMaskFeature>;
+            ignition::physics::sdf::ConstructSdfCollision>;
 
   /// \brief Collision type with collision features.
   public: using ShapePtrType = ignition::physics::ShapePtr<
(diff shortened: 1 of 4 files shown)


src/systems/physics/Physics.cc
```diff
@@ -138,6 +139,7 @@ class ignition::gazebo::systems::PhysicsPrivate
           ignition::physics::GetEntities,
           ignition::physics::GetContactsFromLastStepFeature,
           ignition::physics::RemoveEntities,
+          ignition::physics::CollisionFilterMaskFeature,
```
iche033, line 142: we made a few features optional in pull request #140 and I think we can also make the `CollisionFilterMaskFeature` optional as well. We need to merge 3 -> 4 forward first.

iche033: merging 3 -> 4 in pull request #171

iche033: ok PR merged. Can you merge with `master` and make this feature optional? thanks.

luca-della-vedova: Done, now it is optional, added to `CollisionFeatureList`

src/systems/physics/Physics.cc
```diff
@@ -743,6 +745,7 @@ void PhysicsPrivate::CreatePhysicsEntities(const EntityComponentManager &_ecm)
         sdf::Collision collision = _collElement->Data();
         collision.SetRawPose(_pose->Data());
         collision.SetPoseRelativeTo("");
+        uint16_t collideBitmask = collision.Surface()->Contact()->CollideBitmask();
```
iche033, line 748: one minor codecheck warning about line length:
https://build.osrfoundation.org/job/ignition_gazebo-ci-pr_any-ubuntu_auto-amd64/3550/cppcheckResult/

iche033:
the homebrew build failure looks be unrelated. I think we may need to release a new ign-gui bottle for it to pass

src/systems/physics/Physics.cc
luca-della-vedova: [Done](https://github.com/ignitionrobotics/ign-gazebo/pull/160/commits/fd384eb9827162ae14c681a594a79065e8d62d55)! Figured I might as well use auto which saved a few precious characters.

luca-della-vedova:
> Thanks for removing it from the minimum feature list, but I think this still breaks backwards compatibility.
> 
> TPE for example will lose collision support because it doesn't have the bitmask feature. Do you mind creating a new feature list just for this? I know it's more work, but this feature feels niche enough to be an option in itself.

Done [here](https://github.com/ignitionrobotics/ign-gazebo/pull/160/commits/a58ff5f56f66b55ace995af47dbe5227022a060f), I used a different strategy compared to the parts above for printing the error (not printing any). The reason is that even if the user doesn't specify a collide_bitmask in SDF there will still be a value assigned to it (default of 0xFF), so the warning for the missing feature would be always printed.

chapulina:
> There will still be warnings printed in the entityCast function though if the bitmask filter is not implemented, but I'm not sure how to handle them.

That's a good point. We could just make that function silent and let the caller responsible for printing messages.

> We would need a way to know if the user defined a bitmask

I'd be ok punting on this for now, maybe ticketing an issue.

luca-della-vedova:
> That's a good point. We could just make that function silent and let the caller responsible for printing messages.
> 
Something like [this](https://github.com/ignitionrobotics/ign-gazebo/pull/160/commits/94f60ab187080121cb94697abc0260519b2d15a3)?

chapulina:
> Something like this?

Yup, thanks, that looks good to me!

iche033:
the latest changes look good to me, thanks!

PR #161, opened 2020-05-29
diff_drive system with speed limiter

caguero:
This patch adds extra SDF parameters to the `diff_drive` system to add extra constrains if needed. We can now limit min/max velocity, min/max acceleration and min/max jerk.


src/systems/diff_drive/DiffDrive.cc
```diff
@@ -144,6 +188,57 @@ void DiffDrive::Configure(const Entity &_entity,
   this->dataPtr->wheelRadius = _sdf->Get<double>("wheel_radius",
       this->dataPtr->wheelRadius).first;
 
+  // Parse speed limiter parameters.
+  bool hasVelocityLimits     = false;
+  bool hasAccelerationLimits = false;
+  bool hasJerkLimits         = false;
+  double minVel   = 0;
+  double maxVel   = 0;
```
azeey, line 196: Should the max values be initialized to `inf` or `std::numeric_limits<double>::max()`? Otherwise, if a user specifies a `min_velocity` for example and forgets to set `max_velocity`, the limiter will clamp the max velocity to 0.

caguero: Good point, changed in 8876be62680b442c8049d06471abee7b128c1e69.

caguero:
> Looks good on first pass. Are you planning to add tests?

I tweaked the existing test. Originally it was doing some acceleration control so I removed it and added to SDF test world.

caguero:
> It would be great if we had tests with expectations on max velocity and acceleration. Currently, if I comment out the `<max_velocity>` tag, the tests still pass. Otherwise, this looks great!

Added in ac4cee7. The test should fail if you remove `<max_acceleration>` or `<max_velocity>` from the `diff_drive.sdf`.

PR #164, opened 2020-05-31
Remove motor model from quadcopter model in world file

Jaeyoung-Lim:
The motor model needs to be removed since it is already included in the model in fuel

Fixes the warning: 
```
Warning [Model.cc:210] Non-unique names detected in XML children of model with name[x3].
```


azeey:
Thanks for your contribution @Jaeyoung-Lim. It turns out the issue is actually in sdformat and it is emitting that warning because there are repeated 

```xml
<plugin  filename="libignition-gazebo-multicopter-motor-model-system.so" name="ignition::gazebo::systems::MulticopterMotorModel">
```

elements in `quadcopter.sdf`, but it shouldn't be doing so because that's a valid use case. I've created https://github.com/osrf/sdformat/issues/294 to track this. I'll go ahead and close this and we can continue the discussion there.

Jaeyoung-Lim:
@azeey That makes sense, thanks!

PR #165, opened 2020-06-01
Support joint force commands with the Battery plugin

nkoenig:
Joint force commands were not checked in the battery plugin.

To test:
1. `ign gazebo -v 4 examples/worlds/lift_drag_battery.sdf`
2. Listen to the battery topic: `ign topic -et /model/lift_drag_demo_model/battery/linear_battery/state`
3. Run simulation.
4. The battery should be full and not discharging.
5. Start the propeller: `ign topic -t "/model/lift_drag_demo_model/joint/rod_1_joint/cmd_force" -m ignition.msgs.Double  -p "data: 0.7"`
6. The battery should start discharging.

### DIFF
--- a/src/systems/battery_plugin/LinearBatteryPlugin.cc
+++ b/src/systems/battery_plugin/LinearBatteryPlugin.cc
@@ -36,6 +36,7 @@
 #include "ignition/gazebo/components/BatterySoC.hh"
 #include "ignition/gazebo/components/Name.hh"
 #include "ignition/gazebo/components/World.hh"
+#include "ignition/gazebo/components/JointForceCmd.hh"
 #include "ignition/gazebo/components/JointVelocityCmd.hh"
 #include "ignition/gazebo/components/ParentEntity.hh"
 #include "ignition/gazebo/components/Joint.hh"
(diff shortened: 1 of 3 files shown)


src/systems/battery_plugin/LinearBatteryPlugin.cc
```diff
@@ -333,6 +335,20 @@ void LinearBatteryPlugin::PreUpdate(
...
+
+      const auto *jointForceCmd =
+        _ecm.Component<components::JointForceCmd>(jointEntity);
+      if (jointForceCmd)
+      {
+        for (double jointForce : jointForceCmd->Data())
+        {
+          if (fabsf(static_cast<float>(jointForce)) > 0)
+          {
+            this->dataPtr->startDraining = true;
```
caguero, line 347: I don't see any logic for stop draining the battery if I stop applying force to the joints. Maybe initializing `this->dataPtr->startDraining` to `false` at `PreUpdate()` makes the trick.

caguero:
Just one comment, other than that ready to approve.

src/systems/battery_plugin/LinearBatteryPlugin.cc
nkoenig: I believe that is the intent. There is a trigger to start the battery draining, but we don't monitor the actual forces used by the robot in order to alter how much or how fast the battery drains. The joint velocity command behaves in this manner as well.

caguero: I think it makes more sense to me to stop draining the battery if there are no velocity or force commands but if that's the intent the code looks good to me. We can revisit it later if needed.

nkoenig: How about tackling this feature under the umbrella of your battery recharging work.

caguero: Sounds good!

Issue #31, opened 2019-07-25
Insert model from the GUI


osrf-migration:
**Original report ([archived issue](https://osrf-migration.github.io/ignition-gh-pages/#!/ignitionrobotics/ign-gazebo/issues/31)) by Louise Poubel (Bitbucket: [chapulina](https://bitbucket.org/%7B5cfa2075-477b-4ded-bdb9-8d2479544ec4%7D/), GitHub: [chapulina](https://github.com/chapulina)).**

----------------------------------------

It is convenient for users to construct environments through the UI. To be able to do that, they need to be able to insert new models into simulation. This issue is about creating a GUI plugin which allows users to inspect and insert SDF models from:

* custom local paths (Requires #123)
* Fuel's local cache
* Fuel's online models

A few more nice features:

* Display resource names and thumbnails
* Display resources organized by folders
* Search bar to filter folders and resources
* A button to add new local folders to be listed
* Drag resources from one local folder to another to copy / move them
* Drag resources into a Fuel user to upload to that user's account

# Alternatives

* Since Blueprint, users are already able to drag models directly from Fuel on the browser. That doesn't cover local models though.
* It's common on various programs for the user to drag SDF files from a file browser. We should also support that use case besides the plugin in question.

# Additional resources

* Feature request: https://github.com/osrf/subt/issues/130
* [Create entity service](https://osrf-migration.github.io/ignition-gh-pages/#!/ignitionrobotics/ign-gazebo/pull-requests/143/a-6-user-command-spawn-factory-create/diff)
* Gazebo-classic's [InsertModelWidget](https://github.com/osrf/gazebo/blob/gazebo11/gazebo/gui/InsertModelWidget.hh)
    * Lists model by their name in `model.config`
    * Lists all local paths in `GAZEBO_MODEL_PATH`
    * Lists all models in https://models.gazebosim.org
    * Has partial support for Fuel
    * Interaction:
        1. Click on model name
        1. A temp preview visual is attached to mouse, at Z = 0 height
        1. Click on 3D scene, model is spawned
    ![classic_insert](https://user-images.githubusercontent.com/5751272/82961035-0a3b4a00-9f71-11ea-89db-275194edac38.png)
* GzWeb's [insert menu](https://github.com/osrf/gzweb/blob/9598775be65535fc50555b469d712620e7193129/gz3d/client/index.html#L53)
    * Lists a hardcoded list of models organized by categories
    * Lists models by their name and thumbnail
[...]

PR #172, opened 2020-06-03
Resource env var, with transport interface

chapulina:
* Resolves https://github.com/ignitionrobotics/ign-gazebo/issues/123.
* ~Requires https://github.com/ignitionrobotics/ign-common/pull/70~
* ~Builds on top of #181~ 

---

This allows models and meshes to be found according to the the `IGN_GAZEBO_RESOURCE_PATH` environment variable.

* It's the Ignition Gazebo equivalent of both `GAZEBO_MODEL_PATH` and `GAZEBO_RESOURCE_PATH`.
* That variable is already being used to find worlds, now it's extended to `<include>`s and mesh `<uri>`s.
* The name "resources" is better fitting than "model", since it can load different types of resources.
* We've already been supporting the `SDF_PATH` and `IGN_FILE_PATH` variables with the same behaviour, so this implementation is backwards compatible and keeps those variables in sync with `IGN_GAZEBO_RESOURCE_PATH`.
* I added a transport interface so these paths can be queried and updated.
    * The first consumer will be the new GUI resources plugin, see #31 
    * This keeps the ground truth in the server. Gazebo-classic users often would have issues because `gzserver` and `gzclient` would be in different shells and have access to different environment variables. See this outstanding issue: https://github.com/osrf/gazebo/issues/2403
    * I kept the services in the global namespace because worlds can be found through the paths, so the paths are at a level above worlds.

---

## Try it

Get paths:

`ign service -s /gazebo/get_resource_paths --reqtype ignition.msgs.Empty --reptype ignition.msgs.StringMsg_V --timeout 5000 --req "unused: false"`

Add path:

`ign service -s /gazebo/add_resource_paths --reqtype ignition.msgs.StringMsg_V --reptype ignition.msgs.Empty --timeout 5000 --req "data: '/tmp/lala'"`

---

## TODOs

I'll leave this in draft state until:

- [x] Updates to this tutorial are added to this PR: https://github.com/ignitionrobotics/ign-gazebo/blob/master/tutorials/resources.md
- [x] ~~This design document is updated to reflect this PR: https://github.com/ignitionrobotics/design/pull/1/files~~ I opened a new design PR instead: https://github.com/ignitionrobotics/design/pull/4


chapulina:

chapulina:
@JShep1, I think this is ready to be used by #173 . Let me know if you run into any problems.

src/ServerPrivate.cc
```diff
@@ -384,7 +386,26 @@ void ServerPrivate::CreateEntities()
...
 {
   // Advertise available worlds.
-  this->node.Advertise("/gazebo/worlds", &ServerPrivate::WorldsService, this);
+  std::string worldsService{"/gazebo/worlds"};
+  this->node.Advertise(worldsService, &ServerPrivate::WorldsService, this);
+
+  ignmsg << "Serving world names on [" << worldsService << "]" << std::endl;
+
+  // Resource path management
+  std::string addPathService{"/gazebo/add_resource_paths"};
```
nkoenig, line 395: Suggestion; use `gazebo/resource_paths/add` and `gazebo/resources_paths/get`.

src/Util.cc
```diff
@@ -209,6 +210,109 @@ std::string asFullPath(const std::string &_uri, const std::string &_filePath)
...
+      }), gzPaths.end());
+
+  return gzPaths;
+}
+
+//////////////////////////////////////////////////
+void addResourcePaths(const std::vector<std::string> &_paths)
+{
+  // Environment variable holding resource paths
+  const std::string kResourcePathEnv{"IGN_GAZEBO_RESOURCE_PATH"};
```
nkoenig, line 240: Is it possible to define the `IGN_GAZEBO_RESOURCE_PATH` and `SDF_PATH`  strings once?

src/ServerPrivate.cc
chapulina: I like that pattern, we should use it more often: 0fe84c4f6680296f2099166d2de43890fa295bc9

src/Util.cc
chapulina: For this file, I moved them to local global variables: 0fe84c4f6680296f2099166d2de43890fa295bc9

Were you thinking of something else? Maybe defining them on the header and using throughout the code?

JShep1:
I just ran this and am running into an issue with ign gazebo only showing a black screen with the message
```
[GUI] [Dbg] [Gui.cc:144] GUI requesting list of world names. The server may be busy downloading resources. Please be patient.
```
being printed every few seconds.  I replicated this issue on my laptop as well, are you running into this? @chapulina

chapulina:
@JShep1 , a086574 has a temporary fix to that issue. For some reason the global node works when doing `ign gazebo -s` and `ign gazebo -g` separately, but not for `ign gazebo` :thinking: I'll look better into it tomorrow. For now, be warned that if you add any paths through the add service, the GUI will not know about them automatically, which may cause some models to fail.

src/Util.cc
nkoenig: Yeah, I was thinking of single place. That way the variable could be documented once, and we could eventually have a list of environment variables in a single place. I'm fine saving that for later.

chapulina: Yeah makes sense, moved it to a header: 7e679e960349970cc1af6ab3ba75c2d2dff269bc

chapulina:
Ok, this should be good to be reviewed again. @JShep1 , let me know if you need help using it on #173 .

tutorials/resources.md
```diff
@@ -0,0 +1,158 @@
...
+Ignition will look for URIs (path / URL) in the following, in order:
+
+1. All paths on the `IGN_GAZEBO_RESOURCE_PATH`\* environment variable (if
+   path is URI, scheme is stripped)
+2. Current running path / absolute path
+3. [Ignition Fuel](https://app.ignitionrobotics.org/fuel/models)
+    1. Cache (i.e. `$HOME/.ignition/fuel`)
+    2. Web server
+
+> \* The `SDF_PATH` environment variable also works in some scenarious, but
```
iche033, line 126: `scenarious`: sounds like a serious scenario :)

there is another instance [a few lines below](https://github.com/ignitionrobotics/ign-gazebo/pull/172/files#diff-97df90ddf8e71b07502c2520534c230fR144)

src/gui/PathManager.cc
```diff
@@ -0,0 +1,63 @@
...
+#include "PathManager.hh"
+
+using namespace ignition;
+using namespace gazebo;
+using namespace gazebo::gui;
+
+//////////////////////////////////////////////////
+void onAddResourcePaths(const msgs::StringMsg_V &_msg)
+{
+igndbg << "A" << std::endl;
```
nkoenig, line 34:
```suggestion
```

src/gui/PathManager.cc
```diff
@@ -0,0 +1,63 @@
...
+  auto executed = this->node.Request(service, 5000, res, result);
+
+  if (!executed)
+    ignerr << "Service call timed out for [" << service << "]" << std::endl;
+  else if (!result)
+    ignerr << "Service call failed for [" << service << "]" << std::endl;
+
+  onAddResourcePaths(res);
+
+igndbg << "A" << std::endl;
```
nkoenig, line 60:
```suggestion
```

tutorials/resources.md
nkoenig: Fixed in 956354f7af767d0d5715b8be3ece83fc798535a0

chapulina:
The 3 test failures are not there on `ign-gazebo2`, I'll take a look: https://build.osrfoundation.org/job/ignition_gazebo-ci-pr_any-ubuntu_auto-amd64/3765/testReport/

chapulina:
> The 3 test failures are not there on ign-gazebo2

Not true, they're there on `ign-gazebo2` now.

PR #173, opened 2020-06-03
Insert Local Models

JShep1:
This PR adds the functionality to insert local models similar to the currently existing shapes plugin.  All the functionality is there, but some iterations will need to be made on the interface.

A few minor TODOs:
 - Refine insert model plugin UI
 - ~Integrate IGN_GAZEBO_RESOURCE_PATH (currently the path is hardcoded)~
 - ~Some logic is searching for a forward slash in the path so some work will be needed here to be Windows compatible~
 - ~Maybe a loading screen in placement mode before the visual model is placed into the scene?  The larger models are taking awhile to display and it may confuse the user.~  - I'll leave this out for now, and we can address it later, I don't think it'd be a good use of time to investigate what might be a more complicated issue than it seems (I guess it boils down to how do we know a model has been visualized in the scene or not?)
 - ~Find a no thumbnail image to default to~ - I added one that I created, but we can swap in a better one if desired


JShep1:
Proof of Concept:
localmodel

Obviously there's some GUI work to be done, I just didn't want to invest a huge amount of time into a direction we didn't necessarily want to go.

src/gui/plugins/insert_model/InsertModel.cc
```diff
@@ -0,0 +1,221 @@
...
+
+/////////////////////////////////////////////////
+InsertModel::~InsertModel() = default;
+
+/////////////////////////////////////////////////
+void InsertModel::FindLocalModels(const std::string &_path)
+{
+  std::string path = _path;
+  // Recurse if directory
+  if (common::isDirectory(path))
```
chapulina, line 110: In general it's good to recurse into the directories like this. However in this case, this can result in displaying resources that aren't available.

For example, if `IGN_GAZEBO_RESOURCE_PATH=/tmp/models`, and that folder structure is:

```
└── models
    ├── modelA
    └── subfolder
             └── modelB
``` 

This will work:

```
model://modelA
```

But this won't work:

```
model://modelB
```

So I think we should only display the top-level models.

src/gui/plugins/insert_model/InsertModel.qml
```diff
@@ -0,0 +1,64 @@
...
+*/
+import QtQuick 2.9
+import QtQuick.Window 2.2
+import QtQuick.Controls 2.1
+import QtQuick.Controls.Material 2.2
+import QtQuick.Controls.Material.impl 2.2
+import QtQuick.Layouts 1.3
+import QtQuick.Controls.Styles 1.4
+
+Rectangle {
```
chapulina, line 25: We've already talked about this in person, but I'll just leave some UI notes here so we don't forget:

- [ ] All text should be the same size
- [ ] Use [Text.ElideRight](https://doc.qt.io/qt-5/qml-qtquick-text.html#elide-prop) for names
- [ ] Support light / dark modes
- [ ] Use [Material.elevation](https://doc.qt.io/qt-5/qtquickcontrols2-material.html#material-elevation-attached-prop) to give each item a "card" look
- [ ] Add more padding to the edges of the plugin, and between the items
- [ ] Change the cursor to [Qt.WaitCursor](https://doc.qt.io/qt-5/qml-qtquick-mousearea.html#cursorShape-prop) while the preview is being spawned

src/gui/plugins/insert_model/CMakeLists.txt
```diff
@@ -0,0 +1,4 @@
+gz_add_gui_plugin(InsertModel
```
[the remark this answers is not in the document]
JShep1: f07d872f88652a5600dfcb8a9419a5bb0156461f

src/gui/plugins/insert_model/InsertModel.cc
```diff
@@ -0,0 +1,221 @@
...
+  localModel->setData(QString::fromStdString(_model.sdfPath),
+                      this->roleNames().key("sdf"));
+
+  parentItem->appendRow(localModel);
+}
+
+/////////////////////////////////////////////////
+QHash<int, QByteArray> GridModel::roleNames() const
+{
+  return {
```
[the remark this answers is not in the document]
JShep1: f07d872f88652a5600dfcb8a9419a5bb0156461f

src/gui/plugins/insert_model/InsertModel.cc
```diff
@@ -0,0 +1,221 @@
...
+void InsertModel::LoadConfig(const tinyxml2::XMLElement *)
+{
+  if (this->title.empty())
+    this->title = "InsertModel";
+
+  // For shapes requests
+  ignition::gui::App()->findChild
+    <ignition::gui::MainWindow *>()->installEventFilter(this);
+
+  // TODO(john): create vector of paths from IGN_GAZEBO_RESOURCE_PATH here
```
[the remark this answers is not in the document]
JShep1: f07d872f88652a5600dfcb8a9419a5bb0156461f

src/gui/plugins/insert_model/InsertModel.hh
```diff
@@ -0,0 +1,98 @@
...
+
+    /// \brief Destructor
+    public: ~InsertModel() override;
+
+    // Documentation inherited
+    public: void LoadConfig(const tinyxml2::XMLElement *_pluginElem) override;
+
+    /// \brief Callback in Qt thread when mode changes.
+    /// \param[in] _mode New transform mode
+    public slots: void OnMode(const QString &_mode);
```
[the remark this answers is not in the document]
JShep1: f07d872f88652a5600dfcb8a9419a5bb0156461f

src/gui/plugins/insert_model/InsertModel.hh
```diff
@@ -0,0 +1,98 @@
...
+{
+  class InsertModelPrivate;
+
+  /// \brief Local model used to update the GridModel
+  struct LocalModel
+  {
+    std::string configPath = "";
+    std::string sdfPath = "";
+    std::string thumbnailPath = "";
+    std::string name = "";
```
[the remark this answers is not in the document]
JShep1: f07d872f88652a5600dfcb8a9419a5bb0156461f

src/gui/plugins/insert_model/InsertModel.cc
```diff
@@ -0,0 +1,221 @@
...
+      std::string current(*file);
+      this->FindLocalModels(current);
+    }
+  }
+  else if (common::isFile(path))
+  {
+    std::string fileName = common::basename(path);
+
+    // If we have found model.config, extract thumbnail and sdf
+    if (fileName == "model.config")
```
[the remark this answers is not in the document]
JShep1: f07d872f88652a5600dfcb8a9419a5bb0156461f

src/gui/plugins/insert_model/InsertModel.cc
JShep1: f07d872f88652a5600dfcb8a9419a5bb0156461f

src/gui/plugins/insert_model/InsertModel.cc
```diff
@@ -0,0 +1,221 @@
...
+  class InsertModelPrivate
+  {
+    /// \brief Ignition communication node.
+    public: transport::Node node;
+
+    /// \brief Mutex to protect mode
+    public: std::mutex mutex;
+
+    /// \brief Transform control service name
+    public: std::string service;
```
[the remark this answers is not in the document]
JShep1: 0f1ad24e9c42a527166a409b4257e034e8215beb

src/gui/plugins/insert_model/InsertModel.qml
JShep1: First four points are resolved in 47f214cbf4408a95015531257055d17db667184f

nkoenig:
Can you resolve the conflicts please?

JShep1:
Ready for another review.

A few things that I think need to be tweaked:

- I want the panes to have some top and left margin, I couldn't quite figure out how to make it happen
- There's also currently a bug with the rows not being highlighted when selected, The mouse area seemed to interfere with it.
- Possibly we should have some designed name for the "main" thumbnail, some of the thumbnails are less intuitive and side profiles of the resources
- We need a directory/folder icon, I've pinged @ColeOSRF about this and he is going to send one over

JShep1:
Updated preview:

localmodels

nkoenig:
I see no models when I run `ign gazebo -v 4`, and insert the "Resource Spawning". Is it possible to use the Fuel resource directory by default?

I also see no models when I run `IGN_GAZEBO_RESOURCE_PATH=~/.ignition/fuel/fuel.ignitionrobotics.org/OpenRobotics/models ign gazebo -v 4`. 

Let me know what I'm doing wrong.

JShep1:
> See #245 which has some suggestions.

I'll take a look at these shortly

> I also see no models when I run IGN_GAZEBO_RESOURCE_PATH=~/.ignition/fuel/fuel.ignitionrobotics.org/OpenRobotics/models ign gazebo -v 4.
> Let me know what I'm doing wrong.

Currently, the resource finder looks for a `model.config` only one directory below the provided directory.  Since fuel models follow a name like `model-name-here/1/model.config`, the finder won't locate the model.config file.  I could add a recursive approach that I had at an earlier version that handled these cases.  I believe the reason we transitioned away from it was due to a discussion I had with Louise that indicated that `ign-fuel-tools` would handle this case.  Any thoughts whether we should add the recursive approach? @chapulina @nkoenig   I could be wrong, but I don't quite remember any bugs or issues with it

chapulina:
> Is it possible to use the Fuel resource directory by default?

Yeah, John is right. Fuel's directory format doesn't match the expectations of the environment variable and should be treated as a separate case. (Note that this is also the reason why it's difficult to use Fuel with Gazebo-classic).

The current plan is to add Fuel as a separate item on the left panel, see the [mockup](https://github.com/ignitionrobotics/ign-gazebo/issues/31#issuecomment-639160529). That will use `ign-fuel-tools`.

nkoenig:
> > Is it possible to use the Fuel resource directory by default?
> 
> Yeah, John is right. Fuel's directory format doesn't match the expectations of the environment variable and should be treated as a separate case. (Note that this is also the reason why it's difficult to use Fuel with Gazebo-classic).
> 
> The current plan is to add Fuel as a separate item on the left panel, see the [mockup](https://github.com/ignitionrobotics/ign-gazebo/issues/31#issuecomment-639160529). That will use `ign-fuel-tools`.

Thanks for the links. The Fuel use case is more important than the environment variable use case. The rataionale is that Fuel is the default, or should be the default, for Ignition while environment variables is more of an advanced feature. Also, SubT requires the Fuel use case. How about we get this version in, and start work on the Fuel use-case? I have some comments about the style and UX that I'll put in the issue.

chapulina:
> How about we get this version in, and start work on the Fuel use-case?

Yup, @JShep1 is already on it. 

> The rataionale is that Fuel is the default, or should be the default

John, let's do the reverse of the mockup and put Fuel on top of the local directories then.

---

I'm also working on some UI tweaks for this PR.

src/gui/plugins/resource_spawner/ResourceSpawner.cc
```diff
@@ -0,0 +1,289 @@
...
+
+/////////////////////////////////////////////////
+void ResourceSpawner::LoadConfig(const tinyxml2::XMLElement *)
+{
+  if (this->title.empty())
+    this->title = "Resource Spawner";
+
+  // For resource spawn requests
+  ignition::gui::App()->findChild
+    <ignition::gui::MainWindow *>()->installEventFilter(this);
```
chapulina, line 249: I don't see an `eventFilter` callback, I think this can be removed.

src/gui/plugins/resource_spawner/ResourceSpawner.hh
```diff
@@ -0,0 +1,133 @@
...
+    /// \brief Callback when a resource is selected.
+    /// \param[in] _sdfPath The absolute path to the resource's sdf file
+    public slots: void OnResourceSpawn(const QString &_sdfPath);
+
+    /// \brief Loads a local model from an absolute path to a model.config,
+    /// does nothing if a path not containing model.config is passed in
+    /// \param[in] _path The path to search
+    public: void LoadLocalModel(const std::string &_path);
+
+    public: void AddPath(const std::string &_path);
```
chapulina, line 114: Missing documentation

src/gui/plugins/resource_spawner/ResourceSpawner.cc
JShep1: Resolved 96e39b105f768e4e940c74a2e07c3118c411a73e

src/gui/plugins/resource_spawner/ResourceSpawner.hh
JShep1: Resolved 96e39b105f768e4e940c74a2e07c3118c411a73e

PR #179, opened 2020-06-04
Add odom_topic to diff drive

nkoenig:
This adds a `<odom_topic>` element to the DiffDrive system so that a custom odometry topic can be used.

### DIFF
--- a/test/integration/diff_drive_system.cc
+++ b/test/integration/diff_drive_system.cc
@@ -35,18 +35,6 @@ using namespace ignition;
 using namespace gazebo;
 using namespace std::chrono_literals;
 
-/// \brief Test DiffDrive system
-class DiffDriveTest : public ::testing::TestWithParam<int>
-{
-  // Documentation inherited
-  protected: void SetUp() override
-  {
-    common::Console::SetVerbosity(4);
-    setenv("IGN_GAZEBO_SYSTEM_PLUGIN_PATH",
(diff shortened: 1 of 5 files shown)


nkoenig:
> LGTM. The only small suggestion I have would be to use this in a test. Thanks for the extra docs!

Test added: 676e993721e901e0c19ec7e34f12f54e1dd37d61

test/integration/diff_drive_system.cc
```diff
@@ -87,142 +75,175 @@ class Relay
...
+    // Poses for 4s
+    ASSERT_EQ(4000u, poses.size());
+
+    int sleep = 0;
+    int maxSleep = 30;
+    for (; odomPoses.size() < 150 && sleep < maxSleep; ++sleep)
+    {
+      std::this_thread::sleep_for(std::chrono::milliseconds(100));
+    }
+    ASSERT_NE(maxSleep, sleep);
```
chapulina, line 193: The test refactoring looks good to me, I just think this could continue being an `EXPECT` because the test wouldn't crash if this fails.

nkoenig: Yeah, I just didn't like the crash since it would block the other test in this file from running.

chapulina: Oh so if this line fails the test crashes?

nkoenig: That's right.

PR #180, opened 2020-06-04
Fix segfault in the Breadcrumbs system

azeey:
The segfault occurred because a pointer to a component is invalidated by the creation of new components. This doesn't always happen, but if a component type had 99 items, and we take get a pointer to a component, as soon as new components of the same type are created, the underlying `std::vector` containing the component gets expanded invalidating our pointer. The underlying `std::vector` gets expanded every 100 components.

To test, run the attached world (extension is .txt because github doesn't allow .sdf files). It should crash after 2 sim seconds.

[debug_breadcrumbs_crash.txt](https://github.com/ignitionrobotics/ign-gazebo/files/4732600/debug_breadcrumbs_crash.txt)

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,10 @@
 ## Ignition Gazebo 2.x
 
+### Ignition Gazebo 2.XX.XX (2020-XX-XX)
+
+1. Fix segfault in the Breadcrumbs system
+    * [Pull Request 180](https://github.com/ignitionrobotics/ign-gazebo/pull/180)
+
 ### Ignition Gazebo 2.19.0 (2020-06-02)
 
 1. Use updated model names for spawned models when generating SDFormat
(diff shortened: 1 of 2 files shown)


nkoenig:
> The segfault occurred because a pointer to a component is invalidated by the creation of new components. This doesn't always happen, but if a component type had 99 items, and we take get a pointer to a component, as soon as new components of the same type are created, the underlying `std::vector` containing the component gets expanded invalidating our pointer. The underlying `std::vector` gets expanded every 100 components.

This feels like an issue we should resolve in ECS, correct? I could image this biting us and others in the future. If this is the case, could you create an issue for it?

azeey:
I can create an issue, but I'm not sure what can be done because this is inherent to using a contiguous data container for the components. We can move to using something like `std::list`, but that defeats the performance benefits of ECS.

PR #181, opened 2020-06-04
Move test Relay to its own file

chapulina:
This reduces a lot of code duplication that we currently have in tests.

:sparkles: ` 170 additions and 740 deletions. ` :sparkles:

### DIFF
--- a/test/helpers/Relay.hh
+++ b/test/helpers/Relay.hh
@@ -24,9 +24,10 @@
 
 #include "../plugins/MockSystem.hh"
 
-using namespace ignition;
-using namespace gazebo;
-
+namespace ignition
+{
+namespace gazebo
+{
 namespace test
 {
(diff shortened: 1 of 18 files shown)


test/helpers/Relay.hh
```diff
@@ -0,0 +1,101 @@
...
+
+#include <ignition/gazebo/SystemLoader.hh>
+#include <ignition/gazebo/test_config.hh>
+
+#include "../plugins/MockSystem.hh"
+
+using namespace ignition;
+using namespace gazebo;
+
+namespace test
```
azeey, line 30: This puts the class in `::test`. Wouldn't we want it in `ignition::gazebo::test`?

chapulina: Good call: f1b6451

PR #183, opened 2020-06-05
Battery updates

caguero:
This patch introduces the following battery updates:

* The recharging is no longer triggered by the parameter `socThreshold`. The old behavior implies that a battery automatically starts charging when the battery capacity falls below a certain threshold. I think it's more realistic that the recharging can be started or stopped on demand. I took the liberty of removing the `socThreshold` parameter directly because it wasn't documented in the header file.

* If `<enable_recharging>` is present, two new services are available for starting or stopping the battery recharge.

* If there's no joint velocity/force command, the battery doesn't drain anymore. Before, once the battery started draining, it never stopped even if the joints weren't moving.

* I updated the battery status published. It can now be in `CHARGING`, `DISCHARGING`, `FULL` or `NOT_CHARGING` state.

* I documented the SDF charging parameters and added a few minor tweaks.

### DIFF
--- a/examples/worlds/linear_battery_demo.sdf
+++ b/examples/worlds/linear_battery_demo.sdf
@@ -523,7 +523,7 @@
         <resistance>0.07</resistance>
         <smooth_current_tau>2.0</smooth_current_tau>
         <enable_recharge>true</enable_recharge>
-        <!-- charging I = c / t, discharging I = P / V, 
+        <!-- charging I = c / t, discharging I = P / V,
           charging I should be > discharging I -->
         <charging_time>3.0</charging_time>
         <soc_threshold>0.51</soc_threshold>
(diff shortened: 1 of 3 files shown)


caguero:
> Thanks for improving the battery feature! I wasn't sure how to test the start/stop recharge function. Could we add some documentation on what service to send in the sdf file?

It's a bit harder to test because we're using a one way service, as we don't expect any response. One way services are not supported in `ign` command line tool for now. You can create a publisher similar to the one in https://github.com/ignitionrobotics/ign-transport/blob/master/example/requester_oneway.cc  and publish it to `/vehicle_blue/linear_battery/recharge/start` or `/vehicle_blue/linear_battery/recharge/stop`.

How about I update `ign tools` to support one way services, and then, I update this example in a separate PR?

claireyywang:
> How about I update `ign tools` to support one way services, and then, I update this example in a separate PR?

If that's not too much work then I think that'd be very helpful for demo purposes. And yes, separate PR sounds like a good plan.

PR #202, opened 2020-06-09
Model Preview Fix

JShep1:
This PR fixes a bug that occurred when the user would select multiple shapes from the qml without clicking the scene or pressing Escape.  The fix simply clears any pre-existing visualized models so they do not persist in the scene after they are done being used.


JShep1:
The bug that was occurring:
insertbug

The fix:
insertfix

PR #209, opened 2020-06-17
Disable rendering tests on macOS that are known to fail

azeey:
Tests that use OGRE2 rending have been failing on macOS for a long time. There is an effort to address this from the ign-rendering side in ign-rendering#83, but until that is merged, I thought we should disable these tests. I used gtests ability to disable tests by prepending `DISABLED_` to the test name so that the tests are compiled, but they won't be ran.


chapulina:
I think it would be good to release https://github.com/ignitionrobotics/ign-cmake/pull/102 and use that here.

PR #211, opened 2020-06-18
Scene manager losing header file

FloodShao:
Losing this file will result in "common::SkeletonPtr" not defined problem when using SceneManager


FloodShao:
Changes made for DCO. Anything else needed?

PR #212, opened 2020-06-18
Properly add new models into the scenegraph

nkoenig:
The edge check logic was incorrect because it was checking to see if an edge ID already existed in the `sceneGraph` when it should be checking if a connected between the two vertices already existed.

Without this fix, when a model is spawned it won't be added into the graph properly.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,7 +3,7 @@ cmake_minimum_required(VERSION 3.5.1 FATAL_ERROR)
 #============================================================================
 # Initialize the project
 #============================================================================
-project(ignition-gazebo2 VERSION 2.20.0)
+project(ignition-gazebo2 VERSION 2.20.1)
 
 #============================================================================
 # Find ignition-cmake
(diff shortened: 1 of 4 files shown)


chapulina:
I saw we have a test for when entities are deleted (even though the comment says otherwise):

https://github.com/ignitionrobotics/ign-gazebo/blob/e42ad43901b57a359ea16f030a88d4a7cdb32ce7/test/integration/scene_broadcaster_system.cc#L221-L222

It would be good to also test when entities are added, so we're sure this continues working.

nkoenig:
> I saw we have a test for when entities are deleted (even though the comment says otherwise):
> 
> https://github.com/ignitionrobotics/ign-gazebo/blob/e42ad43901b57a359ea16f030a88d4a7cdb32ce7/test/integration/scene_broadcaster_system.cc#L221-L222
> 
> It would be good to also test when entities are added, so we're sure this continues working.

Test added: 0aa28aed10a484b5a5fbc48845fab4af7948645e

PR #216, opened 2020-06-19
Joint position controller system

caguero:
A system to control joints in position. See the example and test for instructions about how to use it.


chapulina:
I noticed that the implementation assumes that the joint has a single axis. Did you consider generalizing it for joints with more axes?

caguero:
> I noticed that the implementation assumes that the joint has a single axis. Did you consider generalizing it for joints with more axes?

I've added support for multiple axis in 6485a48. In reality I changed my mind a little bit and I decided to keep the controller working for a single axis but the axis can be specified. This way, you can instantiate multiple systems if you want to control multiple axis. The implementation is simpler and I think keeps it easy for the user because most of the time they don't need to set a `revolute2` joint at the same time.

There is one additional reason to justify this design: The `ign_ros` bridge would be more complicated if we'd set all indexes at the same time because instead of using a `std_msgs/Float64` on the ROS side, it should use a `std_msgs/Float64MultiArray`, which is a bit harder to use.

test/integration/joint_position_controller_system.cc
```diff
@@ -0,0 +1,126 @@
...
+  test::Relay testSystem;
+  std::vector<double> currentPosition;
+  testSystem.OnPreUpdate(
+      [&](const gazebo::UpdateInfo &, gazebo::EntityComponentManager &_ecm)
+      {
+        auto joint = _ecm.EntityByComponents(components::Joint(),
+                                             components::Name(jointName));
+        // Create a JointPosition component if it doesn't exist. This signals
+        // physics system to populate the component
+        if (nullptr == _ecm.Component<components::JointPosition>(joint))
```
chapulina, line 81: Since this is a test, we could `EXPECT` that the component doesn't exist in the first iteration. Not super important.

src/systems/joint_position_controller/JointPositionController.hh
```diff
@@ -0,0 +1,67 @@
...
+// Inline bracket to help doxygen filtering.
+inline namespace IGNITION_GAZEBO_VERSION_NAMESPACE {
+namespace systems
+{
+  // Forward declaration
+  class JointPositionControllerPrivate;
+
+  /// \brief Joint position controller which can be attached to a model with a
+  /// reference to a single joint.
+  //  Currently only the first axis of a joint is actuated.
```
chapulina, line 36: Now we can choose the joint index :wink: 

We've been trying to add usage documentation to each system's header. Do you mind documenting the SDF parameters and transport interface here? See an example [here](https://github.com/ignitionrobotics/ign-gazebo/blob/master/src/systems/triggered_publisher/TriggeredPublisher.hh).

caguero: > Now we can choose the joint index 
> 
> We've been trying to add usage documentation to each system's header. Do you mind documenting the SDF parameters and transport interface here? See an example [here](https://github.com/ignitionrobotics/ign-gazebo/blob/master/src/systems/triggered_publisher/TriggeredPublisher.hh).

A good thing to do :) a7ad6ef Thanks!

PR #233, opened 2020-07-06
Joint velocity controller using force

caguero:
This PR extends the functionality of the joint controller.

This controller exposes a velocity interface to control a joint. By default, the controller uses a `JointVelocityCmd` component. This component has some limitations (for now). For example, the joint position limits are not enforced. This patch adds an optional mode (`force mode`) that uses a `JointForceCmd` component instead. To achieve that velocity, we use an internal PID.

To test:

`ign gazebo -r -v 4 ../examples/worlds/joint_controller.sdf`

And then send a target velocity:

`ign topic -t "/model/joint_controller_demo/joint/j1/cmd_vel" -m ignition.msgs.Double -p "data: 0.2"`


src/systems/joint_controller/JointController.cc
```diff
@@ -85,6 +95,71 @@ void JointController::Configure(const Entity &_entity,
     return;
   }
 
+  if (_sdf->HasElement("use_force_commands"))
+  {
+    this->dataPtr->useForceCommands =
+      _sdf->Get<unsigned int>("use_force_commands");
```
azeey, line 101: This is only checking if the element `use_force_commands` exists, not if it is set to true, but code initializes the PID and prints out "Force mode ..." without checking if the parameter is true. Consider doing
```suggestion
    this->dataPtr->useForceCommands = _sdf->Get<bool>("use_force_commands", this->dataPtr->useForceCommands).first;
  if (this->dataPtr->useForceCommands)
  {
```

test/integration/joint_controller_system.cc
```diff
@@ -137,3 +137,80 @@ TEST_F(JointControllerTestFixture, JointVelocityCommand)
   }
 }
 
+/////////////////////////////////////////////////
+// Tests that the JointController accepts joint velocity commands
```
azeey, line 141: Should the comment mention that this is accomplished by using joint force commands?

src/systems/joint_controller/JointController.cc
```diff
@@ -85,6 +95,71 @@ void JointController::Configure(const Entity &_entity,
...
+    double p         =  1;
+    double i         =  0.0;
+    double d         =  0.0;
+    double iMax      =  1;
+    double iMin      = -1;
+    double cmdMax    =  1000;
+    double cmdMin    = -1000;
+    double cmdOffset =  0;
+
+    if (_sdf->HasElement("p_gain"))
```
azeey, line 113: Consider simplifying these by using `sdf::Element::Get` with a default value so that the element name doesn't have to be written twice. I think that's less error prone.

src/systems/joint_controller/JointController.cc
caguero: Good catch! Fixed in bc7e02c.

src/systems/joint_controller/JointController.cc
caguero: Good idea. bc7e02c

test/integration/joint_controller_system.cc
caguero: Done in bc7e02c

src/systems/joint_controller/JointController.cc
```diff
@@ -123,21 +198,58 @@ void JointController::PreUpdate(const ignition::gazebo::UpdateInfo &_info,
   if (_info.paused)
     return;
 
-  std::lock_guard<std::mutex> lock(this->dataPtr->jointVelCmdMutex);
+  // Create joint velocity component if one doesn't exist
```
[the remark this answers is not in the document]
caguero: The only reason I can imagine is if a joint is created dynamically and not at the beginning of the simulation. Not sure if this is even possible in Ignition at this point.

src/systems/joint_controller/JointController.cc
```diff
@@ -85,6 +95,39 @@ void JointController::Configure(const Entity &_entity,
     return;
   }
 
+  if (_sdf->HasElement("use_force_commands") &&
+      _sdf->Get<bool>("use_force_commands"))
+  {
+    this->dataPtr->useForceCommands =
+      _sdf->Get<unsigned int>("use_force_commands");
```
azeey, line 102: Shouldn't this be `bool`?

caguero: Actually, we don't even need to parse it again. c022cb8

src/systems/joint_controller/JointController.cc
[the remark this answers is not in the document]
azeey: Yeah, that would make sense, but I don't think it's possible in Ignition right now. I'm okay with leaving it as is for expediency.

jacobperron:
It looks like this change introduced a test failure in [ignition_gazebo-ci-ign-gazebo2-homebrew-amd64](https://build.osrfoundation.org/job/ignition_gazebo-ci-ign-gazebo2-homebrew-amd64/135)

- [PeerTracker/PeerTrackerStale](https://build.osrfoundation.org/job/ignition_gazebo-ci-ign-gazebo2-homebrew-amd64/135/testReport/(root)/PeerTracker/PeerTrackerStale/)

azeey:
@jacobperron, I think that's a flaky test on macOS. https://github.com/ignitionrobotics/ign-gazebo/issues/8

jacobperron:
@azeey Thanks. The failing line is different then the reported failures in #8, but I've added it to the list.

Issue #112, opened 2020-04-29
Visualize contacts


chapulina:
Gazebo-classic has an option to enable visualization of all contacts in the world as blue spheres with a green line indicating the force's direction and magnitude. These help users debug their model's surface properties and identify contact points while stepping through simulation.

Ignition Gazebo already has a [Contact system](https://github.com/ignitionrobotics/ign-gazebo/tree/master/src/systems/contact) which populates the [ContactSensorData](https://github.com/ignitionrobotics/ign-gazebo/blob/master/include/ignition/gazebo/components/ContactSensorData.hh) component.

[Here](https://github.com/osrf/gazebo/blob/6fd426b3949c4ca73fa126cde68f5cc4a59522eb/gazebo/rendering/ContactVisual.cc)'s Gazebo-classic's `ContactVisual` implementation.

![contacts](https://user-images.githubusercontent.com/5751272/80622143-3d83cb00-89fd-11ea-9cfd-e5e0ec6c6dbd.gif)

mcres:
I'll take this, as this is partially addressed in #229.

My idea is writing a plugin with a checkbox that allows the user to turn on and off the visualization of the contacts. This plugin would be linked to a source file that is continually checking for new [ContactSensorData](https://github.com/ignitionrobotics/ign-gazebo/blob/master/include/ignition/gazebo/components/ContactSensorData.hh) components when the checkbox is on, and using the /marker service to actually visualize the contacts.
However, currently the contact messages only return the position. Work is going on in https://github.com/ignitionrobotics/ign-physics/pull/40.

Is this a good idea?

mabelzhang:
Hey @azeey, since you're the most familiar with the Contact plugin, is there a preference how we want to hook up the contact visualization to it (see suggestions 1-3 above)? Thanks!

azeey:
Option 1 seems reasonable to me. The GUI plugin should be able to get the contact sensor data from the GUI ECM (almost a copy of the server ECM). We might have to use `_contacts->SetData` and `_ecm.SetChanged` in https://github.com/ignitionrobotics/ign-gazebo/blob/9ab1cc3f17173c57028f3c460121a36cbdb0c8d7/src/systems/physics/Physics.cc#L1722

mcres:
[...]

PR #234, opened 2020-07-08
Visualize ContactSensorData

mcres:
Fixes issue #112 

This requires #272 

It currently visualizes the position of the contacts returned by the Contact sensors.
It also allows the user to change the size of the markers (radius and length) through a couple of spin boxes.

In order to show the contacts of all of the `components::Collision`, it appears it's not enough to create a `components::ContactSensorData` as stated [here](https://github.com/ignitionrobotics/ign-gazebo/issues/112#issuecomment-654998572), because this is a GUI system.

issue_112


mcres:
Work is being done on visualizing contacts for `<collision>` elements without Contact sensors.

src/gui/plugins/visualize_contacts/VisualizeContacts.qml
```diff
@@ -0,0 +1,98 @@
...
+    Layout.columnSpan: 1
+    Layout.rowSpan: 15
+    Layout.fillWidth: true
+  }
+
+  CheckBox {
+    Layout.alignment: Qt.AlignHCenter
+    id: visualizeContacts
+    Layout.columnSpan: 4
+    text: qsTr("Show/Hide Contacts")
```
mabelzhang, line 42: I would suggest simply saying `Show Contacts` as opposed to `Show/Hide`. Right now checking the box is show, unchecking is hide. But before I did anything, it was confusing, "what would happen if I check it, would it show or hide?"

src/systems/physics/Physics.cc
```diff
@@ -1942,6 +1978,14 @@ void PhysicsPrivate::UpdateCollisions(EntityComponentManager &_ecm)
...
         }
+
+        auto state = _contacts->SetData(contactsComp,
+          this->contactsEql) ?
+          ComponentState::OneTimeChange :
+          ComponentState::NoChange;
+        _ecm.SetChanged(
+          _collEntity1, components::ContactSensorData::typeId, state);
+
         *_contacts = components::ContactSensorData(contactsComp);
```
azeey, line 1989: Don't think you need this if you already used `SetData` above.

src/gui/plugins/visualize_contacts/VisualizeContacts.hh
```diff
@@ -0,0 +1,76 @@
...
+    /// \brief Callback when checkbox state is changed
+    /// \param[in] _checked indicates show or hide contacts
+    public slots: void OnVisualize(bool _checked);
+
+    /// \brief Update the radius of the contact
+    /// \param[in] _radius new radius of the contact
+    public slots: void UpdateRadius(double _radius);
+
+    /// \brief Update the length of the contact
+    /// \param[in] _length new length of the contact
```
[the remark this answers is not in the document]
mcres: I left a [todo comment](https://github.com/ignitionrobotics/ign-gazebo/blob/1817a9ad193e2076bdb823d32da0151d6ef366d2/src/gui/plugins/visualize_contacts/VisualizeContacts.cc#L241) for that, but changing the variable name and explaining it is definitely a good idea.

mcres:
>  The contacts show up, but the text boxes don't show up correctly for me. I only see 1 digit. So initially I just see 0 and 0 for both radius and length. if I arrow to the left, I'd eventually see the decimal point 0., but I can never see the entire number. I dragged the panel to the left to expand it, but the text boxes didn't expand. Maybe they can be set to a longer length.

Hmm that's weird. Does that happen to you if you load the Grid Config plugin, which also has spin boxes?

src/gui/plugins/visualize_contacts/VisualizeContacts.qml
mabelzhang: I think the user would expect them to disappear when it's unchecked. So deleting seems more reasonable.

mabelzhang:
Hmm yeah actually it does in Grid Config plugin as well:
Screenshot from 2020-07-22 04-49-28

Looking at this, maybe you could center everything in the Visualize contacts panel too for consistency.

I think the spin boxes being too short is a GUI problem. It should probably be fixed elsewhere, but it should be fixed. I'm on a 4k laptop screen, and most windows and menus are automatically scaled up 2x. That may or may not be related to this problem. It's a pretty common laptop so others will run into this as well.

chapulina:
>  the spin boxes being too short is a GUI problem

It's possible to set the width of `IgnSpinBox` directly inside `VisualizeContacts.qml`. I'd recommend making `GridLayout` be as large as it can be with `anchors.fill: parent`, then making each spin box take as much width as possible with `Layout.fillWidth: true`. Let me know how it goes!

mcres:
In order to visualize the contacts for objects without contact sensors, we've encountered some problems and solved them the following way:

1. First of all, we had to 'enable' the `<collision>`. In order to create the `ContactSensorData` inside the `Collision` component, the ECM provided in the `GuiSystem::Update` method is not connected to the physics system. For that, I've created a `/enable_collision` and `/disable_collision` [services](https://github.com/mcres/ign-gazebo/blob/issue_112/src/systems/user_commands/UserCommands.cc), which enable or disable an individual `collision` component. The services checks that there's not a contact sensor connected to the collision. Probably some refactoring can be done there, please let me know if that's the case.
2. While running the [example world](https://github.com/mcres/ign-gazebo/blob/issue_112/examples/worlds/visualize_contacts.sdf), I've realized that the contacts remained when two objects were no longer touching (for instance, when removing the cylinder from the top of the car). The reason is that the `ContactSensorData` field in the `SerializedStepMap` message was empty after moving the cylinder, which caused the `EntityComponentManager::SetState` not to correctly update the info.
As a workaround, we've added some dummy data to emtpy `ContactSensorData` components in the [physics system
](https://github.com/mcres/ign-gazebo/blob/issue_112/src/systems/physics/Physics.cc#L1963), which are skipped when visualizing the contacts.
3. **[Unresolved]** There's a similar problem as no. 2 when disabling a collision: it seems that when [removing](https://github.com/mcres/ign-gazebo/blob/issue_112/src/systems/user_commands/UserCommands.cc#L930) the `ContactSensorData`  in the service, the component stays stored somewhere, but I can't find where. I've checked the physics system and the `SerializedStepMap` message, and they are not updating nor showing the component, respectively. However, the ECM is still giving me that value. I've tried changing the contact value just before being removed, but that also doesn't work. @azeey any ideas on where that can be stored? 

Also, I've intentionally removed come of the contact sensors from the example world for testing the plugin.

mcres:
> It's possible to set the width of IgnSpinBox directly inside VisualizeContacts.qml. I'd recommend making GridLayout be as large as it can be with anchors.fill: parent, then making each spin box take as much width as possible with Layout.fillWidth: true. Let me know how it goes!

@chapulina I've changed that and @mabelzhang should be able to see the spin boxes. The problem is the extra space between the different elements:

visualize_gui

I guess that's because of the `anchors.fill: parent`, I wonder if I could limit the expanding height somehow. The `.qml` file is very similar to that of the `Grid Config` plugin, which doesn't take up all of the space available. I've tried the `Layout.minimumHeight` property without results.

chapulina:
> The problem is the extra space between the different elements:

The trick to press them together is to add an invisible "spacer" item at the bottom of the layout like this:

https://github.com/ignitionrobotics/ign-gazebo/blob/aea1accff6d8daee0358ae9025e2a9797d292844/src/gui/plugins/grid_config/GridConfig.qml#L319-L323

With `fillHeight`, you're telling the layout to give that item as much vertical space as possible.

mabelzhang:
> @mabelzhang should be able to see the spin boxes
Confirming text boxes are automatically expanding for me now after dragging the panel to the left.

src/systems/user_commands/UserCommands.cc
```diff
@@ -683,6 +798,175 @@ bool PoseCommand::Execute()
...
+}
+
+//////////////////////////////////////////////////
+bool EnableCollisionCommand::HasContactSensor(const Entity _collision)
+{
+  auto *ecm = this->iface->ecm;
+
+  auto *linkEntity = ecm->Component<components::ParentEntity>(_collision);
+  auto allLinkSensors =
+    ecm->EntitiesByComponents(components::Sensor(),
```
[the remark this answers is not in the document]
mcres: Hmm I get a vector of size `0` when I try either
```
ecm->EntitiesByComponents(
      components::Sensor(),
      components::ContactSensor(),
      components::ParentEntity(*linkEntity));
```
or
```
ecm->EntitiesByComponents(
      components::ContactSensor(),
      components::ParentEntity(*linkEntity));
```

mcres:
https://github.com/ignitionrobotics/ign-gazebo/pull/234#issuecomment-664414267 (no. 3) is currently being fixed at #272

src/systems/user_commands/UserCommands.cc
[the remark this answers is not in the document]
azeey: My bad. `components::ContactSensor` is a component that holds an `sdf::ElementPtr`, so the comparison will fail if the default constructed `components::ContactSensor()` is used. Please ignore my suggestion.

src/gui/plugins/CMakeLists.txt
```diff
@@ -90,4 +90,5 @@ add_subdirectory(scene3d)
 add_subdirectory(shapes)
 add_subdirectory(transform_control)
 add_subdirectory(video_recorder)
+add_subdirectory(visualize_contacts)
```
chapulina, line 93: Please keep the alphabetical order

src/gui/plugins/visualize_contacts/VisualizeContacts.cc
```diff
@@ -0,0 +1,393 @@
...
+#include <ignition/gui/MainWindow.hh>
+
+#include "ignition/gazebo/components/Name.hh"
+#include "ignition/gazebo/components/World.hh"
+#include "ignition/gazebo/EntityComponentManager.hh"
+#include "ignition/gazebo/gui/GuiEvents.hh"
+#include "ignition/gazebo/rendering/RenderUtil.hh"
+#include "ignition/gazebo/components/Collision.hh"
+#include "ignition/gazebo/components/ContactSensor.hh"
+#include "ignition/gazebo/components/ContactSensorData.hh"
```
chapulina, line 53: Also be sure to alphabetize all the includes within their groups

src/gui/plugins/visualize_contacts/VisualizeContacts.cc
```diff
@@ -0,0 +1,393 @@
...
+VisualizeContacts::~VisualizeContacts() = default;
+
+/////////////////////////////////////////////////
+void VisualizeContacts::LoadConfig(const tinyxml2::XMLElement *)
+{
+  if (this->title.empty())
+    this->title = "Visualize contacts";
+
+  ignition::gui::App()->findChild<
+    ignition::gui::MainWindow *>()->installEventFilter(this);
```
chapulina, line 134: I think this isn't necessary, because there's no `eventFilter` function to receive the filtered events. This is only needed by plugins that receive events.

src/gui/plugins/visualize_contacts/VisualizeContacts.cc
```diff
@@ -0,0 +1,393 @@
...
+            return true;
+
+          // Set marker id, poses and request service
+          this->dataPtr->positionMarkerMsg.set_id(markerID);
+          ignition::msgs::Set(this->dataPtr->positionMarkerMsg.mutable_pose(),
+            ignition::math::Pose3d(contact.position(i).x(),
+              contact.position(i).y(), contact.position(i).z(),
+              0, 0, 0));
+
+          // Placeholder for the force value (see todo comment above)
```
chapulina, line 292: I'm a bit concerned about displaying force data that's not real. I think it's good that we have everything in place to display the force once that's available, but the end user doesn't know that this isn't the real data, so it may cause confusion. How about commenting this out for now?

src/gui/plugins/visualize_contacts/VisualizeContacts.qml
```diff
@@ -0,0 +1,119 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+*/
+import QtQuick 2.9
+import QtQuick.Controls 2.1
+import QtQuick.Dialogs 1.0
```
chapulina, line 19: I think `Dialogs` is not needed

src/systems/physics/Physics.cc
```diff
@@ -2003,31 +2039,49 @@ void PhysicsPrivate::UpdateCollisions(EntityComponentManager &_ecm)
...
-        const auto &contactMap = entityContactMap[_collEntity1];
-
-        msgs::Contacts contactsComp;
 
-        for (const auto &[collEntity2, contactData] : contactMap)
-        {
+          // If there's an empty ContactSensorData, it won't be updated in the
+          // GUI ECM and there could be remaining contacts which shouldn't
+          // exist anymore. We insert dummy data in the empty ContactSensorData
+          // in order to avoid this problem.
```
chapulina, line 2051: Is this still needed now that the GUI knows of removed components? (#272)

src/systems/user_commands/UserCommands.cc
```diff
@@ -683,6 +798,175 @@ bool PoseCommand::Execute()
...
+  this->iface->ecm->
+    RemoveComponent(entityMsg->id(), components::ContactSensorData::typeId);
+
+  igndbg << "Disabled collision [" << entityMsg->id() << "]" << std::endl;
+
+  return true;
+}
+
+//////////////////////////////////////////////////
+bool DisableCollisionCommand::HasContactSensor(const Entity _collision)
```
chapulina, line 938: The `DisableCollisionCommand::HasContactSensor` and `EnableCollisionCommand::HasContactSensor` functions look duplicate. It would be nice to reuse the same function so that if we change one of them, we change both at the same time. Maybe it could be a function in `UserCommandsInterface`?

chapulina (commented):
Works for me! I just have some minor comments.

src/gui/plugins/visualize_contacts/VisualizeContacts.cc
mcres: I don't understand, do you mean commenting out the `force` declaration and using a hardcoded value instead?

src/systems/user_commands/UserCommands.cc
mcres: Yep, they are duplicate. I just wasn't sure where to put that, thanks for the suggestion ;)

src/gui/plugins/visualize_contacts/VisualizeContacts.cc
chapulina: Maybe I didn't understand the implementation then. I thought it was already hardcoded? Now it's always 1 scaled to the value the user inputs. It's not the real force, right? Also, the normal direction also doesn't seem to be correct?

mcres: Yep, all questions are correct. I don't get what's the suggestion for avoiding confusion for the user.

src/systems/physics/Physics.cc
mcres: Hmm for some reason when I delete that code, the forces are only displayed once and then they don't get updated.

Also, when I try to disable a collision by requesting the `disable_collision` service, forces keep showing up most of the times. I'll take a look at the `SerializedState` msg to see what's going on.

src/gui/plugins/visualize_contacts/VisualizeContacts.cc
chapulina: Oh gotcha, so I'd remove the green cylinders completely. We can leave all the code related to them commented out.

chapulina:
I'm going to remove the `beta` label from this PR, so it won't be included on Dome's initial release, but it can come on a minor release after that.

src/gui/plugins/visualize_contacts/VisualizeContacts.cc
mcres: Agree.

mcres:
Addressed the feedback.

[Point no. 3](https://github.com/ignitionrobotics/ign-gazebo/pull/234#issuecomment-664414267) doesn't work every time, even after merging #272. I think the following is happening:

1. In the example world, I call the `disable_collision` service for, say, the caster wheel.
2. The collision entity gets inserted into the `removedComponents` [unordered map](https://github.com/ignitionrobotics/ign-gazebo/blob/master/src/EntityComponentManager.cc#L116).
3. Before the ECM deserialize its state in `SetRemovedComponentsMsgs`, `ClearRemovedComponents` is called and the GUI update is not successful.

mabelzhang:
Sounds to me like most things in this ticket are resolved, other than bullet 3 linked in the previous comment.

@azeey Is this something acceptable that we can create an issue for the future and merge this PR, or does that bullet need to be fixed before this goes in?

azeey:
> Sounds to me like most things in this ticket are resolved, other than bullet 3 linked in the previous comment.
> 
> @azeey Is this something acceptable that we can create an issue for the future and merge this PR, or does that bullet need to be fixed before this goes in?

Yeah, I think we can create an issue for it and merge this PR. @mcres, I haven't been able to reproduce the issue on the `visualize_contacts.sdf` example, do you have an example that shows this behavior?

mcres:
@azeey This is the sequence:

1. I `ign gazebo -v 4 examples/worlds/visualize_contacts.sdf -r` and manually check the contact visualization in the plugin
2. In another terminal, I try to turn off the visualization of a specific collision --- in this case the ground floor and the right wheel, id 7 and 23, respectively--- by disabling it like this: `ign service -s /world/visualize_contacts/disable_collision --reqtype ignition.msgs.Entity --reptype ignition.msgs.Boolean --timeout 1000 --req 'type: 5; id: 7'` and `ign service -s /world/visualize_contacts/disable_collision --reqtype ignition.msgs.Entity --reptype ignition.msgs.Boolean --timeout 1000 --req 'type: 5; id: 23'`

I used to encounter 2 issues:
1. The contact just kept showing (which I'm not able to reproduce anymore, I guess that's good news)
2. The last contact shown immediately before requesting the services keep showing, but there are no new contacts (this still happens, though very few times). In the image below, the car was moving when I turned off the right wheel and floor contact. The weird blue sphere is the position of the right wheel when I turned the contact off and its position won't change.

issue


If this isn't reproducible, merging and opening an issue if it happens again makes sense to me!

azeey:
I haven't been able to reproduce the two issues you mentioned. But I noticed what I think is a performance issue with Markers. When the update period is set to a shorter amount of time, the gui becomes jumpy after a little while

Peek 2020-11-02 11-41
 
This is what I think is going on. When an existing marker is modified, a new material is created for it. This material is then assigned to the existing marker visual. 
https://github.com/ignitionrobotics/ign-gazebo/blob/1030417c25c8d282244050b2c068053c32e642ba/src/rendering/MarkerManager.cc#L329-L332

This causes the existing material for the marker to get removed. Over time, this becomes really expensive for some reason. My guess is Ogre has an ever increasing list of resource names that has to be traversed even if the actual active list of materials is kept small. Profiling it showed a hotspot in `ignition::rendering::v4::OgreMarker::SetMaterial(std::shared_ptr<ignition::rendering::v4::Material>, bool)` which eventually calls `Ogre::ResourceManager::remove(std::string const&)`

mcres:
I see, I didn't notice about that.
Since the problem is in Ogre, is it possible to merge and create another issue? I think the PR is getting pretty big.

azeey:
> Since the problem is in Ogre, is it possible to merge and create another issue? I think the PR is getting pretty big.

That sounds good. Do you mind resolving the merge conflict?

azeey (approved):

azeey (changes requested):
Looks like tests are failing https://build.osrfoundation.org/job/ignition_gazebo-ci-pr_any-ubuntu_auto-amd64/4598/testReport/junit/(root)/ContactSystemTest/MultipleCollisionsAsContactSensors/:
```
/var/lib/jenkins/workspace/ignition_gazebo-ci-pr_any-ubuntu_auto-amd64/ign-gazebo/test/integration/contact_system.cc:118
Value of: contactMsgs.size()
  Actual: 10
Expected: 0u
Which is: 0
```

mcres:
Hmm I have no idea where this could come from and to be honest for me it would probably take more time than I have.
Is it okay if you take a look into this @azeey?

src/EntityComponentManager.cc
```diff
@@ -1164,13 +1164,6 @@ void EntityComponentManager::SetState(
     {
       const auto &compMsg = compIter.second;
 
-      // Skip if component not set. Note that this will also skip components
```
azeey, line 1167: @chapulina I know we've said we should keep this POC in the past, but I can't remember why.  Do you know what the problem was? Removing it was the easiest way I found to fix the test failures because the workaround of this POC was to send dummy messages that caused expectations to fail in other tests.

chapulina: My guess is that we don't want to reset the value of components that can't be serialized. Those components would send an empty value through the wire. We don't run into this much because most of the default components can be serialized.

azeey: I see. Would it be possible to not send components that can't be serialized in the first place instead of sending an empty value?

chapulina: It's valuable to know that the component is present, even if the value isn't serializable. We get around that for `NoData` components by adding dummy data to the serialized message, which isn't great:

https://github.com/ignitionrobotics/ign-gazebo/blob/a5abacfa74db6e8db0218bd32c55980e15cf249b/include/ignition/gazebo/components/Component.hh#L211-L218

azeey: Can you elaborate more? If the "data" of a component is its mere presence, as in `NoData`, I think sending an empty message makes sense and doesn't break anything. But if the "data" of the component is more than its presence, I don't see how sending an empty message is useful. i.e, What's the use case of knowing a component exists if we can't see its data (with the exception of `NoData` components)?

chapulina: There are 2 use cases for the state, so let's look at them separately.

## GUI

Imagine you have a `CustomComponent` with `UnserializableData`. We can't send the data from the server to the client, but we can at least tell the client that an entity has that component. `CustomComponent` will be listed on the Component Inspector, without data.

That's more useful than not displaying it at all, right? Helps debugging to a certain extent. Right now, the assumption for any GUI systems is that they have the complete list of entities and components, even if they may not be able to peek inside some of them.

GUI systems aren't doing much component setting through, so I think we shouldn't expect the situation I mentioned above ("we don't want to reset the value of components that can't be serialized.") to be an important issue

## Distributed sim

In this case, the presence of some component without can be helpful to let systems know that it needs to be populated.

I think the resetting issue could come into play during distributed simulation though. For example:

* Robot A is being simulated by secondary 1. It has `CustomComponent`.
* Parts of Robot A's state are sent to secondary 2.
* Some system in secondary 2 sets the value of `CustomComponent` for Robot A
* Syncing the state from secondary 1 to 2 sends an empty `CustomComponent`, resetting its data by mistake

---

This is very hypothetical, but do you see where I'm going? In summary, it would be nice to be able to tell default data from no data.

azeey: For the first use case, I think the change I made still allows sending components without any data. The change I'm proposing removes
https://github.com/ignitionrobotics/ign-gazebo/blob/2f7f33d4fa34657badec81558d5953fe7643769f/src/EntityComponentManager.cc#L1160-L1163

which was actually skipping components without data.

For the second use case, yes, the `CustomComponent` will end up being reset in secondary 1 instead of having the value set by secondary 2. However, I think this would be an issue without the change I'm proposing as well. Without this change, `CustomComponent` will keep its old value even though it's being updated by secondary 2.

So if there is a component that can't be serialized, but both sides (server/client or 2 secondaries) want to set the value, we'll end up in an error condition.

But I feel like we're giving up a functionality, i.e, the ability to set arbitrary values, for an alternative that's still ends up in an error condition. 

> In summary, it would be nice to be able to tell default data from no data.

What if we used the [iostate](https://en.cppreference.com/w/cpp/io/ios_base/iostate) flags to indicate this. For example, we can set `failbit` in the `DefaultSerializer` when a component cannot be serialized. It's still kind of a hack, but it would allow us to tell the difference between a component that cannot be serialized and a component whose serialization returned an empty string.

chapulina: >  we can set failbit in the DefaultSerializer when a component cannot be serialized

That sounds like a good idea. I wonder if protobuf is happy to serialize that.

> Without this change, CustomComponent will keep its old value even though it's being updated by secondary 2.

I think the difference is that secondary 2 will know that `CustomComponent` has changed, and that it doesn't know what the changes are. So at least it could have a fallback for that case.

> But I feel like we're giving up a functionality, i.e, the ability to set arbitrary values

Yeah you're right, the current situation is still less than ideal.

src/systems/physics/Physics.cc
```diff
@@ -2246,31 +2282,43 @@ void PhysicsPrivate::UpdateCollisions(EntityComponentManager &_ecm)
...
+            _collEntity1, components::ContactSensorData::typeId, state);
           return true;
         }
-
-        const auto &contactMap = entityContactMap[_collEntity1];
-
-        msgs::Contacts contactsComp;
-
-        for (const auto &[collEntity2, contactData] : contactMap)
+        else
```
chapulina, line 2297: I don't think this block needs to be inside an `else`, because the `if` above is returning. We could save some indentation by removing the `else`.

Alternatively, you could remove the block that's inside the `if` and let the `SetData` below set the empty `contactsComp`. Makes sense?

src/systems/user_commands/UserCommands.cc
```diff
@@ -248,6 +332,24 @@ void UserCommands::Configure(const Entity &_entity,
       &UserCommandsPrivate::PoseService, this->dataPtr.get());
 
   ignmsg << "Pose service on [" << poseService << "]" << std::endl;
+
+  // Enable collision service
+  std::string enableCollisionService{
+    "/world/" + worldName + "/enable_collision"};
```
chapulina, line 338: This is something new since this PR was opened. We should pass these service names through the `AsValidTopic` function like was done on #522.

src/EntityComponentManager.cc
chapulina: Ok, so, after some thought, let's go with the current implementation and we can add the `failbit` approach if we have a concrete use case for it.

src/gui/plugins/visualize_contacts/VisualizeContacts.cc
```diff
@@ -0,0 +1,376 @@
...
+
+  {
+    std::lock_guard<std::mutex> lock(this->dataPtr->serviceMutex);
+    if (this->dataPtr->checkboxPrevState && !this->dataPtr->checkboxState)
+    {
+      // Remove the markers
+      this->dataPtr->positionMarkerMsg.set_action(
+        ignition::msgs::Marker::DELETE_ALL);
+      this->dataPtr->forceMarkerMsg.set_action(
+        ignition::msgs::Marker::DELETE_ALL);
```
chapulina, line 219: When I uncheck the "Show Contacts" box, I get this warning:

`[GUI] [Wrn] [MarkerManager.cc:555] Unable to delete all markers in namespace[forces], namespace can't be found.`

Since we're never creating the forces markers,  I recommend removing all the logic around them. The reason is that it's hard to maintain commented out code. Leaving it out makes the plugin simpler and easier to maintain. Whenever we have support for contact normals and forces and want to show them, we can come back to this PR's history and revive the code. What do you think?

chapulina (changes requested):
The contacts between the cylinder and the vehicle seem out of place:

cylinder_contacts

I think the error is coming from DART though, because the GUI isn't adding any offset.

azeey:
> The contacts between the cylinder and the vehicle seem out of place:

Looks like the issue is in ODE, the collision detector we are using for DART. Here's the same world in Gazebo classic.

image

chapulina (approved):
I addressed my final comments in c28fc2053deeea0c179171a90897e2e32d1f9936

This is good to go with happy CI :rocket:

PR #242, opened 2020-07-14
Changes to support physics plugin loading

mjcarroll:
Requires: ignitionrobotics/ign-bazel#21
Requires: ignitionrobotics/ign-physics#81
Requires: ignitionrobotics/ign-sensors#32

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -405,11 +405,10 @@ cc_library(
 )
 
 [
-
 cc_binary(
-    name = "libignition-gazebo-%s-system.so" % system.replace("_", "-"),
+    name = "libignition-gazebo-%s-system.so" % (system.replace("_", "-")),
     srcs = glob([
-            "src/systems/%s/*.cc" % system,
+          "src/systems/%s/*.cc" % system,
           "src/systems/%s/*.hh" % system,
       ],
       allow_empty = False,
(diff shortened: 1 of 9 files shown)


JShep1:
I think the resources under ign_gazebo/src/gui/resources also need to be built, indicated by error `[GUI] [Err] [GazeboGui.cc:151] Failed to instantiate custom drawer, drawer will be empty`

JShep1:
I'm also still getting issues upstream with finding certain qml resources, running `bazel run //ign_tools:ign -- launch /path/to/gazebo_plugins.ign` with branches https://github.com/ignitionrobotics/ign-tools/pull/21 and https://github.com/ignitionrobotics/ign-launch/pull/35 results in two .qml resources unable to be found/instantiated: `qrc:/GzScene3D/GzScene3D.qml` and `qrc:/EntityTree/EntityTree.qml`, the other resources, oddly enough, are found, like TransformControl.  I printed out the available Qt resources at the point right before they are loaded, and I can see
```
:/GzScene3D
:/GzScene3D/GzScene3D.qml
:/EntityTree
:/EntityTree/EntityTree.qml
```
so it's odd that they aren't able to be found.

JShep1:
Can you resolve conflicts?

Issue #159, opened 2020-05-29
Add buoyancy plugin


chapulina:
@nkoenig started a port that was almost merged on BitBucket, but didn't get in in time for the GitHub migration. We should reopen that PR soon so it's not left behind.

https://osrf-migration.github.io/ignition-gh-pages/#!/ignitionrobotics/ign-gazebo/pull-requests/506/page/1

nkoenig:
Ported in pull request #252

PR #252, opened 2020-07-16
Buoyancy

nkoenig:
Redo of the buoyancy PR. Original is [here](https://osrf-migration.github.io/ignition-gh-pages/#!/ignitionrobotics/ign-gazebo/pull-requests/506/page/1).

This will resolve issue #159.


nkoenig:
> LGTM. I think the only unaddressed comment from the original PR is this one from @azeey :
> 
> @azeey
> 
> "It would be good to document the reference frame of the position."
> 
> @nkoenig
> 
> "What language would you recommend? There is no frame attached to this component, so a system can use what ever frame they desire."
> 
> @azeey
> 
> "That's true, but I think that would lead to confusion. I would recommend maybe saying that the position is relative to the pose of the entity this component is attached to. If we need to specify it in world frame, we can create a CenterOfVolumeInWorld component like we did for Pose and WorldPose. "
> 
> Also, I'd appreciate it if you could also take another look, @azeey .

Add position documentation in 9eda1a7946a8b6cb0c24f44e28cd4069eee68568

PR #253, opened 2020-07-17
Some UI tweaks to Resource Spawner

chapulina:
Suggestions to #173.

I could keep tweaking more details, but this is already a large first pass. Some things I can think of:

* Start with the 1st folder selected
* Use folder icons
* Center the grid items when there's extra space on the right

resource_spawner_2


chapulina:
> the right side containing the models is squished until the user manually expands the sidebar, should we stick with this or do something like expanding the minimum size of the sidebar?

I tweaked the minimum sizes a bit so that it's not too big but also not too squished. Let me know what you think: 50d7520bae770484758ca4ee1cba52b98477165d

chapulina:
Added @ColeOSRF's folder icons:

folders

PR #260, opened 2020-07-22
Deactivate PerformerDetector if its parent model gets removed

azeey:
The `PerformerDetector` system currently assumes that its parent model always exists, but when used with levels, the parent model may get unloaded during simulation causing a segfault. This change deactivates a `PerformerDetector` if the model it is associated with gets removed.


nkoenig:
@azeey ,  can you get this merged as soon as possible so that I can role out an update to cloudsim?

nkoenig:
This looks good. I'll merge and start the release process.

Issue #239, opened 2020-07-14
Support names with spaces.


luca-della-vedova:
I have been playing with some models that have spaces in their names, such as [Pine Tree](https://app.ignitionrobotics.org/OpenRobotics/fuel/models/Pine%20Tree) and noticed that when I spawn multiple models in the same world, such as:
```
    <include>
      <name>Pine Tree</name>
      <uri>model://Pine Tree</uri>
      <pose>78.16778723671018 -35.70987539227364 0 0 0 1.5708</pose>
      <static>True</static>
    </include>
    <include>
      <name>Pine Tree_2</name>
      <uri>model://Pine Tree</uri>
      <pose>74.0254778248307 -32.28629661564661 0 0 0 1.5708</pose>
      <static>True</static>
    </include>
```
I get the following error (and the model is spawned black):
```
[ign-8] [GUI] [Err] [SceneManager.cc:159] Visual: [Pine] already exists
```
I tracked down the problem being in the Name component missing the second word, and I believe this is due to the way we are doing serialization and deserialization in the Component template, specifically if I understand correctly we use the operator << for serialization and the operator >> for deserialization.
However, this doesn't work greatly for strings since, in the case of strings with spaces, the >> operator will only operate until the space (hence both models are seen as having the name "Pine").
By contrast this world file works OK for gazebo classic, do you think it should be fixed or should we just impose a "No model names with spaces" policy?

chapulina:
> this world file works OK for gazebo classic

If I recall correctly, classic's support for names with spaces is spotty. I could find a couple issues related to using spaces in resources:

* https://github.com/osrf/gazebo/issues/2056
* https://github.com/osrf/gazebo/issues/2369

> this is due to the way we are doing serialization and deserialization in the Component template

Thank you for tracking down the source of the problem. Regardless of whether we want to support names with spaces, this sounds like a bug on string components. We can create a custom serializer for `std::string` that doesn't use operators. I'll work on that :ok_hand: 

> should we just impose a "No model names with spaces" policy?

Once the serialization issue is solved, I imagine we may run into others, such as [topic names](https://ignitionrobotics.org/api/transport/9.0/nodestopics.html).

I don't think `components::Name` should be opinionated about this kind of detail though. There will be use cases where users would like to use pretty human-readable names for their entities and it won't break anything in their setup. So I think this could be handled at each system's level. For example, systems that use entity names for topics could convert `Pine Tree` to `pine_tree` and document that.

[...]

Issue #85, opened 2020-04-21
Simplify determination of a sensor's topic name 


gerkey:
In the course of migrating a plugin from Gazebo to Ignition (#80), I needed to subscribe to data published via `ign-transport` by an IMU sensor. To determine the fully qualified topic name to subscribe to, I ended up jumping through [quite some hoops](https://github.com/gerkey/ardupilot_gazebo/blob/ignition/src/ArduPilotPlugin.cc#L841-L884) inside `PreUpdate()`:

```
// This lookup is done in PreUpdate() because in Configure() it's not possible to get the fully qualified topic name we want
  if(!this->dataPtr->imuInitialized)
  {
    // Set unconditionally because we're only going to try this once.
    this->dataPtr->imuInitialized = true;
    std::string imuTopicName;
    _ecm.Each<ignition::gazebo::components::Imu, ignition::gazebo::components::Name>(
            [&](const ignition::gazebo::Entity &_imu_entity,
                const ignition::gazebo::components::Imu * /*_imu*/,
                const ignition::gazebo::components::Name *_name)->bool
        {
          if(_name->Data() == this->dataPtr->imuName)
          {
            // The parent of the imu is imu_link
            ignition::gazebo::Entity parent = _ecm.ParentEntity(_imu_entity);
	    this->dataPtr->modelLink = parent;
            if(parent != ignition::gazebo::kNullEntity)
            {
              // The grandparent of the imu is the quad itself, which is where this plugin is attached
              ignition::gazebo::Entity gparent = _ecm.ParentEntity(parent);
              if(gparent != ignition::gazebo::kNullEntity)
              {
                ignition::gazebo::Model gparent_model(gparent);
                if(gparent_model.Name(_ecm) == this->dataPtr->modelName)
                {
                  imuTopicName = ignition::gazebo::scopedName(_imu_entity, _ecm) + "/imu";
                  igndbg << "Computed IMU topic to be: " << imuTopicName << std::endl;
                }
              }
            }
          }
          return true;
        });
    
    if(imuTopicName.empty())
    {
      ignerr << "[" << this->dataPtr->modelName << "] "
            << "imu_sensor [" << this->dataPtr->imuName
            << "] not found, abort ArduPilot plugin." << "\n";
      return;
    }

    this->dataPtr->node.Subscribe(imuTopicName, &ignition::gazebo::systems::ArduPilotPluginPrivate::imuCb, this->dataPtr.get());
  }
```

[...]

PR #266, opened 2020-07-30
Sensor topics available through components and GUI

chapulina:
Addresses https://github.com/ignitionrobotics/ign-gazebo/issues/85 
Requires https://github.com/ignitionrobotics/ign-sensors/pull/33/ and https://github.com/ignitionrobotics/ign-transport/pull/153

---

While working on #239, I thought it would be good to also address #85. This adds a `SensorTopic` component that can be checked from any system and is also displayed on the GUI. It uses the new API on `ign-sensors` to get the real topic that is being published, after all inputs and modifications have been applied to it.

:warning: This PR doesn't add topic components for rendering sensors because the API on this branch doesn't make it easy to access the ECM and the sensor pointer at the same time. The [Sensors::CreateSensor](https://github.com/ignitionrobotics/ign-gazebo/blob/aea1accff6d8daee0358ae9025e2a9797d292844/src/systems/sensors/Sensors.cc#L466-L467) function on Dome takes an `Entity` ID which will make it easier to add topics for those sensors, so my current suggestion is to only support topic components for rendering sensors from Dome. Let me know what you think, @iche033 .


test/integration/air_pressure_system.cc
```diff
@@ -53,6 +58,43 @@ TEST_F(AirPressureTest, AirPressure)
...
   EXPECT_FALSE(*server.Running(0));
 
+  const std::string sensorName = "air_pressure_sensor";
+
+  auto topic = "world/air_pressure_sensor/model/air_pressure_model/link/link/"
+      "sensor/air_pressure_sensor/air_pressure";
+
+  bool updateChecked{false};
+
+  // Create a system that records altimeter data
```
iche033, line 68: `...that checks sensor topic`?

test/integration/logical_camera_system.cc
```diff
@@ -72,17 +73,53 @@ TEST_F(LogicalCameraTest, LogicalCameraBox)
...
   EXPECT_FALSE(*server.Running(0));
 
+  const std::string sensorName = "logical_camera";
+
+  auto topic = "world/logical_camera_sensor/model/logical_camera/link/"
+      "logical_camera_link/sensor/logical_camera/logical_camera";
+
+  bool updateChecked{false};
+
+  // Create a system that records altimeter data
```
iche033, line 83: update comment here

iche033:
> The Sensors::CreateSensor function on Dome takes an Entity ID which will make it easier to add topics for those sensors, so my current suggestion is to only support topic components for rendering sensors from Dome.

sounds good to me.

PR #268, opened 2020-08-04
2 ➡️ 3

chapulina:


chapulina:
@azeey , could the diff drive failure on Focal be due to the friction on that DART version? It looks like the vehicle is going forward but not turning. 

I don't have a good guess about what could be causing the breadcrumb crash yet.

chapulina:
Broke part of this PR into #296

nkoenig:
Is this PR still a draft?

chapulina:
> Is this PR still a draft?

It needs updating, I'll get to it soon

chapulina:
#317

azeey:
> could the diff drive failure on Focal be due to the friction on that DART version? It looks like the vehicle is going forward but not turning.

I believe the upstream version of DART does have the same friction characteristics. The difference is our fork has slip and in the currently released implementation, each object has a very tiny amount of slip. Maybe that made the difference. I think setting smaller friction coefficients on the lateral direction of the wheels in `test/worlds/diff_drive.sdf` would fix the test failures for both DART versions. They currently use the default value, which is 1.0

PR #272, opened 2020-08-05
Sync components removal

mcres:
It sets the `SerializedStateMap` and `SerializedState` messages when a component is removed after `EntityComponentManager::RemoveComponent` is called. This allows a more precise synchronization between the server and GUI `EntityComponentManager`.

This PR is made in order to solve an issue in #234 already described [here](https://github.com/ignitionrobotics/ign-gazebo/pull/234#issuecomment-664414267) (no. 3).

A couple of comments about the PR:

- [This](https://github.com/ignitionrobotics/ign-gazebo/compare/master...mcres:sync_components_removal?expand=1#diff-f62ca607b3aece56b7217a0fb98c1253R864) TODO comment wasn't removed, given that the PR doesn't do any queuing.
- I think the PR should have 2 tests: one that checks the `SerializedStateMap` message after calling `RemoveComponent`, and another one that checks that a component is actually removed when its `remove` flag in the message is set to true. I've implemented the first one and the second should be already done [here](https://github.com/mcres/ign-gazebo/blob/61841ca5548145e364c8aeb832950289ac6b121c/src/EntityComponentManager_TEST.cc#L1628-L1693).

### DIFF
--- a/include/ignition/gazebo/EntityComponentManager.hh
+++ b/include/ignition/gazebo/EntityComponentManager.hh
@@ -534,6 +534,10 @@ namespace ignition
       /// is protected to facilitate testing.
       protected: void ClearNewlyCreatedEntities();
 
+      /// \brief Clear the list of removed components so that a call to
+      /// RemoveComponent doesn't make the list grow indefinitely.
+      protected: void ClearRemovedComponents();
+
       /// \brief Process all entity remove requests. This will remove
       /// entities and their components. This function is protected to
       /// facilitate testing.
--- a/src/EntityComponentManager.cc
+++ b/src/EntityComponentManager.cc
@@ -48,6 +48,18 @@ class ignition::gazebo::EntityComponentManagerPrivate
   /// \return True if created successfully.
   public: bool CreateComponentStorage(const ComponentTypeId _typeId);
 
+  /// \brief Create a message for the removed components
+  /// \param[in] _entity Entity with the removed components
+  /// \param[out] _msg Entity message
+  public: void SetRemovedComponentsMsgs(Entity &_entity,
+                          msgs::SerializedEntity *_msg);
+
+  /// \brief Create a message for the removed components
+  /// \param[in] _entity Entity with the removed components
+  /// \param[out] _msg State message
+  public: void SetRemovedComponentsMsgs(Entity &_entity,
+                        msgs::SerializedStateMap &_msg);
+
   /// \brief Map of component storage classes. The key is a component
   /// type id, and the value is a pointer to the component storage.
   public: std::map<ComponentTypeId,
@@ -75,7 +87,7 @@ class ignition::gazebo::EntityComponentManagerPrivate
   /// \brief The set of components that each entity has
   public: std::map<Entity, std::vector<ComponentKey>> entityComponents;
 
-  /// \brief A mutex to protect newly created entityes.
+  /// \brief A mutex to protect newly created entities.
   public: std::mutex entityCreatedMutex;
 
   /// \brief A mutex to protect entity remove.
@@ -83,6 +95,9 @@ class ignition::gazebo::EntityComponentManagerPrivate
 
   /// \brief A mutex to protect from concurrent writes to views
   public: mutable std::mutex viewsMutex;
+
+  /// \brief A mutex to protect removed components
+  public: mutable std::mutex removedComponentsMutex;
 
   /// \brief The set of all views.
   public: mutable std::map<detail::ComponentTypeKey, detail::View> views;
@@ -99,7 +114,7 @@ class ignition::gazebo::EntityComponentManagerPrivate
   /// which belongs the component, and the value is the component being
   /// removed. This is set when RemoveComponent() is called, so we assume
   /// only one removed component per entity.
-  std::map<Entity, ComponentKey> removedComponents;
+  std::unordered_multimap<Entity, ComponentKey> removedComponents;
 };
 
 //////////////////////////////////////////////////
@@ -160,6 +175,13 @@ void EntityComponentManager::ClearNewlyCreatedEntities()
   {
     view.second.ClearNewEntities();
   }
+}
+
+/////////////////////////////////////////////////
+void EntityComponentManager::ClearRemovedComponents()
+{
+  std::lock_guard<std::mutex> lock(this->dataPtr->removedComponentsMutex);
+  this->dataPtr->removedComponents.clear();
 }
 
 /////////////////////////////////////////////////
@@ -305,7 +327,11 @@ bool EntityComponentManager::RemoveComponent(
   this->UpdateViews(_entity);
 
   // Add component to map of removed components
-  this->dataPtr->removedComponents[_entity] = _key;
+  {
+    std::lock_guard<std::mutex> lock(this->dataPtr->removedComponentsMutex);
+    this->dataPtr->removedComponents.insert(std::make_pair(_entity, _key));
+  }
+
   return true;
 }
 
@@ -743,6 +769,68 @@ void EntityComponentManager::RebuildViews()
 }
 
 //////////////////////////////////////////////////
+void EntityComponentManagerPrivate::SetRemovedComponentsMsgs(Entity &_entity,
+                                        msgs::SerializedEntity *_entityMsg)
+{
+  std::lock_guard<std::mutex> lock(this->removedComponentsMutex);
+  uint64_t nEntityKeys = this->removedComponents.count(_entity);
+  if (nEntityKeys == 0)
+    return;
+
+  auto it = this->removedComponents.find(_entity);
+  for (uint64_t i = 0; i < nEntityKeys; ++i)
+  {
+    auto compMsg = _entityMsg->add_components();
+
+    auto removedComponent = it->second;
+
+    // Empty data is needed for the component to be processed afterwards
+    compMsg->set_component(" ");
+    compMsg->set_type(removedComponent.first);
+    compMsg->set_remove(true);
+
+    it++;
+  }
+
+  // Remove entity from map
+  this->removedComponents.erase(_entity);
+}
+
+//////////////////////////////////////////////////
+void EntityComponentManagerPrivate::SetRemovedComponentsMsgs(Entity &_entity,
+                                    msgs::SerializedStateMap &_msg)
+{
+  std::lock_guard<std::mutex> lock(this->removedComponentsMutex);
+  uint64_t nEntityKeys = this->removedComponents.count(_entity);
+  if (nEntityKeys == 0)
+    return;
+
+  // Find the entity in the message
+  auto entIter = _msg.mutable_entities()->find(_entity);
+
+  auto it = this->removedComponents.find(_entity);
+  for (uint64_t i = 0; i < nEntityKeys; ++i)
+  {
+    auto removedComponent = it->second;
+
+    msgs::SerializedComponent compMsg;
+
+    // Empty data is needed for the component to be processed afterwards
+    compMsg.set_component(" ");
+    compMsg.set_type(removedComponent.first);
+    compMsg.set_remove(true);
+
+    (*(entIter->second.mutable_components()))[
+      static_cast<int64_t>(removedComponent.first)] = compMsg;
+
+    it++;
+  }
+
+  // Remove entity from map
+  this->removedComponents.erase(_entity);
+}
+
+//////////////////////////////////////////////////
 void EntityComponentManager::AddEntityToMessage(msgs::SerializedState &_msg,
     Entity _entity, const std::unordered_set<ComponentTypeId> &_types) const
 {
@@ -775,9 +863,12 @@ void EntityComponentManager::AddEntityToMessage(msgs::SerializedState &_msg,
     compBase->Serialize(ostr);
 
     compMsg->set_component(ostr.str());
-
-    // TODO(anyone) Set component being removed once we have a way to queue it
   }
+
+  // Add a component to the message and set it to be removed if the component
+  // exists in the removedComponents map.
+  // TODO(anyone) Set component being removed once we have a way to queue it
+  this->dataPtr->SetRemovedComponentsMsgs(_entity, entityMsg);
 }
 
 //////////////////////////////////////////////////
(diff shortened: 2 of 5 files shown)


mcres:
@azeey @mabelzhang

src/EntityComponentManager.cc
```diff
@@ -853,6 +862,29 @@ void EntityComponentManager::AddEntityToMessage(msgs::SerializedStateMap &_msg,
...
+      cmp.set_type(removedComponent.first);
+      cmp.set_remove(true);
+      // Empty data is needed for the component to be processed afterwards
+      cmp.set_component(" ");
+
+      (*(entIter->second.mutable_components()))[
+        static_cast<int64_t>(removedComponent.first)] = cmp;
+
+      // Remove component from map
+      this->dataPtr->removedComponents.erase(_entity);
```
azeey, line 886: We should make it possible to clear `removedComponents` outside of this function. For example, `GuiRunner` calls `SetState` to synchronize its ECM and ends up calling `RemoveComponent`, but since it doesn't publish its ECM, `removedComponents` is never cleared and would grow indefinitely. I suggest creating a function `ClearRemovedComponents` and whoever uses the ECM is responsible for calling it after a time step or a sync, just like `ClearNewlyCreatedEntities` and `ProcessRemoveEntityRequests`. 

https://github.com/ignitionrobotics/ign-gazebo/blob/139020c8ec0af219efd848b84b4954c2c65936e9/src/SimulationRunner.cc#L713-L716
https://github.com/ignitionrobotics/ign-gazebo/blob/139020c8ec0af219efd848b84b4954c2c65936e9/src/gui/GuiRunner.cc#L103-L104

src/EntityComponentManager.cc
```diff
@@ -94,6 +94,12 @@ class ignition::gazebo::EntityComponentManagerPrivate
 
   /// \brief Keep track of entities already used to ensure uniqueness.
   public: uint64_t entityCount{0};
+
+  /// \brief Map of removed components. The key is the entity to
+  /// which belongs the component, and the value is the component being
+  /// removed. This is set when RemoveComponent() is called, so we assume
+  /// only one removed component per entity.
+  std::map<Entity, ComponentKey> removedComponents;
```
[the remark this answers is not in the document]
mcres: I don't really know how the sync works, but I supposed that it would occur immediately after `RemoveComponent` is called. So if you want to remove two different components of the same entity you would have to call `RemoveComponent` twice. My question is, in this case sync would take place after the first call or after the second?

src/EntityComponentManager.cc
mcres: I assume that `this->dataPtr->removedComponents.erase(_entity)` wouldn't be needed there, even if the ECM is actually publishing the message? I mean, it would be the responsibility of the server to call that method?

src/EntityComponentManager.cc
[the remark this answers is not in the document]
azeey: No, I believe sync happens once every iteration when the `SceneBroadcaster` publishes the state message in its `PostUpdate` https://github.com/ignitionrobotics/ign-gazebo/blob/139020c8ec0af219efd848b84b4954c2c65936e9/src/systems/scene_broadcaster/SceneBroadcaster.cc#L302 So, calling `RemoveComponent` doesn't cause a sync.

src/EntityComponentManager.cc
azeey: Well, remember that the server and the GUI have their own instances of the `EntityComponentManager` (ECM) class. What the GUI does on its own ECM doesn't affect the server. In the future, we will probably add a way for the GUI's ECM to be sent to the server, but that's not the case right now.

When the GUI calls `RemoveComponent` on its instance of ECM,  the removed component is added to `removeComponents` of the GUI ECM. But since the GUI doesn't call `AddEntityToMessage` (because it doesn't need to publish its state),  the `removedComponents` container inside the GUI ECM will not be cleared.

It would be better if `removedComponents` gets cleared every time step regardless of whether `AddEntityToMessage` is called.

src/EntityComponentManager.cc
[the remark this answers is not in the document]
mcres: 7992720

src/EntityComponentManager.cc
```diff
@@ -853,6 +862,29 @@ void EntityComponentManager::AddEntityToMessage(msgs::SerializedStateMap &_msg,
     compIter->second.set_component(ostr.str());
 
     // TODO(anyone) Set component being removed once we have a way to queue it
+
+    // Add a component to the message and set it to be removed if the component
+    // exists in the removedComponents map.
+
+    bool entityHasRemovedComponent =
```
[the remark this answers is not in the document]
mcres: 7992720

src/EntityComponentManager.cc
```diff
@@ -853,6 +862,29 @@ void EntityComponentManager::AddEntityToMessage(msgs::SerializedStateMap &_msg,
     compIter->second.set_component(ostr.str());
 
     // TODO(anyone) Set component being removed once we have a way to queue it
+
+    // Add a component to the message and set it to be removed if the component
```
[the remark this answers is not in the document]
mcres: Addressed in 7992720

I've created 2 overloaded functions to deal with `msgs::SerializedState` and `msgs::SerializedStateMap`.
Also added a test for the `SerializedState` message.

src/EntityComponentManager.cc
mcres: I think I didn't explain myself very well.

I've added the `ClearRemovedComponents` function in 7992720, but I'm not sure if [this](https://github.com/ignitionrobotics/ign-gazebo/blob/79927200d07cd5bc5953dde14ee15182988b06ff/src/EntityComponentManager.cc#L796) is still necessary or it's enough if `SimulationRunner.cc` and `GuiRunner.cc` call `ClearRemovedComponents`?

azeey: Ah, I see. Yeah, I don't think its necessary. In fact, I think we should remove it because leaving it in would make it so that only one caller of `EntityComponentManager::State` would get the list of removed components. The second caller won't get the removed components because `removedComponents` would have been cleared by the first call.

src/EntityComponentManager_TEST.cc
```diff
@@ -2009,6 +2009,112 @@ TEST_P(EntityComponentManagerFixture, SetChanged)
...
+  {
+    auto iter = newStateMsg.entities().find(e1);
+    const auto &e1Msg = iter->second;
+    EXPECT_EQ(1, e1Msg.components_size());
+    auto compIter = e1Msg.components().begin();
+
+    // First component
+    const auto &e1c0Msg = compIter->second;
+    EXPECT_FALSE(e1c0Msg.remove());
+  }
```
azeey, line 2065: These tests look great! One more check would be to create another instance of `EntityComponentManager` that is synced with `manager` via `SetState` before and after the components are removed. This will test that deserialization is working properly without actually publishing messages.

include/ignition/gazebo/EntityComponentManager.hh
```diff
@@ -534,6 +534,10 @@ namespace ignition
       /// is protected to facilitate testing.
       protected: void ClearNewlyCreatedEntities();
 
+      /// \brief Clear the list of removed components so that a call to
+      /// RemoveComponent doesn't make the list grow indefinitely.
+      protected: void ClearRemovedComponents();
```
azeey, line 539: Now that this exists, can you add a call to `ClearRemovedComponents` from https://github.com/ignitionrobotics/ign-gazebo/blob/139020c8ec0af219efd848b84b4954c2c65936e9/src/SimulationRunner.cc#L713-L716 and https://github.com/ignitionrobotics/ign-gazebo/blob/139020c8ec0af219efd848b84b4954c2c65936e9/src/gui/GuiRunner.cc#L103-L104

src/EntityComponentManager.cc
```diff
@@ -94,6 +109,12 @@ class ignition::gazebo::EntityComponentManagerPrivate
 
   /// \brief Keep track of entities already used to ensure uniqueness.
   public: uint64_t entityCount{0};
+
+  /// \brief Map of removed components. The key is the entity to
+  /// which belongs the component, and the value is the component being
+  /// removed. This is set when RemoveComponent() is called, so we assume
+  /// only one removed component per entity.
```
azeey, line 116: The documentation is out of date now.

src/EntityComponentManager_TEST.cc
azeey: It would also be great if you could add checks that cover `ClearRemovedComponents`? You'll have to add a member function `EntityCompMgrTest` that in turn calls `ClearRemovedComponents`. If you are running out of time, let me know and I can add this test.

mcres: Thank you!
Addressed in 942c6d4.

> It would also be great if you could add checks that cover ClearRemovedComponents

Following up on https://github.com/ignitionrobotics/ign-gazebo/pull/272#discussion_r468292455, I've added a `ClearRemovedComponents` on the first two tests, so I think I don't need to add an extra test for that? Let me know.
I've also added the sync test between managers.

include/ignition/gazebo/EntityComponentManager.hh
mcres: 942c6d4

src/EntityComponentManager.cc
mcres: 942c6d4

src/EntityComponentManager_TEST.cc
```diff
@@ -2009,6 +2013,191 @@ TEST_P(EntityComponentManagerFixture, SetChanged)
...
+
+  // Remove components and synchronize again
+  manager.RemoveComponent(e1, e1c1);
+  manager.RemoveComponent(e1, e1c2);
+
+  msgs::SerializedStateMap newStateMsg;
+  manager.State(newStateMsg);
+
+  guiManager.SetState(newStateMsg);
```
azeey, line 2175: Before calling `guiManager.State()`,  it would be good to check if the components are not there via the `Component` API

```suggestion
  EXPECT_TRUE(nullptr != guiManager.Component<IntComponent>(e1));
  EXPECT_TRUE(nullptr != guiManager.Component<DoubleComponent>(e1));
  EXPECT_TRUE(nullptr != guiManager.Component<StringComponent>(e1));
  guiManager.SetState(newStateMsg);
  EXPECT_TRUE(nullptr != guiManager.Component<IntComponent>(e1));
  EXPECT_TRUE(nullptr == guiManager.Component<DoubleComponent>(e1));
  EXPECT_TRUE(nullptr == guiManager.Component<StringComponent>(e1));
  
```

src/EntityComponentManager_TEST.cc
azeey: That will do. Thanks!

src/EntityComponentManager_TEST.cc
mcres: 2f3cd6e

azeey:
This looks ready to me. @chapulina, mind taking a quick look?

src/EntityComponentManager.cc
```diff
@@ -766,9 +856,12 @@ void EntityComponentManager::AddEntityToMessage(msgs::SerializedState &_msg,
     compBase->Serialize(ostr);
 
     compMsg->set_component(ostr.str());
-
-    // TODO(anyone) Set component being removed once we have a way to queue it
   }
+
+  // Add a component to the message and set it to be removed if the component
+  // exists in the removedComponents map.
+  // TODO(anyone) Set component being removed once we have a way to queue it
```
chapulina, line 863: I think these TODOs can be removed now, right? We're not "queuing" the removal, but for the purposes of this function, keeping track of what components were removed should be all that's needed.

src/SimulationRunner.cc
```diff
@@ -716,6 +716,9 @@ void SimulationRunner::Step(const UpdateInfo &_info)
   // Process entity removals.
   this->entityCompMgr.ProcessRemoveEntityRequests();
 
+  // Process components removals
+  this->entityCompMgr.ClearRemovedComponents();
```
chapulina, line 720: I noticed that the order between `ClearRemovedComponents` and `ProcessRemoveEntityRequests` is swapped on the server and client.

It probably doesn't matter, right? But just for consistency in case the internal implementation of these functions change in the future, it would be nice to be consistent with the ordering.

src/EntityComponentManager.cc
mcres: Should I move them to where they originally were?
I didn't remove them because I though maybe some kind of queuing should need to be done in the future.

chapulina: I'd just remove these lines. The queuing is still needed, but not by this function

src/EntityComponentManager.cc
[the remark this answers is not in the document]
chapulina: I just saw this question. I think we've been keeping it for backwards compatibility. We can deprecate the use of anything non-map. We'll probably have to deprecate on dome and remove on E.

src/EntityComponentManager.cc
mcres: b913c32

src/SimulationRunner.cc
mcres: b913c32

src/EntityComponentManager_TEST.cc
```diff
@@ -2009,6 +2013,198 @@ TEST_P(EntityComponentManagerFixture, SetChanged)
...
+  // Check message
+  {
+    auto iter = stateMsg.entities().find(e1);
+    const auto &e1Msg = iter->second;
+    auto compIter = e1Msg.components().begin();
+
+    // First component
+    const auto &e1c2Msg = compIter->second;
+    compIter++;
+    EXPECT_TRUE(e1c2Msg.remove());
```
chapulina, line 2043: This expectation is failing on macOS, as well as line 2052:

https://build.osrfoundation.org/job/ignition_gazebo-ci-pr_any-homebrew-amd64/3546/testReport/(root)/EntityComponentManagerRepeat_EntityComponentManagerFixture/SerializedStateMapMsgAfterRemoveComponent_1/

I'm not sure yet what could be different there.

mcres: It must be the way I access the components with `compIter`. It seems that the order they are accessed is different in macOS than Linux. This has been solved [like this](https://github.com/ignitionrobotics/ign-gazebo/blob/master/src/EntityComponentManager_TEST.cc#L1528-L1554) in the `State` test. Should I do the same?

azeey: The order of elements in `components()` seems like an implementation detail that we probably don't want to rely on because it depends on the (undetermined) order in `removedComponents` which is an `std::unordered_multimap`. If we want to preserve the order of components, you can use `std::unordered_map<Entity, std::vector<ComponentKey>>` for `removedComponents` instead.

A different solution would be to have the expectations in a map from `component type id` -> `bool expectation` when creating the components at the beginning of the test. The component type id is returned by `CreateComponent` as the first element of a pair. Once you have the map, you can iterate through the components in `stateMsg` and look them up from the map based on their type.

mcres:
Is there anything that needs to be done yet?

chapulina:

Issue #207, opened 2020-06-16
Topics / services without world name


chapulina:
A lot of `ign-gazebo`'s transport API includes the world name, for example, for the `diff_drive` world:

```
$ ign gazebo diff_drive.sdf & 
$ ign topic -l
/clock
/gazebo/resource_paths
/model/vehicle_blue/odometry
/model/vehicle_green/odometry
/stats
/world/diff_drive/clock
/world/diff_drive/dynamic_pose/info
/world/diff_drive/pose/info
/world/diff_drive/scene/deletion
/world/diff_drive/scene/info
/world/diff_drive/state
/world/diff_drive/stats
```

It was initially done this way because we want to eventually support running multiple worlds in parallel. But from the user's point of view, who is using a single world most of the time, this just makes the topics inconvenient to work with. For example, it makes GUI configs less portable because they need topics with world names hardcoded in them.

See relevant discussion on the [pull request](https://osrf-migration.github.io/ignition-gh-pages/#!/ignitionrobotics/ign-gazebo/pull-requests/367/page/1#comment-107384060) that added a `/clock` topic.

## Possible solutions

* Remove the world name from all topics
* Use world-agnostic topics when there's a single world, but namespace them when there are multiple worlds
* Have a [global namespace](https://github.com/osrf/gazebo/blob/cce342310bce53559d7da6ed1456899e42ad27cf/gazebo/transport/Node.cc#L105-L132) like like Gazebo-classic's "~" - this could be just the first namespace provided, for example, by the `/gazebo/worlds` service.

chapulina:
This task is important to allow us to remove all GUI plugins from the world SDF files and still have good usability. And this is important for reusing SDF files across Gazebo-classic and Ignition.

chapulina:
> it makes GUI configs less portable because they need topics with world names hardcoded in them.

#278 resolves the GUI issue

PR #278, opened 2020-08-08
Customizable layouts - fully functional

chapulina:
Related to #207

Since [this PR](https://osrf-migration.github.io/ignition-gh-pages/#!/ignitionrobotics/ign-gazebo/pull-requests/451/page/1) we've been able to load GUI configs from the command line with `--gui-config` and default to `~/.ignition/gazebo/gui.config` when no `<gui>` was specified on SDF.

But because of issue #207, we couldn't really reuse configuration files across worlds that have different names. Which meant that the 2 features above couldn't really be used.

This PR, together with https://github.com/ignitionrobotics/ign-gui/pull/104, allow us to remove custom topics and services from SDF files. This means we can rely on the default `gui.config` all the time!

More advantages:

* This lowers the bar for using Gazebo-classic SDF worlds with Ignition, because it's one less thing for them to set.
* 3,069 deleted lines!

See the new tutorial for more info.

:warning: **Delete your existing default file to see the effects of this PR: `rm ~/.ignition/gazebo/gui.config`**

---

:point_down: that's what loading `ign gazebo shapes.sdf` looks like now.

shapes_default

## Fix

The fix is based on setting the `worldNames` on the `MainWindow` so that any plugin can find them. This means that even pure `ign-gui` plugins like `WorldControl` and `WorldStats` can access world names if they want.

This is not a solution to #207, but it makes it easier to write GUI plugins that will work with whatever world is currently loaded.

## New layout

I tweaked a bit the default `gui.config` file so it includes more plugins. I used a gray background for shapes and transform control because it needs to look ok on both light and dark themes.

## Deleted lines

I went over all SDF files and:

* Deleted the `<topic>`s and `<service>` from `WorldControl` and `WorldStats`
* If the world was using the default camera pose `-6 0 6 0 0.5 0` and wasn't loading any plugin that's not on `gui.config`, I deleted the entire `<gui>` block so the default is used.
* For a couple of worlds, like `rolling_shapes`, I moved things so that they work well with the default camera angle.


chapulina:
> reflect that another reason for removing this file might be to receive the updated version of the default configuration.

Thanks, @caguero . Updated the tutorial in 2cd4b10

Issue #285, opened 2020-08-12
Unable to create GUI marker with parent


pohzhiee:
# Reproduction
The following was ran on Ubuntu 20.04 using source build, with ign-transport reverted by 1 commit.

In terminal 1:
```bash
ign gazebo shapes.sdf -v
```
In terminal 2:
```bash
ign service -s /marker --reqtype ignition.msgs.Marker --reptype ignition.msgs.Empty --timeout 1000 --req 'action: ADD_MODIFY, type: SPHERE, id: 4, scale: {x:2.0, y:2.0, z:2.0}, parent: "cylinder"'
```
This will result in the following shown in terminal 1:
```
...
[Wrn] [Component.hh:144] Trying to serialize component with data type [N3sdf2v95WorldE], which doesn't have `operator<<`. Component will not be serialized.
[Wrn] [Component.hh:144] Trying to serialize component with data type [N3sdf2v95ModelE], which doesn't have `operator<<`. Component will not be serialized.
[GUI] [Wrn] [Application.cc:649] [QT] file::/ComponentInspector/Vector3d.qml:93:9: QML Image: Detected anchors on an item that is managed by a layout. This is undefined behavior; use Layout.alignment instead.
[GUI] [Wrn] [Application.cc:649] [QT] file::/ComponentInspector/Vector3d.qml:93:9: QML Image: Detected anchors on an item that is managed by a layout. This is undefined behavior; use Layout.alignment instead.
[GUI] [Msg] Loading plugin [ignition-rendering3-ogre2]
[GUI] [Wrn] [Application.cc:649] [QT] Qt has caught an exception thrown from an event handler. Throwing
exceptions from an event handler is not supported in Qt.
You must not let any exception whatsoever propagate through Qt code.
If that is not possible, in Qt 5 you must at least reimplement
QCoreApplication::notify() and catch all exceptions there.

terminate called after throwing an instance of 'Ogre::InvalidParametersException'
  what():  OGRE EXCEPTION(2:InvalidParametersException): Node ID: 21, named '' already was a child of Node ID: 8, named ''. in Node::addChild at /var/lib/jenkins/workspace/ogre-2.1-debbuilder/repo/OgreMain/src/OgreNode.cpp (line 440)
```

# Cause
In the [marker creation section of the code](https://github.com/ignitionrobotics/ign-gazebo/blob/ign-gazebo3/src/rendering/MarkerManager.cc#L480-L505),

[...]

PR #290, opened 2020-08-13
Add checks for marker parent

pohzhiee:
Fixes issue #285. 

As for the tests, I am unsure where to add it, since there are no existing rendering tests that I am aware of. 
Also, should the test involve everything including starting a server, launching the GUI, and then using the marker service? Or should the test only involve the server and the rendering by manipulating the ECM directly?

### DIFF
--- a/src/rendering/MarkerManager.cc
+++ b/src/rendering/MarkerManager.cc
@@ -498,7 +498,10 @@ bool MarkerManagerPrivate::ProcessMarkerMsg(const ignition::msgs::Marker &_msg)
       visualPtr->AddGeometry(markerPtr);
 
       // Add visual to root visual
-      this->scene->RootVisual()->AddChild(visualPtr);
+      if (!visualPtr->HasParent())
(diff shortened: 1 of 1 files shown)


src/rendering/MarkerManager.cc
```diff
@@ -498,7 +498,10 @@ bool MarkerManagerPrivate::ProcessMarkerMsg(const ignition::msgs::Marker &_msg)
       visualPtr->AddGeometry(markerPtr);
 
       // Add visual to root visual
-      this->scene->RootVisual()->AddChild(visualPtr);
+      if (!visualPtr->HasParent())
+      {
+          this->scene->RootVisual()->AddChild(visualPtr);
```
JShep1, line 503: Nit: Just two spaces as an indent

JShep1: a3f65c2ec9586816e3a3cb2ecf11ca18469401a8

PR #296, opened 2020-08-15
2 ➡️ 3 (part 1)

chapulina:
Broken off #268 because that is having to many failures


src/gui/Gui_TEST.cc
```diff
@@ -0,0 +1,159 @@
...
+#include "ignition/gazebo/test_config.hh"
+
+int g_argc = 1;
+char **g_argv = new char *[g_argc];
+
+using namespace ignition;
+using namespace gazebo;
+
+/////////////////////////////////////////////////
+TEST(GuiTest, PathManager)
```
azeey, line 36: Is this expected to work on macOS? https://build.osrfoundation.org/job/ignition_gazebo-ci-pr_any-homebrew-amd64/3558/testReport/(root)/UNIT_Gui_TEST/test_ran/

chapulina: Good point, this is also a problem on `ign-gazebo2`. Disabled here and will also disable there.

chapulina:
Ignoring actions now because #269 is not being forward-ported yet.

PR #297, opened 2020-08-16
Allow executing a blocking single Server run in both paused and unpaused states

diegoferigo:
This is the second attempt after https://github.com/ignitionrobotics/ign-gazebo/pull/61 to allow executing paused blocking `Server` runs. The first PR was closed as detailed in https://github.com/ignitionrobotics/ign-gazebo/pull/61#issuecomment-655621591 for further discussion https://github.com/ignitionrobotics/ign-gazebo/issues/51#issuecomment-655979495. 

This PR introduces a new `Server::RunOnce` method that allows executing either unpaused or paused steps that are 1) always blocking and 2) always increase the iteration count.

The following table recaps the differences between `Run` and `RunOnce`:

| `Server::Run` | `Server::RunOnce` | Comment |
| --- | --- | --- |
| `Run(blocking=true, iterations=1, paused=false)` | `RunOnce(paused=false)` | Same behaviour, iterations count is increased by 1. |
| `Run(blocking=true, iterations=1, paused=true)` | - | Simulation loops forever until another thread calls `Server::SetPaused(false)`. Paused runs do not increase the iteration count. When the `Run` call returns, the iteration count is increased by 1. |
| - | `RunOnce(paused=true)` | Iteration count is increased by 1. Systems perform a paused step. Physics does not advance. |

Given that this PR touches core classes of the simulator, I tried to implement this feature by modifying the existing classes the little as possible. Rather than adding new APIs also to the `SimulationRunner` class that would have required further discussion, I preferred marking `Server` a friend class and introducing a new private flag.

### DIFF
--- a/src/Server.cc
+++ b/src/Server.cc
@@ -290,12 +290,12 @@ bool Server::Run(const bool _blocking, const uint64_t _iterations,
 /////////////////////////////////////////////////
 bool Server::RunOnce(const bool _paused)
 {
-    if (_paused) {
-      for (auto &runner : this->dataPtr->simRunners)
-        runner->blockingPausedStepPending = true;
-    }
+  if (_paused) {
+    for (auto &runner : this->dataPtr->simRunners)
+      runner->blockingPausedStepPending = true;
+  }
 
-    return this->Run(true, 1, _paused);
+  return this->Run(true, 1, _paused);
 }
 
 /////////////////////////////////////////////////
--- a/src/Server.cc
+++ b/src/Server.cc
@@ -292,7 +292,7 @@ bool Server::RunOnce(const bool _paused)
 {
   if (_paused) {
     for (auto &runner : this->dataPtr->simRunners)
-      runner->blockingPausedStepPending = true;
+      runner->SetNextStepAsBlockingPaused(true);
   }
 
   return this->Run(true, 1, _paused);
--- a/src/Server.cc
+++ b/src/Server.cc
@@ -288,6 +288,17 @@ bool Server::Run(const bool _blocking, const uint64_t _iterations,
 }
 
 /////////////////////////////////////////////////
+bool Server::RunOnce(const bool _paused)
+{
+    if (_paused) {
+      for (auto &runner : this->dataPtr->simRunners)
+        runner->blockingPausedStepPending = true;
+    }
+
+    return this->Run(true, 1, _paused);
+}
+
+/////////////////////////////////////////////////
 void Server::SetUpdatePeriod(
     const std::chrono::steady_clock::duration &_updatePeriod,
     const unsigned int _worldIndex)
--- a/src/Server.cc
+++ b/src/Server.cc
@@ -290,7 +290,8 @@ bool Server::Run(const bool _blocking, const uint64_t _iterations,
 /////////////////////////////////////////////////
 bool Server::RunOnce(const bool _paused)
 {
-  if (_paused) {
+  if (_paused)
+  {
     for (auto &runner : this->dataPtr->simRunners)
       runner->SetNextStepAsBlockingPaused(true);
   }
@@ -434,4 +435,3 @@ bool Server::RequestRemoveEntity(const Entity _entity,
 
   return false;
 }
-
--- a/src/Server_TEST.cc
+++ b/src/Server_TEST.cc
@@ -483,15 +483,43 @@ TEST_P(ServerFixture, RunOnceUnpaused)
     EXPECT_FALSE(*server.Running(0));
     EXPECT_EQ(0u, *server.IterationCount());
 
+    // Load a system
+    gazebo::SystemLoader systemLoader;
+    auto mockSystemPlugin = systemLoader.LoadPlugin(
+        "libMockSystem.so", "ignition::gazebo::MockSystem", nullptr);
+    ASSERT_TRUE(mockSystemPlugin.has_value());
+
+    // Check that it was loaded
+    const size_t systemCount = *server.SystemCount();
+    EXPECT_TRUE(*server.AddSystem(mockSystemPlugin.value()));
+    EXPECT_EQ(systemCount + 1, *server.SystemCount());
+
+    // Query the interface from the plugin
+    auto system = mockSystemPlugin.value()->QueryInterface<gazebo::System>();
+    EXPECT_NE(system, nullptr);
+    auto mockSystem = dynamic_cast<gazebo::MockSystem*>(system);
+    EXPECT_NE(mockSystem, nullptr);
+
+    // No steps should have been executed
+    EXPECT_EQ(0u, mockSystem->preUpdateCallCount);
+    EXPECT_EQ(0u, mockSystem->updateCallCount);
+    EXPECT_EQ(0u, mockSystem->postUpdateCallCount);
+
     // Make the server run fast
     server.SetUpdatePeriod(1ns);
 
     while (*server.IterationCount() < 100)
       server.RunOnce(false);
 
+    // Check that the server provides the correct information
     EXPECT_EQ(*server.IterationCount(), 100);
     EXPECT_FALSE(server.Running());
     EXPECT_FALSE(*server.Running(0));
+
+    // Check that the system has been called correctly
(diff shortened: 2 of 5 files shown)


diegoferigo:
@chapulina Do you think this PR can target citadel instead of dome? It introduces a new public API so there's no breaking change in this sense, however I'm not sure if in this case also ABI is preserved nor if it's a strict requirement at the current stage of development.

chapulina:
> Do you think this PR can target citadel instead of dome? 

Yeah adding a public function like that is ABI compatible, so feel free to retarget to Citadel :+1:

diegoferigo:
Ok thanks @chapulina, I will rebase this PR as soon as I can.

I used in the past months the closed #61 extensively and I have quite a lot of dowstream unit testing that run successfully also after updating my project to this new API call.

I think this PR is already ready to be reviewed, let me know if any of you have comments.

src/Server.cc
```diff
@@ -216,6 +216,17 @@ bool Server::Run(const bool _blocking, const uint64_t _iterations,
   return false;
 }
 
+/////////////////////////////////////////////////
+bool Server::RunOnce(const bool _paused)
+{
+    if (_paused) {
```
chapulina, line 222: Nit: use 2-space indentation

src/Server_TEST.cc
```diff
@@ -450,6 +450,42 @@ TEST_P(ServerFixture, RunNonBlocking)
...
+    EXPECT_FALSE(server.Running());
+    EXPECT_FALSE(*server.Running(0));
+    EXPECT_EQ(0u, *server.IterationCount());
+
+    // Make the server run fast.
+    server.SetUpdatePeriod(1ns);
+
+    while (*server.IterationCount() < 100)
+        server.RunOnce(false);
+
```
chapulina, line 466: You could `EXPECT` the iteration count to be 100 at the end of the loop

src/Server_TEST.cc
```diff
@@ -450,6 +450,42 @@ TEST_P(ServerFixture, RunNonBlocking)
   EXPECT_FALSE(*server.Running(0));
 }
 
+/////////////////////////////////////////////////
+TEST_P(ServerFixture, RunOnceUnpaused)
+{
+    gazebo::Server server;
```
chapulina, line 456: On both tests, consider adding a system and checking that it was updated as expected.

You can use the mock system like other tests on this file.

src/SimulationRunner.hh
```diff
@@ -491,6 +491,10 @@ namespace ignition
       /// \brief Map from file paths to Fuel URIs.
       private: std::unordered_map<std::string, std::string> fuelUriMap;
 
+      /// \brief True if Server::RunOnce triggered a blocking paused step
+      private: bool blockingPausedStepPending{false};
+
+      friend class Server;
```
chapulina, line 497: Would it be possible to add accessor functions for the private variable instead of making the server a friend? You know, social distancing the classes :wink:

src/Server.cc
diegoferigo: Done in 3981a1c

src/Server_TEST.cc
diegoferigo: Done in 1dc665c

src/Server_TEST.cc
diegoferigo: Sure, done in b7b09b5

src/SimulationRunner.hh
diegoferigo: I considered adding accessor functions to the `SimulationRunner` class while I was developing this PR, however I'm not convinced enough that having setters / getters of a private property that is just an implementation detail would provide any added value. Such methods would become part of the public APIs of the `SimulationRunner` class, with all the consequences of dealing in the future with them. Would any downstream user be interested in this new variable? Unfriending a class in case of a better approach would not require any deprecation process, contrarily than removing public APIs.

If you think that these two new methods would not pollute the class and give me green light, it's a trivial change from my side for adding the following accessors:

```cpp
/// \brief Get whether the next step is going to be executed as paused.
/// \return True if the next step is being executed as paused, false otherwise.
bool NextStepIsBlockingPaused() const;

/// \brief Set the next step to be blocking and paused.
void SetNextStepAsBlockingPaused(const bool value);
```

chapulina: > Such methods would become part of the public APIs 

I appreciate the concern, but note that the `SimulationRunner`'s header is not installed, so we can modify it as needed. My personal opinion is that with the accessors we can keep better control of what parts of the `SimulationRunner` are exposed to the `Server`.

diegoferigo: > `SimulationRunner`'s header is not installed

Ow thanks for pointing it out, I didn't notice it. Done in ff6701b.

src/Server.cc
```diff
@@ -220,6 +220,17 @@ bool Server::Run(const bool _blocking, const uint64_t _iterations,
   return false;
 }
 
+/////////////////////////////////////////////////
+bool Server::RunOnce(const bool _paused)
+{
+  if (_paused) {
```
chapulina, line 226: Style nit

```suggestion
  if (_paused)
  {
```

src/Server_TEST.cc
```diff
@@ -475,6 +475,100 @@ TEST_P(ServerFixture, RunNonBlocking)
...
```
chapulina, line 481: Nit: indentation

```suggestion
  gazebo::Server server;
```

src/SimulationRunner.cc
```diff
@@ -1170,3 +1178,13 @@ void SimulationRunner::AddToFuelUriMap(const std::string &_path,
 {
   this->fuelUriMap[_path] = _uri;
 }
+
+bool SimulationRunner::NextStepIsBlockingPaused() const
```
chapulina, line 1182:
```suggestion
//////////////////////////////////////////////////
bool SimulationRunner::NextStepIsBlockingPaused() const
```

src/SimulationRunner.cc
```diff
@@ -1170,3 +1178,13 @@ void SimulationRunner::AddToFuelUriMap(const std::string &_path,
...
+{
+  return this->blockingPausedStepPending;
+}
+
+void SimulationRunner::SetNextStepAsBlockingPaused(const bool value)
```
chapulina, line 1187:
```suggestion
//////////////////////////////////////////////////
void SimulationRunner::SetNextStepAsBlockingPaused(const bool value)
```

diegoferigo:
> LGTM, thanks for iterating and for the contribution! I just have some final style comments. This branch also needs to be synced with ign-gazebo3, and we should take one last look at CI results.

@chapulina Style fixed in db2989e. Thank you for providing assistance and constant feedback all along the process :)

src/Server_TEST.cc
```diff
@@ -475,6 +475,100 @@ TEST_P(ServerFixture, RunNonBlocking)
...
+  EXPECT_EQ(0u, mockSystem->postUpdateCallCount);
+
+  // Make the server run fast
+  server.SetUpdatePeriod(1ns);
+
+  while (*server.IterationCount() < 100)
+    server.RunOnce(false);
+
+  // Check that the server provides the correct information
+  EXPECT_EQ(*server.IterationCount(), 100);
```
chapulina, line 515: Compiler warning:

```
/var/lib/jenkins/workspace/ignition_gazebo-ci-pr_any-ubuntu_auto-amd64/ign-gazebo/src/Server_TEST.cc:515:3:   required from here
/var/lib/jenkins/workspace/ignition_gazebo-ci-pr_any-ubuntu_auto-amd64/ign-gazebo/test/gtest/include/gtest/gtest.h:1448:16: warning: comparison of integer expressions of different signedness: ‘const long unsigned int’ and ‘const int’ [-Wsign-compare]
```

diegoferigo: Fixed in https://github.com/ignitionrobotics/ign-gazebo/compare/db2989e16fb84ac77780d252a7d75501013bf9be..e991e58e4be16b667f551c2381f0b7ed37e9446c

chapulina:
@osrf-jenkins run tests again

PR #302, opened 2020-08-20
Add marker array service

### LINKED ISSUE #287: Add service for rendering multiple markers
Currently, Ignition Gazebo doesn't have a service that supports rendering multiple markers at once.
While working on #229, we've seen that if several markers are needed at a specific moment, requesting them individually could drop the performance significantly due to transport messages costs.

This could be done using the already existing [Marker_V message](https://github.com/ignitionrobotics/ign-msgs/blob/master/proto/ignition/msgs/marker_v.proto) as the request and most of the existing code in `MarkerManager.cc`, resulting in a [service callback](https://github.com/ignitionrobotics/ign-gazebo/blob/master/src/rendering/MarkerManager.cc#L593-L597) very similar to the `/marker` service, which renders an individual marker.
It would probably be nice to also add an example of the new functionality [here](https://github.com/ignitionrobotics/ign-gazebo/blob/master/examples/standalone/marker/marker.cc).

JShep1:
Closes #287 


src/rendering/MarkerManager.cc
```diff
@@ -601,3 +613,11 @@ void MarkerManagerPrivate::OnMarkerMsg(const ignition::msgs::Marker &_req)
   std::lock_guard<std::mutex> lock(this->mutex);
   this->markerMsgs.push_back(_req);
 }
+
+/////////////////////////////////////////////////
+void MarkerManagerPrivate::OnMarkerMsgArray(const ignition::msgs::Marker_V &_req)
```
iche033, line 618: does changing the return type to `bool` help with the misleading service call time out msg?

JShep1: Good catch, I'll try this out shortly

iche033: @caguero do you know if the return type is the reason for the service call time out?

The cmd @JShep1 tested below got no response after changing to `bool`:

```
ign service -s /marker_array --reqtype ignition.msgs.Marker_V --reptype ignition.msgs.Boolean --timeout 1000 --req 'marker: {action: ADD_MODIFY, type: SPHERE, id: 4, pose: {position: {x:2.0, y:2.0, z:2.0}, orientation: {x:0.0, y:0.0, z:0.0, w:1.0}}, parent: "cylinder"}'
```

iche033: I think I know why. You'll need to use the version of the callback with `msgs::Boolean` response arg, see [this example](https://github.com/ignitionrobotics/ign-gazebo/blob/master/src/gui/plugins/scene3d/Scene3D.cc#L2208)

JShep1: That worked for me!  Updates made [here](29ed047e36833e766e269d0203cd989144840a7c)

examples/standalone/marker/marker.cc
```diff
@@ -226,6 +226,83 @@ int main(int _argc, char **_argv)
...
+  markerMsg3->mutable_material()->mutable_diffuse()->set_a(1);
+  markerMsg3->mutable_lifetime()->set_sec(2);
+  markerMsg3->mutable_lifetime()->set_nsec(0);
+  ignition::msgs::Set(markerMsg3->mutable_scale(),
+                    ignition::math::Vector3d(1.0, 1.0, 1.0));
+  ignition::msgs::Set(markerMsg3->mutable_pose(),
+                      ignition::math::Pose3d(3, 3, 4, 0, 0, 0));
+
+  // Publish the three created markers above simultaneously
+  node.Request("/marker_array", markerMsgs);
```
iche033, line 305: I tried running this example but I don't see the shapes spawned at the end? The `ign service` call worked though

JShep1: Ahh, thanks for catching this, I had to update the service request after the callback return type was updated to bool, it should be good now.

iche033: great works for me now!

PR #303, opened 2020-08-20
Add file extension automatically for record plugin

JShep1:
This PR automatically adds the user selected file extension to the file name entered by the user, so that they do not have to manually add the extension themselves. 

Closes https://github.com/osrf/subt/issues/362


mjcarroll:
As an alternative to this logic, the save dialog also has a property `defaultSuffix` that could be used:

https://doc.qt.io/qt-5/qml-qtquick-dialogs-filedialog.html#defaultSuffix-prop

JShep1:
> As an alternative to this logic, the save dialog also has a property `defaultSuffix` that could be used:
> 
> https://doc.qt.io/qt-5/qml-qtquick-dialogs-filedialog.html#defaultSuffix-prop

Thanks I'll check this out and see if I can condense the logic

nkoenig:
> > As an alternative to this logic, the save dialog also has a property `defaultSuffix` that could be used:
> > https://doc.qt.io/qt-5/qml-qtquick-dialogs-filedialog.html#defaultSuffix-prop
> 
> Thanks I'll check this out and see if I can condense the logic

The might fix the case where the user enters the extension in the file dialog box, which results in `<name>.mp4.mp4`.

JShep1:
> The might fix the case where the user enters the extension in the file dialog box, which results in `<name>.mp4.mp4`.

I initially implemented this logic then hesitated.  The condition that QML specifies for the `defaultSuffix` attribute is that it is appended if no other extension is specified.  This could introduce an issue in which the user could improperly add `.ogv` to the end of their `mp4` video title and then QML would allow it.

I feel if we went this route we would want to compare the entered user extension to the user chosen video type, I realized this logic could become a lot more complicated than needed, and may introduce unexpected or undesired behavior (modifying the user entered extension, which I guess leads to how much we want to trust the user with what they're doing).  All of this is why I simply went with appending the initially chosen video format.  What do you think would be the best route here? @nkoenig

JShep1:
> As an alternative to this logic, the save dialog also has a property `defaultSuffix` that could be used:
> 
> https://doc.qt.io/qt-5/qml-qtquick-dialogs-filedialog.html#defaultSuffix-prop



> As an alternative to this logic, the save dialog also has a property `defaultSuffix` that could be used:
> 
> https://doc.qt.io/qt-5/qml-qtquick-dialogs-filedialog.html#defaultSuffix-prop

Don't think we'll be able to do this, `defaultSuffix` was introduced in 5.10 and Bionic brings in 5.9.

nkoenig:
> > The might fix the case where the user enters the extension in the file dialog box, which results in `<name>.mp4.mp4`.
> 
> I initially implemented this logic then hesitated. The condition that QML specifies for the `defaultSuffix` attribute is that it is appended if no other extension is specified. This could introduce an issue in which the user could improperly add `.ogv` to the end of their `mp4` video title and then QML would allow it since it _technically_ sees a proper file type extension.
> 
> I feel if we went this route we would want to compare the entered user extension to the user chosen video type in order to ensure a correct video extension. I realized this logic could become a lot more complicated than needed, and may introduce unexpected or undesired behavior (modifying the user entered extension, which I guess leads to how much we want to trust the user with what they're doing). All of this is why I simply went with appending the initially chosen video format. What do you think would be the best route here? @nkoenig

I think most applications will use whatever extension is specified, or automatically add an extension. Examples:

1. User input: "test.mp4" --> Created file: "test.mp4"
2. User input "test" --> Created file: "test.mp4"
3. User input "test.garbage" --> Created file: "test.garbage"

JShep1:
> I think most applications will use whatever extension is specified, or automatically add an extension.

Implementation added, ready for another pass.

nkoenig (approved):

chapulina (approved):

PR #330, opened 2020-09-02
Choose Render Engine from Cmd Line

JShep1:
This currently supports only rendering engines native to ignition-rendering (Ogre, Ogre2).  Still need to add `--rendering-engine-client` and `--rendering-engine-server`


chapulina:
I see this is targeting master. Do you think it would be possible to target Citadel?

JShep1:
> I see this is targeting master. Do you think it would be possible to target Citadel?

It will require an update to the `runServer` function signature (https://github.com/ignitionrobotics/ign-gazebo/pull/330/files#diff-c1dfe0042975f8dbb40b802e04cc4bd1R54-R59), I initially was going to set the old `runServer` to be deprecated and overload the function, but since `ign.cc` is technically C, function overloading is not allowed.  Can we simply update the function signature in Citadel?  I wasn't quite sure so I initially targeted at Dome

chapulina:
> It will require an update to the runServer function signature

Oh, `ign.hh` is not installed, so feel free to change it as you need on any version :wink:

JShep1:
> Oh, `ign.hh` is not installed, so feel free to change it as you need on any version

Sounds good, I'll retarget to Citadel

JShep1:
Closing to retarget at Citadel [here](https://github.com/ignitionrobotics/ign-gazebo/pull/331)

PR #335, opened 2020-09-03
Fix Log system tests to work with plugin names without ".so"

azeey:
This fixes a regression from ignitionrobotics/ign-gazebo#279.

### DIFF
--- a/src/ServerPrivate.cc
+++ b/src/ServerPrivate.cc
@@ -42,7 +42,7 @@ struct LoggingPlugin
   public: static std::string &LoggingPluginFileName()
   {
     static std::string recordPluginFileName =
-      std::string("libignition-gazebo") +
+      std::string("ignition-gazebo") +
       IGNITION_GAZEBO_MAJOR_VERSION_STR + "-log-system";
     return recordPluginFileName;
   }
--- a/src/ServerPrivate.cc
+++ b/src/ServerPrivate.cc
@@ -43,7 +43,7 @@ struct LoggingPlugin
   {
     static std::string recordPluginFileName =
       std::string("libignition-gazebo") +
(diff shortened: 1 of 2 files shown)


src/ServerPrivate.cc
```diff
@@ -43,7 +43,7 @@ struct LoggingPlugin
   {
     static std::string recordPluginFileName =
       std::string("libignition-gazebo") +
```
mabelzhang, line 45: Does the `lib` part need to be removed too?

azeey: It doesn't change the test results because this is not used for checking against library names in https://github.com/ignitionrobotics/ign-gazebo/blob/4ebe95cebcd5739dbe7c11e5e06e1cdbcfa5fde1/src/ServerPrivate.cc#L204. I went ahead and removed it anyway though. https://github.com/ignitionrobotics/ign-gazebo/pull/335/commits/5032e60a25e9e875de81fd2e9697679b618190e7

mabelzhang: Thanks! While we're here, could I ask you to change the playback plugin name in `Server.cc` too? :D https://github.com/ignitionrobotics/ign-gazebo/blob/5032e60a25e9e875de81fd2e9697679b618190e7/src/Server.cc#L87-L88

chapulina (approved):
Oops, thanks for catching that. LGTM once @mabelzhang 's last comment is addressed.

src/ServerPrivate.cc
azeey: Done https://github.com/ignitionrobotics/ign-gazebo/pull/335/commits/9fea17a96f4eec2dce413cdffdb5eeabc151f991. Also renamed the other libraries in the file.

mabelzhang (approved):

PR #336, opened 2020-09-03
add 10k actors and generalize actors pose in erb

claireyywang:
minor tweaks to add more descriptive actors and shapes sdf file names, and generalize actors pose in erb file


examples/worlds/actors_population.sdf.erb
```diff
@@ -138,13 +138,14 @@
...
 
     <%
-      # number of population
-      total = 10
-      for i in (0..total-1)
-        for j in (1..total)
+      # population = column * row
+      column = 100
+      for j in (0..column-1)
+        row = 100
```
chapulina, line 144: How about making the number of rows and columns a command line option which defaults to 10? This should give users more flexibility and we don't need to commit all the generated combinations.

Here's a snippet that we've used in the past to parse command line options of the format `[opt]=[arg]`

```
  vars = ARGV.take_while {|arg| arg[/^\w+=/]}
  ARGV.slice!(0, vars.size)
  vars.each do |var|
    k, v = var.split('=', 2)
    TOPLEVEL_BINDING.eval %Q(#{k} = "#{v}")
  end
```

Then you can expose options like `e` and `c`:

`erb r=10 c=5 actors_population.sdf.erb > actors_population.sdf`

And inside the code, use:

```
      column = c.to_i
      row = r.to_i
```

chapulina (commented):
Like was mentioned during the meeting, let's not commit the large generated files because they can be misleading to users, and also significantly increase the size of the source code and install packages.

examples/worlds/actors_population.sdf.erb
claireyywang: I pushed the cli parsing code, but am not sure how to set default command line option to 10. Any pointer's appreciated!

chapulina: How about 

```
column = 10
if (defined? c)
  column = c.to_i
end
```

examples/worlds/actors_population.sdf.erb
```diff
@@ -138,13 +142,26 @@
     </include>
 
     <%
-      # number of population
```
chapulina, line 141: The dude on line 141 seems a bit lost in the crowd, is he needed?

chapulina (approved):
Works well! I just have a suggestion to remove the lonely guy who doesn't follow the crowd :sweat_smile:

examples/worlds/actors_population.sdf.erb
claireyywang: Do you mean this `# number of population` line? I think you are looking at the old file? This is the new change equivalent to that line https://github.com/ignitionrobotics/ign-gazebo/pull/336/files#diff-b1617932cd58db140bca53385d0da9bcR145

chapulina: Oh sorry, I meant the included actor around line 141 of the new file: 

```
    <include>
      <uri>https://fuel.ignitionrobotics.org/1.0/Mingfei/models/actor</uri>
    </include>
```

iche033: I remember the lonely actor used to be able to walk around others without bumping into anyone :)

claireyywang: Ooo that's what it was. I was wondering why there's always an extra actor different from others. Yeah I'll remove it.

iche033 (approved):

claireyywang:
Failed test cases on osx and ubuntu are not introduced by this PR. Merging...

PR #339, opened 2020-09-09
Make sure OpenGL core profile context is used by GzScene3D

peci1:
Should fix ignitionrobotics/ign-rendering#128 .

### DIFF
--- a/src/gui/plugins/scene3d/Scene3D.cc
+++ b/src/gui/plugins/scene3d/Scene3D.cc
@@ -1922,18 +1922,24 @@ QSGNode *RenderWindowItem::updatePaintNode(QSGNode *_node,
     // and makeCurrent down below while setting up our own context.
     current->doneCurrent();
 
-    auto surfaceFormat = QSurfaceFormat();
-    surfaceFormat.setMajorVersion(3);
-    surfaceFormat.setMinorVersion(3);
-    surfaceFormat.setProfile(QSurfaceFormat::CoreProfile);
-    
     this->dataPtr->renderThread->context = new QOpenGLContext();
-    // Although it seems unbelievable, we can request another format for a
-    // shared context; it is needed because Qt selects by default a compat
-    // context which is much less likely to provide OpenGL 3.3 (only closed
-    // NVidia drivers). This means there will be mismatch between what is
-    // reported by QSG_INFO=1 and by OGRE.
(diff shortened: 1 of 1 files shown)


src/gui/plugins/scene3d/Scene3D.cc
```diff
@@ -1922,8 +1922,18 @@ QSGNode *RenderWindowItem::updatePaintNode(QSGNode *_node,
...
+    surfaceFormat.setProfile(QSurfaceFormat::CoreProfile);
+    
     this->dataPtr->renderThread->context = new QOpenGLContext();
-    this->dataPtr->renderThread->context->setFormat(current->format());
+    // Although it seems unbelievable, we can request another format for a
+    // shared context; it is needed because Qt selects by default a compat
+    // context which is much less likely to provide OpenGL 3.3 (only closed
+    // NVidia drivers). This means there will be mismatch between what is
+    // reported by QSG_INFO=1 and by OGRE.
+    this->dataPtr->renderThread->context->setFormat(surfaceFormat);
```
iche033, line 1936: can you wrap the code with a check for `ogre2` since this requirement is engine-dependent?

```
    if (this->RenderUtil()->EngineName() == "ogre2")
    {
      ...
      this->dataPtr->renderThread->context->setFormat(surfaceFormat);
    }
    else
    {
      this->dataPtr->renderThread->context->setFormat(current->format());
    }
```

 I think in the future we should support querying the underlying render engine for the min required opengl version so we don't need to hard code the engine name check.

iche033 (commented):
thanks for testing this and submitting a PR!

Tested on my machine that uses nvidia driver and it's working fine.

src/gui/plugins/scene3d/Scene3D.cc
peci1: You're right, I fixed that.

iche033:
looks good to me. Just need to trigger another round of CI builds once jenkins is back online again..

peci1:
I'm not sure what the test results mean... None of the failures should be related.

iche033:
the test failures should be unrelated and are captured in ignitionrobotics/ign-rendering#129.

peci1:
Aahh... I've just triggered a rebuild :-D If that's possible, you can stop it to save buildfarm time.

iche033 (approved):

chapulina (approved):

PR #342, opened 2020-09-09
Resource spawner: accept relative paths

chapulina:
The resource spawner currently doesn't work for relative paths, because the file is being converted to text before URIs can be resolved.

This PR adds a new event that accepts a file path instead of an SDF string. Notably, this fixes spawning models from this collection:

https://app.ignitionrobotics.org/GoogleResearch/fuel/collections/Google%20Scanned%20Objects

I also went ahead and updated a lot of the terminology on the plugin to be more generic and don't apply just to models. We should add support for actors and lights soon.

### DIFF
--- a/include/ignition/gazebo/gui/GuiEvents.hh
+++ b/include/ignition/gazebo/gui/GuiEvents.hh
@@ -161,13 +161,17 @@ namespace events
     static const QEvent::Type kType = QEvent::Type(QEvent::User + 3);
   };
 
-  /// \brief Event called to spawn a preview model.  Used in shapes plugin.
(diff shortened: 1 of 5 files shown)


JShep1 (approved):
Looks great!  Thanks for the quick fix, I tested it out with some of my favorite resources
items

I think CI is down right now so pending that it's good to go.

chapulina:
@osrf-jenkins run tests, you can do it! :mechanical_arm:

JShep1:
Just wanting to check this is ready to merge @chapulina , I think the homebrew warnings aren't related to this PR.

PR #346, opened 2020-09-11
Support spawning during log playback

nkoenig:
This automatically includes the `usercommands` plugin during log playback in order to support spawning on new models.

I added in a `SetEntityCreateOffset` function to prevent entity ID collisions. Logplayback will directly insert entities with their recorded ids, which doesn't alter the `entityCount`. As a result, a spawn call will attempt to create an entity with an existing ID.

See issue #347 


src/systems/log/LogPlayback.cc
```diff
@@ -209,6 +209,10 @@ void LogPlayback::Configure(const Entity &,
   // Prepend working directory if path is relative
   this->dataPtr->logPath = common::absPath(this->dataPtr->logPath);
 
+  // Set the entity offset.
+  // \todo This number should be included in the log file.
+  _ecm.SetEntityOffset(90000);
```
chapulina, line 214: How about ticketing an issue for keeping track of the max entity count during log record? 

Also, maybe a number like `std::numeric_limits<int64_t>::max() / 2` could be more generic than the current number.

include/ignition/gazebo/EntityComponentManager.hh
```diff
@@ -696,6 +696,12 @@ namespace ignition
           const Entity _entity, const ComponentTypeId _type,
           gazebo::ComponentState _c = ComponentState::OneTimeChange);
 
+      /// \brief All future entities will have an id that starts at _offset.
+      /// This can be used to avoid entity id collisions, such as during log
+      /// playback.
+      /// \param[in] _offset Offset value.
+      public: void SetEntityCreateOffset(uint64_t _offset);
```
chapulina, line 703: Mind adding a unit test for this function? It could help covering the lines on `CreateEntity()` which are currently uncovered.

src/EntityComponentManager.cc
```diff
@@ -1224,3 +1224,9 @@ std::unordered_set<ComponentTypeId> EntityComponentManager::ComponentTypes(
 
   return result;
 }
+
+/////////////////////////////////////////////////
+void EntityComponentManager::SetEntityCreateOffset(uint64_t _offset)
+{
+  this->dataPtr->entityCount = _offset;
```
chapulina, line 1231: I can see this being an issue is `_offset < this->dataPtr->entityCount`.

chapulina (commented):
I think an alternative implementation would be to increase `entityCount` within `EntityComponentManagerPrivate::CreateEntityImplementation` whenever the entity being created has a higher number than the current `entityCount`.

I think this would cover your use case without the need to hardcode an offset for log playback. It would also avoid exposing the ECM's internal counting mechanism in case we ever want to change it.

nkoenig:
> I think an alternative implementation would be to increase `entityCount` within `EntityComponentManagerPrivate::CreateEntityImplementation` whenever the entity being created has a higher number than the current `entityCount`.
> 
> I think this would cover your use case without the need to hardcode an offset for log playback. It would also avoid exposing the ECM's internal counting mechanism in case we ever want to change it.

Yeah, I tried that but it fails when the log file spawns a new entity. For example:

1. Start log playback paused.
2. Spawn a new element.
3. Start playback
4. Log file inserts a new entity. This entity could have have same id as the entity in step 2.

src/systems/log/LogPlayback.cc
nkoenig: The issue is #347.

Added a test in 8a2b44cf1122c67287ac079a20efac4cb8ffd422

include/ignition/gazebo/EntityComponentManager.hh
nkoenig: Added test: 8a2b44cf1122c67287ac079a20efac4cb8ffd422

src/EntityComponentManager.cc
nkoenig: Yup, this function could be dangerous if used improperly.

nkoenig: Warning added in d9001c7c84e3fb8f9d6a0e19c7472601e1515729

chapulina (approved):
> I tried that but it fails when the log file spawns a new entity.

Oh got it, makes sense, I didn't think about the order of operations.

PR #351, opened 2020-09-15
Enhancing log playback performance

adlarkin:
I've changed how log playback works in order to increase RTF and remove jumpy playback.

Previously, when setting updated entity poses in the log file as a `PeriodicChange`, log playback was jumpy. This was because the `SceneBroadcaster` updates at a different rate than when the log file is parsed ([60 Hz](https://github.com/ignitionrobotics/ign-gazebo/blob/fcc3f63425f3add16c9f38168a012b6f05032fbf/src/systems/scene_broadcaster/SceneBroadcaster.cc#L265-L266)), so, if log parsing didn't occur during the same cycle as a publication from the `SceneBroadcaster`, changed poses from the latest log parse call were not captured in the rendering update. In order to get around this, the updated poses were [saved as a `OneTimeChange`](https://github.com/ignitionrobotics/ign-gazebo/commit/fcc3f63425f3add16c9f38168a012b6f05032fbf).

The issue with using `OneTimeChange` is that playback performance is noticeably slower for worlds with a lot of entities. The reason for this is because whenever there is a `OneTimeChange`, the `SceneBroadcaster` will [publish the full state](https://github.com/ignitionrobotics/ign-gazebo/blob/fcc3f63425f3add16c9f38168a012b6f05032fbf/src/systems/scene_broadcaster/SceneBroadcaster.cc#L276-L280), even if only one pose was changed.

In order to maintain the performance that comes with `PeriodicChange` while removing the jumpy playback that arises from different parsing and scene update rates, I cached the changed poses from the latest log parse call as a `PeriodicChange`. Then, whenever it is time for a scene update, only the cached pose updates are rendered instead of the full scene.

Below shows the RTF difference when playing back a log at regular speed for a SubT world, which has a lot of entities. The first video is before caching (`OneTimeChange`), and the second video uses caching (`PeriodicChange`). A few things to note:
* The RTF varies significantly using `OneTimeChange`, when it should stay close to 100% since we omit physics during log playback.
* The RTF stays close to 100% when using caching and `PeriodicChange`, which is expected for log playback at regular speed since we have the pose updates for every step without having to run any physics.

log_playbacl_before_update
log_playback_after_update

Here is an example of log playback at double speed with the simple [double pendulum world](https://github.com/ignitionrobotics/ign-gazebo/blob/21fda4bed0b4e41f5f44f04ebb5cdce5a1f654fd/examples/worlds/log_record_dbl_pendulum.sdf). We still achieve smooth playback and keep an RTF close to the theoretical expectation of 200% for 2x playback speed (on my machine, at least).

dbl_pendulum_double_speed


adlarkin:
After taking a look at the failed tests in Jenkins, I believe that the failed tests can be ignored for the following reasons:
1. I copied the Docker setup from Jenkins over to my machine, and I could not reproduce the failed test from `LogSystemTest.RecordAndPlayback`. Perhaps the failure of this test in Jenkins had something to do with the CI machine?
2. There was a segmentation fault for `INTEGRATION_triggered_publisher.test_ran`, which also appears to be problematic in previous builds (see [build 4138](https://build.osrfoundation.org/job/ignition_gazebo-ci-pr_any-ubuntu_auto-amd64/4138/testReport/)).
3. There was a test failure for `DiffDriveTest.SkidPublishCmd`, which also occurred in the [previous build](https://build.osrfoundation.org/job/ignition_gazebo-ci-pr_any-ubuntu_auto-amd64/4140/testReport/).

Considering that the `DiffDrive` test failure is an issue in other builds, I'm assuming that we can ignore the failure of this test in the GitHub action Ubuntu CI as well.

@chapulina @iche033 what do you guys think?

src/systems/log/LogPlayback.cc
```diff
@@ -133,53 +138,15 @@ LogPlayback::~LogPlayback()
...
-    const msgs::Pose_V &_msg)
+void LogPlaybackPrivate::Parse(const msgs::Pose_V &_msg)
 {
-  // Maps entity to pose recorded
-  // Key: entity. Value: pose
-  std::map <Entity, msgs::Pose> idToPose;
-
-  for (int i = 0; i < _msg.pose_size(); ++i)
+  // save the new entity pose updates
+  this->recentEntityPoseUpdates.clear();
```
nkoenig, line 144: Is it correct to clear the `recentEntityPoseUpdates`? `Parse(const msgs::Pose_V &_msg)` could be called multiple times from a single `Update` iteration, so would clearing the `map` miss some pose updates?

adlarkin: That's a good point. Looking at the `LogPlayback::Update` function, it looks like there are two places where `Parse(const msgs::Pose_V &_msg)` could be called: lines 486 & 563. I haven't seen an issue of missing pose updates in the tests I ran, but maybe that's because the rendering updates at 60Hz, which means the naked eye could miss something that happened for a single frame (and actually, now that you bring this up, maybe this is why the `LogSystemTest.RecordAndPlayback` test that I mentioned in my comment [above](https://github.com/ignitionrobotics/ign-gazebo/pull/351#issuecomment-693677769) is failing).

How about I define a boolean flag in the `LogPlayback::Update` method that decides whether `clear()` should be called or not? `clear()` should only be called in the first `Parse(const msgs::Pose_V &_msg)` call for each `Update`, so the flag will initially be set to true, but once `Parse(const msgs::Pose_V &_msg)` has been called once, then the flag will be set to false for the rest of the `Update` iteration.

adlarkin: Handling multiple `Parse` calls in a single `Update` has been addressed in commit [06ba47b](https://github.com/ignitionrobotics/ign-gazebo/pull/351/commits/06ba47be6a19ecee0e03cb93a8daf80250d22e9a).

nkoenig (approved):

src/systems/log/LogPlayback.cc
iche033: does it work if you just call  `this->dataPtr->recentEntityPoseUpdates.clear()` once before going into the while loop then you don't need to handle clearing the cache in `Parse` with `_clear`?

nkoenig: Done in b07ebddf7d552e56030213228c0ad3c9384bab58

nkoenig:
I'm noticing some jerky playback. I tested with a log file that has five robot models. Can you confirm?

nkoenig:
The problem appears to be on the GUI side. The GUI process is running at over 200% CPU, and the server is running at 48% while maintaining a constant RTF.

adlarkin:
> I'm noticing some jerky playback. I tested with a log file that has five robot models. Can you confirm?

@nkoenig @iche033 I believe the jerky playback has been re-introduced by commit [b07ebdd](https://github.com/ignitionrobotics/ign-gazebo/pull/351/commits/b07ebddf7d552e56030213228c0ad3c9384bab58).

We cannot call clear on every `Update`, because if we clear cached poses but never parse the log file on a given `Update` cycle (which can happen depending on which timestep we are on - sometimes, the [batch](https://github.com/ignitionrobotics/ign-gazebo/blob/b07ebddf7d552e56030213228c0ad3c9384bab58/src/systems/log/LogPlayback.cc#L472-L473) received from the log file is empty), then we have no updated poses to share with the `SceneBroadcaster`, which means that no rendering update will occur until a future log file `Parse` call aligns with the next update from the `SceneBroadcaster` (recall that the `SceneBroadcaster` publishes changes at 60Hz, NOT at every `PreUpdate -> Update -> PostUpdate` step - so, log file parsing and rendering updates don't always line up). 

So, we must keep the poses that are currently cached until we know that we have new poses to overwrite the cache with, or else we run the risk of losing track of recent changes since `Parse` is not called every `Update` cycle (we know that we have new poses to overwrite the cache with whenever `Parse` is called).

Does that make sense? If not, I can try to explain things another way, but essentially, we have to remember that the `EntityComponentManager` keeps track of changed entities for a single `PreUpdate -> Update -> PostUpdate` step only - so, in order to make sure that pose updates from a step at time `t` are captured in a rendering update that occurs at time `t + dt`, we must keep these updates cached and _**only clear them if we have something to replace them with**_ since they are a `PeriodicChange`, not a `OneTimeChange`.

src/systems/log/LogPlayback.cc
adlarkin: > does it work if you just call `this->dataPtr->recentEntityPoseUpdates.clear()` once before going into the while loop then you don't need to handle clearing the cache in `Parse` with `_clear`?

I don't believe that this will work - see my comment [below](https://github.com/ignitionrobotics/ign-gazebo/pull/351#issuecomment-694315698).

adlarkin:
@nkoenig @iche033 as a follow up, I tested the code before and after commit [b07ebdd](https://github.com/ignitionrobotics/ign-gazebo/pull/351/commits/b07ebddf7d552e56030213228c0ad3c9384bab58), and noticed that jumpy playback occurs when calling clear every `Update` instead of in the first `Parse` call - so I think it's safe to say that jumpy playback is occurring because `Parse` isn't necessarily called every `Update`, as I described in my [previous comment](https://github.com/ignitionrobotics/ign-gazebo/pull/351#issuecomment-694315698).

In order to achieve smooth playback, should we revert [b07ebdd](https://github.com/ignitionrobotics/ign-gazebo/pull/351/commits/b07ebddf7d552e56030213228c0ad3c9384bab58) and go back to [06ba47b](https://github.com/ignitionrobotics/ign-gazebo/pull/351/commits/06ba47be6a19ecee0e03cb93a8daf80250d22e9a), or do either of you have any other preferences?

nkoenig:
> @nkoenig @iche033 as a follow up, I tested the code before and after commit [b07ebdd](https://github.com/ignitionrobotics/ign-gazebo/pull/351/commits/b07ebddf7d552e56030213228c0ad3c9384bab58), and noticed that jumpy playback occurs when calling clear every `Update` instead of in the first `Parse` call - so I think it's safe to say that jumpy playback is occurring because `Parse` isn't necessarily called every `Update`, as I described in my [previous comment](https://github.com/ignitionrobotics/ign-gazebo/pull/351#issuecomment-694315698).
> 
> In order to achieve smooth playback, should we revert [b07ebdd](https://github.com/ignitionrobotics/ign-gazebo/pull/351/commits/b07ebddf7d552e56030213228c0ad3c9384bab58) and go back to [06ba47b](https://github.com/ignitionrobotics/ign-gazebo/pull/351/commits/06ba47be6a19ecee0e03cb93a8daf80250d22e9a), or do either of you have any other preferences?

Thanks for the explanation. That makes sense. I'll revert my changes.

nkoenig:
Reverted.

src/systems/log/LogPlayback.cc
iche033: ok got it thanks for the explanation. That makes sense. I think it helps to document this behavior. Could you add a brief inline comment on this, maybe at where  `clearCachedPoseUpdates` is declared?

adlarkin: Yeah, I left a comment in [dee2afe](https://github.com/ignitionrobotics/ign-gazebo/pull/351/commits/dee2afe3fd488201dd7a7aeb2bbd94113246d2f4). There is also another mention of the multiple parse calls in the [`Parse(const msgs::Pose_V &_msg, bool &_clear)` method documentation](https://github.com/ignitionrobotics/ign-gazebo/blob/dee2afe3fd488201dd7a7aeb2bbd94113246d2f4/src/systems/log/LogPlayback.cc#L73-L79). Do those comments seem to be helpful?

iche033: thanks, yeah the comments are helpful.

adlarkin: Great, I'll mark this as resolved (feel free to "unresolve" this conversation if any other issues arise).

iche033:
changes look good. Waiting for one last round of CI builds

nkoenig:
Can I get one more review of this?

chapulina (approved):
Great work, @adlarkin ! Solves 2 problems at once :+1:

iche033 (approved):

Issue #101, opened 2020-04-28
Get/set user camera pose


chapulina:
Add a widget which lets the user set the exact camera pose from the GUI.

This could be an addition to the [view_angle](https://github.com/ignitionrobotics/ign-gazebo/tree/ign-gazebo2/src/gui/plugins/view_angle) plugin.

jennuine:
Implemented in #863

PR #352, opened 2020-09-15
Add support for moving the GUI camera to a pose

nkoenig:
Also added a topic to publish the camera pose.

There are situation where I would like to move the user camera to a pose that is not associated with a model.


chapulina:
Can you fix the conflicts, @nkoenig ?

I'll also mention issue #101 for cross-referencing, since this solves half of that issue.

nkoenig:
> Can you fix the conflicts, @nkoenig ?
> 
> I'll also mention issue #101 for cross-referencing, since this solves half of that issue.

Resolved, and thanks for the link.

src/gui/plugins/scene3d/Scene3D.cc
```diff
@@ -2306,6 +2398,33 @@ bool Scene3D::OnViewAngle(const msgs::Vector3d &_msg,
   return true;
 }
 
+/////////////////////////////////////////////////
+bool Scene3D::OnMoveToPose(const msgs::Pose &_msg, msgs::Boolean &_res)
```
chapulina, line 2402: I think we should be making use of the [GUICamera](https://github.com/ignitionrobotics/ign-msgs/blob/master/proto/ignition/msgs/gui_camera.proto) message for these things. But maybe this will need some deeper refactoring, so probably this is too much to do now...

Another thing I think would be good to include in the message is the animation time. I imagine some use cases may need the camera to teleport directly to the target pose without an animation.

src/gui/plugins/scene3d/Scene3D.cc
```diff
@@ -215,6 +225,9 @@ inline namespace IGNITION_GAZEBO_VERSION_NAMESPACE {
     /// \brief Flag for indicating whether we are in view angle mode or not
     public: bool viewAngle = false;
 
+    /// \brief Flag for indicating whether we are in move to pose mode or not
+    public: bool moveToPose = false;
```
chapulina, line 229: Maybe we don't need this flag if we set the `moveToPoseValue` to NaNs when not in move to pose mode. Just a suggestion.

src/gui/plugins/scene3d/Scene3D.cc
```diff
@@ -2215,6 +2290,21 @@ void Scene3D::LoadConfig(const tinyxml2::XMLElement *_pluginElem)
...
+  this->dataPtr->node.Advertise(this->dataPtr->moveToPoseService,
+      &Scene3D::OnMoveToPose, this);
+  ignmsg << "Move to pose service on ["
+         << this->dataPtr->moveToPoseService << "]" << std::endl;
+
+  // camera position topic
+  this->dataPtr->cameraPoseTopic = "/gui/camera/pose";
+  this->dataPtr->cameraPosePub =
+    this->dataPtr->node.Advertise<msgs::Pose>(this->dataPtr->cameraPoseTopic);
+  ignmsg << "Camera pose topic published on ["
```
chapulina, line 2305: published -> advertised?

chapulina (commented):
Works for me, just some minor comments

src/gui/plugins/scene3d/Scene3D.cc
```diff
@@ -98,6 +98,16 @@ inline namespace IGNITION_GAZEBO_VERSION_NAMESPACE {
         const rendering::NodePtr &_target, double _duration,
         std::function<void()> _onAnimationComplete);
 
+    /// \brief Move the camera to look at the specified target
```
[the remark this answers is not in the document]
nkoenig: Fixed in 251b573c3e83ee444d4d136a08eeb5e514eae91f

src/gui/plugins/scene3d/Scene3D.cc
```diff
@@ -347,11 +363,21 @@ inline namespace IGNITION_GAZEBO_VERSION_NAMESPACE {
...
+    /// \brief Move to pose service
+    public: std::string moveToPoseService;
+
     /// \brief Shapes service
     public: std::string shapesService;
+
+    /// \brief Camera pose topic
+    public: std::string cameraPoseTopic;
+
+    /// \brief Camera pose topic
```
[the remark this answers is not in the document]
nkoenig: Fixed in: 251b573c3e83ee444d4d136a08eeb5e514eae91f

src/gui/plugins/scene3d/Scene3D.cc
```diff
@@ -501,6 +527,32 @@ void IgnRenderer::Render()
...
 
+  // Move to pose
+  {
+    IGN_PROFILE("IgnRenderer::Render MoveToPose");
+    if (this->dataPtr->moveToPose)
+    {
+      if (this->dataPtr->moveToHelper.Idle())
+      {
+        std::vector<Entity> selectedEntities =
+          this->dataPtr->renderUtil.SelectedEntities();
```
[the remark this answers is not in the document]
nkoenig: Removed in: 251b573c3e83ee444d4d136a08eeb5e514eae91f

src/gui/plugins/scene3d/Scene3D.cc
nkoenig: Changed to GUICamera in 251b573c3e83ee444d4d136a08eeb5e514eae91f.
Do you want to capture the animation time idea in an issue?

src/gui/plugins/scene3d/Scene3D.cc
nkoenig: Removed the flag, and used an `std::optional` in 251b573c3e83ee444d4d136a08eeb5e514eae91f

src/gui/plugins/scene3d/Scene3D.cc
nkoenig: Fixed in 251b573c3e83ee444d4d136a08eeb5e514eae91f

src/gui/plugins/scene3d/Scene3D.cc
chapulina: Sure, I can ticket an issue.

Do you think now it's weird that the service is called `/gui/camera/pose` but the message can take more than just pose though?

nkoenig: I think it's fine. I think the idea would be to add new parameters to the GUICamera message, which we couldn't do with the Pose message.

src/gui/plugins/scene3d/Scene3D.cc
```diff
@@ -2276,10 +2358,11 @@ void Scene3D::Update(const UpdateInfo &_info,
           return true;
         });
 
-    auto renderWindow = this->PluginItem()->findChild<RenderWindowItem *>();
     renderWindow->SetWorldName(this->dataPtr->worldName);
   }
 
+  msgs::Pose poseMsg = msgs::Convert(renderWindow->CameraPose());
+  this->dataPtr->cameraPosePub.Publish(poseMsg);
```
iche033, line 2365: convert the pose and publish only if there are connections?

src/gui/plugins/scene3d/Scene3D.cc
```diff
@@ -2252,6 +2318,21 @@ void Scene3D::LoadConfig(const tinyxml2::XMLElement *_pluginElem)
   ignmsg << "View angle service on ["
          << this->dataPtr->viewAngleService << "]" << std::endl;
 
+  // view pose service
```
iche033, line 2321: `view pose service` -> `move to pose service`

iche033 (commented):
looks good to me. Just minor comments.

src/gui/plugins/scene3d/Scene3D.cc
nkoenig: Done in 673ba0fbb1f59383d66ab415a92a768ee1383988

src/gui/plugins/scene3d/Scene3D.cc
nkoenig: Done in 673ba0fbb1f59383d66ab415a92a768ee1383988

iche033 (approved):

chapulina (approved):

PR #355, opened 2020-09-17
Suggestions to scrubber UI (#299)

chapulina:
Targeted at #299 

* Update default `playback_gui.config` so the scrubber is floating and attached to the scene's center bottom (removed `WorldControl` to give it more room)
* Fix problem with passed `--gui-config`s not being able to access `gui::worldNames()` (overlooked at #278)
* Remove `Rectangle`s from QML that were just wrapping other elements, for simplicity

scrubber_qml2

### DIFF
--- a/src/gui/Gui.cc
+++ b/src/gui/Gui.cc
@@ -159,12 +159,6 @@ std::unique_ptr<ignition::gui::Application> createGui(
   // Configuration file from command line
   if (_guiConfig != nullptr && std::strlen(_guiConfig) > 0)
   {
-    if (!app->LoadConfig(_guiConfig))
-    {
-      ignwarn << "Failed to load config file[" << _guiConfig << "]."
-              << std::endl;
-    }
-
     // Use the first world name with the config file
     // TODO(anyone) Most of ign-gazebo's transport API includes the world name,
     // which makes it complicated to mix configurations across worlds.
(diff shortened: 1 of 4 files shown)


JShep1 (commented):
This works great, I do, however, think we might need another line of code checking to see if the `guiConfig` variable is `_playback_` or not.  Given `_guiConfig` was previously only used for a path to a custom gui configuration and not being used as a message passing mechanism to load a separate gui config, it was always assumed to be "" or a path to a config.  Since now it can be `_playback_`, [this line](https://github.com/ignitionrobotics/ign-gazebo/pull/355/files#diff-5fba7fafd836ae3a5c902e852be57f66R173) is failing and providing the error message.  I guess we can hard check if it equals `_playback_` to not execute that line for now, but I feel like there should be a better solution down the road when we possibly have multiple various default gui configs for various scenarios.

chapulina:
Thanks, @JShep1 , I think I've addressed all your feedback.

While working on this I also noticed that whenever I playback a log it also records a new one. I'm not sure if it's caused by your PR, I'll try `ign-gazebo2`.

JShep1:
> While working on this I also noticed that whenever I playback a log it also records a new one. I'm not sure if it's caused by your PR, I'll try `ign-gazebo2`.

Where is the new log file being stored to?

chapulina:
> Where is the new log file being stored to?

A timestamped directory in `~/.ignition/gazebo/log`. What I'm doing is:

1. `rm -rf ~/.ignition/gazebo/log`
1. `ign gazebo --record rolling_shapes.sdf -r`
1. See there's only 1 log
1. `ign gazebo --playback ~/.ignition/gazebo/log/<timestamp>`
1. See there are now 2 logs

chapulina:
> I playback a log it also records a new one

Ahhh nevermind, it's only recording the console logs! Forget that this happened :)

JShep1:
> Ahhh nevermind, it's only recording the console logs! Forget that this happened :)

What are you talking about? :thinking: :wink:

JShep1 (approved):

chapulina:
> it's only recording the console logs!

All the messages that you see if you run `-v 4` get logged to `~/.ignition/gazebo/log/timestamp/server_console.log`, regardless of whether you're recording a `state.log` or not.

PR #356, opened 2020-09-17
3 ➡️ 4

chapulina:
Forward-port Citadel to Dome. This contains bug fixes, so it should be ok to go in during feature freeze.

Diff:

https://github.com/ignitionrobotics/ign-gazebo/compare/master...ign-gazebo3


chapulina:
I believe the homebrew test failures are flaky.

j-rivero (commented):
Long one :) I just checked the .cc and .hh files, can not find any major problem. I did not compile the thing so let's see what CI says.

One question: I still see `find . \( -name '*.sdf*' -o  -name '*.md' -o -name '*.cc' \) -exec grep -nHi 'libignition-gazebo.*.so' {} \` multiple files using the filename="libignition-foo.so" notation in sdf files, documents and cc files. Do they need to be changed?

j-rivero:
> I believe the homebrew test failures are flaky.

I see segfauls in three of them. We can take care of them in different issues or PRs but at least one happens after a warning (that might be a hint):
```
93: Warning [Model.cc:212] Non-unique names detected in XML children of model with name[X3].
 93/130 Test  #93: INTEGRATION_multicopter ..............................***Exception: SegFault  1.35 sec
```

I just relaunch the job to see if we get more information.

chapulina:
> Do they need to be changed?

Good point, removed in 6d0d0b1

j-rivero (approved):

PR #357, opened 2020-09-17
Fix diff_drive_skid test

azeey:
A recent [update](https://github.com/azeey/dart/pull/9) to DART changed the way the slip parameter is implemented. Part of this change removed the small amount of slip each surface had by default. This in turn broke this test because the vehicle would not steer as some amount of slip is needed in the lateral direction. The solution is to lower the coefficient of friction in the lateral direction so the friction in that direction can be overcome by the steering force.

I believe this should also fix test failures on Focal since the DART version used there does not support the slip parameter.


nkoenig (approved):

azeey:
Not sure why the GIthub CI Action is not running for this PR.

chapulina:
Can we update the required `ign-physics` version to 1.9?

azeey:
> Can we update the required `ign-physics` version to 1.9?

Yup. https://github.com/ignitionrobotics/ign-gazebo/pull/357/commits/f9d82ca54c7b2ae334a9ff008a329d75271b129a

azeey:
CI Looks good. The two failing tests are not caused by this PR.

PR #359, opened 2020-09-20
Add search and sort for resource spawner

### LINKED ISSUE #341: Resource Spawner Search Bar and Sort
As the number of fuel models increases, it would probably be helpful to have a search bar at the top of the [Resource Spawner](https://github.com/ignitionrobotics/ign-gazebo/tree/master/src/gui/plugins/resource_spawner) plugin to query the currently selected user's models.

It may also be helpful to have a Sort Downloaded option which will place all of the already downloaded fuel models at the first of the list of resources, a Sort A-Z option., and a Sort Most Recent option (the initially loaded order of resources).

JShep1:
Closes #341 

This PR most of all implements a Search textfield as well as a Sort By Combo Box.  It also introduces some optimizations in loading the models (stores all resources by owner at initialization, previously only the owner data was being gathered and another search for the models was done when the user clicked on each owner). Some slight formatting was changed with the QML as I thought some of it looked a little wonky, I'm still open for suggestions on the QML.

The search bar will search (case insensitive) all the names and owners of all the resources with your submitted keyword, similar to how Fuel's website does.  I also added a sort by Most Recent, A - Z, Z - A, and Downloaded.  The sort and search work together to display the subset of the two. 

Are there any more similar organization features or sorting methods that would be desired while we're here?

Here's a small example:

searchandsort

### DIFF
--- a/src/gui/plugins/resource_spawner/ResourceSpawner.qml
+++ b/src/gui/plugins/resource_spawner/ResourceSpawner.qml
@@ -30,7 +30,7 @@ import QtQml.Models 2.2
 Rectangle {
   id: resourceSpawner
   color: Material.background
-  Layout.minimumWidth: 600
+  Layout.minimumWidth: 700
   Layout.minimumHeight: 500
   anchors.fill: parent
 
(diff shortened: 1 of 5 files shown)


src/gui/plugins/resource_spawner/ResourceSpawner.qml
```diff
@@ -301,11 +312,64 @@ Rectangle {
...
+          Layout.preferredWidth: (searchSortBar.width - 80) / 2
+          Layout.leftMargin: 10
+          TextField {
+            id: searchField
+            anchors.fill: parent
+            placeholderText: "Search"
+            leftPadding: 2
+            topPadding: 15
+            color: Material.theme == Material.Light ? "black" : "white"
+            onAccepted: {
```
chapulina, line 339: While doing some quick testing, it felt a bit awkward to have to press enter for the changes to take effect. That was especially confusing when I erased the search, didn't press enter, then selected another user, and no models showed. I think this could be at least `onEditingFinished` so it's triggered when losing focus.

I see that `TextField` doesn't have a signal that's triggered for every text change, but have you looked into `TextInput`? It looks like it has a [textEdited](https://doc.qt.io/qt-5/qml-qtquick-textinput.html#textEdited-signal) signal. Not sure how updating the results for every new letter would affect performance though.

chapulina (approved):
Works great, the code looks good. I just have a comment about `onAccepted` for your consideration.

src/gui/plugins/resource_spawner/ResourceSpawner.qml
JShep1: I changed the `TextField` to `TextInput` as the re-search for every key entered doesn't noticeably affect performance and search is still quick after every key entered.  The downside is that `TextInput` doesn't have the native underline for the input box nor does it have a `placeholderText` attribute.  I added the search icon next to the `TextInput` box and also made the box an offset color from the top bar that it's on so that it is more visible.  Additionally, I increased the minimum width of this plugin given the 2 pane format is more natural, I think, and the search and sort fits more naturally and doesn't feel squished.  Let me know what you think and if you think we can improve this anymore. 
Screenshot from 2020-10-01 17-49-58

chapulina: Works great!

PR #365, opened 2020-09-23
Fix gui sendEvent memory leaks

iche033:
The leaks mainly come from `sendEvent` calls. According to Qt's [API documentation](https://doc.qt.io/qt-5/qcoreapplication.html#sendEvent), the events are not deleted and so should be created on the stack

With the animation memory fix in ignitionrobotics/ign-common#98, I see fixed memory usage over time when running the `actors_population.sdf` world.

ign-gazebo-gui-mem-usage


chapulina (approved):
Nice! Something to keep in mind as we create more events.

PR #372, opened 2020-09-24
Decrease time step for quadcopter world

chapulina:
With the current time step, the quad penetrates the ground if turned off from very high:

quad_ground

I think users trying out this world may run into this often, so I think it's good to have a good default behaviour. This change doesn't prevent the demo from running at 100% RTF on my machine.


azeey (approved):
Thi

PR #374, opened 2020-09-24
Fix adding performers to entity tree

chapulina:
When running `ign gazebo levels.sdf`, we see these errors, and the performers are never added to the entity tree:

```
[GUI] [Err] [EntityTree.cc:128] Failed to find parent entity [6]                                                                      
[GUI] [Err] [EntityTree.cc:128] Failed to find parent entity [22]
```

Similar errors can be seen whenever a world with a performer is loaded.

This happens because the performer entities come up before their parent models.

This PR puts such entities in a queue so they're processed later once their parents show up.

### DIFF
--- a/src/gui/plugins/entity_tree/EntityTree.cc
+++ b/src/gui/plugins/entity_tree/EntityTree.cc
@@ -125,8 +125,8 @@ void TreeModel::AddEntity(unsigned int _entity, const QString &_entityName,
 
   if (nullptr == parentItem)
   {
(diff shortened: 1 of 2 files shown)


mjcarroll (approved):
LGTM.

Actions CI failure is Diff drive, which is known flaky I believe?

chapulina:
> Actions CI failure is Diff drive, which is known flaky I believe?

It should be fixed when https://github.com/ignitionrobotics/ign-gazebo/pull/357 is merged forward.

PR #377, opened 2020-09-25
Add instructions to bitmask world

chapulina:
All example worlds should have some instructions so the user knows what the world is trying to demonstrate.

I also removed the `<gui>` tag because the default GUI fits this world well.


JShep1 (approved):
LGMT! Didn't realize this feature even existed, it's super cool and a feature that's also pretty relevant in game engines.

Issue #325, opened 2020-09-02
APIs that ease migration from classic


chapulina:
Classes like [ignition::gazebo::Model](https://github.com/ignitionrobotics/ign-gazebo/blob/master/include/ignition/gazebo/Model.hh) are meant to provide helpful wrappers around entities and components which are simpler to use. This is important for users that are not accustomed to C++. This should also ease migration from Gazebo-classic, because the APIs would be more similar.

Not all functions from Gazebo-classic make sense in the new paradigm though. For example, one could query time information like `SimTime`, `RealTime`, `IsPaused` from `gazebo::physics::World`, but on Ignition, these properties don't belong to the world entity and should be obtained from `UpdateInfo` instead.

Here's a preliminary list of some APIs that we could offer on Ignition to mimic Gazebo-classic. The underlying implementation would use calls like `ecm.Component`, `ecm.EntityByComponents`, etc.

* We should create a class for each entity type: `World`, `Model`, `Link`, `Joint`, `Light`, `Actor`, `Sensor`, etc.
* We should have accessors for most properties, such as name, pose...
* Child entities should be accessible by name, such as `World::LightByName()`
* Each class will also have accessors specific to that entity type. For example, `World::Gravity()`, `World::MagneticField()`, etc. 
* We should also add generic APIs that apply to many entity types to [Util.hh](https://github.com/ignitionrobotics/ign-gazebo/blob/master/include/ignition/gazebo/Util.hh)

---

Status:

- [x] World 
- [x] Model
- [x] Link
- [x] Joint
- [x] Light
- [x] Actor
- [x] Sensor

PR #378, opened 2020-09-25
More world APIs, helper function ComponentData

chapulina:
See the added tutorial for the new APIs.

Also added the `EntityComponentManager::ComponentData` helper function that returns `std::nullopt` if the entity doesn't have that component. It's a bit redundant that the user needs to pass 2 template arguments, but I couldn't find a better way.

Part of solving #325


include/ignition/gazebo/EntityComponentManager.hh
```diff
@@ -214,6 +214,16 @@ namespace ignition
...
               ComponentTypeT *Component(const ComponentKey &_key);
 
+      /// \brief Get the data from a component.
+      /// * If the component type doesn't hold any data, this won't compile.
+      /// * If the entity doesn't have that component, it will return nullopt.
+      /// * If the entity has the component, return its data.
+      /// \param[in] _entity The entity.
+      /// \return The data of the component of the specified type assigned to
+      /// specified Entity, or nullptr if the component could not be found.
+      public: template<typename ComponentTypeT, typename DataType>
```
mjcarroll, line 224: Since we have `DataType` defined in our components as `Type`,  would it be possible to drop the second template argument?

I think we could do

```
template<typename ComponentTypeT>
std::optional<ComponentTypeT::Type> ComponentData(const Entity _entity) const;
```

Which I think makes calling cleaner, so you don't have to know the mapping from ComponentTypeT to DataType.

chapulina: Awesome, this is the piece of the puzzle that I was missing. Done in 88ad6319503e8f518a88827d02ba6b2e7b0bffa1

include/ignition/gazebo/World.hh
```diff
@@ -0,0 +1,192 @@
...
+      /// \todo(anyone) Make const
+      public: gazebo::Entity ActorByName(const EntityComponentManager &_ecm,
+          const std::string &_name);
+
+      /// \brief Get the ID of a model entity which is an immediate child of
+      /// this world.
+      /// \param[in] _ecm Entity-component manager.
+      /// \param[in] _name Model name.
+      /// \return Model entity.
+      /// \todo(anyone) Make const
```
mjcarroll, line 146: I just took a quick look and couldn't see what was keeping this from being const?

mjcarroll (commented):
A few questions.

include/ignition/gazebo/World.hh
mjcarroll: I just set it to `const` locally, and it didn't seem to have an issue building, maybe worth going ahead and doing?

chapulina: Nice catch, that was carelessly copied from

https://github.com/ignitionrobotics/ign-gazebo/blob/0d16dc96ec0544440a082f37d2cfa72a97dcbc9f/include/ignition/gazebo/Model.hh#L129-L131

Fixed in 6a8a6d0

mjcarroll (approved):

PR #381, opened 2020-09-26
Fix top-left toolbar layout so magnet shows

chapulina:
The plugin was too small to show it :scream: 

Before:

nognet

After:

magnet

> I had to change the size of `Shapes` for `TransformControl` to grow. See https://github.com/ignitionrobotics/ign-gui/issues/129


JShep1 (approved):
Never even realized it was gone :sweat_smile:

PR #384, opened 2020-09-26
Remove sidebar and put world control in bottom left

JShep1:
I just thought it was slightly weird how all of the other tutorials are in this format and this one had the world control in the side bar, so I just updated it to be like the others.


chapulina (approved):
Nice catch!

PR #386, opened 2020-09-26
Update log tutorials

JShep1:
The justification for the removal of `log_playback.sdf` is in this issue: https://github.com/ignitionrobotics/ign-gazebo/issues/293 .  I think we should probably make a tutorial for logging now.


chapulina (approved):
Good call :+1:

PR #391, opened 2020-09-29
Fix error message on Visualize Lidar plugin

chapulina:
This prevents this message when choosing the first topic:

`Unable to subscribe from topic []`


JShep1 (approved):
Thanks!

PR #394, opened 2020-10-01
Add nullchecks for scene

JShep1:
These nullchecks are needed for a mock custom render engine plugin example I'm currently writing [here](https://github.com/ignitionrobotics/ign-rendering/pull/154).

### DIFF
--- a/src/gui/plugins/scene3d/Scene3D.cc
+++ b/src/gui/plugins/scene3d/Scene3D.cc
@@ -441,6 +441,9 @@ void IgnRenderer::Render()
   if (!this->dataPtr->followTarget.empty())
   {
     rendering::ScenePtr scene = this->dataPtr->renderUtil.Scene();
+    if (!scene)
+      return;
+
     rendering::NodePtr target = scene->NodeByName(this->dataPtr->followTarget);
     if (!target && !this->dataPtr->followTargetWait)
     {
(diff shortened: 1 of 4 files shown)


src/gui/plugins/scene3d/Scene3D.cc
```diff
@@ -441,6 +441,9 @@ void IgnRenderer::Render()
   if (!this->dataPtr->followTarget.empty())
   {
     rendering::ScenePtr scene = this->dataPtr->renderUtil.Scene();
+    if (!scene)
```
chapulina, line 444: How about getting the scene once at the beginning of `IgnRenderer::Render()` and printing a warning? This way, if we add other scene calls to the body of this function, we don't need to remember to check for `nullptr`.

JShep1: Fixed here 3008548703b50ab7916bcba68e047c4bd5c456c8 .  I might need some help on the warning message, wasn't sure what exactly to put

chapulina: That looks good to me

chapulina (approved):
LGTM, let's wait for Jenkins to come back online to check CI

JShep1:
> LGTM, let's wait for Jenkins to come back online to check CI

:+1: I also retargeted at another branch of mine adding support for custom render engine to make the dependency easier to pull when testing the `ign-rendering` functionality

PR #395, opened 2020-10-07
Fix source build instructions for ign-gazebo3

adlarkin:
When trying to build `ign-gazebo3` from source, @chapulina and I noticed a few things:
1. The `qtdeclarative5-models-plugin` third-party dependency seems unnecessary (this package also doesn't exist for Ubuntu Focal - only exists for Ubuntu Bionic)
2. There were missing dependencies for `ign-sensors` and `ign-fuel-tools`

Users should now be able to go through the source build instructions without any issues.

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,7 +140,7 @@ for dependency installation instructions for each supported operating system.
 1. Install third-party libraries:
 
     ```
-    sudo apt-get -y install cmake build-essential curl cppcheck g++-8 libbenchmark-dev libgflags-dev doxygen ruby-ronn libtinyxml2-dev libtinyxml-dev software-properties-common libeigen3-dev qtdeclarative5-models-plugin
+    sudo apt-get -y install cmake build-essential curl cppcheck g++-8 libbenchmark-dev libgflags-dev doxygen ruby-ronn libtinyxml2-dev libtinyxml-dev software-properties-common libeigen3-dev
     ```
 
 2. Install required Ignition libraries:
@@ -162,7 +162,7 @@ for dependency installation instructions for each supported operating system.
     ```
 
     ```
-    sudo apt-get -y install libignition-cmake2-dev libignition-common3-dev libignition-math6-eigen3-dev libignition-plugin-dev libignition-physics2-dev libignition-rendering3-dev libignition-tools-dev libignition-transport8-dev libignition-gui3-dev libignition-msgs5-dev libsdformat9-dev
+    sudo apt-get -y install libignition-cmake2-dev libignition-common3-dev libignition-math6-eigen3-dev libignition-plugin-dev libignition-physics2-dev libignition-rendering3-dev libignition-tools-dev libignition-transport8-dev libignition-gui3-dev libignition-msgs5-dev libsdformat9-dev libignition-sensors3-dev libignition-fuel-tools4-dev
     ```
 
 ### Building from source
--- a/README.md
+++ b/README.md
@@ -99,7 +99,7 @@ are handled for you. The downside of a binary install is that you won't be able 
 the code. See [Source Install](#source-install) for information on
 installing Ignition Gazebo from source.
 
-**Ubuntu Bionic (version 2)**
+**Ubuntu Bionic**
 
 1. Configure package repositories.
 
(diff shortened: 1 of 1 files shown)


adlarkin:
Another thing to note is that the [source install link](https://github.com/ignitionrobotics/ign-gazebo/blob/ign-gazebo3/README.md#source-install) in the `README` is currently broken because the [section it links to](https://github.com/ignitionrobotics/ign-gazebo/blob/ign-gazebo3/README.md#source-install-version-3) is labeled as `Source Install (version 3)` instead of `Source Install`. Should we strip the `(version 3)` from this header to fix this link?

README.md
```diff
@@ -140,7 +140,7 @@ for dependency installation instructions for each supported operating system.
 1. Install third-party libraries:
 
     ```
-    sudo apt-get -y install cmake build-essential curl cppcheck g++-8 libbenchmark-dev libgflags-dev doxygen ruby-ronn libtinyxml2-dev libtinyxml-dev software-properties-common libeigen3-dev qtdeclarative5-models-plugin
+    sudo apt-get -y install cmake build-essential curl cppcheck g++-8 libbenchmark-dev libgflags-dev doxygen ruby-ronn libtinyxml2-dev libtinyxml-dev software-properties-common libeigen3-dev
```
chapulina, line 143: Now that every repository has all the dependencies listed in `.github/ci/packages.apt`, we should start using that in the installation instructions so the instructions are always up-to-date. What do you think of doing that here?

For example, running this after cloning:

```
export SYSTEM_VERSION=bionic
sudo apt -y install \
  $(sort -u $(find . -iname 'packages-'$SYSTEM_VERSION'.apt' -o -iname 'packages.apt') | tr '\n' ' ')
```

chapulina (commented):
> Should we strip the (version 3)

Yeah I think that makes it simpler

README.md
adlarkin: This is definitely a cleaner approach that is easier to maintain. I went ahead and made this change in [a18718a](https://github.com/ignitionrobotics/ign-gazebo/pull/395/commits/a18718a30790768ccf0338798a02731657b5a804).

adlarkin:
> Another thing to note is that the [source install link](https://github.com/ignitionrobotics/ign-gazebo/blob/ign-gazebo3/README.md#source-install) in the `README` is currently broken because the [section it links to](https://github.com/ignitionrobotics/ign-gazebo/blob/ign-gazebo3/README.md#source-install-version-3) is labeled as `Source Install (version 3)` instead of `Source Install`. Should we strip the `(version 3)` from this header to fix this link?

@chapulina I have also noticed that the [Binary Install](https://github.com/ignitionrobotics/ign-gazebo/tree/adlarkin/source_build_fix#binary-install) instructions install `ign-gazebo2`, not `ign-gazebo3`. I'm assuming that this was written at a time when the debian package for `ign-gazebo3` was not yet available - since it's available now, should I fix the binary install instructions to download `ign-gazebo3`?

README.md
chapulina: Thanks. Just note that the user can only run that after cloning the repository, so that instruction should be moved down, or the cloning moved up.

chapulina:
> should I fix the binary install instructions to download ign-gazebo3?

Yes, please, thanks. The `ign-gazebo3` branch should direct the user to install that version. We should also check that `ign-gazebo2` and `ign-gazebo4` are correct.

README.md
adlarkin: Just noticed that as well. I went ahead and made a note about cloning first in [4ae6846](https://github.com/ignitionrobotics/ign-gazebo/pull/395/commits/4ae684628272c11a392ca89254b42d7930987341).

adlarkin:
> We should also check that `ign-gazebo2` and `ign-gazebo4` are correct.

It looks like `ign-gazebo2` and `ign-gazebo4` are correct, but they don't make use of the `packages*.apt` files in the source install instructions - would you like me to update the `README`s for those branches to use `packages*.apt` as well?

chapulina (approved):

chapulina:
> would you like me to update the READMEs for those branches to use packages*.apt as well?

That would be awesome, thanks!

adlarkin:
I know this is already merged, but one other thing I just noticed is that we don't mention installing `cppcheck` in order to run `make codecheck` (https://github.com/ignitionrobotics/ign-gazebo/tree/ign-gazebo3#testing). Should we add this to the `README`? If so, should I just make another PR for this?

PR #397, opened 2020-10-08
Docker development setup for ign-gazebo3

adlarkin:
Here's a proposal to add a Docker setup that would allow for easier development of `ign-gazebo3`. The idea is to have a `Dockerfile` that installs all of the `ign-gazebo3` dependencies, and to load the source code for `ign-gazebo3` into this Docker container as a volume at runtime. Then, users can develop and test their changes inside of Docker without having to worry about configuring source and binary installs on their own machine - users would simply build the source code for `ign-gazebo3` inside of the container using something like [colcon](https://colcon.readthedocs.io/en/released/#).

I've added a [README](https://github.com/ignitionrobotics/ign-gazebo/blob/development_docker/docker/development/README.md), which outlines how to use a source install of `ign-gazebo` within Docker.

A few things to consider:

1. In order to get graphics capabilities inside of docker, `run.bash` configures a few things and then passes this info to the `docker run` command. Right now, `run.bash` can only be used if:
    - You have [nvidia-docker2](https://github.com/NVIDIA/nvidia-docker)
    - Your computer has an nvidia graphics card
2. @iche033 suggested looking into [osrf/rocker](https://github.com/osrf/rocker) as a potential solution for allowing this setup to work with various graphics cards (AMD and NVidia). I haven't had the chance to experiment with this yet.

This is just an initial draft/proposal. I'd love to hear anyone's thoughts if they have any - I'm sure there are ways to improve this setup!

### DIFF
--- a/docker/development/Dockerfile
+++ b/docker/development/Dockerfile
@@ -1,0 +1,59 @@
+# Ubuntu 18.04 with nvidia-docker2 beta opengl support
+FROM nvidia/opengl:1.0-glvnd-devel-ubuntu18.04
+
+# install third-party packages for things like developemnt
+RUN apt-get -qq update && apt-get -y install \
+    git \
+    gnupg \
+    g++-8 \
+    lsb-release \
+    sudo \
+    vim \
+    wget \
+  && rm -rf /var/lib/apt/lists/*
+
+# set up repositories
+RUN sh -c 'echo "deb http://packages.osrfoundation.org/gazebo/ubuntu-stable `lsb_release -cs` main" > /etc/apt/sources.list.d/gazebo-stable.list' \
+  && sh -c 'echo "deb http://packages.osrfoundation.org/gazebo/ubuntu-prerelease `lsb_release -cs` main" > /etc/apt/sources.list.d/gazebo-prerelease.list' \
+  && wget http://packages.osrfoundation.org/gazebo.key -O - | apt-key add -
+
+# install dependencies
+COPY packages.apt .
+COPY packages-bionic.apt .
+RUN apt-get -qq update && apt-get -y install \
+    $(sort -u $(find . -iname 'packages-bionic.apt' -o -iname 'packages.apt') | tr '\n' ' ') \
+  && rm -rf /var/lib/apt/lists/*
+
+# enable colcon for building packages (time zone needs to be enabled)
+RUN sh -c 'echo "deb http://packages.ros.org/ros/ubuntu `lsb_release -cs` main" > /etc/apt/sources.list.d/ros-latest.list' \
+  && apt-key adv --keyserver 'hkp://keyserver.ubuntu.com:80' --recv-key C1CF6E31E6BADE8868B172B4F42ED6FBAB17C654
+RUN echo 'Etc/UTC' > /etc/timezone && ln -s /usr/share/zoneinfo/Etc/UTC /etc/localtime \
+  && apt-get -qq update && apt-get -y install \
+    python3-colcon-common-extensions \
+  && rm -rf /var/lib/apt/lists/*
+
+# configure gcc8
+RUN update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-8 800 --slave /usr/bin/g++ g++ /usr/bin/g++-8 --slave /usr/bin/gcov gcov /usr/bin/gcov-8
+
+# create a non-root user
+# uses same user_id as user outside the container
+# requires docker build argument user_id
+ARG user_id
+ENV USERNAME developer
+RUN useradd -U --uid ${user_id} -ms /bin/bash $USERNAME \
+ && echo "$USERNAME:$USERNAME" | chpasswd \
+ && adduser $USERNAME sudo \
+ && echo "$USERNAME ALL=NOPASSWD: ALL" >> /etc/sudoers.d/$USERNAME
+
+# commands below run as the developer user
+USER $USERNAME
+
+# create a colcon workspace that can build ign-gazebo from source
+# (the ign-gazebo source code will be loaded into the src directory of this workspace)
+RUN mkdir -p ~/ws/src
+
+# start the container where the workspace is
+WORKDIR /home/$USERNAME/ws
+
+# start a bash shell by default
+CMD [ "/bin/bash" ]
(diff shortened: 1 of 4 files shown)


docker/development/Dockerfile
```diff
@@ -0,0 +1,59 @@
...
+FROM nvidia/opengl:1.0-glvnd-devel-ubuntu18.04
+
+# install third-party packages for things like developemnt
+RUN apt-get -qq update && apt-get -y install \
+    git \
+    gnupg \
+    g++-8 \
+    lsb-release \
+    sudo \
+    vim \
```
chapulina, line 11: Not everyone will use `vim`, so I think that could be left out

docker/development/Dockerfile
```diff
@@ -0,0 +1,59 @@
...
+    wget \
+  && rm -rf /var/lib/apt/lists/*
+
+# set up repositories
+RUN sh -c 'echo "deb http://packages.osrfoundation.org/gazebo/ubuntu-stable `lsb_release -cs` main" > /etc/apt/sources.list.d/gazebo-stable.list' \
+  && sh -c 'echo "deb http://packages.osrfoundation.org/gazebo/ubuntu-prerelease `lsb_release -cs` main" > /etc/apt/sources.list.d/gazebo-prerelease.list' \
```
chapulina, line 17: Since this container is meant for building from source, I'd remove the `prerelease` repo

chapulina (commented):
Do you mind adding a `README.md` file to `docker/development` with the instructions that you have on the PR?

How about calling `colcon` from within the docker file? 

I also think that requiring the user to run `build.sh` and `run.sh` from different directories could be a source of confusion. I think we should either run them from the same directory, or clone the `ign-gazebo` repository within it so it doesn't matter where the scripts are run from.

adlarkin:
> Do you mind adding a `README.md` file to `docker/development` with the instructions that you have on the PR?

`README` has been added, and can be found [here](https://github.com/ignitionrobotics/ign-gazebo/blob/development_docker/docker/development/README.md).

> I also think that requiring the user to run `build.sh` and `run.sh` from different directories could be a source of confusion.

`build.bash` and `run.bash` are both run from the `docker/development` directory - or am I misunderstanding what you mean here?

> How about calling `colcon` from within the docker file?

@chapulina I prefer loading the repository into the container as a [volume](https://docs.docker.com/storage/volumes/) at run time and having users build the repository in the container as they wish instead of cloning the repository and pre-building it in the `Dockerfile` for a few reasons:

1. If users want to use different branches of the repository or test changes that were made since image build time, they'll have to re-build the workspace anyways. Also, if the branch of the repository that was cloned in the `Dockerfile` has been updated since the image was built, the image will need to be re-built or the user will need to pull the recent commits/changes and re-build inside of the container, so building the workspace as a part of the `Dockerfile` doesn't really help much (if anything, I see it as a limitation).
2. If users want to clone the repository using SSH, they'd have to set up SSH keys inside of the Dockerfile, which is a bit of a pain.
3. If users want to make changes to the repository (commits and/or new branches) that was cloned as a part of the image, they'd have to set up their GitHub information inside of the Docker container.
4. Most importantly, using a volume means that changes made to files in the volume are persistent. This is important because perhaps a user has a copy of the repository on their machine locally, and they want to test their branch - so, they could load their repository into the container, build the workspace, and then test their changes. If they see that something is wrong, they can make changes, test again, and then kill the container when they're happy with their changes. After the container is destroyed, their local copy of the repository will still have the changes they made, which can then be committed and pushed.
5. Since changes made to files in a volume are synced between the container and the user's own copy, users can go through the development process on their own machine, using whatever development tools/IDEs they already have set up. This way, users don't have to re-configure their development tools inside of Docker - Docker would simply be used for building/testing.

Essentially, volumes allow for greater development and testing flexibility, and remove the need to configure things like Git and development tools inside of Docker.

chapulina (commented):
> build.bash and run.bash are both run from the docker/development directory - or am I misunderstanding what you mean here?

Sorry, I think what I meant is that different paths need to be passed to the scripts. See my comment below on how to make the scripts runnable from any directory and removing the need to pass in extra paths as arguments.

>  the image will need to be re-built or the user will need to pull the recent commits/changes and re-build inside of the container

Ok, fair point. I guess I don't mind much pulling the changes inside the container and re-building, but I think this is subjective. So I'm ok with the way you have it now.

---

You make a lot of good points for building as a manual step. What do you think of documenting the workflow in the README? I for example am used to just having SSH, Git and a clone all inside my containers, so I'm not used to going out of the container to commit and push. I think it would be beneficial to users to have this usage explained.

As a side note, the more I look into this, the more I think it would be more valuable to have a top-level `Dockerfile` for the entire collection than having one per library. The workflow that is being proposed here makes it easy to build a single library from source, but that's often not the case for developers. And if I understand correctly, from the moment that I need to build a dependency from source, this file is not useful without some extra changes to mount more volumes, right?

adlarkin:
> As a side note, the more I look into this, the more I think it would be more valuable to have a top-level Dockerfile for the entire collection than having one per library. The workflow that is being proposed here makes it easy to build a single library from source, but that's often not the case for developers. And if I understand correctly, from the moment that I need to build a dependency from source, this file is not useful without some extra changes to mount more volumes, right?

@chapulina you're correct about this proposal being easy/targeted for building a single library (in this case, `ign-gazebo`) from source. If you need source dependencies from other repositories, then it will require one of the following:
1. Cloning the repositories and checking out the relevant branches inside of the Docker container
2. Mounting these other repositories as volumes
3. Using `docker cp` to temporarily copy dependent repositories into the Docker container

As you have already noticed, doing one of the three things above can be annoying if you need more than just one library from source. So, I am 100% in favor of a top-level Dockerfile that uses the entire collection from source. If we decide to go down the top-level Dockerfile route, there are a few things I think we should discuss first. I've created an [issue](https://github.com/ignitionrobotics/ign-common/issues/112) for this so that we can have a discussion about a top-level Dockerfile approach in the issue, avoiding irrelevant discussion in this PR.

If we decide to use a top-level Dockerfile, is it even worth having Docker setups for each individual library? Perhaps we should close this PR if having library-specific Docker setups is unnecessary.

adlarkin:
I'm closing this because I'm working on a different (and simpler) approach that leverages [osrf/rocker]() for increased hardware compatibility. We have a good conversation about what's being done in https://github.com/ignitionrobotics/ign-common/issues/112.

PR #400, opened 2020-10-09
Tutorial on migrating SDF files from Gazebo classic

chapulina:
This is just an initial tutorial. Over time, I think we should increment it as we identify other specific pain points.

This is in draft mode until #281 gets in, CC @mjcarroll .


tutorials/migration_sdf.md
```diff
@@ -0,0 +1,336 @@
...
+
+This method is not recommended, because the file most likely won't work if
+copied to a different computer. It also becomes inconvenient to move files to
+different directories.
+
+## Plugins
+
+Plugins are binary files compiled to use with a specific simulator. Plugins
+for Gazebo classic and Ignition Gazebo aren't compatible, so plugins will need
+to be specified for each simulator separately.
```
mjcarroll, line 189: It may be important to note that plugins from different versions (citadel vs dome) are also not likely to be compatible.

chapulina: Oh I see what you mean. From the SDF file's point of view, nothing should change though. I'll clarify.

chapulina: See fc5db21

mjcarroll (approved):

PR #412, opened 2020-10-14
Fix a crash in the grid_config plugin

nkoenig:
The grid plugin will crash  when changing properties if a world has `<grid>false</grid>` in the `<scene>` tag.

For example:

```
<?xml version="1.0" ?>
<sdf version="1.6">
  <world name="default">
    <plugin
      filename="ignition-gazebo-user-commands-system"
      name="ignition::gazebo::systems::UserCommands">
    </plugin>
    <plugin
    </plugin>
    <physics name="1ms" type="ode">
      <max_step_size>0.004</max_step_size>
      <real_time_factor>1.0</real_time_factor>
    </physics>

    <scene>
      <ambient>0.1 0.1 0.1 1.0</ambient>
      <background>0 0 0 1.0</background>
      <grid>false</grid>
      <origin_visual>false</origin_visual>
    </scene>

    <include>
      <static>true</static>
      <name>staging_area</name>
      <pose>0 0 0 0 0 0</pose>
      <uri>https://fuel.ignitionrobotics.org/1.0/OpenRobotics/models/Cave Starting Area Type B</uri>
    </include>
  </world>
</sdf>
```

This fixes the problem, but there might be a more elegant solution.

### DIFF
--- a/src/gui/plugins/grid_config/GridConfig.cc
+++ b/src/gui/plugins/grid_config/GridConfig.cc
@@ -131,8 +131,16 @@ void GridConfig::UpdateGrid()
       mat->SetDiffuse(this->dataPtr->gridParam.color);
       mat->SetSpecular(this->dataPtr->gridParam.color);
     }
+    else
+    {
+      ignerr << "Grid visual missing material" << std::endl;
+    }
 
     visual->SetVisible(this->dataPtr->gridParam.visible);
+  }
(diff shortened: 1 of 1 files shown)


chapulina:
> there might be a more elegant solution.

See #413

chapulina (approved):

PR #414, opened 2020-10-16
Add ability to record video based on sim time

iche033:
The current default behavior of video recording in gui camera is to record using wall clock time as timestamp. This PR adds the ability to specify from the gui config to record videos using sim time. This is so that we can still produce smooth videos that run in real time even when the playback RTF is low or fluctuates often. To get perfect video timing, you'll also require this fix in ignitionrobotics/ign-common#105

To test:

1. add a `<record_video><use_sim_time>` flag to the scene3d plugin in your  `~/.ignition/gazebo/gui.config` (and maybe `~/.ignition/gazebo/playback_gui.config` if you're doing playback - for me I just had to update `gui.config`), e.g.

    ```
    <!-- 3D scene -->
    <plugin filename="GzScene3D" name="3D View">
      <ignition-gui>
        <title>3D View</title>
        <property type="bool" key="showTitleBar">false</property>
        <property type="string" key="state">docked</property>
      </ignition-gui>

      <engine>ogre2</engine>
      <scene>scene</scene>
      <ambient_light>0.4 0.4 0.4</ambient_light>
      <background_color>0.8 0.8 0.8</background_color>
      <camera_pose>6 0 6 0 0.5 3.14</camera_pose>
      <record_video>
        <use_sim_time>true</use_sim_time>
      </record_video>
    </plugin>
    ```

1. Launch ign-gazebo with a complex scene so that the RTF is low. Alternatively, artificially increase cpu usage using `stress`, e.g. `stress -c 12`

1. Open the Video Recorder plugin (from top right menu) and start recording. Keep an eye on the duration of sim time that you record.

1. Stop the video recorder and save to file

1. Playback video and verify video length is the same as the duration of sim time that was recorded

Ideally this configuration is specified using a proper video record msg but adding that may introduce breaking changes.


nkoenig:
Loading levels can cause simulation to pause for a bit. Will this change make a resulting video look smooth? I think the answer is yes since it's using sim time.

jennuine:
Pauses seem to be more noticeable now. See https://github.com/ignitionrobotics/ign-common/pull/105#issuecomment-710205479

chapulina:
Is this PR still under consideration?

iche033:
yes this is still needed. Waiting for another review on #419 first.

nkoenig (approved):
I'm good with this and #419 . They have been used to successfully record a few hundred videos.

src/gui/plugins/scene3d/Scene3D.cc
```diff
@@ -2281,6 +2302,24 @@ void Scene3D::LoadConfig(const tinyxml2::XMLElement *_pluginElem)
       }
     }
 
+    if (auto elem = _pluginElem->FirstChildElement("record_video"))
+    {
+      if (auto useSimTimeElem = elem->FirstChildElement("use_sim_time"))
+      {
+        std::string useSimTimeStr =
+            common::lowercase(useSimTimeElem->GetText());
```
chapulina, line 2310: How about using `QueryBoolText`?

chapulina (approved):
LGTM, I just have a suggestion for retrieving the boolean XML value.

src/gui/plugins/scene3d/Scene3D.cc
iche033: yep now using `QueryBoolText`. bb95549

iche033:
CI looks good. Merging.

PR #415, opened 2020-10-16
[Backport] Enable alpha based transparency on PBR materials by default #249

chapulina:
Backporting #249 to improve transparent materials on Citadel.

It changes the default behavior, but most likely no users are using materials with alpha yet. That's the hope :crossed_fingers:

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,6 +2,9 @@
 
 ### Ignition Gazebo 3.X.X (20XX-XX-XX)
 
+1. Enable alpha based transparency on PBR materials by default
+    * [Pull Request 249](https://github.com/ignitionrobotics/ign-gazebo/pull/249)
+
 ### Ignition Gazebo 3.4.0 (2020-10-14)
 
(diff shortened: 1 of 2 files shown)


luca-della-vedova:
Ah in hindsight this needs the `SetAlphaFromTexture` function that was [added](https://github.com/ignitionrobotics/ign-rendering/commits/main/include/ignition/rendering/Material.hh) in `ign-rendering4` for Dome so it wouldn't compile without backporting that PR as well.
Since the BaseMaterial class doesn't use PIMPL I believe the addition of the needed [member variable](https://github.com/ignitionrobotics/ign-rendering/blob/main/include/ignition/rendering/base/BaseMaterial.hh#L317) that was needed would break ABI, so it is probably not possible to backport this PR?

We could consider adding a data pointer to avoid similar issues in the future?

chapulina:
Ahh you're right :cry: That's probably why it initially targeted Dome... :confused:

PR #417, opened 2020-10-16
Combined server gui

nkoenig:
This change `ign gazebo` to run the server and GUI in the same process. The goal is to improve performance (less protobuf serialization), and help with video recording.

### DIFF
--- a/src/cmd/cmdgazebo.rb.in
+++ b/src/cmd/cmdgazebo.rb.in
@@ -448,42 +448,6 @@ class Cmd
             options['file'], options['record-topics'].join(':'),
             options['gui_config'])
 
-        # serverPid = Process.fork do
-        #   ENV['RMT_PORT'] = '1500'
-        #   Process.setpgid(0, 0)
-        #   Process.setproctitle('ign gazebo server')
-        #   Importer.runServer(parsed, options['iterations'], options['run'],
-        #     options['hz'], options['levels'], options['network_role'],
-        #     options['network_secondaries'], options['record'],
-        #     options['record-path'], options['record-resources'],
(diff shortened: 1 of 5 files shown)


jennuine (commented):
Unfortunately, I still see the same pause/jump
after_natefix

chapulina (changes requested):
I don't think we should change the default behavior on Citadel. This is a behavior change that could affect logic that's counting on having separate processes.

How about making it an option?

nkoenig:
> I don't think we should change the default behavior on Citadel. This is a behavior change that could affect logic that's counting on having separate processes.
> 
> How about making it an option?

Do you know what things are dependent on separate processes? I would like to know what all would break.

chapulina:
> Do you know what things are dependent on separate processes?

The main thing that comes to mind is the rendering singleton. The current implementation relies on having 2 separate scenes that are synced through the state topic, and they have separate rendering threads. If the render engine treats these 2 scenes as the same one, it could crash if 2 threads are writing to the scene at the same time.

Using different scenes for GUI and server means that they can be slightly different. For example, right now they can have different background colors. I don't think this is very important though. On the flip side, if we can remove the reliance on the `SceneBroadcaster` for the GUI 3D scene, that system can be significantly throttled, because widgets like the entity tree and component inspector don't need 60Hz.

I can't think of anything else right now. In any case, we never know what users may be doing in their plugins.

nkoenig:
Mentioning rendering brings to mind that this probably also breaks sensor generation.

nkoenig:
Doesn't solve the intended purpose.

PR #423, opened 2020-10-21
No install apt recommends and clear cache in Docker image

joxoby:
To reduce the size.


chapulina (approved):
Thanks! @adlarkin , something to keep in mind for #397

Issue #293, opened 2020-08-14
Log record examples not storing to /tmp/log


JShep1:
Currently, the log playback example ([here](https://github.com/ignitionrobotics/ign-gazebo/blob/ign-gazebo2/examples/worlds/log_playback.sdf)) obtains a log from `/tmp/log` which the other `log_record*.sdf` no longer store to, it seems the behavior has become deprecated as indicated [here](https://github.com/ignitionrobotics/ign-gazebo/blob/ign-gazebo2/examples/worlds/log_record_dbl_pendulum.sdf#L34-L36).

I feel like we should either update the documentation of the `log_playback.sdf` file or modify it in some way so that going forward, confusion is minimized for the user.  I'm not sure the best way given the logs from the record examples are now stored in a timestamped location so we can't specify a static location for these log files in the playback example.  Do you have any thoughts @mabelzhang ?

mabelzhang:
Maybe the way that makes the most sense is to just comment it out with a note saying this tag is valid and can be used.

I originally thought that tag isn't being used again but I checked the code and apparently it is. Looks like the tutorial documents both the `--playback <path>` option and the SDF option, so it should be okay: https://github.com/ignitionrobotics/ign-gazebo/blob/master/tutorials/log.md

mabelzhang:
I looked again, and I think we should just remove `log_playback.sdf`.

The reason is that `<path>` is actually injected by `Server.cc` here https://github.com/ignitionrobotics/ign-gazebo/blob/ign-gazebo2/src/Server.cc#L87-L90
Its value comes from `_config.LogPlaybackPath()`, which is the value specified on the command line with `--playback`. So there's really no need for the user to create an SDF just for playback.

Specifying the path like in `log_playback.sdf` still works, but to be parallel with recording, which deprecated `<path>`, I wonder if we should also deprecate it in the playback. In that case, we should remove the playback SDF example in the log tutorial too.

In the state the file is right now, if we just comment out `<path>`, running the file will result in an error:
```
[Err] [LogPlayback.cc:247] Unspecified log path to playback. Nothing to play.
```
If we leave in the `<path>`, running the file also results in an error:
```
[Err] [LogPlayback.cc:256] Log path invalid. File [/tmp/log/state.tlog] does not exist. Nothing to play.
```
So I don't see a reasonable way to keep the file. The only thing in that file is really those few lines that `Server.cc` populates.

JShep1:
[...]

PR #424, opened 2020-10-21
[Citadel] Document deprecation of playback <path> SDF param

mabelzhang:
Documentation to address #293 for Citadel.

Discussed with @chapulina and we decided to do the following:
Blueprint & Citadel:
- Keep `log_playback.sdf` since the distros have already been released and the file installed.
- Update tutorial to say `<path>` is not recommended.
- Update header comment in example file. I thought it made more sense to update `log_playback.sdf` rather than `log_record_*.sdf`, because it's really the playback file that's being replaced, and they just have to use `--playback [arg]`. Nothing changes for recording.
- Add a note in `Migration.md`

Essentially deprecate `<path>` for Blueprint and Citadel, remove it in Dome.

Will make followup PR for Blueprint.
Dome will be in a separate PR with different items - `log_playback.sdf` has already been removed in Dome, documentation needs to be updated, and some test code needs to be changed, so it'll take a bit longer.

### DIFF
--- a/Migration.md
+++ b/Migration.md
@@ -36,6 +36,9 @@ Gazebo 2+ for playback. [BitBucket pull request
 #257](https://osrf-migration.github.io/ignition-gh-pages/#!/ignitionrobotics/ign-gazebo/pull-requests/257)
 added an SDF message to the start of log files.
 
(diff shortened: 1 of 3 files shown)


chapulina (approved):
Thanks!

PR #426, opened 2020-10-21
View angle icons higher contrast

herobank110:


chapulina (approved):
This looks so much better! Thank you for the contribution!

va_light

va_dark

PR #432, opened 2020-10-29
Remove unneeded if statement

JShep1:
Since the targeted branch now iterates through the set of component types parameter, I don't foresee a case in which an irrelevant entity would be added to the message.

Does anyone see any potential problems with this?


mjcarroll:
Is 755b88d the only commit to review here?

iche033 (approved):
I added a print statement inside the if statement and tested moving objects and also log playback of different worlds but it never gets here. I think it maybe ok to remove. So looks good to me.

looking at the history of commits, it suggests that @nkoenig added this logic. Maybe it would be good to get some confirmation from him or another reviewer to double check if it's safe to remove.

JShep1:
> I added a print statement inside the if statement and tested moving objects and also log playback of different worlds but it never gets here. I think it maybe ok to remove. So looks good to me.
> 
> looking at the history of commits, it suggests that @nkoenig added this logic. Maybe it would be good to get some confirmation from him or another reviewer to double check if it's safe to remove.

Sounds good.  I believe the only case that this if statement would occur would be when the passed in message is already populated with data.  In all cases that this function is used, an empty message is passed into the function call, so this if statement isn't triggered.  This isn't public facing API so I would suppose if we aren't going to be passing in a previously populated message, we can move ahead with deleting that line of code.  I'll leave the decision up to @nkoenig .  Although the check theoretically doesn't take long, it can be called thousands of times every iteration (as a function of the number of entities) and since I'm aiming for optimizing this piece of code, every nanosecond counts.

nkoenig (approved):

PR #433, opened 2020-10-29
Added tests for the AnimationTime component

adlarkin:
I added missing tests for a `std::chrono::steady_clock::duration` component after doing something similar in #401.

### DIFF
--- a/include/ignition/gazebo/components/Actor.hh
+++ b/include/ignition/gazebo/components/Actor.hh
@@ -20,6 +20,7 @@
 #include <ignition/msgs/actor.pb.h>
 
 #include <chrono>
+#include <ratio>
 #include <string>
 
 #include <sdf/Actor.hh>
@@ -50,7 +51,7 @@ namespace serializers
     public: static std::ostream &Serialize(std::ostream &_out,
                 const std::chrono::steady_clock::duration &_time)
     {
-      _out << std::chrono::duration_cast<std::chrono::milliseconds>(
+      _out << std::chrono::duration_cast<std::chrono::nanoseconds>(
           _time).count();
       return _out;
     }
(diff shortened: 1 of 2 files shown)


adlarkin:
A few other things to consider:

1. Should we use `uint64_t` instead of `int64_t` for de-serialization, or can [std::chrono::duration::count()](https://en.cppreference.com/w/cpp/chrono/duration/count) return a negative value if a jump back in time occurs?

https://github.com/ignitionrobotics/ign-gazebo/blob/aea1accff6d8daee0358ae9025e2a9797d292844/include/ignition/gazebo/components/Actor.hh#L58-L71

2. The serialization/de-serialization are only as accurate as milliseconds, since that is what we are currently casting the durations to. If we want to serialize time that is more precise than milliseconds, then the information that exceeds millisecond precision will be lost. Should we use a more precise duration cast? For example, something like `std::chrono::nanoseconds`?

chapulina:
> can std::chrono::duration::count() return a negative value if a jump back in time occurs?

The specific duration type we're using is `std::chrono::steady_clock::duration`, which is equivalent to `std::chrono::nanoseconds`, which under the hood is `std::chrono::duration<int64_t, std::nano>`. Note the `int64_t` there. So it can be negative.

You're right that we use negative durations for jumps back in time. For example, `dt` will be negative when a log playback is rewinded:

https://github.com/ignitionrobotics/ign-gazebo/blob/da2ddbc14bc487d53e49de5b9fdea3272797698c/src/systems/log/LogPlayback.cc#L484-L486

Now, will it ever be negative for `AnimationTime`? I don't think it should. But I was trying to be general when implementing the serialization.

> Should we use a more precise duration cast?

Yeah I think that's a good idea. For the animation it probably isn't necessary, but I think it's good to be general.

adlarkin:
I went ahead and modified the serialization to use nanoseconds instead of milliseconds in [02ad2db](https://github.com/ignitionrobotics/ign-gazebo/pull/433/commits/02ad2db75ee0b6600d7d88c436c156e24b91e8e9). This should be ready for another review now.

chapulina (approved):
Nice, thanks for the test!

Issue #84, opened 2020-04-21
Add helper API to encapsulate the check/create step for components


gerkey:
In the course of migrating a plugin from Gazebo to Ignition (#80), I ended up following a pattern like this when accessing components:
```
if(!_ecm.EntityHasComponentType(this->dataPtr->modelLink, components::WorldPose::typeId))
{
  _ecm.CreateComponent(this->dataPtr->modelLink, ignition::gazebo::components::WorldPose());
}
```
That is, do some sort of lookup and if it fails, then create the component, and subsequently read from or write to it.

This pattern seems common enough that it would be worthwhile providing helper API to encapsulate it into a single `PleaseGiveMeThisComponentButCreateItFirstIfYouNeedTo()` call.

On a related topic, and also related to defining each system's component API (#83), it would be useful to know which components, if any, are guaranteed to exist. I found that sometimes I needed to do the check-create pattern and sometimes I could just use a component blindly without checking its validity. If that's just bad practice, and the right thing to do is to always check for existence of a component before accessing it, then we should say so in the docs (and also clean up internal component handling, e.g., the [IMU system accesses the `Name` component without checking that it's valid](https://github.com/ignitionrobotics/ign-gazebo/blob/master/src/systems/imu/Imu.cc#L161-L162))

peci1:
The example Brian showed gets even more obscure when you need to set some value to the component:

https://github.com/ignitionrobotics/ign-gazebo/blob/9fb057712d5eb1c4965b2755bfb17e7ce6234fe3/src/systems/joint_position_controller/JointPositionController.cc#L231-L241

From the developer's point of view, this is nightmare...

Compare with Gazebo:

```c++
this->dataPtr->joints[iter->first]->SetForce(0, iter->second);
```

I think this design pattern (or rather antipattern?) really needs some simplification. The earlier it gets implemented, the less work will be spent on rewriting other parts of the codebase :) Let's make it a challenge for Edifice? :)

One good solution would be to provide wrappers like https://github.com/ignitionrobotics/ign-gazebo/blob/ign-gazebo4/src/Link.cc for most of the relevant entity types. I always wondered why there is Link.cc and not Joint.cc ...

chapulina:
> One good solution would be to provide wrappers 

That's in the works, see #375 #378 #325

> Let's make it a challenge for Edifice?

The goal is to add these from Citadel onwards :wink:

peci1:
Great to hear that!

[...]

PR #436, opened 2020-10-31
Helper function to set component data

chapulina:
Addresses #84 .

Adds the `SetComponentData` function, which simplifies the code necessary to set data to a component that may not be created yet.

It simplifies code like this:

```
auto poseComp = _ecm.Component<components::Pose>(entity); 
if (nullptr == poseComp) 
{ 
  _ecm.CreateComponent(entity, components::Pose({force})); 
} 
else 
{ 
  poseComp->SetData(pose,
      [](const math::Pose3d &_a, const math::Pose3d &_b)
      {
        return _a == _b;
      };);
} 
```

Into:

```
ecm.SetComponentData<components::Pose>(entity, pose);
```

### DIFF
--- a/include/ignition/gazebo/detail/EntityComponentManager.hh
+++ b/include/ignition/gazebo/detail/EntityComponentManager.hh
@@ -61,7 +61,7 @@ namespace traits
 /// false.
 /// For doubles, `ignition::math::equal` will be used.
 template<typename DataType>
-auto CompareData = [&](const DataType &_a, const DataType &_b) -> bool
+auto CompareData = [](const DataType &_a, const DataType &_b) -> bool
 {
   if constexpr (std::is_same<DataType, double>::value)
   {
--- a/include/ignition/gazebo/detail/EntityComponentManager.hh
+++ b/include/ignition/gazebo/detail/EntityComponentManager.hh
@@ -58,7 +58,7 @@ namespace traits
 /// \brief Helper function to compare two objects of the same type using its
 /// equality operator.
 /// If `DataType` doesn't have an equality operator defined, it will return
-/// true.
+/// false.
 /// For doubles, `ignition::math::equal` will be used.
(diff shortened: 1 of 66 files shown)


include/ignition/gazebo/detail/EntityComponentManager.hh
```diff
@@ -20,15 +20,61 @@
...
+  if constexpr (std::is_same<DataType, double>::value)
+  {
+    return math::equal(_a, _b);
+  }
+  else if constexpr (traits::HasEqualityOperator<DataType>::value)
+  {
+    return _a == _b;
+  }
+
+  return false;
```
peci1, line 75: Doc-comment says it returns true...

chapulina: oops, fixed: cb3e63a229666f88dd433afec03c6f6a49856b02

include/ignition/gazebo/detail/EntityComponentManager.hh
```diff
@@ -20,15 +20,61 @@
...
+  template<typename T>
+  TestEqualityOperator operator == (const T&, const T&);
+
+  /// \brief Type trait that determines if an operator== is defined for `T`.
+  template<typename T>
+  struct HasEqualityOperator
+  {
+    enum
+    {
+      value = !std::is_same<decltype(*(T*)(0) == *(T*)(0)),
```
mjcarroll, line 51: Codecheck doesn't like this because it is a C-style cast.  I don't see an obvious way around it, so I think a `// NOLINT` would be sufficient.

mjcarroll (approved):
This looks good, but will need to pass CI, it isn't currently getting past codecheck.

diegoferigo:
xref https://github.com/ignitionrobotics/ign-gazebo/issues/325

PR #446, opened 2020-11-04
Fix shadow artifacts by disabling double sided rendering

iche033:
Depends on osrf/sdformat#410

Double sided material was enabled by default  in ign-gazebo 4.0.0 through the `SetAlphaFromTexture()` function call, which was originally added to support textures with alpha transparency. However, double sided rendering caused self-shadowing artifacts to appear on some meshes. This PR now makes double sided material an opt-in feature insead. It reads this property from the Material SDF DOM, which defaults to `false`. 

 The screenshots below are from running the `fuel_textured_material.sdf` world and zooming into upper body of the rescue randy mesh. The left image is before the change and shows subtle shadow artifacts lines across the mesh and the right image shows the result with double sided material disabled.

double_sided


chapulina (approved):
Great improvement!

luca-della-vedova:
Just one note, since there is support in sdformat for the double sided rendering tag now, we should update the assets that need it to render correctly in fuel, otherwise they would be rendered incorrectly after this PR is merged:
https://app.ignitionrobotics.org/OpenRobotics/fuel/models/Pine%20Tree
https://app.ignitionrobotics.org/OpenRobotics/fuel/models/Oak%20tree

iche033:
I've updated the two tree models on Fuel. 

While testing, I found out that I missed the msg <-> sdf conversion part. That's added in 7f528184a95fac8b106a25ae3769050e96dd30b9 and `<double_sided>` should work properly now. One caveat is that I had to add the `double_sided` field to the `header` of the `msgs::Material` msg because adding extra proto field would break ABI.

luca-della-vedova (approved):

iche033:

iche033:
CI looks good. Merging

Issue #113, opened 2020-04-29
Unload model plugins when the model is removed from the world


diegoferigo:
As discussed in https://github.com/ignitionrobotics/ign-gazebo/issues/39#issuecomment-613957221, model plugins are not unloaded when the model is removed from the world.

There are few problems that this behaviour could cause:

1. Memory leak. If the simulation is very long and models are inserted and deleted multiple times, the RAM consumption slowly increases. (e.g. https://github.com/ignitionrobotics/ign-gazebo/issues/47)
1. If the plugin access ECM components of the model and it does not handle properly the case when they are not found (because the model has been removed), the behaviour could be undefined.

In #39 the problem was acknowledged but no specific issue was opened to track it.

ejalaa12:
Hi, 
I can confirm that this problem is still present in gazebo Garden.
It's easily reproducable with an Hydrodynamics system which prints ["No linear vel"](https://github.com/gazebosim/gz-sim/blob/gz-sim7/src/systems/hydrodynamics/Hydrodynamics.cc#L446-L450) once its model has been removed from the world. Which means the system is still running, but the entity is not present anymore.

takahashi-e6:
Hi.

This problem seems to occur in Fortress as well, and can be reproduced by simply deleting the plugin-enabled model with the following command:

ign service -s /world/empty/remove --reqtype ignition.msgs.Entity --reptype ignition.msgs.Boolean --timeout 5000 --req 'type: MODEL; name: "my_model"'

How can I properly delete a model?

I looked up the ign command but couldn't find the answer.

MSZ98:
Hi,

In the `gazebo_ros_pkgs` -> `gazebo_ros_openni_kinect.cpp` -> `Load()` method, the `GazeboRosCameraUtils` plugin is loaded and stays active even after the model is removed from the simulation. When you try to add a model with a camera to the simulation again using the same namespace, `gazebo_ros_openni_kinect.cpp` attempts to load the `GazeboRosCameraUtils` plugin once more. However, an error occurs because the first `GazeboRosCameraUtils` plugin already created a dynamic reconfigure ROS service, and there’s an attempt to advertise the same service again.

I could use a different namespace for the new model, but this solution feels messy, and the old camera thread could still impact performance.

PR #454, opened 2020-11-10
Fix segfault in the Breadcrumb system when associated model is unloaded

azeey:
Currently, when a model is unloaded, the model systems associated with that model do not get unloaded. If the systems then try to access the components of the model, they will fail. However, most systems check the validity of the model they are associated with in the `System::Configure` function and assume that if the model is valid then, it will be valid in `PreUpdate/PostUpdate` functions. I think it would be nice for that assumption to hold, but right now it might not.

In the case of the Breadcrumb system, the system retrieves the pose of the model at every time step in `PreUpdate` and assumes that it will be valid. However, if the model is unloaded, the `Pose` component will be a `nullptr` causing a segfault.

The ideal fix would be to make sure systems are unloaded when the model gets unloaded (see #113), but until then, this PR adds a check for the validity of the model in `Breadcrumbs::PreUpdate`.

### DIFF
--- a/src/systems/breadcrumbs/Breadcrumbs.cc
+++ b/src/systems/breadcrumbs/Breadcrumbs.cc
@@ -71,6 +71,12 @@ void Breadcrumbs::Configure(const Entity &_entity,
       _sdf->Get<bool>("allow_renaming", this->allowRenaming).first;
 
   this->model = Model(_entity);
+  if (!this->model.Valid(_ecm))
+  {
+    ignerr << "The Breadcrumbs system should be attached to a model entity. "
+           << "Failed to initialize." << std::endl;
+    return;
+  }
 
   if (!_sdf->HasElement("breadcrumb"))
   {
@@ -162,7 +168,7 @@ void Breadcrumbs::PreUpdate(const ignition::gazebo::UpdateInfo &_info,
     // unloaded by the level manager. Ideally, this system would have been
     // unloaded along with the model, but that is not currently the case. See
     // issue #113
-    if (!model.Valid(_ecm))
+    if (!this->model.Valid(_ecm))
     {
       return;
     }
--- a/src/systems/breadcrumbs/Breadcrumbs.cc
+++ b/src/systems/breadcrumbs/Breadcrumbs.cc
@@ -157,6 +157,15 @@ void Breadcrumbs::PreUpdate(const ignition::gazebo::UpdateInfo &_info,
                 std::back_inserter(cmds));
       this->pendingCmds.clear();
     }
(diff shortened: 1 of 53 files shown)


azeey:
Failing tests are fixed by https://github.com/ignitionrobotics/ign-gazebo/pull/455

test/integration/breadcrumbs.cc
```diff
@@ -547,6 +550,123 @@ TEST_F(BreadcrumbsTest, AllowRenaming)
   EXPECT_TRUE(this->server->HasEntity("B1_0_1"));
 }
 
+/////////////////////////////////////////////////
+/// Return a list of model entities whose names match the given regex
```
chapulina, line 554: This could be eventually useful to move to `EntityComponentManager`. The trick is that it's probably too slow, and users may not realize that.

src/systems/breadcrumbs/Breadcrumbs.cc
```diff
@@ -160,6 +160,15 @@ void Breadcrumbs::PreUpdate(const ignition::gazebo::UpdateInfo &_info,
...
       this->pendingCmds.clear();
     }
+    // Check that the model is valid before continuing. This check is needed
+    // because the model associated with the Breadcrumbs might have been
+    // unloaded by the level manager. Ideally, this system would have been
+    // unloaded along with the model, but that is not currently the case. See
+    // issue #113
+    if (!model.Valid(_ecm))
+    {
+      return;
```
chapulina, line 170: Do we want to print a warning in case `cmds` is not empty?

chapulina (approved):

test/integration/breadcrumbs.cc
azeey: Yeah, we can probably make it generic enough that it would work with any entity that has a name component.

src/systems/breadcrumbs/Breadcrumbs.cc
azeey: I added a check in `Configure` in https://github.com/ignitionrobotics/ign-gazebo/pull/454/commits/9fa5c76c742c7890745bbbaee2c58c5a46773d15 that prints an error if the model is not valid. I'm not sure adding a warning here is useful because the user can't do anything about it. It is a valid use case to send the deploy command if the model is loaded back by the level manager, in which case both the original system and the new system will get the command, but the original system will silently return early here.

chapulina: Oh I see, yeah agreed that a warning would be confusing if there's a valid use case. I guess we can't tell apart when the model was removed by the `LevelManager` or by some other mechanism.

PR #456, opened 2020-11-11
Tape Measure Plugin

JShep1:
An initial prototype of the tape measure plugin.

Hotkey for new measurement: `m`

See also https://github.com/ignitionrobotics/ign-gazebo/pull/458 for right click mid-measure-action cancellation.

I made a few decisions here that I think we can discuss and possibly change:

- I added gui events for a mouse hover and the left click event in `Scene3D`.  My reasoning was primarily that I needed to get the data from the `ScreenToScene()` call within `Scene3D` but didn't want to couple this plugin together with `Scene3D` itself.  I'm not sure if this is the best route to go, but it was the best I could think of for now, are there any other better ideas as to how to do this?  I was thinking we could eventually move the gui events header to `ign-common` so that `ign-gui` can contain plugins like this that don't depend on the ecm but still need data from the `Scene3D`  would this be a reasonable path?  Any foreseen issues?
- I only added singular measurements for now.  There are a couple main reasons.  The first is that text markers are only currently supported in OGRE 1.x, if we had multiple markers in a OGRE 2 scene, the user would have no way of knowing what the distance would be for each measurement.  For this reason, I simply put the distance for the measurement within the gui window.  The second reason is that we have no way of being able to selectively remove specific measurements if multiple existed within the scene.  We will need to add support for user interaction with markers for this to be possible.

~Additionally, @ColeOSRF is working on new icons for the measurement button and the delete button.~

Example:
tapemeasure

Depends on https://github.com/ignitionrobotics/ign-gui/pull/148 for gui events


JShep1:
Design doc for this plugin can be found here https://github.com/ignitionrobotics/design/blob/master/tape_measure_tool.md

src/gui/plugins/tape_measure/TapeMeasure.qml
```diff
@@ -0,0 +1,116 @@
...
+        source: "trashcan.png"
+        sourceSize.width: 36;
+        sourceSize.height: 36;
+      }
+      onClicked: {
+        TapeMeasure.OnReset();
+      }
+    }
+    Text {
+      text: qsTr(" Distance (m): " + distance.toFixed(3))
```
chapulina, line 113: The color of this text is not respecting the light / dark theme.

Also, maybe this is just me, but I think this text could be a bit bigger.

src/gui/plugins/scene3d/Scene3D.cc
```diff
@@ -767,12 +767,44 @@ Entity IgnRenderer::UniqueId()
...
   this->HandleMouseTransformControl();
   this->HandleMouseViewControl();
 }
 
+/////////////////////////////////////////////////
+void IgnRenderer::BroadcastHoverPos()
+{
+  if (this->dataPtr->hoverDirty)
+  {
+    math::Vector3d pos = this->ScreenToScene(this->dataPtr->mouseHoverPos);
```
chapulina, line 783: This is not an issue introduced by this PR, but it looks like the `ScreenToScene` function doesn't handle planes well.

With the `rolling_shapes.sdf` world for example, the points are below the plane:

rolling_tape

And even flat planes don't behave well from below. I think that's because the triangles are facing the other way, so maybe this is expected:

shapes_tape

src/gui/plugins/tape_measure/TapeMeasure.cc
```diff
@@ -0,0 +1,282 @@
...
+  class TapeMeasurePrivate
+  {
+    /// \brief Ignition communication node.
+    public: transport::Node node;
+
+    /// \brief True if currently measure, else false.
+    public: bool measure = false;
+
+    /// \brief The id of the start point marker.
+    public: int startPointId = 1;
```
chapulina, line 42: How about making these IDs const? We also try to use the `k` prefix for const variables.

src/gui/plugins/tape_measure/TapeMeasure.cc
```diff
@@ -0,0 +1,282 @@
...
+{
+  this->DeleteMarker(_id);
+
+  ignition::msgs::Marker markerMsg;
+  markerMsg.set_ns(this->dataPtr->ns);
+  markerMsg.set_id(_id);
+  markerMsg.set_action(ignition::msgs::Marker::ADD_MODIFY);
+  markerMsg.set_type(ignition::msgs::Marker::SPHERE);
+  ignition::msgs::Set(markerMsg.mutable_scale(),
+    ignition::math::Vector3d(0.1, 0.1, 0.1));
```
chapulina, line 174: I know this will probably be complicated to achieve with markers, but I'll just make an observation for the future.

The fixed scale means that the points may be too big or too small depending on what's being simulated and the user's zoom level. Maybe they're too big to measure the distance between fingers of a robot hand, but too small when measuring the size of a road. One option would be to scale it according to the zoom level like it's done for the transform controls:

scale_tape

src/gui/plugins/tape_measure/TapeMeasure.cc
```diff
@@ -0,0 +1,282 @@
...
+  markerMsg.set_action(ignition::msgs::Marker::ADD_MODIFY);
+  markerMsg.set_type(ignition::msgs::Marker::LINE_LIST);
+  markerMsg.mutable_material()->mutable_ambient()->set_r(_color[0]);
+  markerMsg.mutable_material()->mutable_ambient()->set_g(_color[1]);
+  markerMsg.mutable_material()->mutable_ambient()->set_b(_color[2]);
+  markerMsg.mutable_material()->mutable_ambient()->set_a(_color[3]);
+  markerMsg.mutable_material()->mutable_diffuse()->set_r(_color[0]);
+  markerMsg.mutable_material()->mutable_diffuse()->set_g(_color[1]);
+  markerMsg.mutable_material()->mutable_diffuse()->set_b(_color[2]);
+  markerMsg.mutable_material()->mutable_diffuse()->set_a(_color[3]);
```
chapulina, line 208: How about using `ignition::math::Color` for all the colors?

Then you can also use [ignition::msgs::Set](https://github.com/ignitionrobotics/ign-msgs/blob/1c3dfd91463090f7503d710176851c3c722977b5/include/ignition/msgs/Utility.hh#L337-L341) to set the message fields from the colors.

src/gui/plugins/tape_measure/TapeMeasure.qml
```diff
@@ -0,0 +1,116 @@
...
+  Layout.minimumWidth: 200
+  Layout.minimumHeight: 100
+
+  property color snapTitle: (Material.theme == Material.Light) ?
+    Material.color(Material.Grey, Material.Shade900) :
+    Material.color(Material.Grey, Material.Shade200)
+
+  property color snapItem: (Material.theme == Material.Light) ?
+    Material.color(Material.Grey, Material.Shade800) :
+    Material.color(Material.Grey, Material.Shade100)
```
chapulina, line 37: These properties aren't used

src/gui/plugins/tape_measure/TapeMeasure.qml
```diff
@@ -0,0 +1,116 @@
...
+  background: Rectangle {
+    color: "transparent"
+  }
+
+  ButtonGroup {
+    id: group
+  }
+
+  Shortcut {
+    sequence: "M"
```
chapulina, line 66: Can we also use `Esc` to cancel a measurement?

src/gui/plugins/tape_measure/TapeMeasure.qml
```diff
@@ -0,0 +1,116 @@
...
+import QtQuick.Controls 2.1
+import QtQuick.Controls.Material 2.2
+import QtQuick.Controls.Material.impl 2.2
+import QtQuick.Layouts 1.3
+import QtQuick.Controls.Styles 1.4
+import "qrc:/qml"
+
+ToolBar {
+  id: tapeMeasure
+  Layout.minimumWidth: 200
```
chapulina, line 28: It needs to be a little bit bigger to fit the distance completely

chapulina (commented):
Works well for me.

I don't think that this can be avoided if we use the `MarkerManager`, but a user could send a `DELETE_ALL` at any time and mess up the tape measure's markers. I think the least we can do is be prepared in case this happens, by never assuming the markers we placed will be there.

I tried calling the following service while using the tape measure, and the worst I could get is that one of the end markers disappears, but not both. You can also end up with no markers, but the distance is non-zero. I think these aren't big issues, but just something to keep in mind.

`ign service -s /marker --reqtype ignition.msgs.Marker --reptype ignition.msgs.Empty --timeout 2000 --req "action: DELETE_ALL"`

src/gui/plugins/tape_measure/TapeMeasure.cc
```diff
@@ -0,0 +1,282 @@
...
+#include <ignition/common/Console.hh>
+#include <ignition/gui/Application.hh>
+#include <ignition/gui/MainWindow.hh>
+#include <ignition/plugin/Register.hh>
+#include <ignition/transport/Node.hh>
+#include <ignition/transport/Publisher.hh>
+
+#include "ignition/gazebo/gui/GuiEvents.hh"
+
+#include "TapeMeasure.hh"
```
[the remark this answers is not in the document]
JShep1: What version of `codecheck` detects this?  I'm not seeing this issue locally.  I moved it up but am unsure exactly where to put it.

src/gui/plugins/scene3d/Scene3D.cc
JShep1: Agh yeah, I believe there's some logic in the `ScreenToScene` function that assumes a flat plane.  I ticketed an issue for it [here](https://github.com/ignitionrobotics/ign-gazebo/issues/464).  I think the plane from below case is as expected.  Trying to click a back facing polygon may introduce weird behavior especially in a case like a SubT cave world where the user may be selected the ground in a tunnel from outside the tunnel.

src/gui/plugins/tape_measure/TapeMeasure.cc
JShep1: Good point, I believe that functionality is coming from [this line](https://github.com/ignitionrobotics/ign-gazebo/blob/ign-gazebo3/src/gui/plugins/scene3d/Scene3D.cc#L1415)?  After thinking about it for a second, I think the effect comes from the world "moving away" and the visual gizmo staying the same size.  @ColeOSRF do you know how this tool is usually  handled in large vs small scenarios and when the user is zooming in and out?

src/gui/plugins/tape_measure/TapeMeasure.qml
JShep1: I'll create another PR since Shortcuts are unique across an application for each key event and the `Transform Control` already makes use of `Esc`

src/gui/plugins/tape_measure/TapeMeasure.cc
JShep1: dda9038aef5fc0e953aadaea329fca4f37d20695

src/gui/plugins/tape_measure/TapeMeasure.qml
JShep1: dda9038aef5fc0e953aadaea329fca4f37d20695

src/gui/plugins/tape_measure/TapeMeasure.qml
JShep1: dda9038aef5fc0e953aadaea329fca4f37d20695

src/gui/plugins/tape_measure/TapeMeasure.qml
JShep1: 4d41ebe04525263d82069ea2d1d3f31c66711731

src/gui/plugins/tape_measure/TapeMeasure.cc
JShep1: 4d41ebe04525263d82069ea2d1d3f31c66711731

src/gui/plugins/tape_measure/TapeMeasure.qml
JShep1: Added the logic for escape and `m` key press to the cpp side and removed the `Shortcut`  65d122ab3dec6a2f1b23c5c8068501c3eb277129

chapulina:
Just dropping a cross-reference link here. This requires: https://github.com/ignitionrobotics/ign-gui/pull/148

chapulina (approved):
Works for me! Before merging this, we need to:

- [x] Merge https://github.com/ignitionrobotics/ign-gui/pull/148
- [ ] Release `ign-gui` with :point_up: 
- [ ] Update the minimum required `ign-gui` version here

chapulina (approved):
:rocket:

PR #443, opened 2020-11-04
Resolve updated codecheck issues

adlarkin:
This PR is meant to resolve new codecheck issues introduced by https://github.com/ignitionrobotics/ign-cmake/pull/117.

From my understanding, the plan is to forward port the fixes here to all other ignition versions once https://github.com/ignitionrobotics/ign-cmake/pull/117 is merged.

Signed-off-by: Ashton Larkin <ashton@openrobotics.org>

PR #457, opened 2020-11-11
2 ➡️ 3

chapulina:
Forward-porting `ign-gazebo2` to `ign-gazebo3`.

The only extra change I needed to do for the port to compile was moving several `ign-physics` includes from `Physics.cc` to `Physics.hh` to compensate for the header inclusion moved up in #443 . The `entityCast` function needs the features defined before it. CC @adlarkin

### DIFF
--- a/src/LevelManager.cc
+++ b/src/LevelManager.cc
@@ -15,6 +15,10 @@
  *
  */
 
+#include "LevelManager.hh"
+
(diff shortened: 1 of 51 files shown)


adlarkin (approved):
Thanks for fixing the feature includes in `Physics.hh`. Looks good to me!

PR #441, opened 2020-11-03
3 ➡️ 4

chapulina:
Forward-porting `ign-gazebo3` to `ign-gazebo4`.

PR #459, opened 2020-11-13
3 to 4

adlarkin:
Forward port `ign-gazebo3` to `ign-gazebo4`.

One thing to note about this forward port is that `examples/worlds/log_playback.sdf` exists in `ign-gazebo3`, but not `ign-gazebo4`. #386 details why this file was removed. Also, is the `examples/worlds/log_playback.sdf` file in `ign-gazebo3` correct, or does that also need to be updated?

### DIFF
--- a/tutorials/log.md
+++ b/tutorials/log.md
@@ -99,6 +99,9 @@ directory specified to record:
 
 ### From plugin in SDF
 
+This feature is deprecated and will be removed in Ignition Dome.
+Use the command line argument instead.
+
 Alternatively, playback can be specified in an SDF file. See example file
 `examples/worlds/log_playback.sdf`:
 
--- a/tutorials/log.md
+++ b/tutorials/log.md
@@ -96,35 +96,6 @@ directory specified to record:
 
 `ign gazebo -r -v 4 --playback <path>`
 
-
-### From plugin in SDF
-
-This feature is deprecated and will be removed in Ignition Dome.
-Use the command line argument instead.
-
-Alternatively, playback can be specified in an SDF file. See example file
-`examples/worlds/log_playback.sdf`:
-
-```{.xml}
-<world name="default">
-    ...
-    <plugin
-      filename='ignition-gazebo-log-system'
-      name='ignition::gazebo::systems::LogPlayback'>
-      <path>/tmp/log</path>
-    </plugin>
-    ...
-</world>
-```
-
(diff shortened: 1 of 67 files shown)


tutorials/log.md
```diff
@@ -99,6 +99,9 @@ directory specified to record:
 
 ### From plugin in SDF
 
+This feature is deprecated and will be removed in Ignition Dome.
+Use the command line argument instead.
+
```
chapulina, line 104: I think we should leave this out of the forward-port. @mabelzhang is looking into the documentation updates specific to Dome, see #293.

chapulina (commented):
Just a minor comment. Since `ign-rendering` 4.1.0 is out, and this PR depends on it (see conversation on #441 ), do you mind bumping the version requirement here:

https://github.com/ignitionrobotics/ign-gazebo/blob/020a0172308eb5b64e4699b5afb25bb1105800ba/CMakeLists.txt#L117

tutorials/log.md
mabelzhang: I need to continue that task. I'll make a PR to Dome directly when I find out what's up with those tests.

adlarkin: This relates to the comment I left in the description of this PR:

> One thing to note about this forward port is that `examples/worlds/log_playback.sdf` exists in `ign-gazebo3`, but not `ign-gazebo4`. #386 details why this file was removed. Also, is the `examples/worlds/log_playback.sdf` file in `ign-gazebo3` correct, or does that also need to be updated?

Do you guys know the answer to the question about `ign-gazebo3` in that comment?

Also, just to be clear, should I go ahead and remove the lines @chapulina mentioned for this forward port and let @mabelzhang address this when she makes a PR to Dome?

mabelzhang: That SDF file should exist in ign-gazebo3 (with a deprecation note), and should not exist in ign-gazebo4, in other words, leaving it in its current state in both. The lines in the tutorial can be removed. I have a different line for Dome in a branch somewhere. Thanks!

adlarkin: I went ahead and made the change in https://github.com/ignitionrobotics/ign-gazebo/pull/459/commits/08f1a0f646b8b492a3a5bc1cc67136fea0b3b75f, can you guys give it one last look?

chapulina (approved):

PR #461, opened 2020-11-17
Switch to async state service request

iche033:
The scene broadcsater system provides a synchronous state service which blocks until the state is filled in the next iteration. In normal ign gazebo usage this works fine but sometimes causes deadlock when running the [distributed_levels](https://github.com/ignitionrobotics/ign-gazebo/tree/ign-gazebo4/examples/scripts/distributed_levels) demo due to a race condition. This PR adds a new non-blocking async state service in the scene broadcaster system, and switches the GUI to call this async state service instead.

More info:

The NetworkManager classes in distributed simulation rely heavily on ign-transport for synchronizing all secondaries with the primary. The deadlock situation happens when the state service call is invoked while the secondaries are being stepped at the same time by the primary:

1. Primary sends msgs to step all Secondaries
1. A state service request is received by the scene broadcaster system in Primary and blocks until the state service is filled in the next iteration
1. step ack msgs sent back by the Secondaries are never received by the Primary because the ign-transport thread is being blocked.
1. Primary cannot step the simulation without ack msg, which means the scene broadcaster can not fill the state required by the state service call
1. Deadlock! 

Another option is to update the distributed system stepping synchronization strategy - I started with this fix and have a workaround implemented but ultimately I came back to the async fix because I think that it would be good to avoid a blocking service call in general. 

### DIFF
--- a/src/gui/GuiRunner.cc
+++ b/src/gui/GuiRunner.cc
@@ -90,6 +90,13 @@ void GuiRunner::OnStateAsyncService(const msgs::SerializedStepMap &_res)
 {
   this->OnState(_res);
 
+  // todo(anyone) store reqSrv string in a member variable and use it here
+  // and in RequestState()
+  std::string id = std::to_string(gui::App()->applicationPid());
+  std::string reqSrv =
+      this->node.Options().NameSpace() + "/" + id + "/state_async";
+  this->node.UnadvertiseSrv(reqSrv);
+
   // Only subscribe to periodic updates after receiving initial state
   if (this->node.SubscribedTopics().empty())
     this->node.Subscribe(this->stateTopic, &GuiRunner::OnState, this);
--- a/src/gui/GuiRunner.cc
+++ b/src/gui/GuiRunner.cc
@@ -59,7 +59,16 @@ GuiRunner::~GuiRunner() = default;
 /////////////////////////////////////////////////
 void GuiRunner::RequestState()
 {
-  this->node.Request(this->stateTopic, &GuiRunner::OnStateService, this);
+  // set up service for async state response callback
+  std::string id = std::to_string(gui::App()->applicationPid());
+  std::string reqSrv =
+      this->node.Options().NameSpace() + "/" + id + "/state_async";
+  this->node.Advertise(reqSrv, &GuiRunner::OnStateAsyncService, this);
+  ignition::msgs::StringMsg req;
+  req.set_data(reqSrv);
+
+  // send async state request
+  this->node.Request(this->stateTopic + "_async", req);
 }
 
 /////////////////////////////////////////////////
@@ -77,15 +86,8 @@ void GuiRunner::OnPluginAdded(const QString &_objectName)
 }
 
 /////////////////////////////////////////////////
-void GuiRunner::OnStateService(const msgs::SerializedStepMap &_res,
-    const bool _result)
+void GuiRunner::OnStateAsyncService(const msgs::SerializedStepMap &_res)
 {
-  if (!_result)
-  {
-    ignerr << "Service call failed for [" << this->stateTopic << "]"
-           << std::endl;
-    return;
-  }
   this->OnState(_res);
 
   // Only subscribe to periodic updates after receiving initial state
--- a/src/systems/scene_broadcaster/SceneBroadcaster.cc
+++ b/src/systems/scene_broadcaster/SceneBroadcaster.cc
@@ -21,8 +21,8 @@
 
 #include <chrono>
 #include <condition_variable>
-#include <set>
 #include <string>
+#include <unordered_set>
 
 #include <ignition/common/Profiler.hh>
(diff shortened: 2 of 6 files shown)


src/systems/scene_broadcaster/SceneBroadcaster.cc
```diff
@@ -75,6 +76,15 @@ class ignition::gazebo::systems::SceneBroadcasterPrivate
   /// \return True if successful.
   public: bool StateService(ignition::msgs::SerializedStepMap &_res);
 
+  /// \brief Callback for state service - non blocking.
+  /// \param[out] _res Response containing the last available full state.
+  /// \return True if successful.
+  // public: bool StateAsyncService(ignition::msgs::Empty &_req);
+  public: void StateAsyncService(
+      // ignition::msgs::Empty &_req, ignition::msgs::Boolean &_res);
+      // ignition::msgs::Empty &_req);
```
chapulina, line 85: Can the commented out code be removed?

src/systems/scene_broadcaster/SceneBroadcaster.cc
```diff
@@ -294,6 +308,18 @@ void SceneBroadcaster::PostUpdate(const UpdateInfo &_info,
...
     }
 
+    // process async state requests
+    if (!this->dataPtr->stateRequests.empty())
+    {
+      ignition::msgs::SerializedStepMap res;
+      res.CopyFrom(this->dataPtr->stepMsg);
+      for (const auto &reqSrv : this->dataPtr->stateRequests)
+      {
+        this->dataPtr->node->Request(reqSrv, res);
```
chapulina, line 318: I think the GUI would want the full state in this case, right? Maybe we should check `if (!this->dataPtr->stateRequests.empty())` in line 292 so the full state is populated?

src/gui/GuiRunner.cc
```diff
@@ -77,15 +86,8 @@ void GuiRunner::OnPluginAdded(const QString &_objectName)
...
-    const bool _result)
+void GuiRunner::OnStateAsyncService(const msgs::SerializedStepMap &_res)
 {
-  if (!_result)
-  {
-    ignerr << "Service call failed for [" << this->stateTopic << "]"
-           << std::endl;
-    return;
-  }
   this->OnState(_res);
```
chapulina, line 91: At this point, maybe we want to unadvertise the `StateAsyncService`? Just to keep the list of services cleaner and maybe alleviate a bit the transport.

chapulina (commented):
The general idea looks good to me. I just have some questions.

src/gui/GuiRunner.cc
iche033: yep, unadvertised service. 14472ba

src/systems/scene_broadcaster/SceneBroadcaster.cc
iche033: removed. 14472ba

src/systems/scene_broadcaster/SceneBroadcaster.cc
iche033, line 313: when the async request is received, I set the `this->stateServiceRequest` variable to `true` so that should make sure the full state is populated

src/systems/scene_broadcaster/SceneBroadcaster.cc
```diff
@@ -294,6 +303,18 @@ void SceneBroadcaster::PostUpdate(const UpdateInfo &_info,
       this->dataPtr->stateCv.notify_all();
     }
 
+    // process async state requests
+    if (!this->dataPtr->stateRequests.empty())
+    {
+      ignition::msgs::SerializedStepMap res;
```
caguero, line 309: Do we need to do this extra copy? It seems that `stateMutex` protects it from being accessed while we send the requests. Ignition Transport makes its own copy when you call `Request()` and the service provider is on a separate process (it doesn't copy if the service provider is on the same process.

caguero (commented):
It looks good to me. I just left two minor comments.

src/systems/scene_broadcaster/SceneBroadcaster.cc
```diff
@@ -190,6 +195,9 @@ class ignition::gazebo::systems::SceneBroadcasterPrivate
 
   /// \brief Flag used to indicate if the state service was called.
   public: bool stateServiceRequest{false};
+
+  /// \brief A list of async state requests
+  public: std::set<std::string> stateRequests;
```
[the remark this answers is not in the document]
iche033: yep switched to `unordered_set`. ab948fb

src/systems/scene_broadcaster/SceneBroadcaster.cc
iche033: I see, we won't need to copy then. Removed copy in ab948fb

caguero (approved):

chapulina (approved):

PR #463, opened 2020-11-20
Fixes flaky RecordAndPlayback test in INTEGRATION_log_system

azeey:
The flakiness comes from two sources:
1. Poses recorded by the LogRecorder are published by the SceneBroadcaster system throttled at 60 Hz. The throttle mechanism uses real-time instead of sim-time which causes a variance in the number of recorded poses from run to run. However, the expected number of recorded poses was calculated with the assumption that the simulation would run with a 1.0 RTF. If the CPU load is high, there could be a mismatch between the expected and the actual number of recorded poses, which causes the test to fail. This can be checked by running the test with:

```cpulimit -l 20 -f  bin/INTEGRATION_log_system  -- --gtest_filter="*RecordAndPlayback"```

2. An attempt is made to match up played back poses with the closest timestamp in the recorded file. These poses are again published by the SceneBroadcaster, so they are subject to the same kind of timing issues as the recorded poses.

The solution in this patch for the first issue is to determine the expected number of recorded poses by counting them in a separate system that mimics the throttling mechanism of ign-transport. For the second issue, a testing system is added to the playback server bypassing the SceneBroadcaster altogether.

Fixes https://github.com/ignitionrobotics/ign-gazebo/issues/435.

### DIFF
--- a/test/integration/log_system.cc
+++ b/test/integration/log_system.cc
@@ -707,6 +707,9 @@ TEST_F(LogSystemTest, RecordAndPlayback)
   // Create temp directory to store log
   this->CreateLogsDir();
 
+  // Used to count the expected number of poses recorded. Counting is necessary
+  // as the number varied depending on the CPU load.
+  int expectedPoseCount = 0;
   // Record
   {
     // World with moving entities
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: test/integration/log_system.cc [hunk 1]
```diff
@@ -814,98 +834,84 @@ TEST_F(LogSystemTest, RecordAndPlayback)
...
 
-  #if !defined (__APPLE__)
-  /// \todo(anyone) there seems to be a race condition that sometimes cause an
-  /// additional messages to be published by the scene broadcaster
-  // 60Hz
-  EXPECT_TRUE(nTotal == expectedPoseCount || nTotal == expectedPoseCount + 1)
+  // The expectedPoseCount might be off by ±2 because the ign transport's
+  // throttle mechanism (which is used by the SceneBroadcaster when publishing
+  // poses) uses real-time
+  EXPECT_LE(std::abs(nTotal - expectedPoseCount), 2)
```
chapulina, line 912: I guess 2 may still be conservative depending on the RTF, right? Not sure if there's much more we can do here though. Did you consider subscribing to the pose topic while recording?

chapulina (approved):
The test passed on macOS and actions, good sign!

I think it's definitely better to compare against the poses from the ECM during playback, to remove an extra source of uncertainty.

This looks good to me, I just have a question below of something else we could try, for your consideration.

test/integration/log_system.cc
azeey: My hope is that it won't be affected by the RTF. I've tested it with `cpulimit` at 10% several times and it passes locally. Also, the ±2 should actually be ±1 since I'm checking if it's less than 2.

I hadn't considered subscribing to the pose topic while recording because I was trying to find a way to calculate the pose count without relying on ign-transport. This way the expected and actual counts are from two different sources and would be more likely to catch errors if the log recorder, which depends on ign-transport, failed to record the expected number of poses.

azeey: Oops, never mind on the ±1, we're actually checking if it's `<= 2`

chapulina: >  counts are from two different sources and would be more likely to catch errors if the log recorder, which depends on ign-transport, failed to record the expected number of poses.

Fair enough. I think the current solution is good :+1:

PR #466, opened 2020-11-24
Update key event handling

JShep1:
Updates `Scene3D` so that it doesn't consume key events but rather processes them and passes them on.

Also updates the `Transform Control` plugin so that the key handling is done on the cpp side rather than the usage of `Shortcut`s in qml which automatically consumes any key event matching the shortcut specified.

`keyPressEvent` was removed from `Scene3D.cc` since it is only called when the `Scene3D` is focused.  This leads to finnicky behavior if the user wishes to use a hotkey for something within `Scene3D` or `Transform Control` while having a separate area of the Ignition gui focused.  This PR is basically to say that we should recommend all button processing and handling should occur within `eventFilter` rather than the usage of `Shortcut`s in qml, we should likely outright ban `Shortcut`s from being used.

Note: I tested this out with the `IgnSpinBox`s and the `TextField` and it seems like `eventFilter` still captures the inputs regardless of if the user is typing into a text field.  There's not a specific way to manually "capture" the input key presses in qml so we may have to leave it up to each individual plugin to capture the key press (returning `true` from `eventFilter` after the key event is handled).  It also seems like certain qml structures capture key input under the hood.  In particular, `Entity Tree` uses a `TreeView` which uses the escape key under the hood to deselect any currently selected entities in the tree.  It actually captures the escape key press if there is currently entities being deselected via the escape key.  For that reason, we should also advise for plugins to use the key release event for using the escape key since the `TreeView` captures the key press but not the key release.

Solves https://github.com/ignitionrobotics/ign-gazebo/issues/369

### DIFF
--- a/src/gui/plugins/scene3d/Scene3D.cc
+++ b/src/gui/plugins/scene3d/Scene3D.cc
@@ -2567,7 +2567,8 @@ bool Scene3D::eventFilter(QObject *_obj, QEvent *_event)
       renderWindow->HandleKeyRelease(keyEvent);
     }
   }
-  else if (_event->type() == ignition::gazebo::gui::events::EntitiesSelected::kType)
+  else if (_event->type() ==
+      ignition::gazebo::gui::events::EntitiesSelected::kType)
   {
     auto selectedEvent =
         reinterpret_cast<gui::events::EntitiesSelected *>(_event);
--- a/src/gui/plugins/scene3d/Scene3D.cc
+++ b/src/gui/plugins/scene3d/Scene3D.cc
@@ -2347,6 +2347,8 @@ void Scene3D::LoadConfig(const tinyxml2::XMLElement *_pluginElem)
          << this->dataPtr->cameraPoseTopic << "]" << std::endl;
 
   ignition::gui::App()->findChild<
+      ignition::gui::MainWindow *>()->QuickWindow()->installEventFilter(this);
+  ignition::gui::App()->findChild<
       ignition::gui::MainWindow *>()->installEventFilter(this);
 }
 
@@ -2547,7 +2549,25 @@ void RenderWindowItem::SetScaleSnap(const math::Vector3d &_scale)
 /////////////////////////////////////////////////
 bool Scene3D::eventFilter(QObject *_obj, QEvent *_event)
 {
-  if (_event->type() == ignition::gazebo::gui::events::EntitiesSelected::kType)
+  if (_event->type() == QEvent::KeyPress)
+  {
+    QKeyEvent *keyEvent = static_cast<QKeyEvent*>(_event);
+    if (keyEvent)
+    {
+      auto renderWindow = this->PluginItem()->findChild<RenderWindowItem *>();
+      renderWindow->HandleKeyPress(keyEvent);
+    }
+  }
+  else if (_event->type() == QEvent::KeyRelease)
+  {
+    QKeyEvent *keyEvent = static_cast<QKeyEvent*>(_event);
+    if (keyEvent)
+    {
+      auto renderWindow = this->PluginItem()->findChild<RenderWindowItem *>();
+      renderWindow->HandleKeyRelease(keyEvent);
+    }
+  }
+  else if (_event->type() == ignition::gazebo::gui::events::EntitiesSelected::kType)
   {
     auto selectedEvent =
         reinterpret_cast<gui::events::EntitiesSelected *>(_event);
@@ -2817,13 +2837,13 @@ void RenderWindowItem::wheelEvent(QWheelEvent *_e)
 }
 
 ////////////////////////////////////////////////
-void RenderWindowItem::keyPressEvent(QKeyEvent *_e)
+void RenderWindowItem::HandleKeyPress(QKeyEvent *_e)
 {
   this->dataPtr->renderThread->ignRenderer.HandleKeyPress(_e);
 }
 
 ////////////////////////////////////////////////
-void RenderWindowItem::keyReleaseEvent(QKeyEvent *_e)
+void RenderWindowItem::HandleKeyRelease(QKeyEvent *_e)
 {
   this->dataPtr->renderThread->ignRenderer.HandleKeyRelease(_e);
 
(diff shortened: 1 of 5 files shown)


src/gui/plugins/scene3d/Scene3D.cc
```diff
@@ -2547,7 +2549,26 @@ void RenderWindowItem::SetScaleSnap(const math::Vector3d &_scale)
...
+      renderWindow->HandleKeyPress(keyEvent);
+    }
+  }
+  else if (_event->type() == QEvent::KeyRelease)
+  {
+    QKeyEvent *keyEvent = static_cast<QKeyEvent*>(_event);
+    if (keyEvent)
+    {
+      auto renderWindow = this->PluginItem()->findChild<RenderWindowItem *>();
+      renderWindow->HandleKeyRelease(keyEvent);
```
chapulina, line 2567: I think this introduced an issue:

1. Select a shape to insert, like a cylinder
1. Hover over the scene
1. Press escape
1. Crash :boom: 

<details><summary>Backtrace</summary>
<p>

```
/usr/lib/x86_64-linux-gnu/libOgreMain.so.2.1.0(_ZN4Ogre11RenderQueue9renderGL3EbbPNS_9HlmsCacheERKNS0_16RenderQueueGroupEPNS_20IndirectBuffe
rPackedEPhS8_+0x264) [0x7f7f07b0cd04]
/usr/lib/x86_64-linux-gnu/libOgreMain.so.2.1.0(_ZN4Ogre11RenderQueue6renderEPNS_12RenderSystemEhhbb+0x949) [0x7f7f07b0f4a9]
/usr/lib/x86_64-linux-gnu/libOgreMain.so.2.1.0(_ZN4Ogre12SceneManager14_renderPhase02EPNS_6CameraEPKS1_PNS_8ViewportEhhb+0x4f2) [0x7f7f07b56
802]
/usr/lib/x86_64-linux-gnu/libOgreMain.so.2.1.0(_ZN4Ogre6Camera19_renderScenePhase02EPKS0_PNS_8ViewportEhhb+0x36) [0x7f7f07970716]
/usr/lib/x86_64-linux-gnu/libOgreMain.so.2.1.0(_ZN4Ogre12RenderTarget28_updateViewportRenderPhase02EPNS_8ViewportEPNS_6CameraEPKS3_hhb+0x59)
 [0x7f7f07b299b9]
/usr/lib/x86_64-linux-gnu/libOgreMain.so.2.1.0(_ZN4Ogre19CompositorPassScene7executeEPKNS_6CameraE+0x3ba) [0x7f7f07c819ca]
/usr/lib/x86_64-linux-gnu/libOgreMain.so.2.1.0(_ZN4Ogre14CompositorNode7_updateEPKNS_6CameraEPNS_12SceneManagerE+0x187) [0x7f7f07c5cc67]
/usr/lib/x86_64-linux-gnu/libOgreMain.so.2.1.0(_ZN4Ogre19CompositorWorkspace7_updateEv+0x2aa) [0x7f7f07c6f8ca]
/usr/lib/x86_64-linux-gnu/libOgreMain.so.2.1.0(_ZN4Ogre18CompositorManager221_updateImplementationERNS_22SceneManagerEnumeratorEPNS_11HlmsMa
nagerE+0x1a1) [0x7f7f07c573c1]
/usr/lib/x86_64-linux-gnu/libOgreMain.so.2.1.0(_ZN4Ogre4Root23_updateAllRenderTargetsEv+0x35) [0x7f7f07b48f95]
/usr/lib/x86_64-linux-gnu/libOgreMain.so.2.1.0(_ZN4Ogre4Root14renderOneFrameEv+0x9c) [0x7f7f07b490fc]
/home/chapulina/citadel_ws/install/lib/ign-rendering-3/engine-plugins/libignition-rendering-ogre2.so(_ZN8ignition9rendering2v317Ogre2RenderT
arget6RenderEv+0x35) [0x7f7f0c539875] /home/chapulina/citadel_ws/src/ign-rendering/ogre2/src/Ogre2RenderTarget.cc:173
/home/chapulina/citadel_ws/install/lib/ign-rendering-3/engine-plugins/libignition-rendering-ogre2.so(_ZN8ignition9rendering2v310BaseCameraIN
S1_11Ogre2SensorEE6UpdateEv+0x5d) [0x7f7f0c4dc81d] /usr/include/c++/8/ext/atomicity.h:69
/home/chapulina/citadel_ws/install/lib/ign-gazebo-3/plugins/gui/libGzScene3D.so(_ZN8ignition6gazebo2v311IgnRenderer6RenderEv+0xfb) [0x7f7f34
33843b] /home/chapulina/citadel_ws/src/ign-gazebo/src/gui/plugins/scene3d/Scene3D.cc:467
/home/chapulina/citadel_ws/install/lib/ign-gazebo-3/plugins/gui/libGzScene3D.so(_ZN8ignition6gazebo2v312RenderThread10RenderNextEv+0x46) [0x
7f7f34339856] /home/chapulina/citadel_ws/src/ign-gazebo/src/gui/plugins/scene3d/Scene3D.cc:1874
```

</p>
</details>

The problem is probably related to the thread in which the event callback is being called. Surprisingly, I can't reproduce this on the `ign-gazebo3` branch, even though I'd expect to. We need to make sure the shape deletion is handled in the render thread,

chapulina (commented):
LGTM except for the crashing issue reported below.

src/gui/plugins/scene3d/Scene3D.cc
JShep1: I  found that this error does actually occur on `ign-gazebo3`.  It's hard to press escape and have it work for cancelling a shape placement as is since the window needs to be prioritized so the process I took to replicate this in `ign-gazebo3` was:
 1. Select shape to insert
 1. Hover over the plugin menu
 1. Re-enter the mouse into the `Scene3D` (to force focus the window)
 1. Press `<ESC>`
 1. If no crash, then restart at 1

I think no crash was noticeable since the escape button press was hardly working before this PR given the way the keys were being handled.  I feel like it may be better to have a separate PR fixing this issue.  I ticketed an issue [here](https://github.com/ignitionrobotics/ign-gazebo/issues/488) and assigned myself.

chapulina: Interesting, thanks for investigating. I think we should make that issue a high priority. But I agree we can get this in for now.

chapulina (approved):

PR #452, opened 2020-11-07
👩‍🌾 Make PeerTracker test more robust

chapulina:
Various test cases in `PeerTracker_TEST` have been flaking for a while, see the [test history](https://build.osrfoundation.org/job/ignition_gazebo-ci-ign-gazebo3-homebrew-amd64/lastBuild/testReport/(root)/PeerTracker/PeerTrackerStale/history/). This PR is an attempt at making it more robust to diverse timing conditions.

I also included some tweaks to the `physics_system` test.

### Most recent `PeerTracker` failures

```
35: [ RUN      ] PeerTracker.PeerTracker
35: /Users/jenkins/workspace/ignition_gazebo-ci-ign-gazebo3-homebrew-amd64/ign-gazebo/src/network/PeerTracker_TEST.cc:84: Failure
35: Value of: tracker5->NumPeers()
35:   Actual: 4
35: Expected: 5u
35: Which is: 5
35: [  FAILED  ] PeerTracker.PeerTracker (2601 ms)
```

```
35: [ RUN      ] PeerTracker.PeerTrackerStale
35: /Users/jenkins/workspace/ignition_gazebo-ci-ign-gazebo3-homebrew-amd64/ign-gazebo/src/network/PeerTracker_TEST.cc:167: Failure
35: Value of: tracker1->NumPeers()
35:   Actual: 1
35: Expected: 0u
35: Which is: 0
35: [1;36m[Dbg] [PeerTracker.cc:163] [0m[1;36mAttempting to remove peer [[0m[1;36m1aa895e8-ec40-403e-8a5c-1aa2ed5b3fcc[0m[1;36m] from [[0m[1;36m2b497001-2f37-4230-9aa4-8cf6c70173f6[0m[1;36m] but it wasn't connected[0m[1;36m[0m
35: [  FAILED  ] PeerTracker.PeerTrackerStale (294 ms)
```

This one with 2 stale peers is a real mystery to me :monocle_face: 

```
37: [ RUN      ] PeerTracker.PeerTrackerStale
37: /Users/jenkins/workspace/ignition_gazebo-ci-ign-gazebo3-homebrew-amd64/ign-gazebo/src/network/PeerTracker_TEST.cc:163: Failure
37: Value of: stalePeers
37:   Actual: 2
37: Expected: 1
37: /Users/jenkins/workspace/ignition_gazebo-ci-ign-gazebo3-homebrew-amd64/ign-gazebo/src/network/PeerTracker_TEST.cc:167: Failure
[...]

PR #469, opened 2020-12-02
3 ➡️ 4

chapulina:
Forward-port `ign-gazebo3` to `ign-gazebo4`.

Comparison:

https://github.com/ignitionrobotics/ign-gazebo/compare/ign-gazebo4...ign-gazebo3

### DIFF
--- a/src/EntityComponentManager.cc
+++ b/src/EntityComponentManager.cc
@@ -945,15 +945,6 @@ void EntityComponentManager::AddEntityToMessage(msgs::SerializedStateMap &_msg,
   // Add a component to the message and set it to be removed if the component
   // exists in the removedComponents map.
   this->dataPtr->SetRemovedComponentsMsgs(_entity, _msg);
-
-  // Remove the entity from the message if a component for the entity was
-  // not modified or added. This will allow the state message to shrink.
-  if (entIter == _msg.mutable_entities()->end() &&
-      (entIter = _msg.mutable_entities()->find(_entity)) !=
-      _msg.mutable_entities()->end())
-  {
-    _msg.mutable_entities()->erase(entIter);
-  }
 }
 
 //////////////////////////////////////////////////
--- a/include/ignition/gazebo/EntityComponentManager.hh
+++ b/include/ignition/gazebo/EntityComponentManager.hh
@@ -226,6 +226,20 @@ namespace ignition
               std::optional<typename ComponentTypeT::Type> ComponentData(
               const Entity _entity) const;
 
+      /// \brief Set the data from a component.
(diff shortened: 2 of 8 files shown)


chapulina:
Well, the `PeerTracker` test still looks flaky even with #452 :confused: 

```
/Users/jenkins/workspace/ignition_gazebo-ci-pr_any-homebrew-amd64/ign-gazebo/src/network/PeerTracker_TEST.cc:177
Value of: stalePeers
  Actual: 2
Expected: 1
```

src/EntityComponentManager.cc
```diff
@@ -945,15 +945,6 @@ void EntityComponentManager::AddEntityToMessage(msgs::SerializedStateMap &_msg,
...
   // exists in the removedComponents map.
   this->dataPtr->SetRemovedComponentsMsgs(_entity, _msg);
-
-  // Remove the entity from the message if a component for the entity was
-  // not modified or added. This will allow the state message to shrink.
-  if (entIter == _msg.mutable_entities()->end() &&
-      (entIter = _msg.mutable_entities()->find(_entity)) !=
-      _msg.mutable_entities()->end())
-  {
-    _msg.mutable_entities()->erase(entIter);
```
chapulina, line 955: The ABI checker gave what I assumed was a false positive, see https://github.com/ignition-tooling/release-tools/issues/336

I inspected the symbols locally just in case, and was surprised to see that this symbol was actually gone:

`_ZNK8ignition6gazebo2v410components13BaseComponent9SerializeERSo`

I narrowed it down to line 955. When it's removed, the symbol disappears with it :fearful: :woman_shrugging: 

You can see all symbols before and after removing the line here: https://gist.github.com/chapulina/7f53a3d43cb031715f005650e582fcb7

azeey: Looks like that symbol is not used by anything in `libignition-gazebo.so` so it must be getting optimized away. If you build in debug mode, it doesn't get removed.

Since the function is defined in a header, I think this is somewhat of a false positive. Any translation unit that includes the `Component.hh` header will have access to symbol. For example,  `UNIT_Component_TEST` still has the symbol even though this it is missing from `libignition-gazebo.so`.

chapulina: Hum interesting and mysterious how the compiler decides to remove the symbol. Your reasoning about the ABI not being broken makes sense to me. I guess all functions defined in headers may be subject to the same rule.

Makes me wonder about why the symbol would be added in the first place. When the compiler does add the symbol, I guess it can be used directly instead of regenerated when the header is included?

This messes up the poor ABI checker :confused:

nkoenig (approved):

PR #470, opened 2020-12-02
Use a std::promise/std::future to avoid busy waiting the step ack messages in NetworkManagerPrimary

ivanpauno:
Minimal improvement, mainly intended to get used to the development cycle.

This PR avoids busy waiting in a loop until all `stepAck` messages are received by instead using a std::promise/std::future as a notification mechanism.

I tested this with the `distributed_levels` example, seems to be working fine.
`colcon test` was also run locally, `INTEGRATION_diff_drive_system` and `INTEGRATION_wheel_slip` are failing but seemed unrelated (they also fail when I use `ign-gazebo4` branch).

Note: I don't have push access to the repo, so I'm working from a fork. It would be nice to get access, so I can add reviewers/labels appropriately.

maybe @mjcarroll can take a look

### DIFF
--- a/src/network/NetworkManagerPrimary.cc
+++ b/src/network/NetworkManagerPrimary.cc
@@ -200,7 +200,8 @@ std::map<std::string, SecondaryControl::Ptr>
 void NetworkManagerPrimary::OnStepAck(const msgs::SerializedStateMap &_msg)
 {
   this->secondaryStates.push_back(_msg);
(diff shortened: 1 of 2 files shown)


mjcarroll:
CC @chapulina for permissions

chapulina (approved):
Towards the future! LGTM. I'll just make the style change below and get it in. Thanks!

PR #475, opened 2020-12-06
Bazel updates for gazebo4

JShep1:


mahiuchun:
I see a link error due to `copyElement` being defined in both `ign_gazebo/src/ServerConfig.cc` and `ign_launch/plugins/gazebo_server/GazeboServer.cc`. Please use `static` or put them into anonymous namespace if they are meant to be local helper functions.

src/ign.cc
```diff
@@ -357,6 +357,8 @@ extern "C" IGNITION_GAZEBO_VISIBLE int runServer(const char *_sdfString,
 //////////////////////////////////////////////////
 extern "C" IGNITION_GAZEBO_VISIBLE int runGui(const char *_guiConfig)
 {
+  Q_INIT_RESOURCE(gazebo);
```
mahiuchun, line 360: I had to remove this in order to avoid a dynamic link error.

mahiuchun (changes requested):
It appears that this PR has not caught up with https://github.com/ignitionrobotics/ign-msgs/commit/e2709830f295061629eee65b2dcd876e80662556 yet so it does not build right now.

chapulina (changes requested):
The build is broken

chapulina:
@mjcarroll , I'll close this as it seems outdated

PR #480, opened 2020-12-10
Fix tests that use Fuel paths with uppercase letters

azeey:
As of https://github.com/ignitionrobotics/ign-fuel-tools/pull/130, fuel paths use all lowercase letters. This fixes some tests that had uppercase letters.


chapulina (approved):
Thanks for staying on top of this. Looks good to me once the tests come back happy.

PR #481, opened 2020-12-10
Basic async primary-secondary stepping

ivanpauno:
First step towards https://github.com/ignitionrobotics/ign-gazebo/issues/468.

TODOs:

- Add a field in the message the primary send to secondaries that indicates how many steps the secondary can move forward.
  If that number is 1, the behavior will be exactly the same as before the PR.
- Real time is only being updated by the primary, secondaries are only asynchronously updating the simulated time/iteration number.

---

Notes on how the case where performers are interacting will be handled (not implemented in this PR), based on a previous discussion with @mjcarroll:

- Performers are far apart -> secondaries can make steps asynchronously
- Performers are seeing each other -> need perfect lockstep. We could also first run the preUpdate/Update phase, sync secondaries state and then run the postUpdate phase. I'm not sure if that's worth it or not.
- Performers interacting each other -> physics should be reallocated to same secondary. We would need to: run preUpdate phase, sync secondaries state, run Update phase (physics), sync secondaries state, run postUpdate phase (all in a perfect lockstep fashion).

I'm not sure if we want to be able to handle a combination of this cases when having more than two secondaries.
e.g.: supposing one performer per secondary, A is interacting with B and C is not interacting with neither A or B, should C run asynchronously?
In that case, we would need some extra information in the message that the primary sends to the secondaries.

---

Results when running the `distributed_levels` simulation in my computer:
- standalone simulation is completed with RTF of about 800~1200%
- distributed, before this PR, RTF was about 30~40%
- distributed, after this PR, RTF of about 40~80%

@mjcarroll does the approach of this PR seem correct?

It would probably be good to test what's the benefit of asynchronous primary-secondary stepping in a more realistic simulation (not only physics), any recommended world to test that?


mjcarroll:
> It would probably be good to test what's the benefit of asynchronous primary-secondary stepping in a more realistic simulation (not only physics), any recommended world to test that?

As far as platforms, some of the SubT vehicles are probably ideal (longer list here: https://github.com/osrf/subt/wiki/Robots)

* Lidar + Camera: https://app.ignitionrobotics.org/openrobotics/fuel/models/X1%20Config%204
* Really sensor heavy (lidar + 4 kinects): https://app.ignitionrobotics.org/OpenRobotics/fuel/models/EXPLORER_X1_SENSOR_CONFIG_2

I think it would be reasonable to just place 4 of these on a plane 20 meters apart and use that as the "no interaction" case to see what the speedup looks like.

Actual worlds are available in the SubT repo: https://github.com/osrf/subt/tree/master/subt_ign/worlds, but I don't think it's really necessary to use one of those in this stage of development.

ivanpauno:
force-pushed to solve conflicts

ivanpauno:
I have been doing some tests today, it seems to me that allowing the secondaries to make steps asynchronously is not of much benefit and the current performance bottleneck is the primary.

> Really sensor heavy (lidar + 4 kinects): https://app.ignitionrobotics.org/OpenRobotics/fuel/models/EXPLORER_X1_SENSOR_CONFIG_2

I used two of those models replacing the box and the sphere in the distributed levels simulation, and the secondaries where always "far ahead" and the primary was always going "quite behind", so there was almost not speedup when letting the secondaries make steps asynchronously.
The simulation was allowing the secondaries to move ahead 1000 steps, and the primary had always about 500 secondaries iterations cached and was never waiting for a secondary message.

I did some profiling, it seems that `SceneBroadcaster::PostUpdate` step is taking much longer (100x) when the simulation is distributed than when it's standalone.
I'm not sure what's the reason, but I assume that that behaves differently when entity/components are being updated from a serialized message than when they are being updated locally.

@mjcarroll I have a bunch of questions:

- are this results expected?
- does it worth exploring why the `SceneBroadcaster::PostUpdate` is much slower in the distributed case and optimizing it?

If we are not going to make the primary go faster, in the only case where there might be a benefit of allowing the secondaries to make steps asynchronously is if the secondaries are really "heavy loaded", but I'm not sure if this is going to be the case in a subT world. Maybe, the world I was running was still too simple.

ivanpauno:
[Last commit](https://github.com/ignitionrobotics/ign-gazebo/pull/481/commits/04f8c3dc38d3b2620873a9acb968a060f87aff09) has the modified "distributed_levels" example I'm using for testing

ivanpauno:
@mjcarroll I have created some scripts that allows to run N robots in M secondaries, and it records stats using `ign_imgui`.

Typical usage:

```bash
#standalone
cd ~/my_ws/src/ign-gazebo/examples/scripts/distributed_levels
# Run with 8 robots
# Will generate an `stats_standalone_r8.csv` that can be visualized with `ign_imgui -i <PATH_TO_CSV>`
./standalone -r 8 
```

```bash
# distributed, machines running a secondary
cd ~/my_ws/src/ign-gazebo/examples/scripts/distributed_levels
./secondary -r 8  # run with 8 robots
```

```bash
# distributed, machines running the primary
cd ~/my_ws/src/ign-gazebo/examples/scripts/distributed_levels
# Run with 8 robots, 2 secondaries (4 robots each secondary).
# Will generate an `stats_primary_r8_s2.csv` that can be visualized with `ign_imgui -i <PATH_TO_CSV>`
./primary -r 8 -s 2
```

My `ign_imgui` fork is [here](https://github.com/ivanpauno/ign_imgui), and allows recording the histogram and stats to a CSV file and loading them.

@mjcarroll Do you think this is good enough to start testing or should I add something else?
The scripts are easy to run and the results are recorded automatically, so testing should be more or less easy.

Example results in my machine

<details>
  <summary>standalone 2 robots, average RTF 183%</summary>

Screenshot from 2020-12-17 18-37-41

</details>

<details>
  <summary>distributed 2 robots, 2 secondaries, average RTF 118%</summary>

Screenshot from 2020-12-17 18-37-41

</details>

PS: This need empy installed, `python3 -m pip install empy`, I didn't realize that there was already a template to generate the [worlds in the distributed_levels folder](https://github.com/ignitionrobotics/ign-gazebo/blob/ign-gazebo4/examples/scripts/distributed_levels/distributed_levels.sdf.erb) :man_facepalming:, if not I would have edited that one.

mjcarroll:
I think that you may have posted the same image twice in your histogram analysis.

chapulina:
What's the status of this PR, are there plans to continue working on it?

mjcarroll:
This was part of some of our work.  I don't think there are plans to pursue it as part of the current project, but I will land this PR.

src/EntityComponentManager.cc
```diff
@@ -1283,6 +1280,8 @@ void EntityComponentManager::SetState(
         // When above TODO is addressed, uncomment AddModifiedComponent below
         // unless calling SetChanged (which already calls AddModifiedComponent)
         // this->dataPtr->AddModifiedComponent(entity);
+        comp->Deserialize(istr);
+        this->SetChanged(entity, typeId, ComponentState::PeriodicChange);
```
Blast545, line 1284: I believe this fix is not part strictly to the async implementation, should this be targeted to ign-gazebo in another PR? @mjcarroll

src/network/NetworkManagerPrimary.cc
```diff
@@ -109,6 +110,9 @@ bool NetworkManagerPrimary::Ready() const
   return (nSecondary == this->dataPtr->config.numSecondariesExpected);
 }
 
+/// Number of iterations a secondary can move ahead.
+constexpr uint64_t kSecondaryIterations = 1000uLL;
```
Blast545, line 114: Should this be a constant/parameter of the networkPrimary class instead?

src/network/NetworkManagerPrimary.cc
```diff
@@ -129,55 +133,75 @@ bool NetworkManagerPrimary::Step(const UpdateInfo &_info)
...
-  this->secondaryStatesPromise = std::promise<void>{};
-  auto future = this->secondaryStatesPromise.get_future();
-  this->simStepPub.Publish(step);
+  if (_info.iterations >= this->nextIteration && !_info.paused) {
+    // Update state based on secondaries messages.
+    std::vector<private_msgs::SecondaryStep> secondariesSteps;
+    {
+      std::unique_lock<std::mutex> guard{this->secondaryStatesMutex};
 
-  // Block until all secondaries are done
```
Blast545, line 151: Maybe it's good to keep this comment, as the next `for loop` is doing that

src/network/NetworkManagerSecondary.cc
```diff
@@ -96,86 +113,129 @@ bool NetworkManagerSecondary::OnControl(const private_msgs::PeerControl &_req,
...
-    else
-    {
-      auto parent =
-          this->dataPtr->ecm->Component<components::ParentEntity>(entityId);
-      this->dataPtr->ecm->RequestRemoveEntity(parent->Data());
-
-      if (this->performers.find(entityId) != this->performers.end())
+      // Update affinities
+      // TODO(ivanpauno):
+      // This needs to be redone based on if performers can "far apart", "can view each other" or "are interacting".
```
Blast545, line 153: I imagine the logic to decide if "far apart" / "View each other" would be done in the primary, right? The secondaries will receive the performers/affinities

src/network/NetworkManagerSecondary.cc
```diff
@@ -96,86 +113,129 @@ bool NetworkManagerSecondary::OnControl(const private_msgs::PeerControl &_req,
...
 
-  // Step runner
-  this->dataPtr->stepFunction(info);
+      // Update state with all the performer's entities
+      std::unordered_set<Entity> entities;
+      for (const auto &perf : this->performers)
+      {
+        // Performer model
+        auto parent = this->dataPtr->ecm->Component<components::ParentEntity>(perf);
+        if (parent == nullptr)  // TODO(ivanpauno): why is this now needed?
```
Blast545, line 198: I think this isn't needed, if this check is triggered something really bad happened because the this->performers is updated right above.

Blast545 (commented):
Looks good to me to iterate on top of this, PTAL @mjcarroll

src/network/NetworkManagerPrimary.cc
mjcarroll: Yes, and in fact, it may make more sense to have this value be configurable, depending on the application.

src/network/NetworkManagerPrimary.cc
mjcarroll: Agreed, let's restore that.

src/network/NetworkManagerSecondary.cc
mjcarroll: Correct.  The bulk of that should happen in the primary, with only distribution happening to the secondaries.

Based on our later conversations, I don't think that it will be necessary to change the affinities, either.  System plugins will still run in their respective secondaries, but the location that physics is being simulated will change.

src/network/NetworkManagerSecondary.cc
mjcarroll: Good point, I believe that it can be removed.

mjcarroll:
Looks like this will also need a linting pass.  I'm not sure why the Focal version isn't failing.

src/network/NetworkManagerPrimary.cc
Blast545: Let me know what you think of a proposed change in a2e607b. Leaving a TO-DO to later build on top of that how to configure that parameter (later, if desired)

src/network/NetworkManagerSecondary.cc
Blast545: > System plugins will still run in their respective secondaries

Correct me if I'm wrong, aren't the secondaries just running the physics step for the performers with the simulation step? IIUC with the current approach the system plugins are run in the primary.

Blast545:
I think this is ready for review as it is, and we should be able to merge as it extends the features of the base distributed system implementation.

I also created a [design document](https://docs.google.com/document/d/1fCq1mgHD4l_IVzZouDTihNWo3Q_fM3KyZC_SCvXqQZQ/edit?usp=sharing) summarizing the efforts here and the plan for distributed simulation that could help us get in the same page about the previous conversations and possible limitations. 

 PTAL @nkoenig @chapulina @mjcarroll. Let me know if you think I should add someone else to the discussion.

Blast545:
I took some time testing this in my personal PC and using EC2 AWS instances. The main purpose was checking that the PR works using multiple computers, but I went ahead and gathered some RTF metrics and organized them in the attached PDF:
[Distributed simulation RTF testing.pdf](https://github.com/ignitionrobotics/ign-gazebo/files/6826324/Distributed.simulation.RTF.testing.pdf)

Can I ask you to take a another look to the code? :pray: I think it's ready to merge. @chapulina @nkoenig @mjcarroll

src/EntityComponentManager.cc
mjcarroll: I agree, this seems unrelated?

src/network/NetworkManagerSecondary.hh
```diff
@@ -81,6 +92,24 @@ namespace ignition
...
+
+      /// \brief Mutex protecting `this->steps`.
+      private: std::mutex stepsMutex;
+
+      /// \brief Condition variable used to awake the thread doing steps.
+      private: std::condition_variable moreStepsCv;
+
+      /// \brief Boolean used to indicate that the thread
+      /// doing asynchronous steps has to stop.
+      private: bool stopAsyncStepThread {false};
```
mjcarroll, line 110: Should this be an atomic?

src/network/NetworkManagerSecondary.cc
```diff
@@ -96,86 +117,128 @@ bool NetworkManagerSecondary::OnControl(const private_msgs::PeerControl &_req,
...
-    else
-    {
-      auto parent =
-          this->dataPtr->ecm->Component<components::ParentEntity>(entityId);
-      this->dataPtr->ecm->RequestRemoveEntity(parent->Data());
-
-      if (this->performers.find(entityId) != this->performers.end())
+      // Update affinities
+      // TODO(ivanpauno):
+      // This needs to be redone based on if performers can "far apart",
```
mjcarroll, line 157: I think based on our current design, performer affinities should never change.  The plugins specific for a controller (aside from physics) will be executed locally on each secondary, and will not shift.

What will change is that the syncronization of physics will either be more strict or collocated if necessary.

Blast545: I think a bit more on this before our meeting on monday.

src/EntityComponentManager.cc
Blast545: At first I thought it was unrelated, but it seems the added remove component didn't allow the code to get until that part. So I think now they make sense together. Let me know WDYT

Blast545:
I took some time testing with 5 instances in the AWS environment I prepared, I'm attaching the results.

I also added some extra graphic tables so it's easier to see the comparison between implementations before and after the PR, and between the PC vs AWS cloud environment. PTAL @mjcarroll 
[Distributed simulation RTF testing_5instances.pdf](https://github.com/ignitionrobotics/ign-gazebo/files/6871104/Distributed.simulation.RTF.testing_5instances.pdf)

Blast545:
For reference, I'm attaching here a guide with the steps I followed to setup this PR and test it using AWS instances. 
[Setting up AWS cloud environment for testing Ignition’s distributed simulation.pdf](https://github.com/ignitionrobotics/ign-gazebo/files/6903656/Setting.up.AWS.cloud.environment.for.testing.Ignition.s.distributed.simulation.pdf)

src/network/NetworkManagerSecondary.cc
Blast545: This won't be addressed in the current PR

src/network/NetworkManagerSecondary.hh
Blast545: Totally right, this caused segfaults. Addressed with: a5aae74

Blast545:
I changed the testing script to delete older results by default with befd5fc. Otherwise just addressed old comments, I think this is ready to merge. PTAL @mjcarroll

chapulina (changes requested):
`ign-gazebo4` is EOL'ing with Dome in the coming days, please retarget this PR at a supported version.

nkoenig (approved):
I've updated the target branch to `ign-gazebo6`, cleaned up some scripts and style. Everything looks good. Just waiting on CI.

mjcarroll:
This has been open for quite some time, I don't think there is any intention to pick it up in the near future.  I'm going to close it.

PR #482, opened 2020-12-10
Add Light Usercommand and include Light parameters in the componentInspector

ahcorde:
I add a new service called `/world/lights/config` to be able to change the configuration of a light. This was added in the UserCommand system plugin.

As you can see in the gif the configuration can be changed in the Component Inspector.

lights


src/rendering/RenderUtil.cc
```diff
@@ -427,6 +432,60 @@ void RenderUtil::Update()
...
+          l->SetSpecularColor(light.second.Specular());
+        if (ignition::math::equal(
+            l->AttenuationRange(), light.second.AttenuationRange()))
+          l->SetAttenuationRange(light.second.AttenuationRange());
+        if (ignition::math::equal(
+            l->AttenuationLinear(), light.second.LinearAttenuationFactor()))
+          l->SetAttenuationLinear(light.second.LinearAttenuationFactor());
+        if (ignition::math::equal(
+            l->AttenuationConstant(), light.second.ConstantAttenuationFactor()))
+          l->SetAttenuationConstant(light.second.ConstantAttenuationFactor());
```
chapulina, line 457: Let's use `{}` for multiline `if`s to improve readability.

src/systems/user_commands/UserCommands.cc
```diff
@@ -19,6 +19,7 @@
 
 #include <google/protobuf/message.h>
 #include <ignition/msgs/boolean.pb.h>
+#include <ignition/msgs/light.pb.h>
 #include <ignition/msgs/entity_factory.pb.h>
```
chapulina, line 23: Nit: alphabetize

src/systems/user_commands/UserCommands.cc
```diff
@@ -123,6 +125,19 @@ class RemoveCommand : public UserCommandBase
   public: bool Execute() final;
 };
 
+/// \brief Command to modify a light entity from simulation.
+class LightCommand : public UserCommandBase
+{
+  /// \brief Constructor
+  /// \param[in] _msg Message identifying the entity to be removed.
```
chapulina, line 132:
```suggestion
  /// \param[in] _msg Message identifying the entity to be edited.
```

src/systems/user_commands/UserCommands.cc
```diff
@@ -178,6 +193,13 @@ class ignition::gazebo::systems::UserCommandsPrivate
   public: bool RemoveService(const msgs::Entity &_req,
       msgs::Boolean &_res);
 
+  /// \brief Callback for light service
+  /// \param[in] _req Request containing light update of an entity.
+  /// \param[in] _res True if message successfully received and queued.
+  /// It does not mean that the entity will be successfully moved.
```
chapulina, line 199:
```suggestion
  /// It does not mean that the light will be successfully updated.
```

chapulina (changes requested):
Did a first pass :+1: 

Mind adding a test to `test/integration/user_commands.cc`?

src/gui/plugins/component_inspector/Light.qml
```diff
@@ -0,0 +1,1107 @@
...
+          height: 40
+          Layout.preferredWidth: rSpecularText.width + indentation*3
+          Loader {
+            id: loaderSpecularR
+            width: iconWidth
+            height: iconHeight
+            y:10
+            sourceComponent: plotIcon
+          }
+          Component.onCompleted: loaderSpecularR.item.componentInfo = "r"
```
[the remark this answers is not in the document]
ahcorde: I can't run the Plotting plugin. I'm compiling from source Ignition Dome (Ubuntu 20.04):

```bash
[GUI] [Dbg] [Application.cc:305] Loading plugin [Plotting]
[GUI] [Wrn] [Application.cc:649] [QT] QQmlComponent: Component is not ready
[GUI] [Err] [Plugin.cc:136] Failed to instantiate QML file [:/Plotting/Plotting.qml].
* Are you sure it's been added to the .qrc file?
* Are you sure the file is valid QML? You can check with the `qmlscene` tool
[GUI] [Msg] Loaded plugin [Plotting] from path [/home/ahcorde/ignition_dome_oficial/install/lib/ign-gazebo-4/plugins/gui/libPlotting.so]
```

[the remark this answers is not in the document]
ahcorde: There is a signal connected to `UnRegisterChartToComponent` but this method does not exist. There is a method called  `UnRegisterChartFromComponent` maybe is it this one?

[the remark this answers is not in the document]
chapulina: Do you have all QML dependencies installed?

https://github.com/ignitionrobotics/ign-gui/blob/ign-gui4/.github/ci/packages.apt

> UnRegisterChartToComponent

Nice catch, I think that's wrong. Interestingly, it doesn't prevent the plot plugin from working for me.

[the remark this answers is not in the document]
ahcorde: Fixed https://github.com/ignitionrobotics/ign-gazebo/pull/490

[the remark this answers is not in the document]
ahcorde: thank you @chapulina, It was `qml-module-qt-labs-platform`. Should we need to update the install.md or readme.md to include all the required dependencies ?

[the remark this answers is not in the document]
ahcorde: @chapulina  can you try again to plot the values?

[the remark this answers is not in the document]
chapulina: > It was qml-module-qt-labs-platform. Should we need to update the install.md or readme.md to include all the required dependencies ?

Yeah the README is using `packages.apt`, so I think we need to update that:

https://github.com/ignitionrobotics/ign-gazebo/blob/ign-gazebo4/.github/ci/packages.apt

src/rendering/RenderUtil.cc
```diff
@@ -1070,13 +1138,14 @@ void RenderUtilPrivate::UpdateRenderingEntities(
...
-  // lights
+  // update lights
   _ecm.Each<components::Light, components::Pose>(
       [&](const Entity &_entity,
-        const components::Light *,
+        const components::Light *_light,
         const components::Pose *_pose)->bool
       {
         this->entityPoses[_entity] = _pose->Data();
+        this->entityLights[_entity] = _light->Data();
```
iche033, line 1148: I'm thinking of making this more efficient. Currently we are populating `entityLights` for all lights in the scene and going through the update process every iteration. Even though we check to see if any light properties have changed, there is still a bit of computation overhead every iteration. In an upcoming environment we are creating, we plan to put in a few hundred lights so this may add up. 

How about we handle light requests similar to pose commands by adding a new `components::LightCmd`? When the light service request is received, we create a `LightCmd` component and attach it to the light entity in UserCommands, e.g. see `WorldPoseCmd` creation [here](https://github.com/ignitionrobotics/ign-gazebo/blob/ign-gazebo4/src/systems/user_commands/UserCommands.cc#L671). In RenderUtil, we will instead be looping through entities with `components::LightCmd`:

```
_ecm.Each<components::LightCmd>(
      [&](const Entity &_entity,
        const components::LightCmd *_light) -> bool
   this->entityLights[_entity] = _light->Data();
   ...
```` 

We then remove the `components::LightCmd` from the entity afterwards to indicate that we've processed the cmd, e.g. see removing `WorldPoseCmd` [here](https://github.com/ignitionrobotics/ign-gazebo/blob/ign-gazebo4/src/systems/physics/Physics.cc#L1659)

File: src/systems/user_commands/UserCommands.cc [hunk 7]
```diff
@@ -630,6 +678,84 @@ bool RemoveCommand::Execute()
...
+    return false;
+  }
+
+  if (lightMsg->has_pose())
+  {
+    lightPose->Data().Pos() = msgs::Convert(lightMsg->pose()).Pos();
+  }
+
+  this->iface->ecm->SetChanged(entity,
+    components::Pose::typeId, ComponentState::OneTimeChange);
```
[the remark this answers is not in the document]
ahcorde: This call is able to change the `light` properties and the `pose`

src/rendering/RenderUtil.cc
ahcorde: @iche033 I added most of your feedback, but I'm not sure where I should remove the LightCmd. The worldPoseCmd is remove in Physics, but  where is the right place to remove these cmds? renderutils.cc?

ahcorde: @iche033 I removed the LightCmd, I had to convert the const-EntityComponentManager vriable in to a non-const EntityComponentManager variable to be able to remove the LightCmd. is this correct ?

https://github.com/ignitionrobotics/ign-gazebo/blob/19b3c9f06d5c097dc923324501a2a05bb889a2dd/src/rendering/RenderUtil.cc#L1162-L1169

iche033: hmm I see. `UpdateFromECM` is called in the [PostUpdate](https://github.com/ignitionrobotics/ign-gazebo/blob/ign-gazebo4/src/systems/sensors/Sensors.cc#L437) function in the `sensors` system on the server side in the sensors system, which gets a const ref `EntityComponentManager`. The physics system is able to remove the component as it's done in [Update](https://github.com/ignitionrobotics/ign-gazebo/blob/ign-gazebo4/src/systems/physics/Physics.cc#L1659) function which has a mutable reference to ECM. We may have to create a new `UpdateECM` function in RenderUtil and call it from a new `Update` function (which also needs to be added)  in `sensors` system. 

On the gui side, the [UpdateFromECM ](https://github.com/ignitionrobotics/ign-gazebo/blob/ign-gazebo4/src/gui/plugins/scene3d/Scene3D.cc#L2411) is called in a function that has a mutable reference to ECM. So we should be able to just do `UpdateECM()` and followed by `UpdateFromECM`

ahcorde: @iche033 done https://github.com/ignitionrobotics/ign-gazebo/pull/482/commits/0e660f93578f097fd8e72337febb3009eeebacad

include/ignition/gazebo/components/LightCmd.hh
```diff
@@ -0,0 +1,53 @@
+/*
+ * Copyright (C) 2019 Open Source Robotics Foundation
```
iche033, line 2: 2020

iche033 (commented):
nice it's working well for me. One thing I noticed is that in lights.sdf world, modifying the light attached to the sphere link does not work? It looks like other property fields of the link are greyed out so maybe link lights are not supposed to be modifiable?

ahcorde:
> One thing I noticed is that in lights.sdf world, modifying the light attached to the sphere link does not work? It looks like other property fields of the link are greyed out so maybe link lights are not supposed to be modifiable?

@iche033, The component inspector should show the Light component if this is in the Entity (sphere). But right now the sphere only contains these components:
 
 - SelfCollide
 - Pose
 - WindMode
 - ParentEntity
 - Model
 - Static
 - Name
 - SourceFilePath

The light is not attached properly to the Model/Link Entity. I think this should be added in the SdfEntityCreator.cc method like this others fields https://github.com/ignitionrobotics/ign-gazebo/blob/ign-gazebo4/src/SdfEntityCreator.cc#L256-L276

ahcorde:
@iche033 did you try this ?

Selección_067

src/systems/user_commands/UserCommands.cc
```diff
@@ -630,6 +713,125 @@ bool RemoveCommand::Execute()
...
+  auto entity = this->iface->ecm->EntityByComponents(
+      components::Name(lightMsg->name()),
+      components::ParentEntity(this->iface->worldEntity));
+
+  auto lightComp = this->iface->ecm->Component<components::Light>(entity);
+  if (nullptr == lightComp)
+  {
+    entity = kNullEntity;
+    // try to find the light inside a link
+    this->iface->ecm->Each<components::Name>(
```
iche033, line 743: this will loop through all the entities in the world since it's checking for entities that have a name. How about doing a search using `EntityByComponents` like above but only passing the filter `lightMsg->name()`? We'll then just verify that the entity's parent is a link or not:


```
    // try to find the light inside a link
    entity = kNullEntity;
    auto lightEnt = this->iface->ecm->EntityByComponents(
        components::Name(lightMsg->name()));
    if (lightEnt != kNullEntity)
    {
      // check if light parent is a link
      auto parentComp = this->iface->ecm->Component<components::ParentEntity>(
          lightEnt); 
      if (parentComp && this->iface->ecm->Component<components::Link>(
          parentComp->Data()))
      {
        lightComp = this->iface->ecm->Component<components::Light>(lightEnt);
        entity = lightEnt;
      }
```

iche033 (commented):
changing the sphere link's light now works for me! I have a suggestion about simplifying the search for lights attached to links.

iche033 (approved):

ahcorde:
@chapulina I added the integration test. can we merge this PR ?

include/ignition/gazebo/components/LightCmd.hh
```diff
@@ -0,0 +1,53 @@
...
+  using LightCmdSerializer =
+      serializers::ComponentToMsgSerializer<sdf::Light, msgs::Light>;
+}
+
+namespace components
+{
+  /// \brief A component type that contains commanded light of an
+  /// entity in the world frame represented by sdf::Light.
+  using LightCmd =
+    Component<sdf::Light, class LightCmdTag, serializers::LightCmdSerializer>;
```
chapulina, line 46: I'd recommend using `msgs::Light` as the data for this component. This may save some conversions.

src/systems/user_commands/UserCommands.cc
```diff
@@ -630,6 +714,94 @@ bool RemoveCommand::Execute()
...
+  auto lightCmdComp =
+    this->iface->ecm->Component<components::LightCmd>(entity);
+  if (!lightCmdComp)
+  {
+    this->iface->ecm->CreateComponent(
+        entity, components::LightCmd(lightComp->Data()));
+  }
+  else
+  {
+    /// \todo(anyone) Moving an object is not captured in a log file.
```
chapulina, line 794: This comment seems out of place

test/integration/user_commands.cc
```diff
@@ -671,3 +672,227 @@ TEST_F(UserCommandsTest, Pose)
...
   EXPECT_NEAR(500.0, poseComp->Data().Pos().Y(), 0.2);
 }
+
+/////////////////////////////////////////////////
+TEST_F(UserCommandsTest, Light)
+{
+  // Start server
+  ServerConfig serverConfig;
+  const auto sdfFile = std::string(PROJECT_SOURCE_PATH) +
+    "/test/worlds/lights.sdf";
```
chapulina, line 682: Use `common::joinPaths` for cross-platform compatibility

test/integration/user_commands.cc
```diff
@@ -671,3 +672,227 @@ TEST_F(UserCommandsTest, Pose)
...
+
+  server.AddSystem(testSystem.systemPtr);
+
+  // Run server and check we have the ECM
+  EXPECT_EQ(nullptr, ecm);
+  server.Run(true, 1, false);
+  EXPECT_NE(nullptr, ecm);
+
+  msgs::Light req;
+  req.set_name("point");
```
chapulina, line 706: What is this request supposed to do?

test/integration/user_commands.cc
```diff
@@ -671,3 +672,227 @@ TEST_F(UserCommandsTest, Pose)
...
+  transport::Node node;
+  EXPECT_TRUE(node.Request(service, req, timeout, res, result));
+  EXPECT_TRUE(result);
+  EXPECT_TRUE(res.data());
+
+  // Point light
+  auto pointLightEntity = ecm->EntityByComponents(components::Name("point"));
+  EXPECT_NE(kNullEntity, pointLightEntity);
+
+  // Check entity has not been moved yet
```
chapulina, line 722:
```suggestion
  // Check entity has not been edited yet
```

test/integration/user_commands.cc
```diff
@@ -671,3 +672,227 @@ TEST_F(UserCommandsTest, Pose)
...
+  ASSERT_NE(nullptr, pointLightComp);
+
+  EXPECT_EQ(math::Color(0, 1, 1, 0), pointLightComp->Data().Diffuse());
+  EXPECT_EQ(math::Color(0.2, 0.2, 0.2, 0.2),
+    pointLightComp->Data().Specular());
+  EXPECT_NEAR(2.6, pointLightComp->Data().AttenuationRange(), 0.1);
+  EXPECT_NEAR(0.7, pointLightComp->Data().LinearAttenuationFactor(), 0.1);
+  EXPECT_NEAR(0.6, pointLightComp->Data().ConstantAttenuationFactor(), 0.1);
+  EXPECT_NEAR(0.001, pointLightComp->Data().QuadraticAttenuationFactor(), 0.1);
+  EXPECT_TRUE(pointLightComp->Data().CastShadows());
```
chapulina, line 765: You can also check the type

src/gui/plugins/component_inspector/Light.qml
```diff
@@ -0,0 +1,1083 @@
...
+          width: margin + indentation
+        }
+
+        // Left spacer
+        Item {
+          Layout.rowSpan: 3
+          width: margin + indentation
+        }
+
+        Component {
```
chapulina, line 269: How about moving this component up top with the other ones? Having it inside the layout makes the layout contents a bit messier to understand.

src/gui/plugins/component_inspector/Light.qml
```diff
@@ -0,0 +1,1083 @@
...
+      maximumValue: 100000
+      decimals: 6
+      onEditingFinished: {
+        sendLight()
+      }
+    }
+  }
+
+  Component {
+    id: readOnlyNumber
```
chapulina, line 174: This component isn't used, I think you can remove it

src/gui/plugins/component_inspector/Light.qml
```diff
@@ -0,0 +1,1083 @@
...
+
+      Behavior on height {
+        NumberAnimation {
+          duration: 200;
+          easing.type: Easing.InOutQuad
+        }
+      }
+
+
+      GridLayout {
```
chapulina, line 251: I don't think a grid layout works well for this widget. See how the RGBA spin boxes are squished in a tiny cell, leaving a lot of empty space on their left, instead of expanding to use as much space as possible.

image

You can try increasing their `columnSpan` so they occupy that empty space. I recommend you draw the layout in a grid and see how many cells you expect each widget to occupy.

Another option is using a `ColumnLayout` with multiple other layouts inside of it, instead of a large `GridLayout`. This would give each item in the column more freedom, because their columns don't need to match. For example, the `Cast shadows` text is large and stretches that whole column right now. If you had separate layouts, that row wouldn't influence the others.

Also, another thing to consider while working on the layout would be to use sliders instead of spins for the 0-1 bounded values. Or maybe slider + spin.

Annnd another cool thing to do for colors is to use a `ColorDialog` like the `GridConfig` plugin does.

chapulina (changes requested):
I think some of the comments below need to be addressed before merging this.

src/rendering/RenderUtil.cc
```diff
@@ -235,6 +242,30 @@ rendering::ScenePtr RenderUtil::Scene() const
...
 }
+//////////////////////////////////////////////////
+void RenderUtil::UpdateECM(const UpdateInfo &,
+                           EntityComponentManager &_ecm)
+{
+  std::lock_guard<std::mutex> lock(this->dataPtr->updateMutex);
+
+  auto entityLightsCmdToDelete =
+    std::move(this->dataPtr->entityLightsCmdToDelete);
+  this->dataPtr->entityLightsCmdToDelete.clear();
```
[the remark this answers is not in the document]
ahcorde: no we can't, because there are two clases trying to get this entity, `Scene3D` and `RenderUtil`. The other variables are using the same logic have a look to the `Update` method

include/ignition/gazebo/components/LightCmd.hh
ahcorde: @chapulina, I tried to be coherent with the other CMD commands. 

 - AngularVelocityCmd -> math::Vector3d
 - JointForceCmd -> std::vector<double>
 - JointVelocityCmd -> std::vector<double>
 - LinearVelocityCmd -> math::Vector3D
 - PoseCmd -> math::Pose3d

none of them are using the message type

test/integration/user_commands.cc
ahcorde: This just add a light with the default parameters except the name and parent id

include/ignition/gazebo/components/LightCmd.hh
chapulina: The commands aren't using SDF types either :wink: We have a couple of components using messages to hold their data, but it's true that there are more using SDF. I think this is mainly historical though, we started thinking we'd use SDF more widely, but I think that's becoming less true. See #494 for more context.

The reason I suggested a message is that it matches better what the components do. They hold data and are serialized. Right now, using SDF, you're still relying on a message for serialization. If you just used the message directly in the type, then you save the SDF <-> msgs conversion by using a `MsgSerializer` directly.

luca-della-vedova:
Hi @ahcorde!

I was trying your PR and I found a small issue, specifically when a color is picked through the color palette all following manual adjustments through the sliders seem to default to the color picked in the palette, example below:

light_color_glitch

src/gui/plugins/plotting/Plotting.cc
```diff
@@ -382,6 +448,20 @@ void Plotting ::Update(const ignition::gazebo::UpdateInfo &_info,
...
+        msgs::Light lightMsgs = convert<msgs::Light>(comp->Data());
+        this->SetData(component.first, lightMsgs);
+      }
+    }
+    else if (typeId == components::LightCmd::typeId)
+    {
+      auto comp = _ecm.Component<components::LightCmd>(entity);
+      if (comp)
+        this->SetData(component.first, comp->Data());
+    }
```
luca-della-vedova, line 464: Do we need to have `LightCmd` components in Plotting and ComponentInspector? From my understanding they would be created on change request then deleted immediately afterwards?

ahcorde:
Thank you for your feedback @luca-della-vedova! The issue with the palette should be resolved with the last commit

src/gui/plugins/plotting/Plotting.cc
luca-della-vedova: [reopening since the same component was also added in `ComponentInspector`, just to confirm whether it is needed or not]

ahcorde: I think we don't need this in both places, we cannot plot this and this is not going to be shown in the Component inspector.

luca-della-vedova (approved):
Looks good to me thanks! For curiosity do you know what the specular value actually does for lights? I didn't see any noticeable difference when changing it.
Also I noticed that the pose component for lights is still read only but that is probably out of the scope of this PR.

iche033:
> For curiosity do you know what the specular value actually does for lights?

I think it affects how shiny the objects are but the final result also depends on the objects' own material properties too.

ahcorde:
friendly ping @chapulina

src/gui/plugins/component_inspector/ComponentInspector.cc
```diff
@@ -728,6 +779,60 @@ void ComponentInspector::OnPose(double _x, double _y, double _z, double _roll,
...
+
+  // if sdf::LightType::SPOT || sdf::LightType::DIRECTIONAL
+  if (_type == 1 || _type == 2)
+  {
+    ignition::msgs::Set(req.mutable_direction(),
+      ignition::math::Vector3d(_directionX, _directionY, _directionZ));
+  }
+
+  auto lightConfigService = "/world/" + this->dataPtr->worldName +
+    "/light_config";
```
chapulina, line 832: Let's run this through `AsValidTopic` just in case the world name has something weird in it

chapulina (commented):
Sorry, I had a review that I wrote weeks ago but never completed.

src/rendering/RenderUtil.cc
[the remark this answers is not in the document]
ahcorde: When I as coding this I find an issue with the rendering, as you mentioned there is no race conditions, but the `LightCmd` is used in two places: 3DScene and in the Sensor class by the camera sensor. 

If we remove the LightCMD in the 3DScene  then the camera sensor will not get the change. For this reason I added the ISystemUpdate in the Sensor class.

chapulina (commented):
I'm getting crashes when dragging the color spins, does it work for you? The backtrace points to this line:

`#8  0x00007f066472a290 in QV4::QObjectWrapper::getQmlProperty(QQmlContextData*, QV4::String*, QV4::QObjectWrapper::RevisionMode, bool*, bool) const () at /usr/lib/x86_64-linux-gnu/libQt5Qml.so.5`

src/systems/user_commands/UserCommands.cc
```diff
@@ -657,6 +759,115 @@ bool RemoveCommand::Execute()
...
+  auto lightMsg = dynamic_cast<const msgs::Light *>(this->msg);
+  if (nullptr == lightMsg)
+  {
+    ignerr << "Internal error, null light message" << std::endl;
+    return false;
+  }
+
+  Entity lightEntity{kNullEntity};
+
+  if (lightMsg->id() != 0)
```
chapulina, line 782: Can you also compare to `kNullEntity`? Like here:

https://github.com/ignitionrobotics/ign-gazebo/blob/f6a7329ad7a283034a54647d44c4c061666cfeb8/src/systems/user_commands/UserCommands.cc#L681

src/systems/user_commands/UserCommands.cc
```diff
@@ -657,6 +759,115 @@ bool RemoveCommand::Execute()
...
+    return false;
+  }
+
+  auto lightComp = this->iface->ecm->Component<components::Light>(lightEntity);
+  if (nullptr == lightComp)
+  {
+    lightEntity = kNullEntity;
+    // try to find the light inside a link
+    auto tempLightEnty = this->iface->ecm->EntityByComponents(
+      components::Name(lightMsg->name()));
```
chapulina, line 814: How's this different from line 796?

src/systems/user_commands/UserCommands.cc
```diff
@@ -657,6 +759,115 @@ bool RemoveCommand::Execute()
...
+    if (tempLightEnty != kNullEntity)
+    {
+      // check if light parent is a link
```
chapulina, line 817: It's not clear to me why this check is necessary. Lights can also be children of the world.

test/integration/model.cc
```diff
@@ -23,6 +23,8 @@
 #include <ignition/gazebo/EntityComponentManager.hh>
 #include <ignition/gazebo/Model.hh>
 #include <ignition/gazebo/components/Joint.hh>
+#include <ignition/gazebo/components/Light.hh>
+#include <ignition/gazebo/components/LightCmd.hh>
```
chapulina, line 27: Please remove these includes

src/gui/plugins/component_inspector/Light.qml
```diff
@@ -0,0 +1,1099 @@
...
+    font.family: "Roboto"
+  }
+
+  /**
+   * Used to create a spin box
+   */
+  Component {
+    id: sliderZeroOne
+    Slider {
+      id: writableSpin
```
chapulina, line 137: spin -> slider

chapulina (changes requested):
I'm not sure what's causing the crash. If you can't reproduce it locally I can try debugging it here. I'm on a Bionic Docker container.

src/gui/plugins/plotting/Plotting.hh
```diff
@@ -24,6 +24,8 @@
 #include <ignition/math/Vector3.hh>
 #include <ignition/math/Pose3.hh>
 
+#include <ignition/msgs/Utility.hh>
```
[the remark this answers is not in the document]
ahcorde: at least I should include `#include <ignition/msgs/light.pb.h>`

src/rendering/RenderUtil.cc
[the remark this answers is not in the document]
ahcorde: The `UpdateECM` is called in two classes. 
 - `Sensors`
https://github.com/ignitionrobotics/ign-gazebo/blob/5bafe4326cf18d5e4c886f8ae09e4403d0a422a1/src/systems/sensors/Sensors.cc#L416
 - `Scene3D`
 https://github.com/ignitionrobotics/ign-gazebo/blob/5bafe4326cf18d5e4c886f8ae09e4403d0a422a1/src/gui/plugins/scene3d/Scene3D.cc#L2703

When this method is called we need to `keep` the data for the other class o classes that may require a render update.
There are two variables (unfortunately with the same name, I can change this). 
 - The `entityLightsCmdToDelete` inside the pimpl pattern is used to keep the state for the following step in the loop. With this be asure that the LightCMD is visible for all the classes that require an update.
 - The local variable `entityLightsCmdToDelete` is used to remove older LigthCMD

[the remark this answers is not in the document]
ahcorde: I updated the name of the variables. The local variable now is called ` olderEntitiesLightsCmdToDelete`

ahcorde:
the user_command test is failing in GIthub actions because:
```
Error while loading the library [/usr/local/lib/ign-gazebo-4/plugins/libignition-gazebo-sensors-system.so]: libignition-gazebo4-rendering.so.4: cannot open shared object file: No such file or directory
```

But in the buildfarm is working fine. Any thoughts @j-rivero ?

test/integration/user_commands.cc
```diff
@@ -693,6 +694,236 @@ TEST_F(UserCommandsTest, Pose)
...
+  EXPECT_FALSE(directionalLightComp->Data().CastShadows());
+  EXPECT_EQ(sdf::LightType::DIRECTIONAL,
+    directionalLightComp->Data().Type());
+
+  // spot light
+  auto spotLightEntity = ecm->EntityByComponents(
+      components::Name("spot"));
+  EXPECT_NE(kNullEntity, spotLightEntity);
+
+  // Check entity has not been moved yet
```
chapulina, line 863: It's not going to be moved, right?

chapulina (commented):
I sometimes get into some weird situations with this branch:

* The application crashes while I'm dragging the slider to change a light color
* After I drag the slide, the values are updated, but then when I move the mouse away, the values go back to the previous value.

I haven't figured out how to reproduce these consistently though. Have you observed anything like this? I'm on a Bionic Docker container building the entire stack from source.

ahcorde:
> - The application crashes while I'm dragging the slider to change a light color

I'm trying to debug this issue with `gdb ruby` but when ign gazebo failed there is no backtrace. I will try to create a standalone executabale.

> - After I drag the slide, the values are updated, but then when I move the mouse away, the values go back to the previous value.

Can you share a gif of this issue ? I can't reproduce it

ahcorde:
@chapulina this is the backtrace:

```bash
#0  google::protobuf::Map<unsigned long, ignition::msgs::SerializedEntityMap>::iterator::operator* (this=0x7fff917f9840) at /usr/include/google/protobuf/map.h:1003
#1  0x00007ffff2c83e66 in google::protobuf::Map<unsigned long, ignition::msgs::SerializedEntityMap>::iterator::operator-> (this=0x7fff917f9840) at /usr/include/google/protobuf/map.h:1004
#2  0x00007ffff2c7aa3a in ignition::gazebo::v4::EntityComponentManagerPrivate::SetRemovedComponentsMsgs (this=0x5555559d5970, _entity=@0x7fff917f98e8: 28, _msg=...)
    at /home/ahcorde/ignition_dome/src/ign-gazebo/src/EntityComponentManager.cc:824
#3  0x00007ffff2c7ba51 in ignition::gazebo::v4::EntityComponentManager::AddEntityToMessage (this=0x555555a0adb0, _msg=..., _entity=28, _types=std::unordered_set with 1 element = {...}, _full=false)
    at /home/ahcorde/ignition_dome/src/ign-gazebo/src/EntityComponentManager.cc:983
#4  0x00007ffff2c7dfa6 in ignition::gazebo::v4::EntityComponentManager::<lambda(auto:4, auto:5)>::operator()<std::__detail::_Node_iterator<std::pair<long unsigned int const, std::unordered_map<long unsigned int, std::pair<long unsigned int, int> > >, false, false>, std::__detail::_Node_iterator<std::pair<long unsigned int const, std::unordered_map<long unsigned int, std::pair<long unsigned int, int> > >, false, false> >(std::__detail::_Node_iterator<std::pair<unsigned long const, std::unordered_map<unsigned long, std::pair<unsigned long, int>, std::hash<unsigned long>, std::equal_to<unsigned long>, std::allocator<std::pair<unsigned long const, std::pair<unsigned long, int> > > > >, false, false>, std::__detail::_Node_iterator<std::pair<unsigned long const, std::unordered_map<unsigned long, std::pair<unsigned long, int>, std::hash<unsigned long>, std::equal_to<unsigned long>, std::allocator<std::pair<unsigned long const, std::pair<unsigned long, int> > > > >, false, false>) const (__closure=0x7fffc0005dd8, itStart=..., itEnd=...)
    at /home/ahcorde/ignition_dome/src/ign-gazebo/src/EntityComponentManager.cc:1113
#5  0x00007ffff2c7ec09 in std::__invoke_impl<void, ignition::gazebo::v4::EntityComponentManager::State(ignition::msgs::SerializedStateMap&, const std::unordered_set<long unsigned int>&, const std::unordered_set<long unsigned int>&, bool) const::<lambda(auto:4, auto:5)>, std::__detail::_Node_iterator<std::pair<long unsigned int const, std::unordered_map<long unsigned int, std::pair<long unsigned int, int>, std::hash<long unsigned int>, std::equal_to<long unsigned int>, std::allocator<std::pair<long unsigned int const, std::pair<long unsigned int, int> > > > >, false, false>, std::__detail::_Node_iterator<std::pair<long unsigned int const, std::unordered_map<long unsigned int, std::pair<long unsigned int, int>, std::hash<long unsigned int>, std::equal_to<long unsigned int>, std::allocator<std::pair<long unsigned int const, std::pair<long unsigned int, int> > > > >, false, false> >(std::__invoke_other, ignition::gazebo::v4::EntityComponentManager::<lambda(auto:4, auto:5)> &&) (__f=...) at /usr/include/c++/9/bits/invoke.h:60
#6  0x00007ffff2c7eb54 in std::__invoke<ignition::gazebo::v4::EntityComponentManager::State(ignition::msgs::SerializedStateMap&, const std::unordered_set<long unsigned int>&, const std::unordered_set<long unsigned int>&, bool) const::<lambda(auto:4, auto:5)>, std::__detail::_Node_iterator<std::pair<long unsigned int const, std::unordered_map<long unsigned int, std::pair<long unsigned int, int>, std::hash<long unsigned int>, std::equal_to<long unsigned int>, std::allocator<std::pair<long unsigned int const, std::pair<long unsigned int, int> > > > >, false, false>, std::__detail::_Node_iterator<std::pair<long unsigned int const, std::unordered_map<long unsigned int, std::pair<long unsigned int, int>, std::hash<long unsigned int>, std::equal_to<long unsigned int>, std::allocator<std::pair<long unsigned int const, std::pair<long unsigned int, int> > > > >, false, false> >(ignition::gazebo::v4::EntityComponentManager::<lambda(auto:4, auto:5)> &&) (__fn=...) at /usr/include/c++/9/bits/invoke.h:95
#7  0x00007ffff2c7ea87 in std::thread::_Invoker<std::tuple<ignition::gazebo::v4::EntityComponentManager::State(ignition::msgs::SerializedStateMap&, const std::unordered_set<long unsigned int>&, const std::unordered_set<long unsigned int>&, bool) const::<lambda(auto:4, auto:5)>, std::__detail::_Node_iterator<std::pair<long unsigned int const, std::unordered_map<long unsigned int, std::pair<long unsigned int, int>, std::hash<long unsigned int>, std::equal_to<long unsigned int>, std::allocator<std::pair<long unsigned int const, std::pair<long unsigned int, int> > > > >, false, false>, std::__detail::_Node_iterator<std::pair<long unsigned int const, std::unordered_map<long unsigned int, std::pair<long unsigned int, int>, std::hash<long unsigned int>, std::equal_to<long unsigned int>, std::allocator<std::pair<long unsigned int const, std::pair<long unsigned int, int> > > > >, false, false> > >::_M_invoke<0, 1, 2>(std::_Index_tuple<0, 1, 2>) (this=0x7fffc0005dc8) at /usr/include/c++/9/thread:244
#8  0x00007ffff2c7ea26 in std::thread::_Invoker<std::tuple<ignition::gazebo::v4::EntityComponentManager::State(ignition::msgs::SerializedStateMap&, const std::unordered_set<long unsigned int>&, const std::unordered_set<long unsigned int>&, bool) const::<lambda(auto:4, auto:5)>, std::__detail::_Node_iterator<std::pair<long unsigned int const, std::unordered_map<long unsigned int, std::pair<long unsigned int, int>, std::hash<long unsigned int>, std::equal_to<long unsigned int>, std::allocator<std::pair<long unsigned int const, std::pair<long unsigned int, int> > > > >, false, false>, std::__detail::_Node_iterator<std::pair<long unsigned int const, std::unordered_map<long unsigned int, std::pair<long unsigned int, int>, std::hash<long unsigned int>, std::equal_to<long unsigned int>, std::allocator<std::pair<long unsigned int const, std::pair<long unsigned int, int> > > > >, false, false> > >::operator()(void) (this=0x7fffc0005dc8) at /usr/include/c++/9/thread:251
#9  0x00007ffff2c7ea0a in std::thread::_State_impl<std::thread::_Invoker<std::tuple<ignition::gazebo::v4::EntityComponentManager::State(ignition::msgs::SerializedStateMap&, const std::unordered_set<long unsigned int>&, const std::unordered_set<long unsigned int>&, bool) const::<lambda(auto:4, auto:5)>, std::__detail::_Node_iterator<std::pair<long unsigned int const, std::unordered_map<long unsigned int, std::pair<long unsigned int, int>, std::hash<long unsigned int>, std::equal_to<long unsigned int>, std::allocator<std::pair<long unsigned int const, std::pair<long unsigned int, int> > > > >, false, false>, std::__detail::_Node_iterator<std::pair<long unsigned int const, std::unordered_map<long unsigned int, std::pair<long unsigned int, int>, std::hash<long unsigned int>, std::equal_to<long unsigned int>, std::allocator<std::pair<long unsigned int const, std::pair<long unsigned int, int> > > > >, false, false> > > >::_M_run(void) (this=0x7fffc0005dc0) at /usr/include/c++/9/thread:195
#10 0x00007ffff1e09d84 in ?? () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#11 0x00007ffff7a14609 in start_thread (arg=<optimized out>) at pthread_create.c:477
#12 0x00007ffff7b50293 in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:95
```

https://github.com/ignitionrobotics/ign-gazebo/blob/555a84ea54a6f92b46929eb2f1a43daa13b1fadb/src/EntityComponentManager.cc#L1113
https://github.com/ignitionrobotics/ign-gazebo/blob/555a84ea54a6f92b46929eb2f1a43daa13b1fadb/src/EntityComponentManager.cc#L983
https://github.com/ignitionrobotics/ign-gazebo/blob/555a84ea54a6f92b46929eb2f1a43daa13b1fadb/src/EntityComponentManager.cc#L824

Do you think could be related with this PR ?

ahcorde:
@chapulina can you try this patch ? 

```c++
--- a/src/EntityComponentManager.cc
+++ b/src/EntityComponentManager.cc
@@ -809,6 +809,9 @@ void EntityComponentManagerPrivate::SetRemovedComponentsMsgs(Entity &_entity,
   // Find the entity in the message
   auto entIter = _msg.mutable_entities()->find(_entity);
 
+  if (entIter == _msg.mutable_entities()->end())
+    return;
+
   auto it = this->removedComponents.find(_entity);
   for (uint64_t i = 0; i < nEntityKeys; ++i)
   {
```

ahcorde:
I added the patch in the last commit with another segfault

iche033:
I ran into this crash in https://github.com/ignitionrobotics/ign-gazebo/pull/614 as well. There is a fix in https://github.com/ignitionrobotics/ign-gazebo/pull/495. We should try to get that merged.

ahcorde:
friendly ping @iche033 and @chapulina

tutorials/light_config.md
```diff
@@ -0,0 +1,45 @@
...
+cd build
+cmake ..
+make
+```
+
+Then you should open two terminals and run:
+
+ - Terminal one:
+ ```bash
+ ign gazebo -r -v 4 empty.world
```
chapulina, line 39:
```suggestion
 ign gazebo -r -v 4 empty.sdf
```

src/rendering/RenderUtil.cc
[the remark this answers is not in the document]
chapulina: Thank you for the clarifications. `Scene3D` and `Sensors` are running on separate processes and are dealing with different instances of the ECM. But I can see how this can become a problem if `Sensors` deletes the component before `SceneBroadcaster` has a chance to propagate the component to the client.

Thanks for walking through this!

src/systems/sensors/Sensors.hh
```diff
@@ -57,9 +57,9 @@ namespace systems
                            EntityComponentManager &_ecm,
                            EventManager &_eventMgr) final;
 
-    // Documentation inherited
-    public: void Update(const UpdateInfo &_info,
-                        EntityComponentManager &_ecm) final;
+   // Documentation inherited
+   public: void Update(const UpdateInfo &_info,
+                       EntityComponentManager &_ecm) final;
```
chapulina, line 62: This indentation change doesn't look right

chapulina (approved):
Thanks for iterating! I can't reproduce the weird behaviour and crash anymore. Feel free to merge once the last final details have been addressed. Be sure to also  sign all commits for DCO.

Thanks! :bulb:

PR #486, opened 2020-12-14
Use ComponentState::PeriodicChange in UpdateState to avoid forcing full scene update

ivanpauno:
When trying to implement primary-secondary async stepping in https://github.com/ignitionrobotics/ign-gazebo/pull/481, I found out that there was almost no speed-up in doing that because the primary was running quite slowly.

After some profiling I realized that the [SceneBroadcaster::PostUpdate](https://github.com/ignitionrobotics/ign-gazebo/blob/020a0172308eb5b64e4699b5afb25bb1105800ba/src/systems/scene_broadcaster/SceneBroadcaster.cc#L229) step was much slower than in an standalone simulation because a full scene update was [always forced](https://github.com/ignitionrobotics/ign-gazebo/blob/020a0172308eb5b64e4699b5afb25bb1105800ba/src/systems/scene_broadcaster/SceneBroadcaster.cc#L279).

This PR modifies `EntityComponentManager::SetState(const ignition::msgs::SerializedStateMap &_stateMsg)` so when a component is updated a `ComponentState::PeriodicChange` change is set instead of a `ComponentState::OneTimeChange`, avoiding to regenerate a full scene all the time.

With this change, the `distributed_levels` example works ~4 times faster when using a distributed simulation.

I'm not sure if using `ComponentState::PeriodicChange` is always correct, so we can consider the following alternatives:
- `NetworkManagerPrimary::Step` is not the only consumer of `EntityComponentManager::SetState`.
  If for the other consumers doing a `ComponentState::OneTimeChange` is not a problem and for `NetworkManagerPrimary::Step` using `ComponentState::PeriodicChange` is always ok, I could add an extra argument to the method.
- If for some components updates `ComponentState::OneTimeChange` make more sense and for others `ComponentState::PeriodicChange` make more sense, the we should add that information to the `SerializedStateMap` message.


mjcarroll:
> With this change, the distributed_levels example works ~4 times faster when using a distributed simulation.

Four times faster than a standalone sim, or the previous implementation?

ivanpauno:
> Four times faster than a standalone sim, or the previous implementation?

Than the previous implementation.

I don't think that the simple `distributed_levels` example can ever run faster in "distributed mode" than in "standalone mode", because the world is extremely simple.

ivanpauno:
Some actual numbers in my machine using the `distributed_levels` world

- Current master, standalone: RTF ~ 1000%
- Current master, distributed: RTF ~ 40%
- This PR, distributed: RTF ~ 150%

Using the `distributed_levels` world but replacing the sphere and the box with two [EXPLORER_X1_SENSOR_CONFIG_2](https://app.ignitionrobotics.org/OpenRobotics/fuel/models/EXPLORER_X1_SENSOR_CONFIG_2).

- Current master, standalone: RTF ~ 150%
- Current master, distributed: RTF ~ 18%
- This PR, distributed: RTF ~ 70%
- async stepping **without** this change, distributed: RTF ~ 20%
- async stepping **with** this change, distributed: RTF ~ 120%

Note: The RTF numbers above is an "average" I made from visually inspecting the RTF factor in the GUI, so don't get them too seriously ... it's just an rough approximation to understand the performance benefit.

nkoenig:
Could this be changed to target `citadel` --> `ign-gazebo3`?

ivanpauno:
> Could this be changed to target citadel --> ign-gazebo3?

I took a quick look and targeting `ign-gazebo3` branch doesn't look hard.
I will update the branch on Monday.

ivanpauno:
@nkoenig do you mean only this PR or also https://github.com/ignitionrobotics/ign-gazebo/pull/481?

ivanpauno:
Updated base branch to `ign-gazebo3`, I can open another one targeting `ign-gazebo4` if desired.
@nkoenig let me know if you want a version of https://github.com/ignitionrobotics/ign-gazebo/pull/481 targeting `ign-gazebo3` too.

ivanpauno:
@mjcarroll @chapulina friendly ping.
I'm not sure if the approach of this PR is correct or not, so it would be great to have feedback to know if we want to explore some of the alternatives mentioned in the PR description or if this is okay.

chapulina:
Nice catch. My concern is that important one-time changes may be missed if we make all changes period. For example, the color of a light is changed in one secondary, and another secondary misses it and keeps seeing the old color.

> we should add that information to the SerializedStateMap message.

Yeah I think this could work. We'd have to target this PR at Edifice though (`main`), because changing a message breaks ABI.

ivanpauno:
> Yeah I think this could work. We'd have to target this PR at Edifice though (main), because changing a message breaks ABI.

Done.
I only added one `is_periodic_change` field to the whole message, I could've added one field per component but I don't see the benefit of doing that.

Each time I change the base branch I bring more reviewers :joy:

chapulina:
> I only added one is_periodic_change field to the whole message

Is there a PR to `ign-msgs` too?

> I could've added one field per component but I don't see the benefit of doing that.

This is related to something @luca-della-vedova was looking into recently. They have a use case using animated actors which could take advantage of more granular `OneTimeChange`s, so that the state is only updated for the entities / components that had those changes. In very large worlds, it can be quite expensive to update thousands of components because one of them changed. So I think there could be value in having that field at the component level, but that may bring its own overhead.

ivanpauno:
> Is there a PR to ign-msgs too?

Now there is :smile: https://github.com/ignitionrobotics/ign-msgs/pull/119 .

> This is related to something @luca-della-vedova was looking into recently. They have a use case using animated actors which could take advantage of more granular OneTimeChanges, so that the state is only updated for the entities / components that had those changes

But in that case, the changes that aren't needed can be omitted from the state message, right?
I'm not sure if more granularity is useful in the message itself ...

chapulina:
> But in that case, the changes that aren't needed can be omitted from the state message, right?

Yeah good point, they shouldn't be there at all when they aren't needed. I'm just trying to think now of any downsides to marking periodic changes as one time changes. The receiving end of that state message won't be able to tell apart which are the one time changes. Are you concerned about the overhead of marking every single component?

In any case, I think it's worth it trying the current approach. We can always add more granularity later if needed.

chapulina:
Now that this requires `ign-msgs`'s `main` branch, this is needed before it can be merged: https://github.com/ignition-tooling/release-tools/issues/362

ivanpauno:
> The receiving end of that state message won't be able to tell apart which are the one time changes

Yes that's true, if there's a potential benefit of having that information in the message maybe it's better to add it now.
AFAIS there's no code currently telling apart periodic changes from one time changes, the only thing that seems to be used is  [HasOneTimeComponentChanges()](https://github.com/ignitionrobotics/ign-gazebo/blob/efe8ea15e304748c195b46f8697033682cc05d45/src/EntityComponentManager.cc#L428) (there's no code calling [ComponentState()](https://github.com/ignitionrobotics/ign-gazebo/blob/efe8ea15e304748c195b46f8697033682cc05d45/src/EntityComponentManager.cc#L384)).

> Are you concerned about the overhead of marking every single component?

It will add some overhead, but probably minimal. It requires N lookups to [this std::map](https://github.com/ignitionrobotics/ign-gazebo/blob/efe8ea15e304748c195b46f8697033682cc05d45/src/EntityComponentManager.cc#L78) while iterating the [components to add to the message](https://github.com/ignitionrobotics/ign-gazebo/blob/efe8ea15e304748c195b46f8697033682cc05d45/src/EntityComponentManager.cc#L890).

src/network/NetworkManagerSecondary.cc
```diff
@@ -167,6 +167,7 @@ void NetworkManagerSecondary::OnStep(
   msgs::SerializedStateMap stateMsg;
   if (!entities.empty())
     this->dataPtr->ecm->State(stateMsg, entities);
+  stateMsg.set_has_one_time_component_changes(this->dataPtr->ecm->HasOneTimeComponentChanges());
```
chapulina, line 170: This line is too long, the codechecker will probably complain

chapulina (approved):
>  there's no code currently telling apart periodic changes from one time changes

So far :wink: But we may be changing this soon for performance. In any case, I think this may not affect the internals of the `State` / `SetState` calls.

---

This looks ok for me, but we need to wait for the transition to `ign-msgs7` to be complete before merging (https://github.com/ignition-tooling/release-tools/issues/362).

src/network/NetworkManagerSecondary.cc
ivanpauno: Thanks, I thought that `colcon test` was running linters but that doesn't seem to be the case.

nkoenig:
> Updated base branch to `ign-gazebo3`, I can open another one targeting `ign-gazebo4` if desired.
> @nkoenig let me know if you want a version of #481 targeting `ign-gazebo3` too.

Sorry for the late reply, I meant both.

ivanpauno:
> Sorry for the late reply, I meant both.

I have pushed a branch with https://github.com/ignitionrobotics/ign-gazebo/pull/481 rebased for `citadel`
https://github.com/ignitionrobotics/ign-gazebo/tree/ivanpauno/secondary-async-stepping-and-primary-acknowledges-citadel and similar for the `ign-imgui` tool https://github.com/ivanpauno/ign_imgui/tree/citadel.
I'm still testing that though, I will comment again after checking it still works.

The instructions to build, run the simulation and see the results are [here](https://github.com/ignitionrobotics/ign-gazebo/pull/481#issuecomment-747727202), feel free to ask if that's not clear.

PS: #481 already includes an optimization similar to this PR, so this one doesn't need to be rebased to `ign-gazebo3` (and the approach taken here depeneds on https://github.com/ignitionrobotics/ign-msgs/pull/119).

ivanpauno:
> I'm still testing that though, I will comment again after checking it still works.

Tested, the `citadel` branches are working too.

chapulina:
re: citadel: This needs to target Edifice because of the new message field.

chapulina:
Let's get this in! We still have a couple months until the Edifice release to revert in case we notice something weird.

PR #493, opened 2020-12-16
Move deselect and preview termination to render thread

### LINKED ISSUE #488: Crash when pressing escape to cancel a model placement
As found [here](https://github.com/ignitionrobotics/ign-gazebo/pull/466/files#r539864204), if the user continues to select models, most simply just a shape from the top toolbar, and presses escape to cancel the action, a crash will eventually occur.

This is likely due to [these calls](https://github.com/ignitionrobotics/ign-gazebo/blob/ign-gazebo3/src/gui/plugins/scene3d/Scene3D.cc#L2839-L2840) being made from outside the render thread.  If we could have some sort of flag we set when an escape key release is detected and then have the render thread read that flag and perform the deselect and remove operations, this would probably suffice.

JShep1:
Fixes #488

Essentially, this PR moves some rendering logic triggered by a model spawn termination or entity deselection to the rendering thread.  Ignition was crashing every so often when escape was pressed while placing an entity but this should no longer occur.

### DIFF
--- a/src/gui/plugins/scene3d/Scene3D.cc
+++ b/src/gui/plugins/scene3d/Scene3D.cc
@@ -315,6 +315,9 @@ inline namespace IGNITION_GAZEBO_VERSION_NAMESPACE {
     /// \brief Flag to indicate whether the z key is currently being pressed
     public: bool zPressed = false;
 
+    /// \brief Flag to indicate whether the escape key has been released.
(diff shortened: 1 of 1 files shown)


nkoenig (approved):
Works for me. Rapidly clicked and escaped a few hundred times and didn't see a problem.

PR #495, opened 2020-12-17
Fix seg fault in `SetRemovedComponentsMsgs` when TransformControl tool is used

### LINKED ISSUE #483: Segmentation fault when using TransformControl tool in crowded worlds
When using the Transform Control plugin to move objects around, it "often" results in a segmentation fault in https://github.com/ignitionrobotics/ign-gazebo/blob/ign-gazebo4/src/EntityComponentManager.cc#L788 line with the following backtrace:

```
#0  ignition::gazebo::v4::EntityComponentManagerPrivate::SetRemovedComponentsMsgs (this=0x555555b08f60, _entity=@0x7fffce60b888: 22, _msg=...) at /home/luca/ws_citadel/src/ign-gazebo/src/EntityComponentManager.cc:794
#1  0x00007ffff38a0110 in ignition::gazebo::v4::EntityComponentManager::AddEntityToMessage (this=0x555555891650, _msg=..., _entity=<optimized out>, _types=..., _full=false) at /usr/include/c++/9/bits/unique_ptr.h:360
#2  0x00007ffff38a147e in ignition::gazebo::v4::EntityComponentManager::State (this=this@entry=0x555555891650, _state=..., _entities=std::unordered_set with 0 elements, _types=std::unordered_set with 1 element = {...}, _full=_full@entry=false)
    at /home/luca/ws_citadel/src/ign-gazebo/src/EntityComponentManager.cc:1031
#3  0x00007fffce658bd0 in ignition::gazebo::v4::systems::SceneBroadcaster::PostUpdate (this=0x55555cb2b350, _info=..., _manager=...) at /home/luca/ws_citadel/install/include/ignition/msgs6/ignition/msgs/serialized_map.pb.h:724
#4  0x00007ffff3944505 in ignition::gazebo::v4::SimulationRunner::<lambda()>::operator()(void) const (__closure=0x55555cc108e8) at /home/luca/ws_citadel/src/ign-gazebo/src/SimulationRunner.cc:430
#5  0x00007ffff2fdad84 in ?? () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#6  0x00007ffff7a20609 in start_thread (arg=<optimized out>) at pthread_create.c:477
#7  0x00007ffff7b5e293 in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:95

```

To reproduce:

- Run a world with many objects (I used actors_population with 100 actors).
- Select an actor and a transform tool (translation)
- Click on an arrow many times in a sequence.

It seems the reason for the crash is that, at least in this world, the [mutable_entities()](https://github.com/ignitionrobotics/ign-gazebo/blob/ign-gazebo4/src/EntityComponentManager.cc#L774) structure is empty, hence the find returns an iterator to `end()`.
I tried a hacky fix by doing early return if the entity is not found but that didn't sound like a sensible solution so I avoided creating a PR.

Here is an example:
Transform_segfault

I'm not familiar enough with the EntityComponentManager to know why this would happen so any input is welcome!

mrushyendra:
`_msg.mutable_entities()` need not necessarily contain the `_entity` as a key when `SetRemovedComponentsMsgs()` is called, for instance in `AddEntityToMessage()`. In the case where `_entity` has no components in `entityComponents` that changed, but does have some removed Components in `removedComponents`, this could lead to a seg fault when `AddEntityToMessage` calls `SetRemovedComponentsMsgs`.

Secondly, this is only tangentially related, but it looks like `AddEntityToMessage` also takes a list of component types `_types`:
https://github.com/ignitionrobotics/ign-gazebo/blob/2f7f33d4fa34657badec81558d5953fe7643769f/src/EntityComponentManager.cc#L849-L851

But this is currently ignored in the call to `SetRemovedComponentMsgs`. I was wondering if that was intentional, or if it should be modified as well?

Fixes #483


src/EntityComponentManager.cc
```diff
@@ -772,6 +772,15 @@ void EntityComponentManagerPrivate::SetRemovedComponentsMsgs(Entity &_entity,
 
   // Find the entity in the message
   auto entIter = _msg.mutable_entities()->find(_entity);
+  // Add the entity to the message, if not already added
```
iche033, line 775: can you add a small comment briefly explaining what you mentioned in the PR description?

iche033 (commented):
confirmed that this fixed the issue.

As for the second point, is the idea for passing in `_types` to skip adding components to the msg if they are not in _types?

src/EntityComponentManager.cc
mrushyendra: Done, in https://github.com/ignitionrobotics/ign-gazebo/pull/495/commits/8285e15ccfc2c82fa23f21a7eedec72e9d994f2f

mrushyendra:
> As for the second point, is the idea for passing in _types to skip adding components to the msg if they are not in _types?

Yeah, that's what I meant

luca-della-vedova (commented):
Just for another data point it also fixes the issue in our simulations

iche033:
>> As for the second point, is the idea for passing in _types to skip adding components to the msg if they are not in _types?

> Yeah, that's what I meant

I see. I'm not sure if we should apply the filter to removed components or not actually. Maybe @chapulina or @azeey may have a better idea?

azeey:
`SetRemovedComponentMsgs` was added as a [refactor](https://github.com/ignitionrobotics/ign-gazebo/pull/272#discussion_r465791416) of the component removal code in `AddEntityToMessage`. While the code was inside `AddEntityToMessage`, the entity would have been added to the message by the time it got to add removed components. So the fix in this PR is appropriate.

> But this is currently ignored in the call to SetRemovedComponentMsgs. I was wondering if that was intentional, or if it should be modified as well?

I don't believe it was intentional. I think adding the filter would be a good idea.

mrushyendra:
Just added the filter as well

iche033 (approved):
the latest changes look good to me

chapulina (commented):
All changes look good to me. I hope no users were expecting the `_types` filter not to be applied to the removed components, because that would break their expectation.

Do you think you could add some tests to check that the removed types are filtered correctly, @mrushyendra ? It would also be great to add a test for the situation where `SetRemovedComponentsMsgs` is called on an entity that doesn't have changed components. 

Thanks!

azeey (commented):
LGTM! I agree with @chapulina, tests would be great!

mrushyendra:
Added a couple of tests

src/EntityComponentManager_TEST.cc
```diff
@@ -2263,6 +2263,86 @@ TEST_P(EntityComponentManagerFixture, SerializedStateMsgAfterRemoveComponent)
...
+  manager.State(stateMsg);
+
+  // Check message
+  {
+    auto iter = stateMsg.entities().find(e1);
+    const auto &e1Msg = iter->second;
+    auto compIter = e1Msg.components().begin();
+
+    // Check number of components
+    EXPECT_EQ(e1Msg.components().size(), 2u);
```
azeey, line 2293: It's better to use `ASSERT_EQ` here to ensure the rest of the code doesn't run if the expected number of components are not in the container.

src/EntityComponentManager_TEST.cc
```diff
@@ -2263,6 +2263,86 @@ TEST_P(EntityComponentManagerFixture, SerializedStateMsgAfterRemoveComponent)
...
+  manager.State(stateMsg, entitySet, types, false);
+
+  // Check message
+  {
+    auto iter = stateMsg.entities().find(e1);
+    const auto &e1Msg = iter->second;
+    auto compIter = e1Msg.components().begin();
+
+    // Check number of components
+    EXPECT_EQ(e1Msg.components().size(), 1u);
```
azeey, line 2337: `EXPECT_EQ` -> `ASSERT_EQ` per my previous comment.

azeey (commented):
CI fails on macOS with the error
```
9: /Users/jenkins/workspace/ignition_gazebo-ci-pr_any-homebrew-amd64/ign-gazebo/src/EntityComponentManager_TEST.cc:2337: Failure
9: Expected equality of these values:
9:   e1Msg.components().size()
9:     Which is: 0
9:   1u
9:     Which is: 1
  9/156 Test   #9: UNIT_EntityComponentManager_TEST .....................***Exception: SegFault 14.53 sec
test 10
```
I've left comments to change `EXPECT_EQ` to `ASSERT_EQ`, but I'm not sure why the test failed.

mrushyendra:
The tests should be passing now. Had to change the `removedComponents` traversal to make use of `equal_range`, instead of using `find`, since the latter may not return an iterator to the first matching value in the multimap:
https://github.com/ignitionrobotics/ign-gazebo/blob/a82ab6d2c82b1b8858bab746540b4fc21b9aacf9/src/EntityComponentManager.cc#L794-L795

iche033:
#482 and #

adlarkin:
All tests are passing now (the only failure for macOS is for [`INTEGRATION_scene_broadcaster_system`](https://build.osrfoundation.org/job/ignition_gazebo-ci-pr_any-homebrew-amd64/4904/testReport/(root)/INTEGRATION_scene_broadcaster_system/test_ran/), which is unrelated to this PR). I am merging this.

PR #497, opened 2020-12-17
Limit scene broadcast publications when paused

nkoenig:
I don't think the scenebroadcaster needs to send frequent messages when paused.


chapulina:
I haven't tried this yet, but have you checked that manipulating models in the GUI while paused still works?

iche033:
I was able to move the models through the GUI when paused. On the other hand, setting the model pose through `ign service` did not move the model since the scene broadcaster is not publishing the new state to the gui. The model moves only after sim is unpaused.

nkoenig:
The `Physics` system removes the pose change component during `Update`. The `SceneBroadcaster's` `PostUpdate` then misses the `OneTimeChange` state.

I've add a `PreUpdate` function to the `SceneBroadcaster` that captures the change event in: 79c6a4392edb43dd77e549c94346ee31d945cb55. Moving an entity from the CLI should now trigger a broadcast event.

chapulina:
> I've add a PreUpdate function to the SceneBroadcaster that captures the change event

It looks like that broke some things. The problem of checking for changes on pre-update is that systems are adding their changes during that call, and we can't guarantee the order in which systems run. Try for example moving the `SceneBroadcaster` above the `Physics` and `UserCommands` systems on the SDF, and you'll see how the GUI is not working properly anymore, even the component inspector fails to show any data sometimes.

chapulina (changes requested):
> the component inspector fails to show any data sometimes.

This is related to the original change, not the pre-update one.

The GUI runner doesn't run its own loop, it just reacts to changes from the server. So all systems' `Update` functions are only called when there's a new state coming from the server. If we don't publish while paused, systems that rely on periodic `Update`s don't work properly. For example, selecting a new entity won't update the `ComponentInspector` completely until the next `Update` call.

I think this could be solved by changing the way GUI systems' update calls are triggered. Maybe we can add a loop to the `GuiRunner` so that it calls updates periodically even if there aren't new state messages from the server. Unfortunately, the `GuiRunner` isn't PIMPLized, so it will be tricky to change a lot in Citadel. Maybe we can target a change like that to Edifice.

nkoenig:
Before I go any further with this PR, I want to make sure the general direction is aligned with our goals. @chapulina , I not opposed to declining this if you think there is a different or better way to improve performance while paused.

chapulina:
Not publishing periodic changes while paused seems reasonable to me. We just need to adjust the GUI's expectations to match that. There may also be other downstream consumers of the state message who would be affected. I worry about such a big change in a released version, so I would suggest we target this at Edifice.

nkoenig:
> Not publishing periodic changes while paused seems reasonable to me. We just need to adjust the GUI's expectations to match that. There may also be other downstream consumers of the state message who would be affected. I worry about such a big change in a released version, so I would suggest we target this at Edifice.

I think I've addressed the problems in 17bb8f178c5cb97e8c741c5a4cdd10efda90b35a

src/systems/scene_broadcaster/SceneBroadcaster.cc
```diff
@@ -239,6 +242,16 @@ void SceneBroadcaster::Configure(
...
 }
 
+//////////////////////////////////////////////////
+void SceneBroadcaster::PreUpdate(const UpdateInfo &_info,
+    EntityComponentManager &_ecm)
+{
+  bool jumpBackInTime = _info.dt < std::chrono::steady_clock::duration::zero();
+  this->dataPtr->changeEvent =  _ecm.HasEntitiesMarkedForRemoval() ||
+        _ecm.HasNewEntities() || _ecm.HasOneTimeComponentChanges() ||
+        jumpBackInTime;
```
chapulina, line 252: @nkoenig , I think this comment is still valid: https://github.com/ignitionrobotics/ign-gazebo/pull/497#issuecomment-749839626

The `Has` functions aren't meant to be called during `PreUpdate`. Entities may be modified after this system's `PreUpdate`, and the result you get here becomes invalid.

nkoenig: I've tried moving the `scene-broadcaster` plugin to different positions in an sdf world. The component inspector and user commands seem to work for me. Do you have a specific example of the failure case?

nkoenig: @chapulina I think you mentioned that you had more information about this?

chapulina: I just took a quick look at this again. I see that the `changeEvent` was moved back to `PostUpdate` on 17bb8f178c5cb97e8c741c5a4cdd10efda90b35a, and that fixes the concern I had. Is the addition of `PreUpdate` still needed?

nkoenig: I think removing the preupdate will break setting poses of objects via a service call. See https://github.com/ignitionrobotics/ign-gazebo/pull/497#issuecomment-749650067

nkoenig: @chapulina , I confirmed that I need the `PreUpdate` in order to capture change events from service calls.

nkoenig: @chapulina , do you have any additional concerns?

chapulina: Ok, I was able to create a scenario where the current implementation fails. Order the systems as follows on the SDF:

1. `SceneBroadcaster`
2. `UserCommands`
3. `Physics`

What happens:

* `PreUpdate`:
    * `SceneBroadcaster` doesn't see any one time changes
    * `UserCommands` creates the `WorldPoseCmd` component
* `Update`:
    * `Physics` removes `WorldPoseCmd`
* `PostUpdate`:
    * `SceneBroadcaster` doesn't see any one time changes, doesn't sent message

I'm thinking of a workaround.

chapulina: Mind taking a look at https://github.com/ignitionrobotics/ign-gazebo/pull/653?

chapulina (approved):

PR #500, opened 2020-12-18
Update gtest to 1.10.0 for Windows compilation

j-rivero:
Windows fails to compile the old version of gtest inside ign-gazebo. Updating here to 1.10.0, which is the version present in ign-common.


chapulina:
Thanks, @j-rivero ! Do you think it would be possible to retarget to `ign-gazebo3` and we merge it forward?

j-rivero:
> Thanks, @j-rivero ! Do you think it would be possible to retarget to `ign-gazebo3` and we merge it forward?

Editing was not going to work, closing and re-opening.

PR #501, opened 2020-12-18
Compilation fixes for Windows - codecheck patches (part 1)

j-rivero:
This is the first part of Windows fixes to make ign-gazebo compile, commits are self-explanatory but in a summary:

 * Missing headers, including ones reported by codecheck
 * Portable setenv from ignition::common
 * Visibility keywords
 * Portability fixes for cmake

Compilation is not yet there, this is the first part of patches. Let's see how it goes for UNIX ci.


chapulina:
Nice! This one too, do you think it would be possible to target `ign-gazebo3`?

j-rivero:
> Nice! This one too, do you think it would be possible to target `ign-gazebo3`?

I need to check if visibility is affected somehow with these changes on top of `ign-gazebo3`

src/systems/CMakeLists.txt
```diff
@@ -61,12 +61,14 @@ function(gz_add_system system_name)
...
   # files, we also install an unversioned symlink into the same versioned folder.
   set(versioned ${CMAKE_SHARED_LIBRARY_PREFIX}${system_target}${CMAKE_SHARED_LIBRARY_SUFFIX})
-  set(unversioned ${CMAKE_SHARED_LIBRARY_PREFIX}${PROJECT_NAME_NO_VERSION_LOWER}-${system_name}${CMAKE_SHARED_LIBRARY_SUFFIX})
+  set(unversioned_file_name ${PROJECT_NAME_NO_VERSION_LOWER}-${system_name}${CMAKE_SHARED_LIBRARY_SUFFIX})
+  set(unversioned ${CMAKE_SHARED_LIBRARY_PREFIX}${unversioned_file_name})
   if (WIN32)
-    # create_symlink requires cmake 3.13 on windows
-    cmake_minimum_required(VERSION 3.13 FATAL_ERROR)
+    # symlinks on Windows require admin priviledges, fallback to copy
+    install(FILES ${versioned} DESTINATION ${CMAKE_SHARED_LIBRARY_PREFIX} RENAME ${unversioned_file_name})
```
chapulina, line 68: Can we install to `IGNITION_GAZEBO_PLUGIN_INSTALL_DIR` instead of `CMAKE_SHARED_LIBRARY_PREFIX`?

src/systems/multicopter_control/LeeVelocityController.hh
```diff
@@ -49,7 +49,7 @@ namespace multicopter_control
   /// https://github.com/ethz-asl/rotors_simulator/blob/master/rotors_control/include/rotors_control/lee_position_controller.h
   /// The controller can be used to command linear velocity and yaw angle
   /// velocity expressed in the body frame.
-  class LeeVelocityController
+  class IGNITION_GAZEBO_VISIBLE LeeVelocityController
```
chapulina, line 52: This isn't installed, so I think this shouldn't be needed?

include/ignition/gazebo/EventManager.hh
```diff
@@ -47,7 +47,7 @@ namespace ignition
     /// occur.
     ///
     /// See \ref ignition::gazebo::events for a complete list of events.
-    class IGNITION_GAZEBO_VISIBLE EventManager
+    class EventManager
```
chapulina, line 50: I guess this is related to https://github.com/ignitionrobotics/ign-gazebo/issues/545. I think we need a different fix, because Windows users will need to see and use this class.

The same goes for the other places where `IGNITION_GAZEBO_VISIBLE` is being removed.

chapulina (commented):
The bump to `ign-msgs7` and `ign-transport10` is being done in #546

src/systems/multicopter_control/LeeVelocityController.hh
j-rivero: Uh, I don't think that binary symbols work in the same way than development headers.

This file is part of the `multicopter-control` system library since MulticopterVelocityControl.hh -> #include "LeeVelocityController.hh". MulticopterVelocityControl.cc -> #include "MulticopterVelocityControl.hh" so the class symbols are going to be in the public ABI of the library. 

```
/usr/lib/x86_64-linux-gnu🔒 ❯ nm -D --demangle  libignition-gazebo3-multicopter-control-system.so | grep LeeVelocityController
00000000000796a0 T ignition::gazebo::v3::systems::multicopter_control::LeeVelocityController::MakeController(ignition::gazebo::v3::systems::multicopter_control::LeeVelocityControllerParameters const&, ignition::gazebo::v3::systems::multicopter_control::VehicleParameters const&)
00000000000781b0 T ignition::gazebo::v3::systems::multicopter_control::LeeVelocityController::InitializeParameters()
0000000000077d40 T ignition::gazebo::v3::systems::multicopter_control::LeeVelocityController::CalculateRotorVelocities(ignition::gazebo::v3::systems::multicopter_control::FrameData const&, ignition::gazebo::v3::systems::multicopter_control::EigenTwist const&, Eigen::Matrix<double, -1, 1, 0, -1, 1>&) const
0000000000077640 T ignition::gazebo::v3::systems::multicopter_control::LeeVelocityController::ComputeDesiredAngularAcc(ignition::gazebo::v3::systems::multicopter_control::FrameData const&, ignition::gazebo::v3::systems::multicopter_control::EigenTwist const&, Eigen::Matrix<double, 3, 1, 0, 3, 1> const&) const
00000000000774a0 T ignition::gazebo::v3::systems::multicopter_control::LeeVelocityController::ComputeDesiredAcceleration(ignition::gazebo::v3::systems::multicopter_control::FrameData const&, ignition::gazebo::v3::systems::multicopter_control::EigenTwist const&) const
```

j-rivero:
> The bump to `ign-msgs7` and `ign-transport10` is being done in #546

Thanks, I've merge main into this PR.

src/systems/CMakeLists.txt
j-rivero: In fact, in my latest branch I've removed completely the this from Windows since I liked to assume  from comment
`for portability of SDF files, we also install an unversioned symlink into the same versioned folder.` that Windows don't need any portability since it was not working on Windows before. Is this correct?

chapulina: The portability mentioned there is between Ignition versions in the past and future. For example, if we don't have unversioned libraries, this SDF file would need to hardcode the Gazebo version as `ignition-gazebo4-physics-system`:

https://github.com/ignitionrobotics/ign-gazebo/blob/bccea9087dd4355fc6d273d41a191d5c16563175/examples/worlds/empty.sdf#L60-L63

src/systems/multicopter_control/LeeVelocityController.hh
chapulina: We're not expecting users to be linking against these plugin libraries, so we don't need to care about their ABI, right?

j-rivero: It wouldn't be the first time I see people linking against plugins but it can be consider as an unsupported use case. I don't have too much experience with plugins to know if there are more implications but with my knowledge I agree on what you say.

j-rivero:
> Nice! This one too, do you think it would be possible to target `ign-gazebo3`?

I ported the changes to ign-gazebo3 branch and If I did not fail during the process [there are problems with ABI stability](https://build.osrfoundation.org/job/ignition_gazebo-abichecker-any_to_any-ubuntu_auto-amd64/2983/API_5fABI_20report/). I'm afraid that changes are too intrusive to be backported.

src/systems/multicopter_control/LeeVelocityController.hh
chapulina: We aren't installing the headers, so I'm not sure how a downstream project would be able to use it. The intention is not to support this use case. So far, all Ignition plugins aren't meant to be extended.

chapulina:
> I'm afraid that changes are too intrusive to be backported.

Ah ok, thanks for trying. Let's keep targeting Edifice then.

src/systems/multicopter_control/LeeVelocityController.hh
j-rivero: removed visibility from all the plugins in a different PR. I did the same here.

src/systems/CMakeLists.txt
j-rivero: Ah thanks. I've modified the code to make it to copy versioned plugins to unversioned in 2a9836f

chapulina (approved):

PR #503, opened 2020-12-18
Re-enable focal and attempt to use tooling images

mjcarroll:
Leaving this draft, as I'm expecting some errors

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,16 +11,16 @@ jobs:
         uses: actions/checkout@v2
       - name: Compile and test
         id: ci
-        uses: ignition-tooling/action-ignition-ci@master
+        uses: ignition-tooling/action-ignition-ci@ign-tooling-bionic
         with:
           codecov-token: ${{ secrets.CODECOV_TOKEN }}
   # TODO(anyone) Enable Focal CI and fix failing tests
-  # focal-ci:
-  #   runs-on: ubuntu-latest
-  #   name: Ubuntu Focal CI
-  #   steps:
-  #     - name: Checkout
-  #       uses: actions/checkout@v2
-  #     - name: Compile and test
-  #       id: ci
-  #       uses: ignition-tooling/action-ignition-ci@focal
+  focal-ci:
+    runs-on: ubuntu-latest
+    name: Ubuntu Focal CI
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Compile and test
+        id: ci
+        uses: ignition-tooling/action-ignition-ci@ign-tooling-focal


.github/workflows/ci.yml
```diff
@@ -7,20 +7,32 @@ jobs:
...
+        with:
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.CR_PAT }}
       - name: Checkout
         uses: actions/checkout@v2
       - name: Compile and test
         id: ci
-        uses: ignition-tooling/action-ignition-ci@master
+        uses: ignition-tooling/action-ignition-ci@ign-tooling-bionic
```
chapulina, line 20: I noticed that this branch is installing Citadel dependencies:

https://github.com/ignition-tooling/action-ignition-ci/compare/master...ign-tooling-bionic

But this PR is targeted at Dome. Maybe we'll need to create separate branches for each distro? I think it would be good to keep using the `master` / `focal` branches if we could, so that all libraries will benefit from the change without us having to open PRs for all of them.

chapulina (commented):
Thanks for remembering that Focal CI needs to be re-enabled! I think that we also need to fix all the Focal warnings / tests.

> This also makes use the docker container actions that are set up here: https://github.com/ignition-tooling/setup-ign-docker

Mind adding some information to that repository's README? It's hard to know what it's doing without digging through the code.

One thing I'm wondering about is how often these images are updated. For example, if we made a release 1 hour ago, will it be picked up?

mjcarroll:
> Mind adding some information to that repository's README? It's hard to know what it's doing without digging through the code.

Yep! Can do.

> One thing I'm wondering about is how often these images are updated. For example, if we made a release 1 hour ago, will it be picked up?

They are currently built hourly.  There is a chance that a release could be missed, but it should be picked up in the next build.

.github/workflows/ci.yml
mjcarroll: I can open a PR in action-ignition-ci to incorporate those changes in the `master` and `focal` branches.

As far as the dependencies, I don't think that the distribution has to exactly match, since all we are doing is installing `apt` packages.  Since we do `apt update` and `install` during the action itself, it should pick up any different dependencies that are necessary.

The point of these containers is to just serve as a cache to speed up a bulk of the apt installs that we have.  They are still flexible to be extended upon.

chapulina: Cool, that sounds good to me :+1:

chapulina:
Regular Focal CI was added in #646 and is being ported forward in #792. Once that's in, this PR can be updated to just use the tooling images (no need to handle the addition of Focal).

.github/workflows/ci.yml
chapulina: > I think it would be good to keep using the master / focal branches if we could, so that all libraries will benefit from the change without us having to open PRs for all of them.

Come to think of it, we'll need to open PRs to all libraries anyway, to add the `Login to GitHub Container Registry` step, right? Maybe keeping separate `ign-tooling-*` branches makes the transition easier. We just need to keep those in sync with `bionic` and `focal`

.github/workflows/ci.yml
```diff
@@ -7,19 +7,31 @@ jobs:
     runs-on: ubuntu-latest
     name: Ubuntu Bionic CI
     steps:
+      - name: Login to GitHub Container Registry
```
chapulina, line 10: Is there a way we can skip this step? Can the container be publicly available, or shared with the `ignitionrobotics` org? This way, we just need to update the `action-ignition-ci` repository and don't need to add this step to all branches of all libraries.

PR #513, opened 2020-12-23
Allow to create light with the create service

ahcorde:
This PR will allow to create lights with the service `/world/<world_name>/create` 

I'm writting a tutorial to control lights (create, modify and delete).


ahcorde:
By the way, I know there are some `Convert` funtions on `ign-msgs` but it does not depend on `sdformat`. The conversion that I need is between `ignition::msgs::Light` and `sdf::Light`. where is the right place to locate this funtion ? Some can idea can be applied to `Set`.

iche033:
> By the way, I know there are some `Convert` funtions on `ign-msgs` but it does not depend on `sdformat`. The conversion that I need is between `ignition::msgs::Light` and `sdf::Light`. where is the right place to locate this funtion ? Some can idea can be applied to `Set`.

Is this what you are looking for?
https://github.com/ignitionrobotics/ign-gazebo/blob/ign-gazebo4/src/Conversions.cc#L444

ahcorde:
> Is this what you are looking for?
> https://github.com/ignitionrobotics/ign-gazebo/blob/ign-gazebo4/src/Conversions.cc#L444

yes, https://github.com/ignitionrobotics/ign-gazebo/pull/513/commits/a677a912a51952580133401c4c21d4b0e8594201

src/systems/user_commands/UserCommands.cc
```diff
@@ -414,9 +414,31 @@ bool CreateCommand::Execute()
...
+          components::Name(lightMsg.name()),
+          components::ParentEntity(lightMsg.parent_id())))
+      {
+        ignwarn << "Light entity named [" << lightMsg.name() << "] already"
+                << " exists. Entity not spawned."
+                << std::endl;
+        return false;
+      }
+
+      sdf::Light lightSDF = convert<sdf::Light>(lightMsg);
```
chapulina, line 429: How about just populating `root` with `lightSDF` and letting the rest of this function handle things like naming and pose? You'll probably need to set `isLight` to true. This way there's less duplication and fields like `allow_renaming` are also used.

chapulina (changes requested):

src/systems/user_commands/UserCommands.cc
```diff
@@ -414,9 +414,81 @@ bool CreateCommand::Execute()
...
+        "<cast_shadows>" +
+        std::to_string(lightMsg.cast_shadows()) +
+        "</cast_shadows>" +
+        "<pose>" +
+        std::to_string(sdfLight.RawPose().Pos().X()) + " " +
+        std::to_string(sdfLight.RawPose().Pos().Y()) + " " +
+        std::to_string(sdfLight.RawPose().Pos().Z()) + " " +
+        std::to_string(sdfLight.RawPose().Rot().Roll()) + " " +
+        std::to_string(sdfLight.RawPose().Rot().Pitch()) + " " +
+        std::to_string(sdfLight.RawPose().Rot().Yaw()) +
```
iche033, line 448: nothing to change here but just a note that many ign math classes have the `<<` operator so if you use `std::stringstream`, you can just do `ss << sdfLight.RawPose()`

src/systems/user_commands/UserCommands.cc
```diff
@@ -414,9 +414,81 @@ bool CreateCommand::Execute()
...
+                << std::endl;
+        return false;
+      }
+
+      sdf::Light sdfLight = convert<sdf::Light>(lightMsg);
+
+      std::string lightType = "point";
+      if (lightMsg.type() == msgs::Light::SPOT)
+        lightType = "spot";
+      if (lightMsg.type() == msgs::Light::DIRECTIONAL)
```
iche033, line 434: nit: `else if`

iche033 (commented):
Looks like the `UserCommndsTest.Create` integration test is failing in the latest build:
https://build.osrfoundation.org/job/ignition_gazebo-ci-pr_any-ubuntu_auto-amd64/5097/testReport/junit/(root)/UserCommandsTest/Create/

ahcorde:
@iche033 I fixed the test https://github.com/ignitionrobotics/ign-gazebo/pull/513/commits/2d20881652e3b4df2bf1f81144db2c9121f9d394

But I'm curious why is not possible to add more than one light. The SDF format allows to have more than one light as a top-level element. Do you know why this rectriction is here ?

https://github.com/ignitionrobotics/ign-gazebo/blob/66536bb27b1e2a7c57730ce1aa8d49c7777f2b37/src/systems/user_commands/UserCommands.cc#L528-L548

iche033 (commented):
> But I'm curious why is not possible to add more than one light.
 I recall @brawner and @azeey mentioning that we are now limiting the sdf Root to have only contain 1 model/actor/light: https://github.com/osrf/sdformat/pull/444


I also noticed one issue with the light sdf string. Adding quotes around the sdf version number seems to have removed one sdf parsing error msg and increased the entity count in the test.

brawner:
>> But I'm curious why is not possible to add more than one light.

> I recall @brawner and @azeey mentioning that we are now limiting the sdf Root to have only contain 1 model/actor/light: osrf/sdformat#444

Just to summarize some of the thinking...

Worlds can continue to contain more than 1 of all those items, this just affects `sdf::Root`. The motivation is that including an sdf file using the `<include>` tag leads to ambiguity for the elements in the `<include>` that override elements in the included sdf. For example, which name would the `name` element in the `<include>` tag override if there were more than one model, light or actor in the included file? 

In many places in the code base it was already assumed there was only one model/light/actor element in an sdf root element or included file. For now, we've deprecated that functionality with warnings and in the future it will become a hard error.

iche033:
I think we can now revert the change 2d20881652e3b4df2bf1f81144db2c9121f9d394 for the test now the light is being spawned after. This would fix the latest [failing test](https://build.osrfoundation.org/job/ignition_gazebo-ci-pr_any-ubuntu_auto-amd64/5116/testReport/)

iche033:
> I think we can now revert the change 2d20881 for the test now the light is being spawned after. This would fix the latest failing test

I just did this in de02da50d751fb0c435156ac39a9ccc009c93c16. Let me know if that's not correct. Approving if builds turn green

ahcorde:
@iche033 CI is green, there are some flaky tests in the buildfarm

iche033 (approved):

src/systems/user_commands/UserCommands.cc
```diff
@@ -424,9 +424,86 @@ bool CreateCommand::Execute()
...
+          components::Name(lightMsg.name()),
+          components::ParentEntity(lightMsg.parent_id())))
+      {
+        ignwarn << "Light entity named [" << lightMsg.name() << "] already"
+                << " exists. Entity not spawned."
+                << std::endl;
+        return false;
+      }
+
+      sdf::Light sdfLight = convert<sdf::Light>(lightMsg);
```
chapulina, line 439: Hum, I think this conversion doesn't make sense if we can't use `sdfLight` directly. We could construct the string below using the fields from the message, there's no need to convert to SDF first.

I now realize that there's no way to add a light to `sdf::Root` without converting it to string, and there's no way to get a  string if a light didn't come from an element (https://github.com/osrf/sdformat/issues/468) :confused:

chapulina (changes requested):
I have some suggestions on #569 on how to address my comments below.

chapulina (approved):

Issue #494, opened 2020-12-17
Data storage classes


chapulina:
One of the challenges with Ignition's modular architecture is to create data structures and convert between them across all libraries.

## Types everywhere

For example, Ignition Math is used by pretty much all libraries, from SDFormat to Rendering and Physics, to store math data types like vectors. But that's not the only place where those types exist. For most Ignition Math types, we also have equivalent Ignition Msgs types, so that the data can be transmitted over the wire.

Sure, that's not a terrible situation, and each class has a different use case. But let's look at something more complex, like a 3D mesh.

We start from [sdf::Mesh](https://github.com/osrf/sdformat/blob/master/include/sdf/Mesh.hh) which stores some basic metadata like file path and scale. That's pretty much the same as what [ignition::msgs::MeshGeom](https://github.com/ignitionrobotics/ign-msgs/blob/ign-msgs6/proto/ignition/msgs/meshgeom.proto) holds. But the actual 3D data, like vertices and normals, is held by [ignition::common::Mesh](https://github.com/ignitionrobotics/ign-common/blob/ign-common3/graphics/include/ignition/common/Mesh.hh), which is used by both Physics and Rendering. But by itself, it doesn't include all the data that those libraries need. Ignition Rendering has a wrapper that adds extra information: [ignition::rendering::MeshDescriptor](https://github.com/ignitionrobotics/ign-rendering/blob/ign-rendering4/include/ignition/rendering/MeshDescriptor.hh). And Ignition Physics needs more information to be passed alongside `common::Mesh` to its [ignition::physics::mesh::AttachMeshFeature](https://github.com/ignitionrobotics/ign-physics/blob/5845793f6c431dd7f11e6a33a0bd01fcadd65fb5/mesh/include/ignition/physics/mesh/MeshShape.hh#L125-L130).

[...]

PR #516, opened 2020-12-23
Particle system - Part 1

caguero:
It depends on https://github.com/ignitionrobotics/ign-msgs/pull/127 .

The goal is to integrate a particle system that can use the particle emitters available in Ignition Rendering.

This patch focuses on the creation of a new `ParticleEmitter` component.


include/ignition/gazebo/components/ParticleEmitter.hh
```diff
@@ -0,0 +1,186 @@
...
+
+  /// \brief Properties of a particle emitter.
+  /// A particle emitter also has:
+  ///    * a pose, which can be stored as a component of
+  ///      an emitter entity via ignition::gazebo::components::Pose
+  ///    * a name, which can be stored as a component of
+  ///      an emitter entity via ignition::gazebo::components::Name
+  ///    * a material, which can be stored as a component of
+  ///      an emitter entity via ignition::gazebo::components::Material
+  struct Emitter
```
chapulina, line 50: So far we haven't been creating entirely new classes to hold component data because there's usually already a class in some other library that can be used here and we don't duplicate information. I'd be curious to see the rest of the implementation to understand how this will fit with the rest of the code. Are we creating a message to hold this data? How about an SDF class? Another thing to keep in mind is that this header-only struct can't be extended with new fields without breaking ABI.

All this questioning is related to #494 :upside_down_face:

caguero: @iche033 and I brainstormed about it when we were thinking about the particle integration and we thought this was the best alternative. I followed the same approach that we did with `LogicalAudio` in pull request #401 . 

The plan is to use an Ignition Message to store this data (there's already a [branch](https://github.com/ignitionrobotics/ign-msgs/blob/particle_emiter/proto/ignition/msgs/particle_emiter.proto) for it) and use a component that can be serialized to reach the clients. Then, in the rendering thread, use the Ignition Rendering API to create the particle emitter based on the request.

I wasn't planning to create an SDF class. Our idea is th

iche033: We discussed whether or not we want to commit to adding this feature to the official sdf spec and create the DOM classes. But in the end, we thought adding to ign gazebo first seems like a less disruptive option that we could do in Dome for subt. 

I'm not opposed to adding to sdf spec. This could be a chance for something like `<visual><geometry><particles>` but we may want to include the sdf folks in the discussion / design

chapulina: >  I followed the same approach that we did with LogicalAudio

Oh ok I hadn't realized that was the approach there. My suggestion would be to use `ignition::msgs::ParticleEmitter` as the data for the component. That should be more efficient to serialize, and we don't need to worry about keeping the message and the component in sync. It looks like the new class isn't doing more than the message would, so I think there isn't a need for it.

> adding this feature to the official sdf spec

Yeah I think it makes sense to configure this feature through the `<plugin>` tag for now.

caguero: Thanks for the suggestion, it looks better in the code to me. How about now? Is this what you were thinking?

chapulina: Much closer! Now that all the data is contained in the message, I think we could get rid of the `ignition::gazebo::particles::Emitter` class and just use the message directly in the component, right? Even if we think we may need to extend the class later, I think it would break ABI just like extending the message, so we can just extend the message.

When you use the message as the data, you don't need the custom operators, the custom serializer... It should be possible to define the component in one line like this:

https://github.com/ignitionrobotics/ign-gazebo/blob/a830b443989877b1e83170390f483b6c02952b25/include/ignition/gazebo/components/ExternalWorldWrenchCmd.hh#L41-L43

(also, let's not forget to bump the required `ign-msgs` version before merging this PR)

caguero: Done in 4cbc9dd! How about now?

chapulina (approved):
Thanks for iterating!

chapulina:
@caguero , would you prefer merging these PRs as they are approved, or should we wait until all the functionality is working and approved before merging all tutorials? Note that this is targeting a stable release branch, and whatever is merged in an intermediary state may be released at any point.

caguero:
> @caguero , would you prefer merging these PRs as they are approved, or should we wait until all the functionality is working and approved before merging all tutorials? Note that this is targeting a stable release branch, and whatever is merged in an intermediary state may be released at any point.

We can review part1, part2 and part3 and merge them all at once when we're happy with it. These three PRs offer the basic integration. Part4 (coming up) is more independent.

chapulina:
> We can review part1, part2 and part3 and merge them all at once when we're happy with it.

Got it. I converted this to draft so that we don't merge it by mistake.

adlarkin:
I'm noticing the following error in CI:

```
  /github/workspace/include/ignition/gazebo/components/ParticleEmitter.hh:20:10: fatal error: ignition/msgs/particle_emitter.pb.h: No such file or directory
   #include <ignition/msgs/particle_emitter.pb.h>
```

I believe that we need to make an `ign-msgs6` release. cc @chapulina @iche033

chapulina:
> ign-msgs6 release.

I'm fine with a release, my only request is that we make sure the current particle message has all we need, because this will go into a stable release, and we won't be able to tweak any fields afterwards. So maybe it's best to wait until all parts are out beforehand. But if @iche033 thinks the implementation is already settled, than go for it.

adlarkin:
> my only request is that we make sure the current particle message has all we need, because this will go into a stable release, and we won't be able to tweak any fields afterwards.

Good point. I am still reviewing the other particle system PRs, so I will make a release once we know all of the other PRs are good to go :+1:

iche033:
yeah let's hold off until we have everything reviewed and ready to go. There's still some work left that need to be on on particles this week.

adlarkin:
@iche033 there was a merge conflict in https://github.com/ignitionrobotics/ign-gazebo/pull/516/commits/5885e45de53dea0833e8dea440e83e154eef2fc9 that I resolved. It looks like #482 added some code to `RenderUtil::UpdateECM` to update lights, so I kept that addition along with the additions we made to `RenderUtil::UpdateECM`.

src/systems/particle_emitter/ParticleEmitter.cc
```diff
@@ -0,0 +1,335 @@
...
+      int counter = 0;
+      while (emitterNames.find(name) != emitterNames.end())
+      {
+        name = emitterName + "_" + std::to_string(++counter);
+      }
+      ignmsg << "Entity named [" << emitterName
+             << "] already exists. Renaming it to " << name << std::endl;
+    }
+  }
+  this->dataPtr->emitter.set_name(name);
```
adlarkin, line 151: We should probably leave a note that the `name` field in the emitter message never seems to be used past this point (at least, I don't see the message's `name` being used anywhere else in this PR).

test/integration/particle_emitter.cc
```diff
@@ -0,0 +1,170 @@
...
+                    msgs::Convert(_emitter->Data().pose()));
+                EXPECT_EQ(math::Vector3d(2, 2, 2),
+                    msgs::Convert(_emitter->Data().size()));
+                EXPECT_DOUBLE_EQ(5.0, _emitter->Data().rate());
+                EXPECT_DOUBLE_EQ(1.0, _emitter->Data().duration());
+                EXPECT_TRUE(_emitter->Data().emitting());
+                EXPECT_EQ(math::Vector3d(3, 3, 3),
+                    msgs::Convert(_emitter->Data().particle_size()));
+                EXPECT_DOUBLE_EQ(2.0, _emitter->Data().lifetime());
+                // TODO(anyone) add material check here
```
adlarkin, line 106: I wasn't sure how to load/test for materials, so I left this check out. If anyone wants to add this check in before merging, that'd be great, but if not, we can do this later.

test/integration/particle_emitter.cc
```diff
@@ -0,0 +1,170 @@
...
+                    msgs::Convert(_emitter->Data().pose()));
+                EXPECT_EQ(math::Vector3d(1, 1, 1),
+                    msgs::Convert(_emitter->Data().size()));
+                EXPECT_DOUBLE_EQ(10.0, _emitter->Data().rate());
+                EXPECT_DOUBLE_EQ(0.0, _emitter->Data().duration());
+                EXPECT_FALSE(_emitter->Data().emitting());
+                EXPECT_EQ(math::Vector3d(1, 1, 1),
+                    msgs::Convert(_emitter->Data().particle_size()));
+                EXPECT_DOUBLE_EQ(5.0, _emitter->Data().lifetime());
+                // TODO(anyone) add material check here
```
adlarkin, line 142: I wasn't sure how to load/test for materials, so I left this check out. If anyone wants to add this check in before merging, that'd be great, but if not, we can do this later.

test/integration/particle_emitter.cc
```diff
@@ -0,0 +1,170 @@
...
+  EXPECT_TRUE(updateDefaultChecked);
+}
+
+/////////////////////////////////////////////////
+/// Main
+int main(int _argc, char **_argv)
+{
+  ::testing::InitGoogleTest(&_argc, _argv);
+  return RUN_ALL_TESTS();
+}
```
adlarkin, line 170: Do we need this? I don't think we use this in other integration tests.

adlarkin (approved):
A few minor comments that would be nice to address before merging, but I'm fine with merging this PR as-is.

src/systems/particle_emitter/ParticleEmitter.cc
nkoenig: The name is also used in the `component` and topic name. I'll leave this code as-is.

test/integration/particle_emitter.cc
nkoenig: Leaving it for later.

test/integration/particle_emitter.cc
nkoenig: Leaving it for later.

test/integration/particle_emitter.cc
nkoenig: It's not required. I removed it in 8a1db3d3012bfdcfb673c266c998952d77475716

PR #519, opened 2020-12-30
GPS sensor in Gazebo

Dotrar:
Related issue:
https://github.com/ignitionrobotics/ign-sensors/issues/23
My pull requests: (This PR would depend on the following: )
https://github.com/ignitionrobotics/ign-sensors/pull/72
(which then depends on)
https://github.com/osrf/sdformat/pull/453

Use the ign-sensor::Gps in the gazebo sim by creating a Gps-system 

This was by far the hardest one to get going, not least due to the "binary/source" issue as mentioned by https://github.com/ignitionrobotics/ign-tools/issues/8 

As mentioned on the sdformat issue, I've only made two simple noise systems that get applied unilaterally to each of the components.. If this doesn't suite I could change it.

I'm also really hoping that the use of `sphericalCoords.PositionTransform()` was correct because I wasn't quite sure on how it's meant to be used; or what a better way to use it _( ie: should the GPS store it's data in a SphericalCoords Type? )_

Some notes:
* I was thinking of putting in a "reference lat/lon" into the sdformat, so that one could specify where the gps starts out in the world file. but I couldn't see that on gazebo classic and I wasn't too sure how the gazebo classic gps function actually worked.


test/integration/gps_system.cc
```diff
@@ -0,0 +1,209 @@
+/*
+ * Copyright (C) 2019 Open Source Robotics Foundation
```
ahcorde, line 2:
```suggestion
 * Copyright (C) 2020 Open Source Robotics Foundation
```

test/integration/gps_system.cc
```diff
@@ -0,0 +1,209 @@
...
+  // vertical position = world position - intial position
+  // so since gps is falling, vertical position should be negative
+  EXPECT_GT(firstMsg.vertical_position(),
+      lastMsg.vertical_position());
+  EXPECT_LT(lastMsg.vertical_position(), 0.0);
+  // vertical velocity should be negative
+  EXPECT_GT(firstMsg.vertical_velocity(),
+      lastMsg.vertical_velocity());
+  EXPECT_LT(lastMsg.vertical_velocity(), 0.0);
+  */
```
ahcorde, line 170: Remove these lines

ahcorde (changes requested):

chapulina:
Thank you again for all the work, @Dotrar . Since the fork has been deleted, I opened a new PR in #1248.

Issue #520, opened 2020-12-30
make doc fails with segmentation fault in doxygen 1.9.0 on macOS


scpeters:
Since doxygen 1.9.0 was released into homebrew-core in https://github.com/Homebrew/homebrew-core/pull/67817, `make doc` has been failing for ign-gazebo with a segmentation fault in `doxygen` itself:

* [![Build Status](https://build.osrfoundation.org/buildStatus/icon?job=ignition_gazebo-ci-ign-gazebo4-homebrew-amd64&build=13)](https://build.osrfoundation.org/view/os_homebrew/job/ignition_gazebo-ci-ign-gazebo4-homebrew-amd64/13/) https://build.osrfoundation.org/view/os_homebrew/job/ignition_gazebo-ci-ign-gazebo4-homebrew-amd64/13/

~~~
make[2]: *** [CMakeFiles/doc] Segmentation fault: 11
~~~

Running `make doc VERBOSE=1` shows that the following command is used: `doxygen build/api_tagfile.dox`. By editing `build/api_tagfile.dox` so that `QUIET = NO`, you can get verbose output from `make doc`. I see the following lines just before the failure:

~~~
Preprocessing <prefix>/ign-gazebo/include/ignition/gazebo/Util.hh...
Parsing file <prefix>/ign-gazebo/include/ignition/gazebo/Util.hh...
Building macro definition list...
Building group list...
Building directory list...
Building namespace list...
Segmentation fault: 11
~~~

I've built doxygen from source in debug mode and have isolated the segmentation fault to [a specific line in doxygen.cpp](https://github.com/doxygen/doxygen/blob/Release_1_9_0/src/doxygen.cpp#L1569). I will submit a bug report to doxygen. In the meantime, we may want to disable generation of docs in this repository.

scpeters:
I uncommented a printf from doxygen and get the following output:

~~~
diff --git a/src/doxygen.cpp b/src/doxygen.cpp
index 3472f8f1..d4546ff3 100644
--- a/src/doxygen.cpp
+++ b/src/doxygen.cpp
@@ -1558,7 +1558,7 @@ static void buildNamespaceList(const Entry *root)
           tagName     = tagInfo->tagName;
           tagFileName = tagInfo->fileName;
         }
-        //printf("++ new namespace %s lang=%s tagName=%s\n",fullName.data(),langToString(root->lang).data(),tagName.data());
[...]

PR #521, opened 2020-12-30
[don't merge] Test re-ordering namespaces

scpeters:
See if this fixes doxygen (see #520). I expect it will break ABI, so I've only opened it as a draft.


scpeters:
doesn't fix it

PR #522, opened 2020-12-30
Generate valid topics everywhere (support names with spaces)

chapulina:
This wraps up #239.

I went through the codebase and updated any topics and services I found to convert topics to valid ones if possible.

The `spaces.sdf` world has some entities with empty spaces, and wouldn't work correctly without this PR, mainly because of the spaces in the world name.

I also added the `validTopic` helper function which is handy to use when we have fallbacks for the topic. For example, when we accept a `<topic>` option from SDF, and fallback to a name-generated topic in case none is provided on SDF. Both these topics need to be validated, so the function helps with that.

We need to be mindful of this when adding new topics and plugins!

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -75,7 +75,7 @@ set(IGN_FUEL_TOOLS_VER ${ignition-fuel_tools5_VERSION_MAJOR})
 
 #--------------------------------------
 # Find ignition-gui
-ign_find_package(ignition-gui4 REQUIRED VERSION 4.1)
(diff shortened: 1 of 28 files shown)


ahcorde (approved):
LGTM

chapulina:

JShep1 (approved):
Looks good with tests passing locally.  There's a failure in CI, it may be a flaky test.

chapulina:
I'll wait for the `ign-gui 4.1.1` release before merging this, otherwise `spaces.sdf` won't work well.

PR #530, opened 2021-01-05
Model creation tutorial using services

ahcorde:
I splitted this tutorial https://github.com/ignitionrobotics/ign-gazebo/pull/515 in two parts as suggested by @chapulina 

Include: 
  - Tutorial
  - Example


examples/standalone/model_creation/model_creation.cc
```diff
@@ -0,0 +1,193 @@
...
+      std::cout << "Service call failed" << std::endl;
+      return;
+    }
+  }
+  else
+    std::cerr << "Service call timed out" << std::endl;
+//! [call service create]
+}
+
+void createModelFromStr(const std::string modelStr)
```
chapulina, line 75:
```suggestion
void createEntityFromStr(const std::string modelStr)
```

This also works for lights, so the name can be more generic

examples/standalone/model_creation/model_creation.cc
```diff
@@ -0,0 +1,193 @@
...
+void createLight()
+{
+  bool result;
+  ignition::msgs::Boolean rep;
+//! [create light]
+  ignition::msgs::EntityFactory entityFactoryRequest;
+
+  entityFactoryRequest.mutable_light()->set_name("point");
+  entityFactoryRequest.mutable_light()->set_id(22);
+  entityFactoryRequest.mutable_light()->set_parent_id(1);
```
chapulina, line 43: I think we shouldn't encourage users to hardcode IDs for their entities. Gazebo assigns the IDs when the entities are created, which is hard to predict. I think this kind of code should just use names.

IDs are helpful to uniquely identify entities at runtime more quickly, like when the server and client talk to each other.

tutorials/model_creation.md
```diff
@@ -0,0 +1,92 @@
+\page model_creation Model creation
+
+This tutorial gives an introduction to the Ignition Gazebo's service `/world/<world name>/create`.
+This service will allow to create models in the scene such us spheres, lights, etc.
+
+Ignition Gazebo creates many services. You can check this typing:
```
chapulina, line 6: This could be a bit misleading, because some of these services are only created if certain plugins are loaded. I think it would be good to mention that the `UserCommands` plugin needs to be loaded to offer the `create` service.

tutorials/model_creation.md
```diff
@@ -0,0 +1,92 @@
...
+
+We will open an empty Ignition Gazebo world, let's start creating a sphere in the world.
+In the next snippet you can see how to create models based on strings.
+
+\snippet examples/standalone/model_creation/model_creation.cc create sphere
+
+The variable `modelStr` contains the SDF of the model that we want to create in the world.
+In this case we are creating a sphere of 1 meter of radius in the position: `x: 0 y:0 z:0.5 roll: 0 pitch: 0 yaw: 0`.
+
+**NOTE**: You can insert here all kind of models that you can described using a SDF string.
```
chapulina, line 61:
```suggestion
**NOTE**: You can insert here all kinds of models that you can described using an SDF string.
```

tutorials/model_creation.md
```diff
@@ -0,0 +1,92 @@
...
+
+Then we need to call the service `/world/<world_name>/create`:
+
+\snippet examples/standalone/model_creation/model_creation.cc call service create sphere
+
+**NOTE**: If the entity name does not exist then the entity will be created in the world or if the entity already
+exist then nothing will happens. You may see some traces in the console showing this information.
```
chapulina, line 68:
```suggestion
exist then nothing will happen. You may see some traces in the console showing this information.
```

How about setting `allow_renaming` to true on the example message, so the user can run this multiple times and a new entity will always be created?

tutorials/model_creation.md
```diff
@@ -0,0 +1,92 @@
...
+\snippet examples/standalone/model_creation/model_creation.cc create light
+
+Or we can create a SDF string:
+
+\snippet examples/standalone/model_creation/model_creation.cc create light str
+
+Please check the API to know which fields are available for each type of light. There are three types of lights:
+ - Point [rendering API](https://ignitionrobotics.org/api/rendering/4.1/classignition_1_1rendering_1_1Light.html) [light msgs API](https://ignitionrobotics.org/api/msgs/6.2/classignition_1_1msgs_1_1Light.html)
+ - Directional [rendering API](https://ignitionrobotics.org/api/rendering/4.1/classignition_1_1rendering_1_1DirectionalLight.html)[light msgs API](https://ignitionrobotics.org/api/msgs/6.2/classignition_1_1msgs_1_1Light.html)
+ - Spot [rendering API](https://ignitionrobotics.org/api/rendering/4.1/classignition_1_1rendering_1_1SpotLight.html)[light msgs API](https://ignitionrobotics.org/api/msgs/6.2/classignition_1_1msgs_1_1Light.html)
```
chapulina, line 88: I still think that a user can get confused looking at this API, because they can't call any of those functions from the example. I think the message API should be more useful here:

https://ignitionrobotics.org/api/msgs/6.2/classignition_1_1msgs_1_1Light.html

chapulina (commented):
Nice! I just have some minor suggestions.

chapulina (approved):
I pushed some tweaks. Thanks for the tutorial!

ahcorde:
@chapulina  Thank you for the fixes! Merging!

PR #531, opened 2021-01-05
Visualize collisions

jennuine:
Closes https://github.com/ignitionrobotics/ign-gazebo/issues/105

Ability to visualize collision (basic shapes or meshes) similar to Gazebo-classic, which are displayed in orange.

Collision visualization can be toggled from (in the scene or in the entity tree):
* A model's right-click context menu, which toggle all collisions in the model
* A link's right-click context menu, which toggles all collisions in the link

Collision visuals can be created for basic shapes and meshes but depending on the render order when a collision (visual) is coplaner with the shape/mesh visual there are times when the collision visual can not be seen. This should be resolved in Edifice from this PR https://github.com/ignitionrobotics/ign-rendering/pull/188 

Here's an example:
Peek 2021-01-05 17-44


src/rendering/RenderUtil.cc
```diff
@@ -457,6 +520,37 @@ void RenderUtil::Update()
       visual->SetUserData("temperature", temp.second);
     }
   }
+
+  // create new collision visuals
+  {
+    if (!newCollisionLinks.empty())
+      DeselectAllEntities();
```
iche033, line 527: is this done so that the object does not get highlighted with bright white emissive material?

src/rendering/RenderUtil.cc
```diff
@@ -1009,6 +1141,15 @@ void RenderUtilPrivate::RemoveRenderingEntities(
         this->removeEntities[_entity] = _info.iterations;
         return true;
       });
+
+  // collisions
+  _ecm.EachRemoved<components::Collision>(
+    [&](const Entity &_entity, const components::Collision *)->bool
+      {
+        this->removeEntities[_entity] = _info.iterations;
+        this->viewingCollisions.erase(_entity);
```
iche033, line 1150: do you also need to update `entityCollisions` and `linkToCollisionEntities`, and probably also `modelToLinkEntities` when links are removed? hmm there's a bit of maintenance that need to be done to keep these maps update-to-date.

src/rendering/RenderUtil.cc
```diff
@@ -457,6 +520,37 @@ void RenderUtil::Update()
...
+
+          sdf::Material material;
+          material.SetAmbient(math::Color(1, 0.5088, 0.0468, 1));
+          material.SetDiffuse(math::Color(1, 0.5088, 0.0468, 1));
+          material.SetSpecular(math::Color(0.5, 0.5, 0.5, 1));
+
+          sdf::Visual visual;
+          visual.SetGeom(*collision.Geom());
+          visual.SetMaterial(material);
+          this->dataPtr->sceneManager.CreateVisual(colEntity, visual, link);
```
iche033, line 548: The code here for creating collision visuals would be a good addition to the  `SceneManager` class, i.e. a new `CreateCollision` function

and we just have to call `rendering::VisualPtr colVisual = this->dataPtr->sceneManager.CreateCollision(colEntity, collision, link)`

iche033 (commented):
Works well for me. Did a first review pass and left some comments.

src/rendering/RenderUtil.cc
jennuine: This was done to avoid encountering this error after a new collision was created and clicking elsewhere (deselecting an entity) https://github.com/ignitionrobotics/ign-gazebo/blob/4a1cba6d2e32e493c5dc0f7233db01b406681adf/src/rendering/RenderUtil.cc#L1427-L1428

include/ignition/gazebo/rendering/SceneManager.hh
```diff
@@ -96,6 +96,11 @@ inline namespace IGNITION_GAZEBO_VERSION_NAMESPACE {
     public: rendering::VisualPtr CreateVisual(Entity _id,
         const sdf::Visual &_visual, Entity _parentId = 0);
 
+    /// \brief Retrieve visual
+    /// \param[in] _id Unique visual (entity) id
+    /// \return Pointer to requested visual
+    public: rendering::VisualPtr GetVisual(Entity _id);
```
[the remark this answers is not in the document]
jennuine: Done 704af0dcc31636fa5f0a736752e55a1ec59bfd9a

src/gui/plugins/modules/EntityContextMenu.cc
```diff
@@ -146,6 +152,12 @@ void EntityContextMenu::OnRequest(const QString &_request, const QString &_data)
     req.set_data(_data.toStdString());
     this->dataPtr->node.Request(this->dataPtr->followService, req, cb);
   }
+  else if (request == "viewCollisions")
```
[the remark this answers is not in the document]
jennuine: 704af0dcc31636fa5f0a736752e55a1ec59bfd9a

src/gui/plugins/modules/EntityContextMenu.qml
```diff
@@ -37,6 +37,29 @@ Item {
...
+      text: "View >"
+      MouseArea {
+        id: viewSubMouseArea
+        anchors.fill: parent
+        hoverEnabled: true
+        onEntered: secondMenu.open()
+      }
+    }
+  }
+  Menu {
```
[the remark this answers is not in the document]
jennuine: 704af0dcc31636fa5f0a736752e55a1ec59bfd9a

src/gui/plugins/scene3d/Scene3D.cc
```diff
@@ -803,6 +809,42 @@ void IgnRenderer::Render()
...
+  {
+    IGN_PROFILE("IgnRenderer::Render ViewCollisions");
+    if (!this->dataPtr->viewCollisionsTarget.empty())
+    {
+      rendering::NodePtr targetNode =
+          scene->NodeByName(this->dataPtr->viewCollisionsTarget);
+
+      if (!targetNode)
+      {
+        // workaround for selecting collision
```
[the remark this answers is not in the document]
jennuine: Nope, removed in 704af0dcc31636fa5f0a736752e55a1ec59bfd9a

src/rendering/RenderUtil.cc
```diff
@@ -457,6 +520,37 @@ void RenderUtil::Update()
...
+
+      for (const auto& colEntity : colEntities)
+      {
+        if (!this->dataPtr->sceneManager.HasEntity(colEntity))
+        {
+          sdf::Collision collision = this->dataPtr->entityCollisions[colEntity];
+
+          sdf::Material material;
+          material.SetAmbient(math::Color(1, 0.5088, 0.0468, 1));
+          material.SetDiffuse(math::Color(1, 0.5088, 0.0468, 1));
```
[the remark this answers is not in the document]
jennuine: 704af0dcc31636fa5f0a736752e55a1ec59bfd9a I set alpha to 0.7 because it was difficult to see at 0.5 (main visual color blended with collision color). Let me know if you would like me to update it to 0.5 instead

src/rendering/RenderUtil.cc
```diff
@@ -457,6 +520,37 @@ void RenderUtil::Update()
...
+          material.SetSpecular(math::Color(0.5, 0.5, 0.5, 1));
+
+          sdf::Visual visual;
+          visual.SetGeom(*collision.Geom());
+          visual.SetMaterial(material);
```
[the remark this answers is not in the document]
jennuine: 704af0dcc31636fa5f0a736752e55a1ec59bfd9a

src/rendering/RenderUtil.cc
jennuine: Done in 85223984950c065ecaf711efd05c13048fd0a547. Since only a model or a light can be removed from the gui (not a link or a collision), I didn't have to iterate through the maps to keep them updated.

src/rendering/RenderUtil.cc
jennuine: 704af0dcc31636fa5f0a736752e55a1ec59bfd9a

src/rendering/RenderUtil.cc
iche033: I tried commenting the above two lines out to see what's causing this error msg appear but I couldn't reproduce it by selecting the model and toggling View > Collision. Can you list the steps to reproduce this?

iche033 (commented):
New changes look good. I couldn't visualize collision for the ground plane. Does it work for you?

src/rendering/RenderUtil.cc
jennuine: 1. Start up Ignition with a sample world
2. Select a model in Scene3D so that it is highlighted
3. Then right click the model, View > Collisions
4. Then click anywhere else in the scene and the error should appear

The error only appears once, after the collision visual has been created. If you toggle off the collision then run through the above steps again then the error doesn't appear again.

jennuine:
> New changes look good. I couldn't visualize collision for the ground plane. Does it work for you?

No, but it's because it's missing the `<size>` element in the geometry of the SDF. So for example, in the `diff_drive.sdf` example the `ground_plane` has https://github.com/ignitionrobotics/ign-gazebo/blob/b432081c6b9f1926285f448c5ff3a24a528cfcba/examples/worlds/diff_drive.sdf#L55-L61

and if `<size>` is added:

```xml
 <collision name="collision"> 
   <geometry> 
     <plane> 
       <normal>0 0 1</normal> 
       <size>100 100</size>
     </plane> 
   </geometry> 
 </collision> 
```

The collision for the ground plane is viewable. Should I add the `<size>` element for all the examples?

src/rendering/RenderUtil.cc
iche033: Thanks I was able to reproduce the problem now. I think the issue is that It was trying to update the emissive property of a new geometry (collision visual's geom) that it does not previously know about. The current code works around this problem by deselecting all entities before creating the new collision visual. But there would now a mismatch between what's selected in scene 3d and the entity tree (and other plugins) for example. It's more of a usability issue now. I would recommend seeing if there's another way around this without deselecting all entities, e.g. updating the logic in the `LowlightNode` function or simply removing that error msg.

iche033:
> No, but it's because it's missing the <size> element in the geometry of the SDF.

that's interesting because I thought if a required field is missing, sdf parser would complain. Even if it went through the parser, the default value should 1x1m which is not the behavior we're seeing. So I think we should add `<size>` to `<plane>` in the ign-gazebo worlds. Can you do that in a separate PR?

jennuine:
> Can you do that in a separate PR?

Yes, I'm assuming branched from and merged to `ign-gazebo3`?

iche033:
> Yes, I'm assuming branched from and merged to ign-gazebo3?

yes, thanks!

src/rendering/RenderUtil.cc
jennuine: Fixed the issue by adding the collision visual's geometry emissive property to the `originalEmissive` map after creation of the collision visual: df7f00942033bc20b6416f76df048df648daca2f

iche033 (approved):

chapulina (changes requested):
I haven't finished reviewing everything yet, but I think there's a problem with the collision poses:

Peek 2021-01-20 18-58

include/ignition/gazebo/rendering/RenderUtil.hh
```diff
@@ -21,6 +21,7 @@
 #include <string>
 #include <vector>
 
+#include <sdf/Collision.hh>
```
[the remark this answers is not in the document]
jennuine: Yup, done dbf7a0c067c434c779e429d08cac07ae701b504b

jennuine:
> I think there's a problem with the collision poses

Oops, thank you for catching that. It is fixed in dbf7a0c067c434c779e429d08cac07ae701b504b

src/gui/plugins/modules/EntityContextMenu.qml
```diff
@@ -37,6 +37,40 @@ Item {
...
+    id: secondMenu
+    x: menu.x + menu.width
+    y: menu.y + viewSubmenu.y
+    MenuItem {
+      id: viewCollisionsMenu
+      text: "Collisions"
```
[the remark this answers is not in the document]
jennuine: This may need to be addressed in the future. So far I have not been able to come up with a reliable way to show whether collisions are on or off

nkoenig:
Works for me.

nkoenig (approved):

chapulina (approved):
Works great!

PR #532, opened 2021-01-06
Add support for topic statistics on breadcrumb deployments

nkoenig:
This is aimed at helping to address https://github.com/osrf/subt/issues/655. 

When `topic_statistics` are enabled on a breadcrumb, then dropped message information will be logged.

I'm targeting Dome instead of Citadel since SubT is using Dome and I want to limit the propagation of what might be debug code.


src/systems/breadcrumbs/Breadcrumbs.cc
```diff
@@ -385,6 +400,30 @@ void Breadcrumbs::OnDeploy(const msgs::Empty &)
...
   {
     std::lock_guard<std::mutex> lock(this->pendingCmdsMutex);
+
+    // Check topic statistics for dropped messages
+    if (this->topicStatistics)
+    {
+      ignmsg << "Received breadcrumb deployment for " <<
+        this->modelRoot.ModelByIndex(0)->Name() << std::endl;
+      std::optional<transport::TopicStatistics> stats =
+        this->node.TopicStats(this->topic);
```
chapulina, line 410: Can we bump the required `ign-transport9` version to 9.1.0 on `CMakeLists` so a user on an older version catches the requirement early before this fails to compile?

caguero (commented):
It looks good to me, just waiting for CI.

src/systems/breadcrumbs/Breadcrumbs.cc
```diff
@@ -385,6 +400,30 @@ void Breadcrumbs::OnDeploy(const msgs::Empty &)
   IGN_PROFILE("Breadcrumbs::PreUpdate");
   {
     std::lock_guard<std::mutex> lock(this->pendingCmdsMutex);
+
+    // Check topic statistics for dropped messages
+    if (this->topicStatistics)
```
azeey, line 405: Can this be done after the mutex is released?

src/systems/breadcrumbs/Breadcrumbs.cc
```diff
@@ -385,6 +400,30 @@ void Breadcrumbs::OnDeploy(const msgs::Empty &)
...
+    {
+      ignmsg << "Received breadcrumb deployment for " <<
+        this->modelRoot.ModelByIndex(0)->Name() << std::endl;
+      std::optional<transport::TopicStatistics> stats =
+        this->node.TopicStats(this->topic);
+      if (stats)
+      {
+        if (stats->DroppedMsgCount() > 0)
+        {
+        ignwarn << "Dropped message count of " << stats->DroppedMsgCount()
```
azeey, line 415: nit: indentation

azeey (approved):
LGTM!

src/systems/breadcrumbs/Breadcrumbs.cc
nkoenig: Done in 2f3d81042795e9fd78757bff8479d50f50ab2d80

src/systems/breadcrumbs/Breadcrumbs.cc
nkoenig: Moved in 8f861132968765420150dfa6d7e165c7aa4b7dc4

src/systems/breadcrumbs/Breadcrumbs.cc
nkoenig: Fixed in 8f861132968765420150dfa6d7e165c7aa4b7dc4

PR #534, opened 2021-01-08
Add joint position controller GUI, also enable tests for GUI plugins

chapulina:
You can try it out with this demo world:

`ign gazebo -v 4 "https://fuel.ignitionrobotics.org/1.0/OpenRobotics/worlds/NAO joint control"`

nao_joint_control4


nkoenig (commented):
Is it possible to have the widget consume the available space? Not sure if this is related to this plugin, or ign-gui in general. See image:

joint_controller_fill

I also tried this with the double pendulum, and I wasn't able to control the joints. I assume this is because the pendulum model doesn't have the joint position controller plugin. Is it possible for this widget to know if the plugin has been specified, and display some kind of warning when the plugin is missing?

chapulina:
> Is it possible to have the widget consume the available space? Not sure if this is related to this plugin, or ign-gui in general. 

Yeah that annoys me too. That's on `ign-gui`'s side. I'll take a stab at this soon, and if I can't fix it, I'll ticket an issue.

> Is it possible for this widget to know if the plugin has been specified, and display some kind of warning when the plugin is missing?

At the moment I think that would be difficult. Some thoughts:

* Once https://github.com/ignitionrobotics/ign-gazebo/issues/191 is addressed, the GUI should have a way to tell if an entity has certain plugins attached. Once that information is available, this plugin could use it to determine if the joints have the correct controllers.
* Once https://github.com/ignitionrobotics/ign-gazebo/issues/190 is addressed, we should have a mechanism to add system plugins on demand from the GUI. In this case, the joint controller could be spawned as needed.
* One temporary solution could be to check if the required topics have subscribers. But I think that's not possible at the moment due to https://github.com/ignitionrobotics/ign-transport/issues/39.

chapulina:
> > Is it possible to have the widget consume the available space? Not sure if this is related to this plugin, or ign-gui in general.

> Yeah that annoys me too. That's on ign-gui's side. I'll take a stab at this soon, and if I can't fix it, I'll ticket an issue.

By "soon" I meant in 2 months :grimacing: Here we go: https://github.com/ignitionrobotics/ign-gui/pull/194

chapulina:
> display some kind of warning when the plugin is missing?

For now, I added a help popup explaining to users that they should add the joint controllers:

nao_help

src/gui/plugins/joint_position_controller/Joint.qml
```diff
@@ -0,0 +1,112 @@
...
+
+    IgnSpinBox {
+      id: spin
+      value: model.value
+      minimumValue: model.min
+      maximumValue: model.max
+      decimals: 2
+      onEditingFinished: {
+        joint.value = spin.value
+      }
```
jennuine, line 82:
```suggestion
      stepSize: 0.1
      onValueChanged: {
        joint.value = spin.value
      }
```
Currently, the `stepSize` is `1` but I think `0.1` would be a better increment. Also, I think using `onValueChanged` instead of `onEditingFinished` is more user friendly, the user doesn't have to hit enter every time they want to update a joint when using the spinbox

jennuine (approved):
LGTM! I left a minor suggestion for the spinbox.

What do you think about having a reset button that resets all the joints back to default?

src/gui/plugins/joint_position_controller/Joint.qml
chapulina: Both are very good suggestions. I started with both of them, but I had to revert to using `onEditingFinished` when I worked on the reset button.

While working on reset, I noticed that the values on the GUI were not being updated to reflect the actual state of the robot. When I fixed that, I realized I couldn't use `onValueChanged`, because that signal doesn't differentiate the value being changed by the user, or from the ECM. So what was happening was that the intermediate state of a joint moving would be sent as a new command, and the joint would tend to return to its original position in a negative feedback loop. It's hard to explain with words, see what happened with `onValueChanged`:

jointfeedback


This is all on 08904d0, let me know what you think.

chapulina:
> What do you think about having a reset button that resets all the joints back to default?

Good idea, added on 08904d0, and also fixed updating the GUI according to the robot's real state. Let me know what you think:

reset

jennuine (approved):
LGTM :smile:

mjcarroll:
This caused a test failure in CI: https://build.osrfoundation.org/job/ignition_gazebo-ci-ign-gazebo4-bionic-amd64/54/testReport/(root)/JointPositionControllerGui/Load/

```
/home/jenkins/workspace/ignition_gazebo-ci-ign-gazebo4-bionic-amd64/ign-gazebo/src/gui/plugins/joint_position_controller/JointPositionController_TEST.cc:90
  plugin->ModelName()
    Which is: {}
  QString("No model selected")
    Which is: { 2-byte object <4E-00>, 2-byte object <6F-00>, 2-byte object <20-00>, 2-byte object <6D-00>, 2-byte object <6F-00>, 2-byte object <64-00>, 2-byte object <65-00>, 2-byte object <6C-00>, 2-byte object <20-00>, 2-byte object <73-00>, 2-byte object <65-00>, 2-byte object <6C-00>, 2-byte object <65-00>, 2-byte object <63-00>, 2-byte object <74-00>, 2-byte object <65-00>, 2-byte object <64-00> }
     ```

Issue #489, opened 2020-12-15
Configure physics properties


iche033:
Support configuring physics properties like max step size, real time update rate, gravity, etc (see [Physics](https://github.com/osrf/sdformat/blob/master/src/Physics.cc) SDF DOM and perhaps also some of the fields in [physics.proto](https://github.com/ignitionrobotics/ign-msgs/blob/ign-msgs6/proto/ignition/msgs/physics.proto)) through ign gazebo components and/or ign-transport

Additionally a component inspector to let users edit these properties through the gui

mrushyendra:
I made some initial changes [here](https://github.com/ignitionrobotics/ign-gazebo/compare/ign-gazebo4...mrushyendra:feature/configure_physics_properties#) to add a component that allows users to change the step size or target rtf. While it works in practice, I was wondering where the best place to actually update the `SimulationRunner`'s `stepSize` and `desiredRtf` properties would be. Currently I'm doing it in `step` here:

https://github.com/mrushyendra/ign-gazebo/blob/56f800e073db107120f924f196fcb92e818b38f6/src/SimulationRunner.cc#L740

but was wondering if that is the right approach. Thank you!

ahcorde:
Hi @mrushyendra 

Thank you for picking this issue. I think it would be great if you can create a PR because we can discuss better the code and see easily the new code.

I think you can also consider two more things:

 - It would also be great to make the `component inspector` call those new topics to edit physics from the GUI.
 - the PhysicCmd should probably be handled by the `UserCommands` system.

This PR https://github.com/ignitionrobotics/ign-gazebo/pull/482 is adding light control, you can get some inspiration.

luca-della-vedova:
I started to build on top of the work by @mrushyendra and should be ready to open a draft PR for component inspector control of physics parameters soon.

PR #536, opened 2021-01-08
Add service and GUI to configure physics parameters (step size and real time factor)

luca-della-vedova:
This PR addresses #489 introducing a service in UserCommands advertised under `/world/world_name/set_physics` to set physics parameters (for now only step size and maximum real time factor), as well as updates to the ComponentInspector GUI to set them.

An example of the GUI is below:

Physics_gui

There are a few parts of the implementation that I found suspicious while writing, specifically:

* Is the placement of the parameters updates in `SimulationRunner` OK? It is currently placed in the main `Step()` loop.
* There is a lot of code duplication for creating the physics parameters, they are currently initialized in the `SimulationRunner` constructor, `SdfEntityCreator` world creation function and `LevelManager` initialization.
* I created both a `Physics` and a `PhysicsCmd` component to hold the current physics data (the first) and send updates (the second), similar to what is done in #482 with `Light` and `LightCmd`, as well as `Pose` and `PoseCmd`.
* I added the plot icon but there is no plotting feature yet, I suspect for these parameters it doesn't make much sense to plot them so I'll probably remove it.
* Also there was some discussion about allowing reconfiguration of `Gravity` and `MagneticField`. Right now however, they have separate components, and they are not stored in the `sdf::Physics` object, hence I suspect it might not be possible if we stick to having the `Physics` component store the `sdf::Physics` object.

Most importantly, it seems the physics component that is now updated in `SimulationRunner` isn't updated properly when the `ComponentInspector` field is deselected by moving the mouse away, instead of pressing return. It seems the value in the physics simulation is updated but the one in the GUI isn't, and there is a delay between the value displayed in the GUI and the one being actually used in the simulation (shown below):

Physics_glitch

I could only get this to work by updating the Physics component in the `UserCommands` service callback but that sounded like a very hacky fix, I left it commented just to have an idea of what it would look like.


include/ignition/gazebo/components/Physics.hh
```diff
@@ -0,0 +1,56 @@
+/*
+ * Copyright (C) 2020 Open Source Robotics Foundation
```
ahcorde, line 2:
```suggestion
 * Copyright (C) 2021 Open Source Robotics Foundation
```

include/ignition/gazebo/components/PhysicsCmd.hh
```diff
@@ -0,0 +1,46 @@
+/*
+ * Copyright (C) 2020 Open Source Robotics Foundation
```
ahcorde, line 2:
```suggestion
 * Copyright (C) 2021 Open Source Robotics Foundation
```

src/gui/plugins/component_inspector/Physics.qml
```diff
@@ -0,0 +1,262 @@
+/*
+ * Copyright (C) 2020 Open Source Robotics Foundation
```
ahcorde, line 2:
```suggestion
 * Copyright (C) 2021 Open Source Robotics Foundation
```

ahcorde (changes requested):
You have included the two elements as drawable (at least I can see the icon close to the name). You should include these new parameters to the Plotting class (src/gui/plugins/plotting/Plotting.hh, src/gui/plugins/plotting/Plotting.cc)

luca-della-vedova:
> You have included the two elements as drawable (at least I can see the icon close to the name). You should include these new parameters to the Plotting class (src/gui/plugins/plotting/Plotting.hh, src/gui/plugins/plotting/Plotting.cc)

Yea it was one of the parts I was not super sure about, I was not sure whether it would be useful to have the parameters drawable since, unlike other drawable parameters like `Pose` they might not be that significant to have on a plot. I was leaning more towards removing the plot icon but if you think there might be some cases in which a user might want to plot the maximum RTF / physics step size over time I can add the changes to the plotting system.

chapulina:
> cases in which a user might want to plot the maximum RTF / physics step size over time

I can imagine someone plotting that together with other data to see how it's affected. I don't think it's extremely necessary though, so it's up to you if you want to remove the plot icons or add full plot support.

luca-della-vedova:
@mrushyendra can you sync your fork to this branch and sign-off your commits for the [DCO error](https://github.com/ignitionrobotics/ign-gazebo/pull/536/checks?check_run_id=1686018016)? I'll resync this branch afterwards.

mrushyendra:
> @mrushyendra can you sync your fork to this branch and sign-off your commits for the [DCO error](https://github.com/ignitionrobotics/ign-gazebo/pull/536/checks?check_run_id=1686018016)? I'll resync this branch afterwards.

@luca-della-vedova Yep sure! Will do by the end of today

src/systems/user_commands/UserCommands.cc
```diff
@@ -248,6 +272,13 @@ void UserCommands::Configure(const Entity &_entity,
       &UserCommandsPrivate::PoseService, this->dataPtr.get());
 
   ignmsg << "Pose service on [" << poseService << "]" << std::endl;
+
+  // Physics service
+  std::string physicsService{"/world/" + worldName + "/set_physics"};
```
ahcorde, line 277: check topic name with transport::TopicUtils::AsValidTopic()

ahcorde (changes requested):
Do you mind to add some tests to `test/integration/user_commands.cc` ?

luca-della-vedova:
> Do you mind to add some tests to `test/integration/user_commands.cc` ?

[Done](https://github.com/ignitionrobotics/ign-gazebo/pull/536/commits/5225857ec5c3596b7a625a10c1e6beb749492a42)! A simple "create world, check physics properties, set them through service and check them again".

src/gui/plugins/component_inspector/ComponentInspector.cc
```diff
@@ -728,6 +747,24 @@ void ComponentInspector::OnPose(double _x, double _y, double _z, double _roll,
...
+      [](const ignition::msgs::Boolean &/*_rep*/, const bool _result)
+  {
+    if (!_result)
+        ignerr << "Error setting physics parameters" << std::endl;
+  };
+
+  ignition::msgs::Physics req;
+  req.set_max_step_size(_stepSize);
+  req.set_real_time_factor(_realTimeFactor);
+  auto physicsCmdService = "/world/" + this->dataPtr->worldName
```
ahcorde, line 763: check the topic -> transport::TopicUtils::AsValidTopic()

ahcorde (changes requested):
LGTM, just a minor comment

src/gui/plugins/component_inspector/ComponentInspector.cc
luca-della-vedova: Done!

src/gui/plugins/component_inspector/ComponentInspector.cc
```diff
@@ -762,6 +762,7 @@ void ComponentInspector::OnPhysics(double _stepSize, double _realTimeFactor)
   req.set_real_time_factor(_realTimeFactor);
   auto physicsCmdService = "/world/" + this->dataPtr->worldName
       + "/set_physics";
+  physicsCmdService = transport::TopicUtils::AsValidTopic(physicsCmdService);
```
ahcorde, line 765: This can return an empty string, 

```
  if (physicsCmdService.empty())
  {
    ignerr << "..." << std::endl;
    return;
  }

**Review by ahcorde — CHANGES_REQUESTED** [review]

File: src/gui/plugins/component_inspector/ComponentInspector.cc (hunk 6 above)
Comment (luca-della-vedova) [reply src/gui/plugins/component_inspector/ComponentInspector.cc:765]: Done!

**Review by ahcorde — APPROVED** [review]
LGTM

File: src/systems/user_commands/UserCommands.cc [hunk 7]
```diff
@@ -698,6 +747,44 @@ bool PoseCommand::Execute()
...
+  if (!this->iface->ecm->EntityHasComponentType(worldEntity,
+    components::PhysicsCmd().TypeId()))
+  {
+    auto physics = convert<sdf::Physics>(*physicsMsg);
+    this->iface->ecm->CreateComponent(worldEntity,
+        components::PhysicsCmd(physics));
+    // HACK, the component is meant to be updated in SimulationRunner
+    // auto physicsComponent =
+    //  this->iface->ecm->Component<components::Physics>(worldEntity);
+    // physicsComponent->Data() = components::Physics(physics).Data();
```
luca-della-vedova, line 783: I guess the last part I'm not sure on my side is if we are OK with the glitch in the main post (when the step size is updated without pressing enter) or if we should add this hack to fix it but at the expense of creating possible glitches if the `PhysicsCmd` component is not processed in `SimulationRunner`, since with this hack the `Physics` component would still be updated and have incorrect information.

src/gui/plugins/plotting/Plotting.cc
```diff
@@ -96,6 +97,11 @@ PlotComponent::PlotComponent(const std::string &_type,
   }
   else if (_type == "double")
     this->dataPtr->data["value"] = std::make_shared<PlotData>();
+  else if (_type == "Physics")
```
chapulina, line 100: I'm concerned that this won't scale well if we add a new type for each component type. For example, we plot `Vector` and `Pose`, instead of `Gravity` and `WorldPose`.

Could the 2 fields be sent independently as `double`s, so the plot doesn't need to know specifically about physics?

src/systems/user_commands/UserCommands.cc
chapulina: Fishy. I'd expect the hack not to be necessary. I'd expect `sendPhysics` to be called when the mouse moves away, just as if you pressed enter.

chapulina (commented):
I haven't tried it out yet, but here's some feedback :wink: 

---

> There is a lot of code duplication for creating the physics parameters

Yeah it would be nice to clean this up a bit... It would take some refactoring. I think for now this PR is ok.

> discussion about allowing reconfiguration of Gravity and MagneticField. 

Yeah I think these can be handled separately

src/SdfEntityCreator.cc
```diff
@@ -202,6 +203,17 @@ Entity SdfEntityCreator::CreateEntities(const sdf::World *_world)
...
       components::Gravity(_world->Gravity()));
 
+  // Physics
+  // \todo(anyone) Support picking a specific physics profile
+  auto physics = _world->PhysicsByIndex(0);
+  if (!physics)
+  {
+    physics = _world->PhysicsDefault();
+  }
+  ignerr << "Creating physics component in SDF creator" << std::endl;
```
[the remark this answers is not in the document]
luca-della-vedova: Leftover of debugging, removed!

src/systems/user_commands/UserCommands.cc
```diff
@@ -698,6 +747,44 @@ bool PoseCommand::Execute()
...
+  if (worldEntity == kNullEntity)
+  {
+    ignmsg << "Failed to find world entity" << std::endl;
+    return false;
+  }
+
+  if (!this->iface->ecm->EntityHasComponentType(worldEntity,
+    components::PhysicsCmd().TypeId()))
+  {
+    auto physics = convert<sdf::Physics>(*physicsMsg);
```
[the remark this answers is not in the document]
luca-della-vedova: Good point! Done

src/SimulationRunner.cc
```diff
@@ -727,6 +770,9 @@ void SimulationRunner::Step(const UpdateInfo &_info)
   // Update all the systems.
   this->UpdateSystems();
 
+  // Update the step size and desired rtf
+  this->UpdatePhysicsParams();
```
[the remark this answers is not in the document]
luca-della-vedova: Yea we weren't sure in the first place where this would go. I thought about just moving it up to before systems are updated but realized it might result in incorrect UpdateInfo being sent over so I went for updating them as early as possible in the [Run loop](https://github.com/ignitionrobotics/ign-gazebo/pull/536/commits/6972fed90c1f351ffee1e81a7eb8fa5edaed9f69#diff-528565fd9b23fec7ea46e2f169d2098361634a77a74cac62166310c4624ca40dR693-R696).

src/systems/user_commands/UserCommands.cc
luca-della-vedova: This was fixed when moving the updating of the physics parameters to before the systems are updated. I'm guessing that the GUI system was sent the old parameters in its `Update()` call, then immediately after updated to the new one in the `UpdatePhysicsParams()` call, resulting in this glitch.

src/gui/plugins/plotting/Plotting.cc
luca-della-vedova: I have been struggling a bit with the implementation of this since I'm not super familiar with the code base. I guess a similar feedback would apply to #482? (that I referenced for the implementation of the component_inspector / plotting tool parts) if that is the case I'll just keep an eye for the best fix there.

src/SimulationRunner.cc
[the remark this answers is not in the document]
chapulina: Ah interesting. The new location makes sense to me, even if we end up with the 2-step process. Thanks!

src/gui/plugins/plotting/Plotting.cc
chapulina: > similar feedback would apply to #482?

chapulina: Ok, I looked closer at the code now. I see that `Plotting` needs to know about every single component that it's going to plot, because it needs to know how to access its data. So I think there isn't an easy way of doing what I suggested above, so just ignore it.

src/gui/plugins/component_inspector/Physics.qml
```diff
@@ -0,0 +1,262 @@
...
+   */
+
+  Component {
+    id: writablePositiveNumber
+    IgnSpinBox {
+      id: writableSpin
+      value: writableSpin.activeFocus ? writableSpin.value : numberValue
+      minimumValue: minPhysParam
+      maximumValue: maxPhysParam
+      decimals: 6
```
chapulina, line 79: I recommend setting a small step size like `stepSize: 0.001`. The default is 1 and it causes things to explode fast :sweat_smile:

chapulina (approved):
This is a really useful feature! I just have a small usability suggestion, and then feel free to merge :wink:

peci1:
Thanks for the feature. One more usability problem is that you cannot set rtf without also setting time step via the service. I.e. a service call like this

    ign service -s /world/simple_cave_01/set_physics --reqtype ignition.msgs.Physics --reptype ignition.msgs.Boolean --timeout 2000 --req "real_time_factor: 1"

results in stopped (or broken?) simulation, because time step is read as zero from the request.

Setting both rtf and step size works, but I see this in the console:

    [Dbg] [SimulationFeatures.cc:48] Simulation timestep set to: 0.004
    Warning [World.cpp:139] [World] Attempting to set negative timestep. Ignoring this request because it can lead to undefined behavior.

Which is weird because I can't figure out where does the World.cpp file come from (I haven't found it anywhere).

Would it be possible to check whether both rtf and time step are non-zero before using them? I think a zero value is wrong for both of them.

chapulina:
> I can't figure out where does the World.cpp file come from

That's [coming from DART](https://github.com/dartsim/dart/blob/f0ba239607b3a7e79f57b856789e7329c0e9d771/dart/simulation/World.cpp#L139-L140).

> Would it be possible to check whether both rtf and time step are non-zero before using them?

+1 - On Gazebo classic a zero "real time update rate" means "run as fast as possible", but we don't have that on Ignition, so I think it's safe to assume a value of zero is unset or user error.

luca-della-vedova:
> Would it be possible to check whether both rtf and time step are non-zero before using them? I think a zero value is wrong for both of them.

Hi @peci1 and thanks for the feedback!
This should be fixed in #740 , can you check if it works for you?

PR #537, opened 2021-01-08
Port: Automatically load a subset of world plugins (#281)

mjcarroll:
Feature to allow loading of a default set of system plugins from a file. This behavior will trigger when a world sdf file is loaded with no plugins defined.  In this case, the simulator will load the plugins from a series of locations including environment variable, the users home folder, and finally in the installation directory.

This should allow users to not have to specify the same set of plugins in every world sdf file.

Co-authored-by: Louise Poubel <louise@openrobotics.org>


nkoenig (approved):

mjcarroll:
Somewhat expected failures because of how the logging subsystem has changed a bit from 3 to 4, I'm working on them.

mjcarroll:
Looks like the examples build test timed out, but was still running successfully.  Do we need to increase the timeout? 

```
73: Scanning dependencies of target joy_to_twist
73: [ 50%] Building CXX object CMakeFiles/joy_to_twist.dir/joy_to_twist.cc.o
 73/154 Test  #73: INTEGRATION_examples_build ...........................***Timeout 240.01 sec
test 74
        Start  74: check_INTEGRATION_examples_build
```

chapulina:
> Do we need to increase the timeout?

Yeah either that or split the test somehow...

mjcarroll:
I'm going to address the flaky/failing tests in a follow-up since they are unrelated to this PR.

PR #538, opened 2021-01-08
Improved Parallel ECM State

JShep1:


JShep1:
This PR was meant to build upon the multi-threaded approach to the `State(...)` call in the `EntityComponentManager` seen here https://github.com/ignitionrobotics/ign-gazebo/pull/451

The reason this approach was attempted was due to the fact that in the aforementioned PR, new threads are spawned every `State` call and run to the end of their execution with the function call as well.  This PR attempted to create a thread pool approach which would spawn threads to wait on a condition variable until the main `State` calling thread would notify the other threads that work is ready to be done.

This approach requires a few things which the PR above didn't:
 * Shared locks, mutexs and condition variables - used primarily for the fact that the region that the threads need to process isn't a critical region given they are each reading their own specific data, but there still needs to be a method of synchronization between these threads and the main calling thread
    * The worker threads need to [wait on a condition variable](https://github.com/ignitionrobotics/ign-gazebo/pull/538/files#diff-c389137bef478f45ceec75126820d09230b7d39135279f54b56b0c0488f76f94R955) which is a boolean unique to each thread that the main thread[ will set and notify all](https://github.com/ignitionrobotics/ign-gazebo/pull/538/files#diff-c389137bef478f45ceec75126820d09230b7d39135279f54b56b0c0488f76f94R1089-R1093) when data is ready to process.
    * After the threads are done with their work, they will [increment a threads done atomic int and notify all](https://github.com/ignitionrobotics/ign-gazebo/pull/538/files#diff-c389137bef478f45ceec75126820d09230b7d39135279f54b56b0c0488f76f94R987-R988), and since the main thread is spinning on a condition that [compares the threads done with the current amount of threads](https://github.com/ignitionrobotics/ign-gazebo/pull/538/files#diff-c389137bef478f45ceec75126820d09230b7d39135279f54b56b0c0488f76f94R1095-R1097) that are processing data, the main thread will only continue when all of the worker threads are done processing their data.
 * A way of each thread being able to recognize what data it's supposed to operate on
    * Since each thread will only run one function provided at their construction, we have no way of knowing at the thread's construction what work they will need to process.  So, I provided each thread an id and later stored each thread's workload (start and end iterators) in a vector where the index is that thread's respective id.  They are accessed within the thread functor [here](https://github.com/ignitionrobotics/ign-gazebo/pull/538/files#diff-c389137bef478f45ceec75126820d09230b7d39135279f54b56b0c0488f76f94R965-R966) 
* Global data structures to transfer the parameters from the `State` function to the `AddEntityToMessage` function that each thread will call.
  * Seen [here](https://github.com/ignitionrobotics/ign-gazebo/pull/538/files#diff-c389137bef478f45ceec75126820d09230b7d39135279f54b56b0c0488f76f94R1085), the `_state` data (and the other parameters, but just referencing `_state` as an example) which is provided as a parameter to `State(...)`, is needed from within each thread's functor seen [here](https://github.com/ignitionrobotics/ign-gazebo/pull/538/files#diff-c389137bef478f45ceec75126820d09230b7d39135279f54b56b0c0488f76f94R975).  So a global variable is used to pass the reference to the thread.
  * This above described solution introduces a significant regression, if another `State(...)` call is made synchronously, the data will become corrupted and undefined behavior will occur.  If this approach is to be implemented later, special care will need to be made with some sort of queue'ing mechanism where each `State(...)` will queue the work load into a data structure protected with a mutex and the threads will process the data accordingly
  
The last bullet point, above all, suggests a devoted data structure or class to manage the `State(...)` thread requests.  This is sort of where I stopped given this relatively simple code will quickly become a lot more messy and less maintainable, and the cons begin to outweigh the one pro (no thread creation cost per `State(...)` call).  So I'm closing this prototype PR for now and documented it in case we ever should want to come back to this approach in the future.

PR #539, opened 2021-01-08
[FIX] Handle multiple logical cameras

JaldertVicarious:
I found a small issue with running multiple logical cameras: only the first one that's defined in the SDF reports seeing any models.

The proposed fix is simple, though there might be better solutions (e.g. not moving `modelPoses` into the `setModelPoses` but modifying that function to take a const reference and changing the logic in the sensor implementation accordingly).

[Here](https://gist.github.com/JaldertVicarious/2800fda7fc8d52f019c6f58757b150cb) is a simple SDF to test it out.

Without the proposed fix, you'll only see one camera (`logical_camera_1`) publishing the pose of the box (`box-1`) it sees, and with the fix both cameras work as expected.

```
ign gazebo multi_logical_camera.sdf -r
```

```
#  In a second terminal
ign topic -t /logical_camera_1 -e
```

```
#  In a third terminal
ign topic -t /logical_camera_2 -e
```

### DIFF
--- a/src/systems/logical_camera/LogicalCamera.cc
+++ b/src/systems/logical_camera/LogicalCamera.cc
@@ -202,7 +202,9 @@ void LogicalCameraPrivate::UpdateLogicalCameras(
         {
           const math::Pose3d &worldPose = _worldPose->Data();
           it->second->SetPose(worldPose);
-          it->second->SetModelPoses(std::move(modelPoses));
+          // Make a copy of modelPoses s.t. SetModelPoses can take ownership
+          auto modelPoses_ = modelPoses;
+          it->second->SetModelPoses(std::move(modelPoses_));
         }
         else
         {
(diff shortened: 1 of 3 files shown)


chapulina:
Thanks for the fix and the example world.

How do you feel about adding a 2nd logical camera to the test world and updating [test/integration/logical_camera_system.cc](https://github.com/ignitionrobotics/ign-gazebo/blob/ign-gazebo4/test/integration/logical_camera_system.cc) to also check it receives messages from it?

JaldertVicarious:
@chapulina Absolutely, will add that.

JaldertVicarious:
Added test into the existing one. To avoid clever code in tests, I just duplicated the logic and assertions where applicable. In the test system, the flow is changed slightly to account for multiple sensors.

JaldertVicarious:
@chapulina A few tests failing, but they seem unrelated.

https://build.osrfoundation.org/job/ignition_gazebo-ci-pr_any-homebrew-amd64/4538/testReport/
https://build.osrfoundation.org/job/ignition_gazebo-ci-pr_any-ubuntu_auto-amd64/5049/testReport/

chapulina (approved):
Thanks for the fix and the test!

JaldertVicarious:
> Thanks for the fix and the test!

My pleasure! Thanks for the quick review & merge!

PR #541, opened 2021-01-08
Added missing version namespace

nkoenig:
Adding this to help fix downstream users of ign-gazebo, such as ign-launch.

### DIFF
--- a/include/ignition/gazebo/detail/EntityComponentManager.hh
+++ b/include/ignition/gazebo/detail/EntityComponentManager.hh
@@ -34,7 +34,6 @@ namespace gazebo
 {
 // Inline bracket to help doxygen filtering.
 inline namespace IGNITION_GAZEBO_VERSION_NAMESPACE {
-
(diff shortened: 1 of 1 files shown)


ahcorde (approved):

j-rivero:
That one was an interesting change. inline namespaces are specially designed to modify the binary symbols names so changing it in a released branch should ring the alarms. However, the abichecker is happy with it. In fact looking for some code modified here does not appear anywhere in binaries:

```
/usr/lib/x86_64-linux-gnu🔒 ❯ nm -D --demangle  libignition-gazebo3-* | grep EntityComponentManager | grep First
/usr/lib/x86_64-linux-gnu🔒 ❯ 
```

My hypothesis is that the file belongs to a private part (details) and that makes the real work of being private and not exposed to public ABI. I might be wrong and the abichecker could also be wrong and weird problems might appear somehow in the new release. Keep an eye on these change if that happen.

PR #543, opened 2021-01-09
Fix examples in migration plugins tutorial

nlamprian:


chapulina (approved):
Thanks for all the fixes! I should have actually tried to compile the examples before putting them on the tutorial.

chapulina:
Just need DCO signed to merge. Thanks!

Issue #529, opened 2021-01-04
OneTimeChange components update and SceneBroadcaster performance


luca-della-vedova:
We have an environment in which we need specific components, other than the standard [Pose](https://github.com/ignitionrobotics/ign-gazebo/blob/ign-gazebo4/src/systems/scene_broadcaster/SceneBroadcaster.cc#L291-L297), to be updated at every iteration, specifically [TrajectoryPose](https://github.com/ignitionrobotics/ign-gazebo/blob/ign-gazebo4/include/ignition/gazebo/components/Pose.hh#L46) for actors since we want to control their movement in a plugin.
If the component doesn't have [OneTimeChange](https://github.com/ignitionrobotics/ign-gazebo/blob/ign-gazebo4/include/ignition/gazebo/Types.hh#L75) specified some updates will be missed and the components will not be updated at every iteration (i.e. actor movements will be "jerky"). On the other hand, if any component is set to OneTimeChange [the whole state of the simulation](https://github.com/ignitionrobotics/ign-gazebo/blob/ign-gazebo4/src/systems/scene_broadcaster/SceneBroadcaster.cc#L286-L290) will be serialized and published, causing a major performance loss.

I thought of at least three possible ways to deal with the issue:

1. Exposing critical components to userspace, with an API where a user can specify (in SDF or a plugin) "For this specific world, I need component X to be accurate (i.e. `Temperature` if they want to simulate temperature, `AnimationTime` if they want smooth animations).
2. Making `OneTimeChange` only update the changed component and not the whole simulation state.
3. Adding a new `OneTimeChange` type that marks only a specific component as changed (i.e. `OneTimeComponentChange`), to preserve the current behavior.

chapulina:
Since #544 all `PeriodicChange`s are published regularly, not just pose. So periodic components don't need to be marked as `OneTimeChange`s to be published.

PR #544, opened 2021-01-11
Publish all periodic change components in Scene Broadcaster

luca-della-vedova:
As reported in #529, the only way users have to have components published regularly is to specify them as `OneTimeChanged`, but this incurs in the large overhead of re-serializing and publishing the whole simulation state.

When looking at the code I noticed that a component being marked as `PeriodicChange` didn't seem to have any effect on the SceneBroadcaster since it only published `Pose` components at regular intervals.
Also, I ran a quick `grep` in the code:

```
luca@luca-focal:~/ws_dome/src/ign-gazebo$ grep -r . -e PeriodicChange
./include/ignition/gazebo/Types.hh:      PeriodicChange = 1,
./src/EntityComponentManager_TEST.cc:  manager.SetChanged(e1, c1.first, ComponentState::PeriodicChange);
./src/EntityComponentManager_TEST.cc:  EXPECT_EQ(ComponentState::PeriodicChange,
./src/EntityComponentManager.cc:    result = ComponentState::PeriodicChange;
./src/EntityComponentManager.cc:  if (_c == ComponentState::PeriodicChange)
./src/systems/wheel_slip/WheelSlip.cc:                      ComponentState::PeriodicChange);
./src/systems/physics/Physics.cc:                ComponentState::PeriodicChange);
./src/systems/physics/Physics.cc:                ComponentState::PeriodicChange);
./src/systems/physics/Physics.cc:                ComponentState::PeriodicChange :
./src/systems/physics/Physics.cc:                  ComponentState::PeriodicChange :
./src/systems/physics/Physics.cc:                ComponentState::PeriodicChange :
./src/systems/physics/Physics.cc:                ComponentState::PeriodicChange :
./src/systems/physics/Physics.cc:                ComponentState::PeriodicChange :
./src/systems/physics/Physics.cc:                ComponentState::PeriodicChange :
./src/systems/physics/Physics.cc:                ComponentState::PeriodicChange :
./src/systems/physics/Physics.cc:                ComponentState::PeriodicChange :
./src/systems/physics/Physics.cc:                ComponentState::PeriodicChange :
./src/systems/multicopter_control/MulticopterVelocityControl.cc:                     ? ComponentState::PeriodicChange
./src/systems/log/LogPlayback.cc:        ComponentState::PeriodicChange);

```

 I noticed that the vast majority of PeriodicChange components are indeed `Pose` components in [Physics](https://github.com/ignitionrobotics/ign-gazebo/blob/ign-gazebo4/src/systems/physics/Physics.cc#L1806-L1807)  (unless users create Linear / Angular Velocity / Acceleration components), with a few exceptions in [MulticopterVelocityControl](https://github.com/ignitionrobotics/ign-gazebo/blob/ign-gazebo4/src/systems/multicopter_control/MulticopterVelocityControl.cc#L439-L441) (Actuator component) and [WheelSlip](https://github.com/ignitionrobotics/ign-gazebo/blob/ign-gazebo4/src/systems/wheel_slip/WheelSlip.cc#L289-L290) (SlipComplianceCmd component).

With this in mind, I decided to just change the periodic publishing of SceneBroadcaster from only `Pose` components to all the components that have been marked as PeriodicChange.

From a performance point of view, this should have no impact in a default simulation, since entities by default only have a `Pose` component. The overhead will appear when the user marks any component as `PeriodicChange`, loads a `MulticopterVelocityControl` or `WheelSlip` system or creates a `LinearVelocity`, `AngularVelocity`, `LinearAcceleration`, `AngularAcceleration` or their `World` variant components, since any of those will also be added at the 60Hz SceneBroadcaster publisher.

### DIFF
--- a/include/ignition/gazebo/EntityComponentManager.hh
+++ b/include/ignition/gazebo/EntityComponentManager.hh
@@ -487,6 +487,12 @@ namespace ignition
       /// \return True if there are any components with one-time changes.
       public: bool HasOneTimeComponentChanges() const;
 
+      /// \brief Get the components that are marked as periodic changes.
+      /// \return All the components that at least one entity marked as
+      /// periodic changes.
+      public: std::unordered_set<ComponentTypeId>
+          GetPeriodicChangedComponents() const;
+
       /// \brief Set the absolute state of the ECM from a serialized message.
       /// Entities / components that are in the new state but not in the old
       /// one will be created.
--- a/include/ignition/gazebo/EntityComponentManager.hh
+++ b/include/ignition/gazebo/EntityComponentManager.hh
@@ -487,11 +487,11 @@ namespace ignition
       /// \return True if there are any components with one-time changes.
       public: bool HasOneTimeComponentChanges() const;
 
-      /// \brief Get the components that are marked as periodic changes.
+      /// \brief Get the components types that are marked as periodic changes.
       /// \return All the components that at least one entity marked as
       /// periodic changes.
       public: std::unordered_set<ComponentTypeId>
-          GetPeriodicChangedComponents() const;
+          ComponentTypesWithPeriodicChanges() const;
 
       /// \brief Set the absolute state of the ECM from a serialized message.
       /// Entities / components that are in the new state but not in the old
--- a/src/EntityComponentManager.cc
+++ b/src/EntityComponentManager.cc
@@ -398,7 +398,7 @@ bool EntityComponentManager::HasOneTimeComponentChanges() const
 
 /////////////////////////////////////////////////
 std::unordered_set<ComponentTypeId>
-    EntityComponentManager::GetPeriodicChangedComponents() const
+    EntityComponentManager::ComponentTypesWithPeriodicChanges() const
 {
   std::unordered_set<ComponentTypeId> periodicComponents;
   for (const auto& compPair : this->dataPtr->periodicChangedComponents)
--- a/src/EntityComponentManager.cc
+++ b/src/EntityComponentManager.cc
@@ -397,6 +397,18 @@ bool EntityComponentManager::HasOneTimeComponentChanges() const
 }
 
 /////////////////////////////////////////////////
+std::unordered_set<ComponentTypeId>
+    EntityComponentManager::GetPeriodicChangedComponents() const
+{
+  std::unordered_set<ComponentTypeId> periodicComponents;
+  for (const auto& compPair : this->dataPtr->periodicChangedComponents)
+  {
+    periodicComponents.insert(compPair.first);
+  }
+  return periodicComponents;
+}
+
+/////////////////////////////////////////////////
 bool EntityComponentManager::HasEntity(const Entity _entity) const
 {
   auto vertex = this->dataPtr->entities.VertexFromId(_entity);
--- a/src/EntityComponentManager_TEST.cc
+++ b/src/EntityComponentManager_TEST.cc
@@ -2082,6 +2082,7 @@ TEST_P(EntityComponentManagerFixture, SetChanged)
   auto c2 = manager.CreateComponent<IntComponent>(e2, IntComponent(456));
 
   EXPECT_TRUE(manager.HasOneTimeComponentChanges());
+  EXPECT_EQ(0u, manager.ComponentTypesWithPeriodicChanges().size());
   EXPECT_EQ(ComponentState::OneTimeChange,
       manager.ComponentState(e1, c1.first));
   EXPECT_EQ(ComponentState::OneTimeChange,
@@ -2093,6 +2094,7 @@ TEST_P(EntityComponentManagerFixture, SetChanged)
   // updated
   manager.RunSetAllComponentsUnchanged();
   EXPECT_FALSE(manager.HasOneTimeComponentChanges());
+  EXPECT_EQ(0u, manager.ComponentTypesWithPeriodicChanges().size());
   EXPECT_EQ(ComponentState::NoChange,
       manager.ComponentState(e1, c1.first));
   EXPECT_EQ(ComponentState::NoChange,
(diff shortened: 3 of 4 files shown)


ahcorde:
Something went wrong with the `EntityComponentManagerPerfrormance.Each` test:

```bash
  [ RUN      ] EntityComponentManagerPerfrormance.Each
  /github/workspace/test/performance/each.cc:128: Failure
  Expected: (cacheEntityAvg) < (cachelessEntityAvg), actual: 6346.98 vs 4991.98
  Matching Entity Count =		1
  Nonmatching Entity Count =	1
  Each Iterations =		100
  Cache total =			634698 ns
  Cache avg per iter =		6346.9799999999996 ns
  Cache avg per iter*entity =	6346.9799999999996 ns
  Cacheless total =		499198 ns
  Cacheless avg per iter=		4991.9799999999996 ns
  Cacheless avg per iter*entity=	4991.9799999999996 ns
  
  [  FAILED  ] EntityComponentManagerPerfrormance.Each (41120 ms)
```

iche033:
hmm I've seen that test fail before but I don't know if it's due to the same issue. We can try rerunning to see if it's a flaky test.

include/ignition/gazebo/EntityComponentManager.hh
```diff
@@ -487,6 +487,12 @@ namespace ignition
       /// \return True if there are any components with one-time changes.
       public: bool HasOneTimeComponentChanges() const;
 
+      /// \brief Get the components that are marked as periodic changes.
+      /// \return All the components that at least one entity marked as
+      /// periodic changes.
+      public: std::unordered_set<ComponentTypeId>
+          GetPeriodicChangedComponents() const;
```
chapulina, line 494: The function name is a bit misleading, because it's returning the types of the components, not the components.

Also, a small style nit is that we avoid using the `Get` prefix.

How about

```suggestion
          ComponentTypesWithPeriodicChanges() const;
```

chapulina (commented):
The idea looks good to me, I think it's always good to make the logic not specific to some component type.

My only concern is the granularity. If just one component of a given type was periodically changed, all other components of that type will also be sent. I can't think of a situation like that right now, so I think this can be addressed if / when the need arises.

luca-della-vedova:
> The function name is a bit misleading, because it's returning the types of the components, not the components.

Done!

> The idea looks good to me, I think it's always good to make the logic not specific to some component type.
> 
> My only concern is the granularity. If just one component of a given type was periodically changed, all other components of that type will also be sent. I can't think of a situation like that right now, so I think this can be addressed if / when the need arises.

I agree with the granularity concern. I'd even say that the previous behavior was not fully ideal because at every iteration it published the Pose of every `Entity` in the simulation, even though many typical environments would have just a few moving objects and a lot of static ones.
From what I understand the `SceneBroadcaster` system is somewhat of a bottleneck so I can have a look at this and see if the implementation would be OK or if it gets too complicated, from my current understanding I think it would require:
* Changing the `EntityComponentManager::State(...)` function (or more likely add a new one) to only serialize components if they are attached to the entity that flagged them as `PeriodicChange`.
* The information is already in the `periodicChangedComponents` set since its key is a pair of `ComponentTypeId` and `ComponentId`. It would probably need to be changed to a `[unordered_]map<[unordered_]set>` though, so we can easily find for each component type, if an entity set it.
* Make sure every downstream system correctly uses the `SetChanged` API and doesn't assume that just because a `ComponentTypeId` has been flagged for a single entity all the Components of that type will be published.

From what I can see the `Physics` system already sets `PeriodicChange` to its components, there might be some minor tunings needed in the way it processes `WorldPoseCmd` components but the rest should be OK.

Anyway all of this would be work for a performance improvement followup PR. It should be possible to reduce the size of the state message by a few orders of magnitude if (for example) environments have thousands of static objects and only a few robots but I'll try to benchmark the possible performance gain before looking too much into it.

My last question for this PR is whether you think I should add an integration test in scene_broadcaster for this feature, something like "create a system that flags a component as periodic change, wait for the publishing, make sure the component is there". It might need a bit of work to implement but it sounds like it might be good to have as well.

chapulina (commented):
> I should add an integration test

That would be great, but I wouldn't block on it if it turns out to be too convoluted.

I'd ask however that we at least add some unit tests for `ComponentTypesWithPeriodicChanges`

luca-della-vedova:
> I'd ask however that we at least add some unit tests for `ComponentTypesWithPeriodicChanges`

Added the unit tests [931d58a](https://github.com/ignitionrobotics/ign-gazebo/pull/544/commits/931d58adaaa9ffddb6cce0dd35a7176c96b32893). I just added them to the SetChanged unit test since it had the setup logic I needed for component creation / deletion.

iche033:
> Make sure every downstream system correctly uses the SetChanged API and doesn't assume that just because a ComponentTypeId has been flagged for a single entity all the Components of that type will be published.

Just want to check my understanding of the current behavior. If 2 entities (e.g. e1, and e2) both have a `Pose` component and we only call `SetChanged` on e1's `Pose` component with `PeriodicChange`, I should expect that only e1's Pose component would be serialized to msg when `State` is called right?

iche033 (approved):

luca-della-vedova:
> > Make sure every downstream system correctly uses the SetChanged API and doesn't assume that just because a ComponentTypeId has been flagged for a single entity all the Components of that type will be published.
> 
> Just want to check my understanding of the current behavior. If 2 entities (e.g. e1, and e2) both have a `Pose` component and we only call `SetChanged` on e1's `Pose` component with `PeriodicChange`, I should expect that only e1's Pose component would be serialized to msg when `State` is called right?

True, there is a [check to skip publishing the component if another entity flagged it](https://github.com/ignitionrobotics/ign-gazebo/blob/ign-gazebo3/src/EntityComponentManager.cc#L835-L843), I got confused with the behavior of OneTimeChange that forces serialization of the whole state.

chapulina (approved):

PR #560, opened 2021-01-14
Suggestions to #451

chapulina:
Some suggestions to simplify #451 a bit.

* Moved calculation function to private class to better hide that implementation detail
* Removed a member variable that was only used in the scope of a function
* Removed the `componentsPerThread` variable
* Added a debug message whenever the iterators change

Let me know what you think, @JShep1


JShep1 (approved):
LGTM, thanks for the improvements!

PR #563, opened 2021-01-15
 Add missing IGNITION_GAZEBO_VISIBLE macros

j-rivero:
Missing visibility from different auxiliary functions and classes

### DIFF
--- a/src/Conversions.cc
+++ b/src/Conversions.cc
@@ -61,12 +61,14 @@
 #include <string>
 
 #include "ignition/gazebo/Conversions.hh"
+#include "ignition/gazebo/Export.hh"
 #include "ignition/gazebo/Util.hh"
 
 using namespace ignition;
 
 //////////////////////////////////////////////////
 template<>
+IGNITION_GAZEBO_VISIBLE
 msgs::Entity_Type ignition::gazebo::convert(const std::string &_in)
 {
   msgs::Entity_Type out = msgs::Entity_Type_NONE;
@@ -104,6 +106,7 @@ msgs::Entity_Type ignition::gazebo::convert(const std::string &_in)
 
 //////////////////////////////////////////////////
 template<>
+IGNITION_GAZEBO_VISIBLE
 math::Pose3d ignition::gazebo::convert(const msgs::Pose &_in)
 {
   math::Pose3d out(_in.position().x(),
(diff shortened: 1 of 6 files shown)


src/Conversions.cc
```diff
@@ -61,12 +61,14 @@
...
 
 #include "ignition/gazebo/Conversions.hh"
+#include "ignition/gazebo/Export.hh"
 #include "ignition/gazebo/Util.hh"
 
 using namespace ignition;
 
 //////////////////////////////////////////////////
 template<>
+IGNITION_GAZEBO_VISIBLE
```
chapulina, line 71: Just curious, is this really needed in the CC files?

src/LevelManager.hh
```diff
@@ -74,7 +74,7 @@ namespace ignition
     ///   when the level is reloaded. Likewise, they should not be deleted.
     /// * Entities spawned during simulation are part of the default level.
     ///
-    class LevelManager
+    class IGNITION_GAZEBO_VISIBLE LevelManager
```
chapulina, line 77: This isn't installed, so I think it shouldn't be needed.

src/SdfGenerator.hh
```diff
@@ -44,6 +44,7 @@ namespace sdf_generator
   /// \input[in] _config Configuration for the world generator
   /// \returns Generated world string if generation succeeded.
   /// Otherwise, nullopt
+  IGNITION_GAZEBO_VISIBLE
```
chapulina, line 47: Same about the header not being installed.

chapulina (approved):
This looks good to me. I have some questions about whether some of the changes are necessary, but they don't break anything so I think it's ok to get them in.

You can ignore homebrew CI for now, we're in an intermediate stage there.

src/Conversions.cc
j-rivero: How the code is right now, it is needed. There is an external library (network) using the function in `NetworkManagerPrimary.cc`. Once compiled in ignition-gazebo5.lib, the symbols need to exported to be linked in a different object file.  Otherwise:

```
NetworkManagerPrimary.obj : error LNK2001: unresolved external symbol "class ignition::msgs::WorldStatistics __cdecl ignition::gazebo::v5::convert<class ignition::msgs::WorldStatistics>
```

src/SdfGenerator.hh
j-rivero: The header is not installed but it host declarations of functions that will go into ignition-gazebo5.lib. Once the binary symbols are generated and go in that lib any external binary/lib using the functions will need them to be exported, in this case SdfGenerator_TEST. So yes, we need these ones.

src/LevelManager.hh
j-rivero: same that in https://github.com/ignitionrobotics/ign-gazebo/pull/563#discussion_r560574124

PR #564, opened 2021-01-15
Fix visibility macro names when used by a different component (Windows)

j-rivero:
The PR solves a critical problem: when building a new component (new shared object) the visibility keyword can not be the same than a library it links to. This translated means: our systems and components can not use `IGNITION_GAZEBO_VISIBLE` but the custom name created by ign-cmake in the form of `IGNITION_GAZEBO_${COMPONENT_NAME}_VISIBLE`. Otherwise there is a nice list of fanky linker errors all around.

### DIFF
--- a/src/systems/breadcrumbs/Breadcrumbs.hh
+++ b/src/systems/breadcrumbs/Breadcrumbs.hh
@@ -17,8 +17,8 @@
 #ifndef IGNITION_GAZEBO_SYSTEMS_BREADCRUMBS_HH_
 #define IGNITION_GAZEBO_SYSTEMS_BREADCRUMBS_HH_
 
-#include <optional>
 #include <memory>
+#include <optional>
 #include <set>
 #include <unordered_map>
 #include <vector>
--- a/src/systems/breadcrumbs/Breadcrumbs.hh
+++ b/src/systems/breadcrumbs/Breadcrumbs.hh
@@ -30,7 +30,6 @@
 #include <ignition/transport/Node.hh>
 #include <ignition/math/Pose3.hh>
 
-#include <ignition/gazebo/breadcrumbs-system/Export.hh>
 #include "ignition/gazebo/Model.hh"
 #include "ignition/gazebo/SdfEntityCreator.hh"
 #include "ignition/gazebo/System.hh"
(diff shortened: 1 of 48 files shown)


src/systems/breadcrumbs/Breadcrumbs.hh
```diff
@@ -17,6 +17,7 @@
 #ifndef IGNITION_GAZEBO_SYSTEMS_BREADCRUMBS_HH_
 #define IGNITION_GAZEBO_SYSTEMS_BREADCRUMBS_HH_
 
+#include <optional>
```
chapulina, line 20: Nit: alphabetize

chapulina (commented):
Wow I didn't know those macros even existed! I'm still not sure if the plugins need the visibility to be set though.

Codechecker is sad - otherwise the PR looks good to me.

```
  /github/workspace/include/ignition/gazebo/gui/Gui.hh:63:  Lines should be <= 80 characters long  [whitespace/line_length] [2]
  /github/workspace/src/systems/camera_video_recorder/CameraVideoRecorder.hh:44:  Lines should be <= 80 characters long  [whitespace/line_length] [2]
  /github/workspace/src/systems/joint_position_controller/JointPositionController.hh:75:  Lines should be <= 80 characters long  [whitespace/line_length] [2]
  /github/workspace/src/systems/logical_audio_sensor_plugin/LogicalAudioSensorPlugin.hh:132:  Lines should be <= 80 characters long  [whitespace/line_length] [2]
  /github/workspace/src/systems/multicopter_control/MulticopterVelocityControl.hh:149:  Lines should be <= 80 characters long  [whitespace/line_length] [2]
  /github/workspace/src/systems/multicopter_motor_model/MulticopterMotorModel.hh:37:  Lines should be <= 80 characters long  [whitespace/line_length] [2]
  /github/workspace/src/systems/optical_tactile_plugin/OpticalTactilePlugin.hh:69:  Lines should be <= 80 characters long  [whitespace/line_length] [2]
  /github/workspace/src/systems/triggered_publisher/TriggeredPublisher.hh:157:  Lines should be <= 80 characters long  [whitespace/line_length] [2]
  Total errors found: 8
```

azeey:
> I'm still not sure if the plugins need the visibility to be set though.

Yeah, I don't think we want them to be visible. I would in fact vote for making them `final` classes.

src/systems/breadcrumbs/Breadcrumbs.hh
j-rivero: ea7502a

j-rivero:
> I'm still not sure if the plugins need the visibility to be set though.

I've removed visibility from all plugins. 

> Yeah, I don't think we want them to be visible. I would in fact vote for making them `final` classes.

I agree. I would leave the change of moving them to final classes in a different PR.

chapulina (approved):

PR #565, opened 2021-01-15
Remove invalid visibility macros to fix Windows compilation

j-rivero:
The PR removes some not-needed/invalid visibility macros found in different situations:

 * Extern C functions
 * Executables, test or other binaries not designed to link against
 * Header only classes
 * Convert system to be a header only class


chapulina (approved):
CI is happy, thanks for all the pointers, these are good things to keep in mind for the entire stack.

PR #571, opened 2021-01-19
Change SelectedEntities to return a const ref

brawner:
Instead of returning a copy of the member variable, this PR changes this getter to return a const ref. In `Scene3D.cc`, a couple of places call the method without retaining ownership of the copied vector. This was specifically an issue here: https://github.com/ignitionrobotics/ign-gazebo/blob/bccea9087dd4355fc6d273d41a191d5c16563175/src/gui/plugins/scene3d/Scene3D.cc#L1846

In this line, because the vector is a copy, the `back()` ref operator returns a reference to an immediately deleted object, which will cause a memory read error and a subsequent segfault (in the gazebo gui).

I don't think this should break API too bad. But if someone was actually mutating the copy that was returned, then they could potentially be affected by this. I don't think it can be backported, since it breaks API and ABI. 

Signed-off-by: Stephen Brawner <brawner@gmail.com>

### DIFF
--- a/include/ignition/gazebo/rendering/RenderUtil.hh
+++ b/include/ignition/gazebo/rendering/RenderUtil.hh
@@ -138,7 +138,7 @@ inline namespace IGNITION_GAZEBO_VERSION_NAMESPACE {
 
     /// \brief Get the entities currently selected, in order of selection.
     /// \return Vector of currently selected entities
-    public: std::vector<Entity> SelectedEntities() const;
+    public: const std::vector<Entity>& SelectedEntities() const;
 
     /// \brief Clears the set of selected entities and lowlights all of them.
     public: void DeselectAllEntities();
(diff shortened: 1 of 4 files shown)


chapulina (approved):
LGTM. Could you add a note to the [migration guide](https://github.com/ignitionrobotics/ign-gazebo/blob/main/Migration.md) before merging?

include/ignition/gazebo/rendering/RenderUtil.hh
```diff
@@ -138,7 +138,7 @@ inline namespace IGNITION_GAZEBO_VERSION_NAMESPACE {
 
     /// \brief Get the entities currently selected, in order of selection.
     /// \return Vector of currently selected entities
-    public: std::vector<Entity> SelectedEntities() const;
+    public: const std::vector<Entity> &SelectedEntities() const;
```
brawner, line 141: @chapulina Does ignition style suggest that the `&` be next to the type or the method here, do you know? I just switched it to this after seeing some other getters that return references above.

chapulina: It goes next to the variable, see the [style guide](https://ignitionrobotics.org/docs/all/contributing#style-guides). This is something carried over from Gazebo classic and we have no linters to catch yet.

chapulina (approved):

PR #572, opened 2021-01-20
👩‍🌾 Fix deprecation warnings

chapulina:
Fix warnings introduced in #546.

Also fixed https://github.com/ignitionrobotics/ign-launch/issues/88, which was caused by a missing inline namespace.

I recommend viewing the diff without whitespace.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -59,7 +59,7 @@ set(IGN_MSGS_VER ${ignition-msgs7_VERSION_MAJOR})
 #--------------------------------------
 # Find ignition-common
 # Always use the profiler component to get the headers, regardless of status.
-ign_find_package(ignition-common3 VERSION 3.5
+ign_find_package(ignition-common3 VERSION 3.6
   COMPONENTS
     profiler
     events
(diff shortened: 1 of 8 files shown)


include/ignition/gazebo/detail/EntityComponentManager.hh
```diff
@@ -32,6 +32,8 @@ namespace ignition
 {
 namespace gazebo
 {
+// Inline bracket to help doxygen filtering.
+inline namespace IGNITION_GAZEBO_VERSION_NAMESPACE {
```
azeey, line 36: nit: This was fixed in https://github.com/ignitionrobotics/ign-gazebo/pull/541

CMakeLists.txt
```diff
@@ -59,7 +59,7 @@ set(IGN_MSGS_VER ${ignition-msgs7_VERSION_MAJOR})
 #--------------------------------------
 # Find ignition-common
 # Always use the profiler component to get the headers, regardless of status.
-ign_find_package(ignition-common3 VERSION 3.5
+ign_find_package(ignition-common3 VERSION 3.6
```
azeey, line 62: Is this version bump also for fixing warnings?

azeey (approved):
LGTM!

CMakeLists.txt
chapulina: It's actually failing the build fast... I used a new machine and noticed that building against common 3.5 fails because `ign-gazebo` is using the API from https://github.com/ignitionrobotics/ign-common/pull/70

PR #573, opened 2021-01-20
Updated SDF examples

jennuine:
Added `<size>` element to ground_plane collisions since it is a [required spec of SDFormat](http://sdformat.org/spec?ver=1.7&elem=geometry#geometry_plane) in all example and test SDF files.

This also provides collision visualization for the ground plane: https://github.com/ignitionrobotics/ign-gazebo/pull/531

### DIFF
--- a/examples/scripts/distributed/secondary.sdf
+++ b/examples/scripts/distributed/secondary.sdf
@@ -28,6 +28,7 @@
           <geometry>
             <plane>
               <normal>0 0 1</normal>
+              <size>100 100</size>
             </plane>
           </geometry>
         </collision>
--- a/examples/scripts/distributed/standalone.sdf
+++ b/examples/scripts/distributed/standalone.sdf
@@ -108,6 +108,7 @@
           <geometry>
             <plane>
               <normal>0 0 1</normal>
+              <size>100 100</size>
             </plane>
           </geometry>
         </collision>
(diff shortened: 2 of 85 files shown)


chapulina (approved):
Nice! Even though DART is ignoring the size and creating an infinite plane, we should still specify it because it is required. Other physics engines may use it.

azeey:
Isn't the size of the plane only meaningful for visualization? i.e, I thought all physics engines should create an infinite plane. If a plane is not infinite, why is it needed? You can just use a box shape.

jennuine:
@azeey when trying to view collisions (i.e., visuals) of the ground plane (https://github.com/ignitionrobotics/ign-gazebo/pull/531) when the geometry is a `//plane`, the collision visual could not be viewed because it was missing the `//size` tag. I did notice in some of the examples used a box shape instead of a plane.

chapulina:
> I thought all physics engines should create an infinite plane. If a plane is not infinite, why is it needed? You can just use a box shape.

I think that's a fair question, but I'm not sure we can make this assumption. A box has 6 sides, while a plane can be interpreted to have 1 or 2 at most. So I can imagine a physics engine that implements a finite 2-sided plane without sides, for example.

That could be something interesting to iterate on the SDF spec.

> the collision visual could not be viewed because it was missing the //size tag

That's an interesting use case, because the collision visual may be misleading. Currently, even though the visual will be 100x100, the actual collision created by DART is infinite.

PR #578, opened 2021-01-23
Set thermal camera sensor ambient temperature range using temperature gradient

iche033:
depends on pull request #498 and https://github.com/ignitionrobotics/ign-rendering/pull/211

Objects without a temperature value or heat signature were all given a fixed ambient temperature value. This PR adds support for setting a temperature range so that all other non-heat source objects will now have some variations in their temperature readings.

Here's an example of inserting a backpack to the thermal_camera.sdf world. It relies on a new feature in https://github.com/ignitionrobotics/ign-rendering/pull/211 that simulates some variation in temperature values based on object's color.

rgb_to_temperature

### DIFF
--- a/examples/worlds/thermal_camera.sdf
+++ b/examples/worlds/thermal_camera.sdf
@@ -121,7 +121,14 @@
     </light>
 
     <atmosphere type="adiabatic">
-      <temperature>310</temperature>
+      <temperature>290</temperature>
+      <!--
+        This is a more exaggerated temperature gradient, which produces a
+        temperature range of ~11.5 kelvin for objects in the thermal camera
+        view. Typical temperature gradient is -0.0065 K/m which produces a
+        temperature range of 0.75 kelvin
+      -->
+      <temperature_gradient>0.1</temperature_gradient>
     </atmosphere>
 
     <model name="ground_plane">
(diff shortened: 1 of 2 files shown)


adlarkin (commented):
Does the example picture you shared in the PR [comment](https://github.com/ignitionrobotics/ign-gazebo/pull/578#issue-560322208) use https://github.com/ignitionrobotics/ign-gazebo/pull/498 and https://github.com/ignitionrobotics/ign-rendering/pull/211? I just ran a simple test with this PR and https://github.com/ignitionrobotics/ign-rendering/pull/211. I used the `thermal_camera.sdf` file in this PR, adding a backpack to the file, and got the following results (it's also important to note that the backpack didn't become visible until I started simulation -- i.e., if you run `ign gazebo` without `-r`, the backpack isn't visible until pressing play. In this example, I started simulation and then paused it after the backpack appears):
pr_test

iche033:
> Does the example picture you shared in the PR comment use #498 and ignitionrobotics/ign-rendering#211?

yes I just modified the rescue randy and phone heat signature range values. These are the [min, max] values I used for testing:
rescue randy: [290, 308]
phone: [290, 315]

PR #581, opened 2021-01-25
Added Ellipsoid and Capsule geometries

ahcorde:
This PR allows to visualize in Gazebo capsules and ellipsoides.

As you can see in the images I still have some issues with the collisions:

Selección_096

Related with 

 - ign-rendering https://github.com/ignitionrobotics/ign-rendering/pull/203 and https://github.com/ignitionrobotics/ign-rendering/pull/200
 - ign-msgs https://github.com/ignitionrobotics/ign-msgs/pull/128
 - ign-physics
    - dartsim: https://github.com/ignitionrobotics/ign-physics/pull/203
    - tpe: https://github.com/ignitionrobotics/ign-physics/pull/203


chapulina:
@ahcorde , I pushed commit ac55ced directly to this branch to add an ellipsoid and a capsule to the markers example. It works for me with https://github.com/ignitionrobotics/ign-rendering/pull/276:

ellipse_marker
capsule_marker

chapulina:
The capsule is currently rotated 90 degrees, this fixes it: https://github.com/ignitionrobotics/ign-common/pull/186

chapulina (commented):
Working well! I think the only thing left to solve is not requiring that all physics engines support these shapes. I think we may be able to separate the features on `ign-physics` and add some logic here to conditionally support these features.

shapes_new

rolling_new

chapulina:
> the only thing left to solve is not requiring that all physics engines support these shapes

Looking into it again now, I see that the current implementation doesn't prevent physics engines that don't support capsules and ellipsoids from being loaded. So I think this is good for now. We can always come back and move these shapes into a separate feature if we see the need.

Just pending the `ign-physics` PR :+1:

chapulina (approved):

chapulina:
We're running out of space for so many features! I updated the top toolbar to look like this:

image

chapulina:
There are no related failures on CI, merging!

PR #583, opened 2021-01-26
Backport state update changes from pull request #486

iche033:
pull request #486 had to be targeted at Edifice because of the new `has_one_time_component_changes` field added in ign-msgs7. This PR uses a hacky workaround for storing that field in the msg header so we can backport to Citadel. 

With this change, I'm getting close to ~100% RTF running the distributed_levels.sdf example locally in distributed sim mode, compared to ~30-40% before the change.

On thing I noticed is that the [SetChanged](https://github.com/ignitionrobotics/ign-gazebo/blob/b6b57322252a645b9d7fa7a0122936e9077400d0/src/EntityComponentManager.cc#L1254) call in EntityComponentManager.cc is not reached in standalone mode so this change may not be affect normal ign-gazebo usage (?)

### DIFF
--- a/src/EntityComponentManager.cc
+++ b/src/EntityComponentManager.cc
@@ -1236,8 +1236,22 @@ void EntityComponentManager::SetState(
       {
         std::istringstream istr(compMsg.component());
         comp->Deserialize(istr);
-        this->SetChanged(entity, compIter.first,
-            ComponentState::OneTimeChange);
+        // \todo(anyone) Note on merging forward:
+        // `has_one_time_component` field is available in Edifice so this
+        // workaround can be removed
+        auto flag = ComponentState::PeriodicChange;
+        for (int i = 0; i < _stateMsg.header().data_size(); ++i)
+        {
+          if (_stateMsg.header().data(i).key() ==
+              "has_one_time_component_changes")
+          {
+            int v = stoi(_stateMsg.header().data(i).value(0));
+            if (v)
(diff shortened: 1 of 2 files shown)


chapulina (approved):
Nice hack :nerd_face: 

> SetChanged call in EntityComponentManager.cc is not reached in standalone mode so this change may not be affect normal ign-gazebo usage (?)

During regular simulation, `SetState` is only called on the GUI, when it receives a new state from the server:

https://github.com/ignitionrobotics/ign-gazebo/blob/afb228d6582341c9813986b73c0994c16ee0dde4/src/gui/GuiRunner.cc#L111

src/EntityComponentManager.cc
```diff
@@ -1236,8 +1236,22 @@ void EntityComponentManager::SetState(
       {
         std::istringstream istr(compMsg.component());
         comp->Deserialize(istr);
-        this->SetChanged(entity, compIter.first,
-            ComponentState::OneTimeChange);
+        // \todo(anyone) Note on merging forward:
```
mjcarroll, line 1239: I don't know if this is a todo, since this is already the case in `ign-gazebo4` forward.  Maybe just a note is sufficient?

mjcarroll (approved):
Super small nit, for those of us who search `\todo` occasionally :)

src/EntityComponentManager.cc
iche033: yep done. 03446c6

PR #585, opened 2021-01-27
Final Windows fixes to complete compilation

j-rivero:
The PR include different fixes:

 * Remove unneeded Qt support from GuiSystem
 * Visibility fixes
 * Disable TestSystem on Windows. Needs more inspection about an error with ign-plugin.


src/systems/CMakeLists.txt
```diff
@@ -33,6 +33,17 @@ function(gz_add_system system_name)
...
   set(system_name ${system_name}-system)
 
+
+  if (MSVC)
+    # Warning #4251 is the "dll-interface" warning that tells you when types used
+    # by a class are not being exported. These generated source files have private
+    # members that don't get exported, so they trigger this warning. However, the
+    # warning is not important since those members do not need to be interfaced
+    # with.
+    set_source_files_properties(${sources} COMPILE_FLAGS "/wd4251 /wd4146")
```
chapulina, line 43: I know we suppress 4146 in some other libraries, but it doesn't seem to be related to the dll interface, right?

https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-2-c4146?view=msvc-160

test/plugins/CMakeLists.txt
```diff
@@ -13,12 +13,20 @@ set (test_plugins
...
   TestModelSystem
   TestSensorSystem
-  TestSystem
   TestWorldSystem
   MockSystem
   Null
 )
 
+# TODO: someone with knowledge of ign-plugin please resolve:
+# TestSystem.obj : error LNK2001: unresolved external symbol IgnitionPluginHook
```
chapulina, line 22: This looks like the visibility problem I was having in https://github.com/ignitionrobotics/ign-rendering/pull/180, it may be a visibility flag needing to be fixed in `ign-plugin`. I think leaving this note here is fine for now.

include/ignition/gazebo/gui/GuiSystem.hh
```diff
@@ -43,8 +41,6 @@ namespace gazebo
   /// separate process that is stepped by updates coming through the network
   class GuiSystem : public ignition::gui::Plugin
   {
-    Q_OBJECT
```
chapulina, line 46: Just checking, is this not needed because `ignition::gui::Plugin` already has it?

chapulina:
Nice! I see the build passing here: [Build Status](https://build.osrfoundation.org/job/ign_gazebo-pr-win/2175/) :tada:

src/systems/CMakeLists.txt
j-rivero: uh that is right, I copy paste the chunk of code without looking well into it. What do you prefer, remove it or fix the comment?

include/ignition/gazebo/gui/GuiSystem.hh
j-rivero: If my memory serves me well I removed it since it does not fit into `Notice that the Q_OBJECT macro is mandatory for any object that implements signals, slots or properties` described here https://doc.qt.io/qt-5/qobject.html

src/systems/CMakeLists.txt
chapulina: I'm leaning towards removing and addressing that warning on the code

include/ignition/gazebo/gui/GuiSystem.hh
chapulina: See this:

_We strongly recommend the use of this macro in all subclasses of QObject regardless of whether or not they actually use signals, slots and properties, since failure to do so may lead certain functions to exhibit strange behavior._

We're using the fact that this is a `QObject` in places like this:

https://github.com/ignitionrobotics/ign-gazebo/blob/8fd9b2647a5c37b1dde5509b5a741c6c955ee10f/src/gui/GuiRunner.cc#L93

So I think it should be left there, but I don't know what problem that was causing on Windows.

j-rivero: 79dff1c reverted.

src/systems/CMakeLists.txt
j-rivero: d38c8f3

j-rivero:
New test on Windows [Build Status](https://build.osrfoundation.org/job/ign_gazebo-pr-win/2176/)

j-rivero:
New test after fix visibility [Build Status](https://build.osrfoundation.org/job/ign_gazebo-pr-win/2177/)

chapulina (approved):
Great! Feel free to merge with happy CI.

PR #586, opened 2021-01-27
Use a custom data structure to manage entity feature maps

azeey:
This ensures that when a model is removed, all the related physics entities are removed as well causing the physics engine to clean up the memory used by the model and its children.

This also fixes a segfault that occurs when ign-gazebo is terminated after a model is unloaded by the level manager.  The segfault seems to be due to a bug in DART (https://github.com/dartsim/dart/issues/1534), but the bug is exposed due to the way children
physics entity pointers outlived their parent models because they weren't removed from their entity map when the parent model was unloaded.

### DIFF
--- a/src/systems/physics/EntityFeatureMap.hh
+++ b/src/systems/physics/EntityFeatureMap.hh
@@ -1,0 +1,170 @@
+/*
+ * Copyright (C) 2021 Open Source Robotics Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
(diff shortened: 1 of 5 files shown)


src/systems/physics/EntityFeatureMap.hh
```diff
@@ -0,0 +1,291 @@
...
+namespace ignition::gazebo
+{
+inline namespace IGNITION_GAZEBO_VERSION_NAMESPACE {
+namespace systems::physics_system
+{
+  // \brief Helper class that associates Gazebo entities with Physics entities
+  // with required and optional features. It can be used to cast a physics
+  // entity with the required features to another physics entity with one of
+  // the optional features. This class was created to keep all
+  // physics entities are kept in one place so that when a gazebo entity is
```
nkoenig, line 41:
```suggestion
  // physics entities in one place so that when a gazebo entity is
```

nkoenig:
Levels and subt work for me. Just one documentation comment.

nkoenig (approved):

src/systems/physics/EntityFeatureMap.hh
azeey: https://github.com/ignitionrobotics/ign-gazebo/pull/586/commits/3f02ad8ccd420484bd4f2864bccf85ba5b707cfc

azeey:
The macOS failure seems like an infrastructure issue. Here's the previous build [Build Status](https://build.osrfoundation.org/job/ignition_gazebo-ci-pr_any-homebrew-amd64/5105/). The only change after that was documentation.

PR #587, opened 2021-01-27
Plugin to spawn lights

ahcorde:
I created a plugin to spawn lights in ignition-gazebo.

lights_spawn

I have some questions and issues:
 - Do you want to see the visual in the component inspector ? the same way we are visualizing link, collisions or lights for models ?
 - I'm getting sometimes this error:
 ```bash
terminate called after throwing an instance of 'Ogre::RenderingAPIException'
  what():  OGRE EXCEPTION(3:RenderingAPIException): Fragment Program 537068800PixelShader_ps failed to compile. See compile log above for details. in GLSLShader::compile at /var/lib/jenkins/workspace/ogre-2.1-debbuilder/repo/RenderSystems/GL3Plus/src/GLSL/OgreGLSLShader.cpp (line 305)
```
 - Where and how is the best place to create the `LightVisual` ? Right now it's in the `RenderUtil.cc` but when I remove the light from Gazebo in the component inspector the visual it's still there. Probably I should attach the Visual to the Light. Is this possible ?
https://github.com/ignitionrobotics/ign-gazebo/blob/139e552a2fa2869794020313bea975143c7f273c/src/rendering/RenderUtil.cc#L404-L417

### CODE REVIEW

**Comment by iche033** [comment]
> Do you want to see the visual in the component inspector ? the same way we are visualizing link, collisions or lights for models ?

you mean the light visuals? I think it's fine to leave them out

> I'm getting sometimes this error:

I see this when spawning a directional light. It could be that we can't have more than one directional lights in the scene. If that's the case, we can print a warning that only 1 directional light is supported for now

> Where and how is the best place to create the LightVisual ? Right now it's in the RenderUtil.cc but when I remove the light from Gazebo in the component inspector the visual it's still there. Probably I should attach the Visual to the Light. Is this possible ?

yes you can attach LightVisual to Light. Since the they are all ign-rendering nodes, you can just do `light->AddChild(lightVisual)`

**Comment by ahcorde** [comment]
@iche033 I improved the implementation.

The issue with the lights is the following: When point lights and directionals are mixed the segfault appears

error2

error1

**Comment by iche033** [comment]
ok narrowed the the point vs directional light issue a bit more. Disabling the point light shadows makes them work now:

```
--- a/src/gui/plugins/lights/Lights.cc
+++ b/src/gui/plugins/lights/Lights.cc
@@ -86,7 +86,7 @@ void Lights::OnMode(const QString &_mode)
                                  "<sdf version=\"1.6\">"
                                  "<light type='point' name='pointlight'>"
                                    "<pose>0 0 2 0 0 0</pose>"
-                                   "<cast_shadows>true</cast_shadows>"
+                                   "<cast_shadows>false</cast_shadows>"
                                    "<diffuse>0.5 0.5 0.5 1</diffuse>"
                                    "<specular>0.5 0.5 0.5 1</specular>"
                                    "<attenuation>"
```

**Comment by ahcorde** [comment]
@iche033 It's ready for a review, thank you for the catch with the lights.

- Related with the PR https://github.com/ignitionrobotics/ign-rendering/pull/202

File: include/ignition/gazebo/components/LightType.hh [hunk 1]
```diff
@@ -0,0 +1,46 @@
...
+namespace ignition
+{
+namespace gazebo
+{
+// Inline bracket to help doxygen filtering.
+inline namespace IGNITION_GAZEBO_VERSION_NAMESPACE {
+namespace components
+{
+  /// \brief This component holds an entity's type. The component has no concept
+  /// of scoped types nor does it care about uniqueness.
```
chapulina, line 37: The documentation doesn't look right

src/SdfEntityCreator.cc
```diff
@@ -379,6 +380,32 @@ Entity SdfEntityCreator::CreateEntities(const sdf::Light *_light)
...
+  }
+  else if (_light->Type() == sdf::LightType::DIRECTIONAL)
+  {
+    this->dataPtr->ecm->CreateComponent(lightEntity,
+        components::LightType(std::string("Directional")));
+  }
+  else if (_light->Type() == sdf::LightType::SPOT)
+  {
+    this->dataPtr->ecm->CreateComponent(lightEntity,
+        components::LightType(std::string("Spot")));
```
chapulina, line 396: How about creating a `convert` function from `sdf::LightType` to `std::string` in `Conversions.hh`? It would be better to reuse the same logic in many places. I'm concerned that different plugins may use different strings (i.e. `spot`, `SPOT`...).

I can even see this function living in libSDFormat.

Also, I think that we should store the types as lowercase. The GUI can capitalize them as needed.

src/SdfEntityCreator.cc
```diff
@@ -379,6 +380,32 @@ Entity SdfEntityCreator::CreateEntities(const sdf::Light *_light)
...
+  }
+
+  // Entity
+  Entity visualEntity = this->dataPtr->ecm->CreateEntity();
```
chapulina, line 400: I'm not sure that creating a visual entity for the light is the best approach here. The light visuals exist for visualization purposes. They're the 3D representation of a light, like the orange boxes are the 3D representation of collisions. They're not simulated entities with their own properties. I recommend removing these entities.

src/gui/plugins/component_inspector/ComponentInspector.cc
```diff
@@ -500,6 +501,15 @@ void ComponentInspector::Update(const UpdateInfo &,
...
         this->dataPtr->worldName = comp->Data();
     }
+    else if (typeId == components::LightType::typeId)
+    {
+      auto comp = _ecm.Component<components::LightType>(this->dataPtr->entity);
+      if (comp)
+        setData(item, comp->Data());
+
+      if (this->dataPtr->entity == this->dataPtr->worldEntity)
+        this->dataPtr->worldName = comp->Data();
```
chapulina, line 511: This if block looks out of place, the world name is not a light type :wink:

src/gui/plugins/lights/Lights.cc
```diff
@@ -0,0 +1,163 @@
...
+
+  // For shapes requests
+  ignition::gui::App()->findChild
+    <ignition::gui::MainWindow *>()->installEventFilter(this);
+}
+
+/////////////////////////////////////////////////
+void Lights::OnMode(const QString &_mode)
+{
+  std::string modelSdfString = _mode.toStdString();
```
chapulina, line 79: It's weird to prefix all of these variables with `model`, because there are no models here, these are lights. I think this could just be `sdfString` to save confusion.

I know that the event has `Model` in the name, but that's changing to be more generic in #595

src/gui/plugins/lights/Lights.cc
```diff
@@ -0,0 +1,163 @@
...
+
+/////////////////////////////////////////////////
+void Lights::LoadConfig(const tinyxml2::XMLElement *)
+{
+  if (this->title.empty())
+    this->title = "Lights";
+
+  // For shapes requests
+  ignition::gui::App()->findChild
+    <ignition::gui::MainWindow *>()->installEventFilter(this);
```
chapulina, line 73: This plugin doesn't have the `eventFilter` callback, so this call isn't doing anything.

src/gui/plugins/lights/Lights.cc
```diff
@@ -0,0 +1,163 @@
...
+  class LightsPrivate
+  {
+    /// \brief Ignition communication node.
+    public: transport::Node node;
+
+    /// \brief Mutex to protect mode
+    public: std::mutex mutex;
+
+    /// \brief Transform control service name
+    public: std::string service;
```
chapulina, line 48: None of these variables is used.

src/gui/plugins/lights/Lights.cc
```diff
@@ -0,0 +1,163 @@
...
+ * limitations under the License.
+ *
+*/
+
+#include "Lights.hh"
+
+#include <ignition/msgs/boolean.pb.h>
+#include <ignition/msgs/stringmsg.pb.h>
+
+#include <algorithm>
```
chapulina, line 23: Do you need `algorithm`?

src/gui/plugins/lights/Lights.hh
```diff
@@ -0,0 +1,57 @@
...
+    public: Lights();
+
+    /// \brief Destructor
+    public: ~Lights() override;
+
+    // Documentation inherited
+    public: void LoadConfig(const tinyxml2::XMLElement *_pluginElem) override;
+
+    /// \brief Callback in Qt thread when mode changes.
+    /// \param[in] _mode New transform mode
```
chapulina, line 47: The documentation doesn't match the implementation. I think this was copied from `Shapes`, I see the same problem there. In fact, even the `OnMode` function name feels strange in this context.

src/rendering/SceneManager.cc
```diff
@@ -848,6 +849,98 @@ rendering::VisualPtr SceneManager::CreateActor(Entity _id,
...
+      auto it2 = this->dataPtr->lights.find(_parentId);
+      if (it2 != this->dataPtr->lights.end())
+      {
+        lightParent = it2->second;
+      }
+    }
+    if (!visualParent && !lightParent)
+    {
+      ignerr << "Parent entity with Id: [" << _parentId << "] not found. "
+             << "Not adding model visual with ID[" << _id
```
chapulina, line 889:
```suggestion
             << "Not adding light visual with ID[" << _id
```

src/rendering/SceneManager.cc
```diff
@@ -848,6 +849,98 @@ rendering::VisualPtr SceneManager::CreateActor(Entity _id,
...
+
+  std::string name = _light.Name().empty() ? std::to_string(_id) :
+      _light.Name();
+
+  rendering::VisualPtr visualParent;
+  rendering::LightPtr lightParent;
+  if (_parentId != this->dataPtr->worldId)
+  {
+    auto it = this->dataPtr->visuals.find(_parentId);
+    if (it != this->dataPtr->visuals.end())
```
chapulina, line 874: Shouldn't a light visual always be attached to a light? In which situation would a light visual be attached to another visual? Looking at the code, it looks like `_parentId` will either be the ID of a light or a world.

chapulina (changes requested):

src/rendering/SceneManager.cc
ahcorde: When there is an object with a light inside (like the point light in the sphere of the `light.sdf` example) we need to use this logic, otherwise the light visual is not visualized. We need to add the light visual child to the upper visual, in this case the sphere visual.

src/SdfEntityCreator.cc
ahcorde: The idea behind this it to let the user know that there is a visual attached to the light. The same way we include visual or collisions to a model. But I fine removing it

src/gui/plugins/lights/Lights.cc
ahcorde: `std::transform`  is used on `OnNewLightClicked`

src/SdfEntityCreator.cc
chapulina: I think the word "visual" has been a source of confusion since Gazebo classic. There are visuals in the SDF, which are those inside a `<visual>` tag, and there are visuals for the rendering scene, which can represent various different SDF tags, like models, links, collisions, lights... This often leads to very confusing terminology like [visual visual](https://github.com/osrf/gazebo/blob/92248a75919603293fbe67bc2716eb59010e7eec/test/integration/visual_pose.cc#L63-L64) for the "rendering visual of an SDF visual".

This is to say that the "visual entities" on Ignition Gazebo correspond to the SDF visuals. While what you're adding here is a visualization of a `<light>` entity. Since they're meant for visualization purposes, they're not part of the simulation, and the ECM doesn't need to know about them. That's the same way as we have link visuals, model visuals and collision visuals - all which are not represented as separate entities.

Let me know if this makes things clearer.

ahcorde: Thank you for the clarification, then we don't need this portion of code. It was removed

src/rendering/SceneManager.cc
iche033: I started looking at the idea of creating LightVisuals always as a child of light. I think this may work: #608

ahcorde:
@iche033 Failing tests are unrelated:

 - Ubuntu
```
INTEGRATION_depth_camera.test_ran

```
 - MacOS
```
INTEGRATION_level_manager.test_ran
INTEGRATION_physics_system.test_ran
INTEGRATION_touch_plugin.test_ran
```

iche033:
Just one unaddressed comment from @chapulina:
> Also, I think that we should store the types as lowercase. The GUI can capitalize them as needed.

otherwise this works for me. I don't know why windows CI is failing - it's happening in other PRs too.

ahcorde:
@iche033, I addressed the comment from @chapulina https://github.com/ignitionrobotics/ign-gazebo/pull/587/commits/2a79677fac5062f04365cfc13666890886c55030

iche033 (approved):

src/rendering/RenderUtil.cc
```diff
@@ -400,6 +400,20 @@ void RenderUtil::Update()
     {
       this->dataPtr->sceneManager.CreateLight(
           std::get<0>(light), std::get<1>(light), std::get<2>(light));
+
+      // create a new id for the light visual
+      auto timeout = 100000u;
+      for (auto i = 0u; i < timeout; ++i)
```
chapulina, line 406: I don't think it's a good idea to block the render thread. If this may take time to be completed, I suggest we queue it and process it when possible.

src/rendering/SceneManager.cc
```diff
@@ -925,6 +926,76 @@ rendering::VisualPtr SceneManager::CreateActor(Entity _id,
...
+           << "]  and name [" << name << "] to the rendering scene."
+           << std::endl;
+    return rendering::VisualPtr();
+  }
+
+  name = lightParent->Name() +  "::" + name + "Visual";
+
+  if (this->dataPtr->scene->HasVisualName(name))
+  {
+    ignerr << "Visual: [" << name << "] already exists" << std::endl;
```
chapulina, line 965: I'm hitting this line when I do:

1. Insert a spot light
2. Delete the spot light
3. Insert a new spot light, and I see:
    `[GUI] [Err] [SceneManager.cc:965] Visual: [spotlight::spotlightVisual] already exists`

I believe the light visual isn't being deleted with the light, and this can cause large memory leaks down the road. We should make sure the light visual is deleted together with the light.

chapulina (commented):
Works great with the event change we talked about.

Besides the comment below, I have 2 requests:

* We should add the `Lights` plugin to the top toolbar next to the shapes in [gui.config](https://github.com/ignitionrobotics/ign-gazebo/blob/766bb23eaca85ff60941e2ef852c4f286356fbcc/src/gui/gui.config#L107).
* Now that the lights have a visual, most example worlds will have a green square near the origin for the sun. This may confuse users. On Gazebo classic, I've seen many users confused about the floating green square, imagine if it's on the ground :sweat_smile:  I suggest we move all suns in the examples to z=10, so it's in a more "sun-like" position. I think it's ok if this comes in a future PR though.

src/rendering/RenderUtil.cc
ahcorde: the right name for this variable is `attempts`, this loop has no relation with time or sleep. It's looking for an available entity ID.

ahcorde:
> * Now that the lights have a visual, most example worlds will have a green square near the origin for the sun. This may confuse users. On Gazebo classic, I've seen many users confused about the floating green square, imagine if it's on the ground sweat_smile  I suggest we move all suns in the examples to z=10, so it's in a more "sun-like" position. I think it's ok if this comes in a future PR though

@chapulina I made a quick review in the example folder. The suns' poses are already `z=10` 

 ```xml
    <light type="directional" name="sun">
      <cast_shadows>true</cast_shadows>
      <pose>0 0 10 0 0 0</pose>
      <diffuse>1 1 1 1</diffuse>
      <specular>0.5 0.5 0.5 1</specular>
      <attenuation>
        <range>1000</range>
        <constant>0.9</constant>
        <linear>0.01</linear>
        <quadratic>0.001</quadratic>
      </attenuation>
      <direction>-0.5 0.1 -0.9</direction>
    </light>
```

File: src/rendering/RenderUtil.cc (hunk 12 above)
Comment (chapulina) [reply src/rendering/RenderUtil.cc]: Ohh ok understood! I wonder if there's a chance of entity ID collision when new entities are inserted into the simulation from the backend. I didn't notice problems in my manual testing earlier.

Comment (ahcorde) [reply src/rendering/RenderUtil.cc]: In this method I can't access the ECM to call `CreateEntity`. Is there any other way to get an empty ID ?

Comment (chapulina) [reply src/rendering/RenderUtil.cc]: For the entity preview when spawning, there's this method in the `Scene3D` plugin:

https://github.com/ignitionrobotics/ign-gazebo/blob/766bb23eaca85ff60941e2ef852c4f286356fbcc/src/gui/plugins/scene3d/Scene3D.cc#L927-L937

That plugin isn't accessible from `RenderUtil`. But I think it could make sense to move the light visual logic from `RenderUtil` to `Scene3d`. The light visuals are only supposed to show on the client scene, not on the server. But `RenderUtil` is used by both scenes. I think this would be a big change and it may take longer to merge this PR though :grimacing:

Comment (ahcorde) [reply src/rendering/RenderUtil.cc]: I can open an issue and merge this PR. Waht do you think @chapulina ?

Comment (ahcorde) [reply src/rendering/RenderUtil.cc]: friendly ping @chapulina

Comment (chapulina) [reply src/rendering/RenderUtil.cc]: An issue sounds good, thanks!

Comment (ahcorde) [reply src/rendering/RenderUtil.cc]: issue created https://github.com/ignitionrobotics/ign-gazebo/issues/633

File: src/gui/gui.config [hunk 14]
```diff
@@ -120,6 +120,22 @@
...
+      <line own="top" target="top"/>
+    </anchors>
+    <property key="resizable" type="bool">false</property>
+    <property key="width" type="double">250</property>
+    <property key="height" type="double">50</property>
+    <property key="state" type="string">floating</property>
+    <property key="showTitleBar" type="bool">false</property>
+    <property key="cardBackground" type="string">#666666</property>
+  </ignition-gui>
+</plugin>
```
chapulina, line 137: This currently looks like this:

image

How about not using the anchors (because of https://github.com/ignitionrobotics/ign-gui/issues/129), like this:

<details>
  <summary>diff</summary>
  
```.diff
diff --git a/src/gui/gui.config b/src/gui/gui.config
index fe87f0d5..8f1b33a4 100644
--- a/src/gui/gui.config
+++ b/src/gui/gui.config
@@ -91,12 +91,10 @@
 <plugin filename="TransformControl" name="Transform control">
   <ignition-gui>
     <title>Transform control</title>
-    <anchors target="3D View">
-      <line own="left" target="left"/>
-      <line own="top" target="top"/>
-    </anchors>
     <property key="resizable" type="bool">false</property>
-    <property key="width" type="double">230</property>
+    <property key="x" type="double">0</property>
+    <property key="y" type="double">0</property>
+    <property key="width" type="double">250</property>
     <property key="height" type="double">50</property>
     <property key="state" type="string">floating</property>
     <property key="showTitleBar" type="bool">false</property>
@@ -107,12 +105,10 @@
 <!-- Insert simple shapes -->
 <plugin filename="Shapes" name="Shapes">
   <ignition-gui>
-    <anchors target="Transform control">
-      <line own="left" target="right"/>
-      <line own="top" target="top"/>
-    </anchors>
     <property key="resizable" type="bool">false</property>
-    <property key="width" type="double">250</property>
+    <property key="x" type="double">250</property>
+    <property key="y" type="double">0</property>
+    <property key="width" type="double">150</property>
     <property key="height" type="double">50</property>
     <property key="state" type="string">floating</property>
     <property key="showTitleBar" type="bool">false</property>
@@ -123,12 +119,10 @@
 <!-- Insert lights -->
 <plugin filename="Lights" name="Lights">
   <ignition-gui>
-    <anchors target="Shapes">
-      <line own="left" target="right"/>
-      <line own="top" target="top"/>
-    </anchors>
     <property key="resizable" type="bool">false</property>
-    <property key="width" type="double">250</property>
+    <property key="x" type="double">400</property>
+    <property key="y" type="double">0</property>
+    <property key="width" type="double">150</property>
     <property key="height" type="double">50</property>
     <property key="state" type="string">floating</property>
     <property key="showTitleBar" type="bool">false</property>
```

</details>

It looks like this:

image

chapulina (approved):
It's working for me! One last thing I noticed is that the directional light is spawned higher than the others, which may confuse users depending on their zoom level:

Peek 2021-02-17 13-07

I think this is good to merge when the last 2 details are addressed, and if CI is happy.

ahcorde:
@chapulina I fixed the two suggestions:

 - https://github.com/ignitionrobotics/ign-gazebo/pull/587/commits/a984f2e5bf18bd570d7a8c064b14d8146ca468ca 
 - https://github.com/ignitionrobotics/ign-gazebo/pull/587/commits/73eb8eb79dbd42147c3213c90bb4a60f4ed77450

Feel free to squash and merge it

ahcorde:
Failing tests are unrelated:

```
INTEGRATION_touch_plugin.test_ran
INTEGRATION_depth_camera.test_ran
```

Merging!

PR #588, opened 2021-01-28
Added screenshot to toolbar

jennuine:
~~Closes https://github.com/ignitionrobotics/ign-gui/issues/95
Added a screenshot service which saves an image of the current 3D scene from the user camera.
This depends on the ign-gui screenshot plugin https://github.com/ignitionrobotics/ign-gui/pull/170~~

Depends on https://github.com/ignitionrobotics/ign-gui/pull/170

Updated [gui.config](https://github.com/ignitionrobotics/ign-gazebo/blob/ign-gazebo3/src/gui/gui.config) toolbar layout to use x / y positions instead of anchors and to contain the screenshot plugin.

image


chapulina (commented):
With the suggestions on https://github.com/ignitionrobotics/ign-gui/pull/171 and #598, the `ign-gui` plugin can be used with both `ign-gazebo` and `ign-gui`, so we don't need to duplicate the logic here.

If you think those changes are ok, I'd suggest we update this pull request to:

* Remove all the changes to `Scene3D`
* Add the screenshot plugin from `ign-gui` to [gui.config](https://github.com/ignitionrobotics/ign-gazebo/blob/ign-gazebo3/src/gui/gui.config) next to the shapes on the grey toolbar.

chapulina (approved):
LGTM! We just need an `ign-gui` release before merging: https://github.com/ignitionrobotics/ign-gui/pull/188

chapulina:
All the test failures seem to be due to a problem on Ignition Fuel's server:

PR #591, opened 2021-01-28
Depend on ignition-utils1

chapulina:
Part of https://github.com/ignitionrobotics/ign-gazebo/issues/568

I didn't bother linking to the library in this PR, so this can focus on adding the dependency and getting CI turning.

I'll make a follow up PR that will actually use `ign-utils`.

PR #595, opened 2021-01-29
Deprecate some gazebo::gui events in favor of ign-gui events

chapulina:
Builds on top of #591 

---

Deprecate events that have been ported to Ignition GUI.

* Deprecate warnings on v5, remove on v6
* Emitting side emits both old and new events on v5. This should allow plugins using these events to keep working, with warnings.
* Receiving side only processes new events. So if any user plugin is emitting the old events expecting our plugins to react, that won't work anymore. I don't think this is likely though.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -129,6 +129,11 @@ ign_find_package(ignition-tools
                  PKGCONFIG "ignition-tools")
 
 #--------------------------------------
+# Find ignition-utils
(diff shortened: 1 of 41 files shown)


caguero (approved):

caguero (changes requested):
CI isn't happy. There's a style error.

chapulina:
> There's a style error.

Fixed in f95d3b3

ahcorde (approved):

PR #600, opened 2021-01-30
Fix topLevelModel method

adlarkin:
Currently, the [`topLevelModel`](https://github.com/ignitionrobotics/ign-gazebo/blob/f6a7329ad7a283034a54647d44c4c061666cfeb8/src/Util.cc#L397-L419) method doesn't work when starting at an entity that has a parent entity that isn't a model entity. For example, consider the following entity tree:
```
model
    - link
        - collision
        - visual
```

If we start at either the `collision` or `visual` entity, the issue is that since the parent entity (`link`) is not a model, [this line](https://github.com/ignitionrobotics/ign-gazebo/blob/f6a7329ad7a283034a54647d44c4c061666cfeb8/src/Util.cc#L411) returns true, so the method breaks from the while loop, returning the starting entity as the "top level model".

Signed-off-by: Ashton Larkin <ashton@larkinfam.com>

### DIFF
--- a/include/ignition/gazebo/Util.hh
+++ b/include/ignition/gazebo/Util.hh
@@ -134,7 +134,8 @@ namespace ignition
     /// \brief Get the top level model of an entity
     /// \param[in] _entity Input entity
     /// \param[in] _ecm Constant reference to ECM.
-    /// \return Entity of top level model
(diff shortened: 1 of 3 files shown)


chapulina (approved):
Nice, thanks for catching the problem and for the fix :+1:

PR #606, opened 2021-02-03
Add /image suffix to thermal camera topic name

iche033:
Adding `/image` suffix to make the topic name consistent with a normal camera sensor. This also fixes the `camera_info` topic name that's generated on the ign-sensors side. 

Topic changes:
    * `/<prefix>/<sensor_name>` -> `/<prefix>/<sensor_name>/image`
    * `/<prefix>/camera_info` -> `/<prefix>/<sensor_name>/camera_info`

### DIFF
--- a/Migration.md
+++ b/Migration.md
@@ -11,6 +11,13 @@ release will remove the deprecated code.
   `const std::vector<Entity> &` instead of forcing a copy. The calling code
   should create a copy if it needs to modify the vector in some way.
 
+* Default generated topic name for thermal cameras now includes the `/image`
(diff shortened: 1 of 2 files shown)


chapulina (commented):
Does the example need to be updated?

https://github.com/ignitionrobotics/ign-gazebo/blob/d50acebae4b65df18928b0a7a9e62a252be1c608/examples/worlds/thermal_camera.sdf#L104

**Update**: No, it doesn't - it's using a custom topic.

chapulina (approved):

PR #607, opened 2021-02-03
Fix build with latest sdformat11 branch

chapulina:
The const-correctness introduced in https://github.com/osrf/sdformat/pull/474 broke `ign-gazebo`. I originally thought this was a Windows-only issue, but it turns out it manifested first on Windows because CI builds SDFormat from source.

### DIFF
--- a/src/Conversions.cc
+++ b/src/Conversions.cc
@@ -277,8 +277,8 @@ msgs::Material ignition::gazebo::convert(const sdf::Material &_in)
   auto pbr = _in.PbrMaterial();
   if (pbr)
   {
-    msgs::Material::PBR *pbrMsg = out.mutable_pbr();
-    sdf::PbrWorkflow *workflow = pbr->Workflow(sdf::PbrWorkflowType::METAL);
+    auto pbrMsg = out.mutable_pbr();
+    auto workflow = pbr->Workflow(sdf::PbrWorkflowType::METAL);
     if (workflow)
       pbrMsg->set_type(msgs::Material_PBR_WorkflowType_METAL);
     else
--- a/src/Conversions.cc
+++ b/src/Conversions.cc
@@ -274,7 +274,7 @@ msgs::Material ignition::gazebo::convert(const sdf::Material &_in)
   out.set_lighting(_in.Lighting());
   out.set_double_sided(_in.DoubleSided());
 
-  sdf::Pbr *pbr = _in.PbrMaterial();
+  auto pbr = _in.PbrMaterial();
   if (pbr)
   {
     msgs::Material::PBR *pbrMsg = out.mutable_pbr();
--- a/src/Model.cc
+++ b/src/Model.cc
@@ -33,7 +33,8 @@ class ignition::gazebo::ModelPrivate
   public: Entity id{kNullEntity};
 };
 
-using namespace ignition::gazebo;
+using namespace ignition;
+using namespace gazebo;
 
 //////////////////////////////////////////////////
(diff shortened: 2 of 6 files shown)


src/Conversions.cc
```diff
@@ -274,11 +274,11 @@ msgs::Material ignition::gazebo::convert(const sdf::Material &_in)
   out.set_lighting(_in.Lighting());
   out.set_double_sided(_in.DoubleSided());
 
-  sdf::Pbr *pbr = _in.PbrMaterial();
+  auto pbr = _in.PbrMaterial();
   if (pbr)
   {
-    msgs::Material::PBR *pbrMsg = out.mutable_pbr();
-    sdf::PbrWorkflow *workflow = pbr->Workflow(sdf::PbrWorkflowType::METAL);
+    auto pbrMsg = out.mutable_pbr();
```
sloretz, line 280: This type's const-ness was already correct.

src/Model.cc
```diff
@@ -33,7 +33,8 @@ class ignition::gazebo::ModelPrivate
   public: Entity id{kNullEntity};
 };
 
-using namespace ignition::gazebo;
+using namespace ignition;
+using namespace gazebo;
```
sloretz, line 37: Will anything this file includes ever be in the `gazebo` namespace? If so, it looks like splitting the using declaration would cause a compile error.

```
#include <iostream>

namespace gazebo
{
  void say_hello() { std::cout << "Hello from gazebo::say_hi()\n";}
}

namespace ignition
{
namespace gazebo
{
  void say_hi() { std::cout << "Hi from ignition::gazebo::say_hi()\n";}
}
}


// using namespace ignition::gazebo;
using namespace ignition;
using namespace gazebo;

int main()
{

  say_hello();
  say_hi();
  return 0;
}
```

```console
$ g++ -std=c++14 test.cpp
test.cpp:19:17: error: reference to ‘gazebo’ is ambiguous
 using namespace gazebo;
[...]
```

src/systems/user_commands/UserCommands.cc
```diff
@@ -542,9 +542,9 @@ bool CreateCommand::Execute()
   }
   else if (isLight && isRoot)
   {
-    auto light = root.Light();
-    light->SetName(desiredName);
-    entity = this->iface->creator->CreateEntities(light);
+    auto light = *root.Light();
```
sloretz, line 545: Hard to tell, does this cause a copy where there wasn't one before, or is `root.Light()` returning a reference?

src/Model.cc
chapulina: I don't think that's a concern, we don't use any `gazebo` namespace outside of `ignition`.

src/systems/user_commands/UserCommands.cc
chapulina: Yeah the idea is to make a copy, because `Root::Light()` returns a const pointer. I think this is only a problem on Windows. As far as I can tell, that has been returning a const pointer since it was implemented and didn't cause issues on Linux or macOS. That line was introduced in #572, before we enabled Windows CI.

chapulina:
I don't know why the Windows CI result is not coming back, but it built correctly. Test failures and warnings are expected.

[Build Status](https://build.osrfoundation.org/job/ign_gazebo-pr-win/2235/)

src/Conversions.cc
chapulina: Oh you're right. I'm ok letting this through to save another round of CI.

caguero (approved):
Looks good to me.

Issue #371, opened 2020-09-24
About dialog with versions and links to online resources


chapulina:
Most applications will have an "About" window that shows the exact version that the user is running. Here's Gazebo-classic's one for example:

![image](https://user-images.githubusercontent.com/5751272/94183012-a3fb0100-fe56-11ea-9340-7a94460f5e78.png)

This is handy for users that are not on a terminal or aren't running with enough verbosity to see the version printed.

I also think it would be handy to print the version of all Ignition libraries that Gazebo is using.

And we should have links to online resources too.

Ignition GUI has a [silly About dialog](https://github.com/ignitionrobotics/ign-gui/blob/22434e23832760130c88d38d7c569a049dcec65c/include/ignition/gui/qml/Main.qml#L221-L245) that could be updated as well:

![image](https://user-images.githubusercontent.com/5751272/94183430-44512580-fe57-11ea-879e-9ca75fa97c83.png)

chapulina:
Some tips on getting started with this issue:

1. I recommend targeting Ignition Citadel, which is branch `ign-gazebo3`
1. The dialog code can go directly into [GazeboDrawer.qml](https://github.com/ignitionrobotics/ign-gazebo/blob/ign-gazebo3/src/gui/resources/GazeboDrawer.qml)
1. We can create a new C++ class similar to [GuiFileHandler](https://github.com/ignitionrobotics/ign-gazebo/blob/ign-gazebo3/src/gui/GuiFileHandler.hh) that will provide version information to the QML code. To start with, it could just provide [IGNITION_GAZEBO_VERSION_FULL](https://github.com/ignitionrobotics/ign-gazebo/blob/ign-gazebo3/include/ignition/gazebo/config.hh.in).

atharva-18:
@chapulina 
I have created a dialog box inside `GazeboDrawer.qml`

![Screenshot from 2021-01-30 13-15-51](https://user-images.githubusercontent.com/39728574/106350749-552a1000-62fd-11eb-9a6b-cde09c8f3eb2.png)


Currently, I'm providing the version number and header using `IGNITION_GAZEBO_VERSION_HEADER`, and the dialog content is:
```
std::string(IGNITION_GAZEBO_VERSION_HEADER)
<div>
  <table>
    ...
  </table>
</div>
```
However, the hyperlink to the tutorials is specific to Citadel, which might cause an issue in a future version, when the version name changes. Is there a general way to provide the hyperlinks?

[...]

PR #609, opened 2021-02-03
Add AboutDialogHandler

atharva-18:
Reference to #371 

Summary:
1. Added about dialog to `GazeboDrawer.qml`
2. Added handler for the above dialog. 
3. Changed `ListElement` QML type's `action` attribute to `actionElement`.

Screenshots:
Screenshot from 2021-02-03 22-30-54
Screenshot from 2021-02-03 22-31-20
Screenshot from 2021-02-03 22-32-04


src/gui/AboutDialogHandler.cc
```diff
@@ -0,0 +1,85 @@
+/*
+ * Copyright (C) 2020 Open Source Robotics Foundation
```
ahcorde, line 2:
```suggestion
 * Copyright (C) 2021 Open Source Robotics Foundation
```

src/gui/AboutDialogHandler.hh
```diff
@@ -0,0 +1,58 @@
+/*
+ * Copyright (C) 2020 Open Source Robotics Foundation
```
ahcorde, line 2:
```suggestion
 * Copyright (C) 2021 Open Source Robotics Foundation
```

ahcorde (changes requested):
Please review the cpplint errors https://github.com/ignitionrobotics/ign-gazebo/pull/609/checks?check_run_id=1824526153

src/gui/CMakeLists.txt
```diff
@@ -1,5 +1,6 @@
 set (gui_sources
   Gui.cc
+  AboutDialogHandler.cc
```
chapulina, line 3: Please place it in alphabetic order

src/gui/resources/GazeboDrawer.qml
```diff
@@ -165,6 +172,34 @@ Rectangle {
...
+   * About dialog
+   */
+  Dialog {
+    id: aboutDialog
+    title: "Ignition Gazebo"
+
+    modal: true
+    focus: true
+    parent: ApplicationWindow.overlay
+    width: parent.width / 4
```
chapulina, line 185: How about we make this `/ 3` and give it a maximum width so it can't get too large? It looks like this for me with a smaller window:

image

src/gui/resources/GazeboDrawer.qml
```diff
@@ -165,6 +172,34 @@ Rectangle {
...
+
+    Text {
+      anchors.fill: parent
+      id: aboutMessage
+      wrapMode: Text.Wrap
+      verticalAlignment: Text.AlignVCenter
+      color: Material.theme == Material.Light ? "black" : "white"
+      textFormat: Text.RichText
+      text: AboutDialogHandler.getVersionInformation()
+      onLinkActivated: AboutDialogHandler.openURL(link)
```
chapulina, line 199: I haven't investigated why, but the links are not clickable for me. Are they working for you?

src/gui/AboutDialogHandler.cc
```diff
@@ -0,0 +1,89 @@
...
+                 "</tr>"
+                 "<tr>"
+                   "<td style='padding-right: 10px;'>"
+                     "Ignition Messages:"
+                   "</td>"
+                   "<td>"
+                     "<a href='https://ignitionrobotics.org/libs/msgs' "
+                     "style='text-decoration: none; color: #f58113'>"
+                       "https://ignitionrobotics.org/libs/msgs"
+                     "</a>"
```
chapulina, line 73: Gazebo classic listed SDFormat and Msgs because those were the only libraries outside Gazebo. But with Ignition, we have many more libraries. So it feels weird to list only SDF and messages. How about removing these 2 lines and leaving only the "Documentation" and "Tutorials" sections?

src/gui/AboutDialogHandler.cc
```diff
@@ -0,0 +1,89 @@
...
+AboutDialogHandler::AboutDialogHandler()
+{
+  aboutText += std::string(IGNITION_GAZEBO_VERSION_HEADER);
+  aboutText += "<table class='nostyle'>"
+                 "<tr>"
+                   "<td style='padding-right: 10px;'>Documentation:"
+                   "</td>"
+                   "<td>"
+                     "<a href='https://ignitionrobotics.org/libs/gazebo' "
+                     "style='text-decoration: none; color: #f58113'>"
```
chapulina, line 38: I don't know if there's a way, but it would be cool if these links used the `Material.accent` color. Don't worry if that's not straightforward, I think it's also ok to leave it as is.

chapulina (commented):
Thank you for the PR!

src/gui/resources/GazeboDrawer.qml
atharva-18: I believe that it is an issue related to QtQuick. The links aren't styled as expected and weren't opening in the default browser. I have made a workaround to this issue by adding the `linkActivated` signal.

src/gui/resources/GazeboDrawer.qml
atharva-18: `width: parent.width / 3 > 500 ? 500 : parent.width / 3`

Something like this?

chapulina: Yes, that looks good to me, thanks!

src/gui/resources/GazeboDrawer.qml
chapulina: Got it! I realized that the link wasn't working for me because I was inside a Docker container :upside_down_face: I tried this now on the host machine and the links work correctly! Thanks!

chapulina (approved):
Thank you for the contribution, the dialog looks great and this also fixes the menu, which was broken! :+1:

PR #615, opened 2021-02-05
Particle system - Part4

caguero:
This pull request adds the ability to change the properties of an existing particle emitter at runtime.

See `examples/worlds/particle_emitter.sdf` for testing.

particles_mod


include/ignition/gazebo/rendering/SceneManager.hh
```diff
@@ -149,13 +149,20 @@ inline namespace IGNITION_GAZEBO_VERSION_NAMESPACE {
     public: rendering::LightPtr CreateLight(Entity _id,
         const sdf::Light &_light, Entity _parentId);
 
-    /// \brief Create a particle emitter
+    /// \brief Create a default particle emitter
```
adlarkin, line 152:
```suggestion
    /// \brief Create a default particle emitter.
    /// UpdateParticleEmitter should be called after this method in order to set the particle emitter's parameters.
```

include/ignition/gazebo/rendering/SceneManager.hh
```diff
@@ -149,13 +149,20 @@ inline namespace IGNITION_GAZEBO_VERSION_NAMESPACE {
...
-        const msgs::ParticleEmitter &_emitter, Entity _parentId);
+    /// \return Default particle emitter object created
+    public: rendering::ParticleEmitterPtr CreateDefaultParticleEmitter(
+        Entity _id, const msgs::ParticleEmitter &_emitter, Entity _parentId);
+
+    /// \brief Update an existing particle emitter
+    /// \brief _id Emitter id to update
+    /// \brief _emitter Data to update the particle emitter
+    /// \return Particle emitter updated
+    public: rendering::ParticleEmitterPtr UpdateParticleEmitter(Entity _id,
```
adlarkin, line 164: Minor comment, but I wonder if re-naming this funtion to `ConfigureParticleEmitter` would be clearer.

src/rendering/RenderUtil.cc
```diff
@@ -963,6 +981,31 @@ void RenderUtilPrivate::CreateRenderingEntities(
...
+          // Find the particle emitter requested to be updated.
+          _ecm.Each<components::ParticleEmitter>(
+              [this, &_entityCmd, &_emitterCmd](const Entity &_entity,
+                  const components::ParticleEmitter *_emitter) -> bool
+              {
+                if (_emitterCmd->Data().name() == _emitter->Data().name())
+                {
+                  this->newParticleEmittersCmd.push_back(
+                    std::make_tuple(_entity, _emitterCmd->Data(), _entityCmd));
+                }
```
adlarkin, line 998:
```suggestion
                if (_emitterCmd->Data().name() == _emitter->Data().name())
                {
                  this->newParticleEmittersCmd.push_back(
                    std::make_tuple(_entity, _emitterCmd->Data(), _entityCmd));
                  return false;
                }
```
Once we have found the emitter that needs to be updated, we shouldn't need to continue searching for other emitters that match this emmiter command. We should just get the next emitter command message (if there are any more) and then run the search again.

src/rendering/RenderUtil.cc
```diff
@@ -963,6 +981,31 @@ void RenderUtilPrivate::CreateRenderingEntities(
...
+                return true;
+              });
+
+          // ToDo: Remove the ParticleEmitterCmd component.
```
adlarkin, line 1002:
```suggestion
          // TODO(anyone) Remove the ParticleEmitterCmd component.
```

src/rendering/RenderUtil.cc
```diff
@@ -963,6 +981,31 @@ void RenderUtilPrivate::CreateRenderingEntities(
...
+          // Do it when https://github.com/ignitionrobotics/ign-gazebo/pull/482
+          // is merged, as it has an UpdateECM() function where we can do it.
```
adlarkin, line 1004: It looks like something similar is being done in https://github.com/ignitionrobotics/ign-gazebo/pull/614. We should probably make sure there are no conflicts between #482 and #614

src/rendering/RenderUtil.cc
```diff
@@ -162,6 +163,12 @@ class ignition::gazebo::RenderUtilPrivate
   public: std::vector<std::tuple<Entity, msgs::ParticleEmitter, Entity>>
       newParticleEmitters;
 
+  /// \brief New particle emitter commands to be requested.
+  /// The elements in the tuple are: [0] entity id of the particle emitter to
+  /// update, [1] particle emitter, [2] entity id of the particle emitter cmd.
+  public: std::vector<std::tuple<Entity, msgs::ParticleEmitter, Entity>>
+      newParticleEmittersCmd;
```
adlarkin, line 170: Higher-level design question: would it be possible to make `particleEmitterCmd` a component of the particle emitter entity it is referring to, instead of having it as a separate entity? It seems to me like the command(s) for updating an emitter should be a component of the emitter to be updated since the commands are associated with the original emitter entity.

src/rendering/SceneManager.cc
```diff
@@ -1343,6 +1363,16 @@ void SceneManager::RemoveEntity(Entity _id)
     }
   }
 
+  {
+    auto it = this->dataPtr->particleEmitters.find(_id);
+    if (it != this->dataPtr->particleEmitters.end())
+    {
+      // ToDo: Destroy the particle emitter.
```
adlarkin, line 1370:
```suggestion
      // TODO(anyone) Destroy the particle emitter.
```

src/systems/particle_emitter/ParticleEmitter.cc
```diff
@@ -40,10 +44,36 @@ using namespace systems;
 // Private data class.
 class ignition::gazebo::systems::ParticleEmitterPrivate
 {
-  /// \brief The particle emitter.
+  /// \brief Callback for receoving particle emitter commands.
```
adlarkin, line 47:
```suggestion
  /// \brief Callback for receiving particle emitter commands.
```

src/systems/particle_emitter/ParticleEmitter.cc
```diff
@@ -40,10 +44,36 @@ using namespace systems;
...
+  public: void OnCmd(const ignition::msgs::ParticleEmitter &_msg);
+
+  /// \brief The particle emitter parsed from SDF.
   public: ignition::msgs::ParticleEmitter emitter;
+
+  /// \brief The transport node.
+  public: ignition::transport::Node node;
+
+  /// \brief Model interface.
+  public: Model model{kNullEntity};
```
adlarkin, line 58: It looks to me like this is only being used in the `Configure` method, so can we make this a local variable in that method instead?

src/systems/particle_emitter/ParticleEmitter.cc
```diff
@@ -186,13 +224,52 @@ void ParticleEmitter::Configure(const Entity &_entity,
...
+
+  // Advertise the topic to receive particle emitter commands.
+  const std::string kDefaultTopic =
+    "/model/" + this->dataPtr->model.Name(_ecm) + "/particle_emitter/" + name;
+  std::string topic = _sdf->Get<std::string>("topic", kDefaultTopic).first;
+  if (!this->dataPtr->node.Subscribe(
+         topic, &ParticleEmitterPrivate::OnCmd, this->dataPtr.get()))
+  {
+    ignerr << "Error subscribing to topic [" << topic << "]" << std::endl;
+    return;
```
adlarkin, line 236:
```suggestion
    ignerr << "Error subscribing to topic [" << topic << "]. "
      << "Particle emitter will not receive updates." << std::endl;
    return;
```

src/systems/particle_emitter/ParticleEmitter.cc
```diff
@@ -186,13 +224,52 @@ void ParticleEmitter::Configure(const Entity &_entity,
...
+  auto entity = _ecm.CreateEntity();
+  if (entity == kNullEntity)
+  {
+    ignerr << "Failed to create a particle emitter entity command" << std::endl;
+    return;
+  }
+
+  _ecm.CreateComponent(
+      entity,
+      components::ParticleEmitterCmd({this->dataPtr->userCmd}));
```
adlarkin, line 270: As I mentioned in my other comment, I think it would be better to make the particle emitter command a component of the original emitter entity defined in `Configure` instead of creating a separate entity just for the command. I'm not sure what difficulties this change would introduce, though. I'm assuming that there may be difficulty deleting the component once the update has been applied and re-defining the component if new commands are given.

test/integration/components.cc
```diff
@@ -1540,3 +1541,23 @@ TEST_F(ComponentsTest, ParticleEmitter)
...
+  auto comp1 = components::ParticleEmitterCmd(emitter1);
+
+  // Stream operators.
+  std::ostringstream ostr;
+  comp1.Serialize(ostr);
+
+  std::istringstream istr(ostr.str());
+  components::ParticleEmitter comp3;
+  comp3.Deserialize(istr);
+  EXPECT_EQ(comp1.Data().emitting(), comp3.Data().emitting());
```
adlarkin, line 1562: Can we add an `EXPECT_EQ` check for the name as well?

src/systems/particle_emitter/ParticleEmitter.hh
```diff
@@ -96,7 +96,10 @@ namespace systems
   ///                        therefore only the horizontal dimension of the
   ///                        image is used.
   ///                        Note that this function overrides the particle
-  /// colors set with <color_start> and <color_end>.
+  ///                        colors set with <color_start> and <color_end>.
+  /// `<topic>`: Topic used to update particle emitter properties at runtime.
+  ///            The default topic is
+  ///            /model/<model_name>/particle_emitter/<emitter_name>
```
adlarkin, line 102: Should we add some documentation about the `particleEmitterCmd` option for updating the particle emitter at runtime? In particular, I think we should document at least the following:
1. How to assign a `particleEmitterCmd` to a particular particle emitter (I'm assuming the `name` field of the command should match the `name` field used when creating the original emitter)
2. Which properties of the emitter can be changed at runtime (in other words, which fields can be set in `particleEmitterCmd` that would modify the emitter associated with the command)

include/ignition/gazebo/components/ParticleEmitterCmd.hh
```diff
@@ -0,0 +1,45 @@
+/*
```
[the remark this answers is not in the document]
iche033: merged into `ParticleEmitter.hh`. 468b02ad76b3df9e24b862be4379d0ce25be5a3f

src/rendering/RenderUtil.cc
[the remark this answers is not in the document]
iche033: 468b02ad76b3df9e24b862be4379d0ce25be5a3f

include/ignition/gazebo/rendering/SceneManager.hh
iche033: 468b02ad76b3df9e24b862be4379d0ce25be5a3f

src/rendering/RenderUtil.cc
iche033: 468b02ad76b3df9e24b862be4379d0ce25be5a3f

include/ignition/gazebo/rendering/SceneManager.hh
iche033: I think the convention for using Configure in ign-gazebo is generally when some logic is executed called only once on startup but here it's updating the params after launch and can be done multiple times so it's using Update instead.

src/systems/particle_emitter/ParticleEmitter.cc
iche033: removed. 468b02a

src/systems/particle_emitter/ParticleEmitter.cc
iche033: done. a07c2ba39162191bacfe2fb74f1e1236124c0dac

I ran into an issue with removing the ParticleEmitterCmd component from the entity, which still needs to be resolved

src/systems/particle_emitter/ParticleEmitter.cc
iche033: 468b02ad76b3df9e24b862be4379d0ce25be5a3f

src/rendering/RenderUtil.cc
iche033: a07c2ba

src/rendering/RenderUtil.cc
iche033: a07c2ba

src/rendering/RenderUtil.cc
iche033: a07c2ba

src/rendering/SceneManager.cc
iche033: 468b02ad76b3df9e24b862be4379d0ce25be5a3f

test/integration/components.cc
iche033: 468b02ad76b3df9e24b862be4379d0ce25be5a3f

src/systems/particle_emitter/ParticleEmitter.cc
iche033: 468b02ad76b3df9e24b862be4379d0ce25be5a3f

test/integration/particle_emitter.cc
```diff
@@ -112,7 +112,12 @@ TEST_F(ParticleEmitterTest, SDFLoad)
                 EXPECT_EQ(math::Color::Green,
                     msgs::Convert(_emitter->Data().color_end()));
                 EXPECT_DOUBLE_EQ(10.0, _emitter->Data().scale_rate());
-                EXPECT_EQ("/path/to/dummy_image.png",
+
+                // color range image is empty because the emitter system
+                // will not be able to find a file that does not exist
+                // \todo(anyone) this should return  "/path/to/dummy_image.png"
+                // and let rendering does the findFile instead
```
adlarkin, line 119:
```suggestion
                // TODO(anyone) this should return  "/path/to/dummy_image.png"
                // and let rendering do the findFile instead
```

src/systems/particle_emitter/ParticleEmitter.hh
```diff
@@ -76,25 +76,15 @@ namespace systems
...
-  ///                   * red
-  ///                   * green
-  ///                   * blue
-  ///                   * yellow
-  ///                   * magenta
-  ///                   * cyan
-  ///                   * white
-  ///                  Default color is white.
+  ///                  Color::White is the default color for the particles
+  ///                  unless a specific function is used.
```
adlarkin, line 80: Can we give example input so that users know what `specific function` means? For example, I am assuming that input for red would be `<color_start>1 0 0</color_start>`.

We should also add this information in to the `color_end` documentation below.

src/systems/particle_emitter/ParticleEmitter.cc
adlarkin: > I ran into an issue with removing the ParticleEmitterCmd component from the entity, which still needs to be resolved

Interesting, so the component is never removed? Maybe the new msgCmd component wasn't added to the entity yet, and we are trying to remove the component too soon? For example, would it make more sens to remove components after updating the emitters with the msgCmd in `RenderUtil::Update()`?:
https://github.com/ignitionrobotics/ign-gazebo/blob/a07c2ba39162191bacfe2fb74f1e1236124c0dac/src/rendering/RenderUtil.cc#L573-L577

src/systems/particle_emitter/ParticleEmitter.hh
adlarkin: https://github.com/ignitionrobotics/ign-gazebo/pull/615/commits/4f64296d87f3f3b37bf22b763a0d61dde0b78355

test/integration/particle_emitter.cc
adlarkin: https://github.com/ignitionrobotics/ign-gazebo/pull/615/commits/4f64296d87f3f3b37bf22b763a0d61dde0b78355

src/systems/particle_emitter/ParticleEmitter.cc
```diff
@@ -280,62 +286,32 @@ void ParticleEmitter::PreUpdate(const ignition::gazebo::UpdateInfo &_info,
...
   if (!this->dataPtr->newDataReceived)
     return;
 
   // Nothing left to do if paused.
   if (_info.paused)
     return;
 
-  // Create component.
-  std::lock_guard<std::mutex> lock(this->dataPtr->mutex);
+   this->dataPtr->newDataReceived = false;
```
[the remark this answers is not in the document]
adlarkin: https://github.com/ignitionrobotics/ign-gazebo/pull/615/commits/4f64296d87f3f3b37bf22b763a0d61dde0b78355

src/systems/particle_emitter/ParticleEmitter.cc
```diff
@@ -220,12 +221,16 @@ void ParticleEmitter::Configure(const Entity &_entity,
     _sdf->Get<double>("max_velocity", 1).first);
 
   // Color start.
-  ignition::msgs::Set(this->dataPtr->emitter.mutable_color_start(),
-      this->dataPtr->GetColor(_sdf->Get<std::string>("color_start")));
+  ignition::math::Color color = ignition::math::Color::White;
+  if (_sdf->HasElement("color_start"))
+    color = _sdf->Get<ignition::math::Color>("color_start");
```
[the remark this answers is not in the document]
adlarkin: I was able to test and verify that the proper way to specify the color values is with RGB as I mentioned above. https://github.com/ignitionrobotics/ign-gazebo/pull/615/commits/4f64296d87f3f3b37bf22b763a0d61dde0b78355

include/ignition/gazebo/components/ParticleEmitterCmd.hh
```diff
@@ -31,12 +31,6 @@ namespace gazebo
 inline namespace IGNITION_GAZEBO_VERSION_NAMESPACE {
 namespace components
 {
-  /// \brief A component that contains a particle emitter command.
-  using ParticleEmitterCmd = Component<msgs::ParticleEmitter,
-        class ParticleEmitterCmdTag,
-        serializers::MsgSerializer>;
-  IGN_GAZEBO_REGISTER_COMPONENT("ign_gazebo_components.ParticleEmitterCmd",
-      ParticleEmitterCmd)
```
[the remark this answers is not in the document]
adlarkin: https://github.com/ignitionrobotics/ign-gazebo/pull/615/commits/4f64296d87f3f3b37bf22b763a0d61dde0b78355

src/rendering/RenderUtil.cc
```diff
@@ -172,9 +171,9 @@ class ignition::gazebo::RenderUtilPrivate
 
   /// \brief New particle emitter commands to be requested.
   /// The elements in the tuple are: [0] entity id of the particle emitter to
-  /// update, [1] particle emitter, [2] entity id of the particle emitter cmd.
-  public: std::vector<std::tuple<Entity, msgs::ParticleEmitter, Entity>>
-      newParticleEmittersCmd;
+  /// update, [1] particle emitter
+  public: std::vector<std::tuple<Entity, msgs::ParticleEmitter>>
+      newParticleEmittersCmds;
```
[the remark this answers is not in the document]
iche033: done. f612ec8

src/systems/particle_emitter/ParticleEmitter.hh
iche033: added some brief doc and referred the user to look at the example world file. f612ec8

src/gui/plugins/scene3d/Scene3D.cc
```diff
@@ -2752,6 +2752,7 @@ void Scene3D::Update(const UpdateInfo &_info,
     msgs::Pose poseMsg = msgs::Convert(renderWindow->CameraPose());
     this->dataPtr->cameraPosePub.Publish(poseMsg);
   }
+  this->dataPtr->renderUtil->UpdateECM(_info, _ecm);
   this->dataPtr->renderUtil->UpdateFromECM(_info, _ecm);
```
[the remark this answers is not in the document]
iche033: thermal camera is server side only so the call in Scene3D (gui side) should not affect thermal cameras.

src/systems/particle_emitter/ParticleEmitter.cc
iche033: I think it's an issue with updating components on the gui side. Adding a OneTimeChange in the particle system when setting component data seem to made sure the new values are propagated to the gui. Updating the emitter properties should work now after f8a971d

adlarkin (approved):

adlarkin:
Everything seems good. Quick question before merging: is the [one time change fix for updating components](https://github.com/ignitionrobotics/ign-gazebo/pull/615#discussion_r578020805) a temporary solution, or is it a permanent fix?

iche033:
> Quick question before merging: is the one time change fix for updating components a temporary solution, or is it a permanent fix?

I'm not exactly sure if it's the right fix so I added a note in 4142eea

PR #617, opened 2021-02-06
Upgrade ign-sensors and support custom sensors

chapulina:
Part of https://github.com/ignitionrobotics/ign-sensors/issues/18

## Summary

Requires https://github.com/ignitionrobotics/ign-plugin/pull/35 and https://github.com/ignitionrobotics/ign-sensors/pull/90.

This is an alternative to #304.

tl;dr: Loading symbols globally is a way of making we find them after they've been registered by another library.

---

TODO: write up a summary of what I think is happening.

## Test it

* Use all the PRs listed above
* Run worlds that contain sensors, like `ign gazebo sensors.sdf` or `ign gazebo sensors_demo.sdf`
* Sensor data should be available through topics, camera images should be visible on widgets

## Checklist
- [x] Added tests
- [x] Added example world and/or tutorial
- [x] While waiting for a review on your PR, please help review
[another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Aignitionrobotics+archived%3Afalse+)
to support the maintainers


src/SystemLoader.cc
```diff
@@ -126,6 +128,7 @@ class ignition::gazebo::SystemLoaderPrivate
 SystemLoader::SystemLoader()
   : dataPtr(new SystemLoaderPrivate())
 {
+  this->dataPtr->loader.SetFlags(RTLD_LAZY | RTLD_GLOBAL);
```
ahcorde, line 131: This line here (I think) is going to generate some conflcts. We are openning all system libraries with this flag which will generate the same problem than my PRs.

We should change the flag inside the system plugin code. In particular `sensor` which is the system plugin that load the rendering sensors (camera, rgbd, etc).

ahcorde (changes requested):
I made a comment about the place to change the flag. I tested locally and all system plugins are failing.

Don't merge the other related PR yet until we are sure we can use this solution.

src/SystemLoader.cc
chapulina: How are the system plugins failing for you? I tried running `sensors_demo` and `rolling_shapes` and they seem to be working for me. Be sure to recompile all libraries from plugin to gazebo.

chapulina:
Removing beta label, we won't have time to wrap this up before code freeze. Let's retarget at Ignition-F.

src/systems/physics/Physics.cc
```diff
@@ -663,10 +663,13 @@ void Physics::Configure(const Entity &_entity,
     return;
   }
 
-  auto classNames = pluginLoader.AllPlugins();
+  auto classNames = pluginLoader.PluginsImplementing<
```
azeey, line 666: Can you comment on why this is necessary?

src/SystemLoader.cc
chapulina: Removed

src/systems/physics/Physics.cc
chapulina: When we were using `RTLD_GLOBAL`, we could find other plugins that have been loaded in the same process before. We're not using that anymore, so this is not required, but I think it's a nice extra filter just in case?

azeey:
With this scheme, I'm not sure how it would be possible to have two different custom sensors. Maybe we can add a custom attribute to differentiate them. e.g.: `<sensor name='my_sensor' type='custom' ignition:sensor_type='custom_imu'>`.

src/systems/physics/Physics.cc
azeey: I see. Thanks!

chapulina:
> With this scheme, I'm not sure how it would be possible to have two different custom sensors. Maybe we can add a custom attribute to differentiate them. e.g.: <sensor name='my_sensor' type='custom' ignition:sensor_type='custom_imu'>.

Yup good point. I was planning to add an example with a 2nd sensor. Is there an advantage to using an attribute instead of an element? 

My biggest concern was the `CustomSensor` component actually. Every system using custom sensors will need to iterate over all custom sensors and dig out the one they want. Maybe `ignition:sensor_type` can be in its own component?

azeey:
> Is there an advantage to using an attribute instead of an element? 

I think it would be more readable since it's right next to the regular `type` attribute where one expects to see the type of sensor.

> My biggest concern was the CustomSensor component actually. Every system using custom sensors will need to iterate over all custom sensors and dig out the one they want. Maybe ignition:sensor_type can be in its own component?

Would making it its own component help? Wouldn't you still have to iterate through all the custom sensors sense they will all have the `ignition:sensor_type` component? I think if we want to avoid systems having to iterate over custom sensors, we would need to provide some sort of registration mechanism.

chapulina:
> Would making it its own component help? 

Yeah good point, I don't think it solves the issue. 

> I think it would be more readable since it's right next to the regular type attribute where one expects to see the type of sensor.

Fair enough. I think it ends up being redundant in case the user also adds a custom element, but I agree that it's more readable.

> we would need to provide some sort of registration mechanism.

I'd like to keep this simple since custom sensors will be an advanced feature that most users don't need to touch. Creating a registration mechanism just for that feels a bit overkill :confused:

azeey:
> I'd like to keep this simple since custom sensors will be an advanced feature that most users don't need to touch. Creating a registration mechanism just for that feels a bit overkill

Agreed.

examples/plugin/custom_sensor_system/OdometerSystem.cc
```diff
@@ -0,0 +1,137 @@
...
+        // Get sensor's scoped name without the world
+        auto sensorScopedName = ignition::gazebo::removeParentScope(
+            ignition::gazebo::scopedName(_entity, _ecm, "::", false), "::");
+        sdf::Sensor data = _custom->Data();
+        data.SetName(sensorScopedName);
+
+        // Default to scoped name as topic
+        if (data.Topic().empty())
+        {
+          std::string topic = scopedName(_entity, _ecm) + "/odometer";
```
ahcorde, line 63: should we check here ig the topic is valid ?

ahcorde (changes requested):

examples/plugin/custom_sensor_system/OdometerSystem.cc
chapulina: `ign-sensors` does that for us internally through the `SetTopic` function, so I think it's not necessary to do it here too:

https://github.com/ignitionrobotics/ign-sensors/blob/ba89178b6f4c75934ba608ab0ceb5348749b51f9/src/Sensor.cc#L218-L235

I do think that not all users would know that though

src/systems/physics/Physics.cc
chapulina: I separated this into another PR just in case: https://github.com/ignitionrobotics/ign-gazebo/pull/965

examples/plugin/custom_sensor_system/OdometerSystem.cc
```diff
@@ -0,0 +1,137 @@
...
+
+//////////////////////////////////////////////////
+void OdometerSystem::RemoveSensorEntities(
+    const ignition::gazebo::EntityComponentManager &_ecm)
+{
+  _ecm.EachRemoved<ignition::gazebo::components::CustomSensor>(
+    [&](const ignition::gazebo::Entity &_entity,
+        const ignition::gazebo::components::CustomSensor *)->bool
+      {
+        auto sensorId = this->entitySensorMap.find(_entity);
```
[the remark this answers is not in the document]
chapulina: Sweet :lollipop:  bd0153d

ahcorde (approved):

CMakeLists.txt
```diff
@@ -99,16 +99,23 @@ set(IGN_PHYSICS_VER ${ignition-physics5_VERSION_MAJOR})
 #--------------------------------------
 # Find ignition-sensors
 ign_find_package(ignition-sensors6 REQUIRED
+  # component order is important
```
ahcorde, line 102: I can see this [warning](https://build.osrfoundation.org/job/ignition_gazebo-ci-pr_any-ubuntu_auto-amd64/6782/cmake/) on Ubuntu 

```
CMake 3.11.0 or higher is required. You are running version 3.10.2
```

ahcorde:
The `INTEGRATION_examples_build` is failing on MacOS becuase it's not able to find `ignition/sensors/Util.hh`.

CMakeLists.txt
chapulina: ebe8bdf

chapulina:
The 23 test failures on Jenkins Ubuntu are due to https://github.com/ignition-tooling/release-tools/pull/494 and can be ignored for now. https://github.com/ignitionrobotics/ign-gazebo/pull/971 should fix them if we don't fix it in the infra. The rest of CI looks good. Merging!

PR #584, opened 2021-01-26
Make topics configurable for joint controllers

chapulina:
This should help with https://github.com/ignitionrobotics/ign-gazebo/issues/447

Note that I didn't bother using `AsValidTopic` because that's only useful from Dome.

PR #623, opened 2021-02-10
3 to 4

adlarkin:
Forward port `ign-gazebo3` to `ign-gazebo4` in preparation for an `ign-gazebo4` release.

Comparison between `ign-gazebo3` and `ign-gazebo4`: https://github.com/ignitionrobotics/ign-gazebo/compare/ign-gazebo4...ign-gazebo3

### DIFF
--- a/examples/worlds/joint_position_controller.sdf
+++ b/examples/worlds/joint_position_controller.sdf
@@ -4,9 +4,9 @@
 
   Try sending joint position commands:
 
-    ign topic -t "/model/joint_position_controller_demo/joint/j1/0/cmd_pos" -m ignition.msgs.Double -p "data: -1.0"
+    ign topic -t "/rotor_cmd" -m ignition.msgs.Double -p "data: -1.0"
 
-    ign topic -t "/model/joint_position_controller_demo/joint/j1/0/cmd_pos" -m ignition.msgs.Double -p "data: 1.0"
+    ign topic -t "/rotor_cmd" -m ignition.msgs.Double -p "data: 1.0"
 -->
 <sdf version="1.6">
   <world name="default">
@@ -45,7 +45,7 @@
           <property type="double" key="height">72</property>
           <property type="double" key="width">121</property>
           <property type="double" key="z">1</property>
-          
+
           <property type="string" key="state">floating</property>
           <anchors target="3D View">
             <line own="left" target="left"/>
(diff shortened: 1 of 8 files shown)


adlarkin:
It looks like there's a mismatch between the author and sign-off emails in https://github.com/ignitionrobotics/ign-gazebo/pull/623/commits/87b2f4ba27cc10a1a586bda8d635aa929e1070ce, causing DCO to fail. I don't think I can fix this commit without re-writing history - if this is true, should I just set DCO to pass manually?

nkoenig (approved):

adlarkin:
> should I just set DCO to pass manually?

I set DCO to pass manually. I'm just waiting for checks to finish so that I can merge forward.

chapulina (changes requested):
I think the changes from #584 got lost in the merge forward. When merging that forward, we need to combine the use of `if (_sdf->HasElement("topic"))` with `AsValidTopic`

adlarkin:
> I think the changes from #584 got lost in the merge forward. When merging that forward, we need to combine the use of `if (_sdf->HasElement("topic"))` with `AsValidTopic`

I noticed that there was a check for `if (topic.empty())` after the `AsValidTopic` line, which I kept. Should we keep both the `if (topic.empty())` AND `if (_sdf->HasElement("topic"))` lines? I thought we only needed one, but perhaps I misread/misunderstood the code.

chapulina:
> Should we keep both the if (topic.empty()) AND if (_sdf->HasElement("topic")) lines?

Yup, so the `HasElement` is checking if the user provided a topic through the SDF. And `topic.empty()` is checking if a valid topic could be created.

I just remembered there's a nice helper function to help with this use case:

https://github.com/ignitionrobotics/ign-gazebo/blob/e647570f25f962d63af75cf669ff72731d57bd5e/include/ignition/gazebo/Util.hh#L143-L156

Here's an example use:

https://github.com/ignitionrobotics/ign-gazebo/blob/e647570f25f962d63af75cf669ff72731d57bd5e/src/systems/detachable_joint/DetachableJoint.cc#L98-L106

chapulina (approved):
Thanks for updating the merge to get the topic from SDF. Like I mentioned offline, it would be nice to run the topic that comes from `_sdf->Get<std::string>("topic")` through `AsValidTopic`, possibly using `validTopic`. But I don't want to block this PR on this, it can be addressed later.

adlarkin:
> Thanks for updating the merge to get the topic from SDF. Like I mentioned offline, it would be nice to run the topic that comes from `_sdf->Get<std::string>("topic")` through `AsValidTopic`, possibly using `validTopic`. But I don't want to block this PR on this, it can be addressed later.

I created an issue for this so that we don't forget: https://github.com/ignitionrobotics/ign-gazebo/issues/624

PR #625, opened 2021-02-11
4.4.0

adlarkin:
Prepare for `4.4.0` release.

Comparison to `4.3.0`: https://github.com/ignitionrobotics/ign-gazebo/compare/ignition-gazebo4_4.3.0...ign-gazebo4

---

**Question for reviewers:** (cc @iche033) since #614 doesn't require users to have the recent `ign-gui4.2.0`, `ign-sensors4.1.0`, and `ign-rendering4.4.0` releases (users can keep using 16-bit thermal cameras as they have been prior to this release if they wish), we don't need to bump the dependencies on those libraries, right? Or do we need to bump those dependencies since users can't use the 8-bit option introduced in #614 without `ign-gui4.2.0`, `ign-sensors4.1.0`, and `ign-rendering4.4.0`?

---

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,7 +3,7 @@ cmake_minimum_required(VERSION 3.5.1 FATAL_ERROR)
 #============================================================================
 # Initialize the project
 #============================================================================
-project(ignition-gazebo4 VERSION 4.3.0)
+project(ignition-gazebo4 VERSION 4.4.0)
 
 #============================================================================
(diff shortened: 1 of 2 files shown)


nkoenig:
We don't need to bump versions in this case. Building ign-gazebo will not be broken if the older libraries are used.

nkoenig (approved):

iche033 (approved):

PR #627, opened 2021-02-12
Added link to HW-accelerated video recording

peci1:
Added link to the tutorial created in https://github.com/ignitionrobotics/ign-common/pull/169 (not yet merged in).

I tried to find a way to make the tutorial link always use the default branch of ign-common, but haven't found a way. So currently, the link would have to be updated with each change of ign-common major version (or, assuming the tutorial would not change much, it could even stay pointing to ign-common3).


mjcarroll:
> I tried to find a way to make the tutorial link always use the default branch of ign-common, but haven't found a way. So currently, the link would have to be updated with each change of ign-common major version (or, assuming the tutorial would not change much, it could even stay pointing to ign-common3).

I think it's probably okay to point to the `common` release that corresponds to this `gazebo` release.  For `edifice`, we can have it point to the `common4` variant.  This can be sorted out in the forward-porting process.

tutorials/video_recorder.md
```diff
@@ -105,3 +105,10 @@ going to the world SDF file, locate the
 
 * **bitrate**: Video encoding bitrate in bps. This affects the quality of the
 generated video. The default bitrate is 2Mbps.
+  
+## Hardware-accelerated encoding
+
+Since Ignition Common 3.10.2, there is support for utilizing the power of GPUs
+to speed up the video encoding process. See the
+[Hardware-accelerated Video Encoding tutorial](https://github.com/ignitionrobotics/ign-common/blob/ign-common3/tutorials/install.md)
```
chapulina, line 113: The canonical location for our tutorials is the http://ignitionrobotics.org website, i.e. https://ignitionrobotics.org/api/common/3.10/tutorials.html. The GitHub markdown preview usually looks ok, but it doesn't render all the doxygen features properly (like cross-links, images, etc), so we shouldn't treat that preview as a reference.

We have nightly doc upload jobs, so the tutorial merged today in https://github.com/ignitionrobotics/ign-common/pull/169 should be live tomorrow. I believe its link will be https://ignitionrobotics.org/api/common/3.10/hw-encoding.html (I know the hardcoded minor version is annoying, we should eventually improve that (https://github.com/ignitionrobotics/docs/issues/52).

peci1: Thanks, I fixed that link.

chapulina (approved):

PR #629, opened 2021-02-12
Add a convenience function for getting possibly non-existing components.

peci1:
# New feature

Continues in the line of #436 and #378.

## Summary
Adds a convenience function `EntityComponentManager::ComponentDefault()` which works as `Component()`, but if the component isn't found, it automatically creates it (using a user-supplied default value).

I have an example usage of the function in https://github.com/osrf/subt/pull/551 (which uses a shim until this PR is merged).

I'm not really sure whether `ComponentDefault` is the best name. If you come up with something better, I'll be glad. I was also thinking about adding optional parameters to `Component()` function, but that would actually need two of them - one boolean saying whether the component should be auto-constructed, and then the supplied default value. I felt that wouldn't really be nice to use and it would be less self-descriptive than having a function whose name explicitly states that it can auto-create the component if needed.

## Test it
A unit test was added.

## Checklist
- [x] Added tests
- [ ] Added example world and/or tutorial
- [x] While waiting for a review on your PR, please help review
[another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Aignitionrobotics+archived%3Afalse+)
to support the maintainers


include/ignition/gazebo/EntityComponentManager.hh
```diff
@@ -214,6 +214,19 @@ namespace ignition
...
+      /// component type. If the component doesn't exist, create it and
+      /// initialize with the given default value.
+      /// \param[in] _entity The entity.
+      /// \param[in] _default The value that should be used to construct
+      /// the component in case the component doesn't exist.
+      /// \return The component of the specified type assigned to the specified
+      /// entity.
+      public: template<typename ComponentTypeT>
+              ComponentTypeT *ComponentDefault(Entity _entity,
+              const typename ComponentTypeT::Type& _default =
```
chapulina, line 227: Style nit

```suggestion
              const typename ComponentTypeT::Type &_default =
```

include/ignition/gazebo/detail/EntityComponentManager.hh
```diff
@@ -127,6 +127,20 @@ ComponentTypeT *EntityComponentManager::Component(const ComponentKey &_key)
       this->ComponentImplementation(_key));
 }
 
+//////////////////////////////////////////////////
+template<typename ComponentTypeT>
+ComponentTypeT *EntityComponentManager::ComponentDefault(Entity _entity,
+    const typename ComponentTypeT::Type& _default)
```
chapulina, line 133:
```suggestion
    const typename ComponentTypeT::Type &_default)
```

src/EntityComponentManager_TEST.cc
```diff
@@ -437,6 +437,31 @@ TEST_P(EntityComponentManagerFixture, EntitiesAndComponents)
   EXPECT_FALSE(manager.EntityHasComponentType(entity, DoubleComponent::typeId));
   EXPECT_FALSE(manager.EntityHasComponentType(entity2, IntComponent::typeId));
 
+  // Query non-existing component, the default value is default-constructed
+  BoolComponent* boolComp = manager.ComponentDefault<BoolComponent>(entity);
```
chapulina, line 441: Style nit, also applies to others below.

```suggestion
  BoolComponent *boolComp = manager.ComponentDefault<BoolComponent>(entity);
```

chapulina (commented):
Thanks for the PR. This is an interesting use case that didn't occur to me before, but thinking of it now it makes sense. It's interesting how similar it is to `SetComponentData` but has an important difference.

It initially reminded me of [rclcpp::Node::get_parameter_or](https://github.com/ros2/rclcpp/blob/18b112fcce90ef41fb2a312dda59508f2164dc5e/rclcpp/include/rclcpp/node.hpp#L593-L613): "_Get the parameter value, or the "alternative_value" if not set, and assign it to "parameter"._" (even though that does not "create" (i.e. declare) the parameter if not existent). 

And then I remembered [sdf::Element::Get](https://github.com/osrf/sdformat/blob/83fe4246b186ab05cac6a596d06c384a4bcae366/include/sdf/Element.hh#L263-L271), which has pretty much the same behaviour as the function here. So my naming suggestion would be to make this an overload of `Component` that accepts a default value.

peci1:
Thanks for your comment, Louise. If I understood you correctly, adding an overload of `Component()` with the default parameter would mean that at call time, the user would always need to provide "something" for the default value. I first thought the user would need to type in either a value or `Type()` for default-constructed value, which would make the API a little more complicated to use (always needing to search which particular type does the component use for storage), but then I found out you can provide just `{}` as the default value and it works :) 

I think it might be a bit unintuitive for others, because I myself came to know initializers only recently, but it would allow for calling in the fashion of:

    ecm.Component<component::Double>(entity, {});

It might be a good compromise between ease of use and compactness.

If there are no other ideas, I'll have a look at it over the weekend...

chapulina:
> the user would always need to provide "something" for the default value. 

Good point, I can see how that could make the API a bit less intuitive. I'm not opposed to `ComponentDefault` though, if there are no other ideas and you think that's clearer, I'm ok with that.

peci1:
Okay, this was actually two weekends, but as no new ideas came, I vouch for the more intuitive API with `ComponentDefault`.

I applied the style nits.

chapulina (approved):
Thanks, looks good to me :+1:

diegoferigo:
xref https://github.com/ignitionrobotics/ign-gazebo/issues/325

PR #656, opened 2021-02-26
Cache top level and static to speed up physics system

chapulina:
# 🦟 Bug fix

## Summary

Profiling a world as follows:

* 3000 non-static simple shapes sitting on a static ground plane
* running with unbounded RTF (`-z 1000000000`)
* running headless (with `-s` and only the `Physics` system)
* Using TPE as the physics engine

I noticed that the `PhysicsPrivate::UpdateSim` was taking way too long for what it does. All that function is doing is getting resulting data from the `ign-physics` step and populating the ECM, and that takes ~70% of an iteration:

Before:

image

I think there's a lot of room for improvement there, but here are just a couple of little improvements. Caching a couple of values that never change (top-level model and static), instead of reevaluating them at every iteration, we can reduce that to closer to 60%:

After:

topmodel_static_cached

For static, one thing to consider is leveraging the ECM and adding `Static` components to links that are static, so that the `Each` call can iterate only over static links.

## Checklist
- [ ] Added tests
- [] `codecheck` passed (See [contributing](https://ignitionrobotics.org/docs/all/contributing#contributing-code))

### DIFF
--- a/src/systems/physics/Physics.cc
+++ b/src/systems/physics/Physics.cc
@@ -196,6 +196,13 @@ class ignition::gazebo::systems::PhysicsPrivate
   /// associated with a physics Link
   public: std::unordered_map<LinkPtrType, Entity> linkEntityMap;
 
+  /// \brief Cache the top-level model for each entity.
+  /// The key is an entity and the value is its top level model.
+  public: std::unordered_map<Entity, Entity> topLevelModelMap;
+
+  /// \brief Keep track of what entities are static (models and links).
+  public: std::unordered_set<Entity> staticEntities;
+
   /// \brief A map between model entity ids in the ECM to whether its battery
   /// has drained.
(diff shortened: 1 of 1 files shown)


nkoenig:
This worked great on SubT. Performance went from 65-70% to around 95% when running a single robot in a world.

nkoenig (approved):

adlarkin (commented):
Changes seem okay to me. It seems like there wasn't much of an improvement if I ran this with the 3000 shapes world with static models instead of non-static, which was a little surprising to me. But I think that this is a step in the right direction, and we will probably need to continue making changes like this.

Also, I didn't see any issues in the code, but I'm not too familiar with the physics system, so if someone else wants to take a look at the changes as a sanity check, that might be a good idea (I think @nkoenig may have already done this).

PR #662, opened 2021-03-03
Check empty world name in Scene3d

iche033:

# 🦟 Bug fix

## Summary

When loading a world with large complex meshes, I get a spam of the following msg:

```
[GUI] [Dbg] [Scene3D.cc:2745] RenderEngineGuiPlugin component not found, render engine won't be set from the ECM
```

This happens because the server takes a while to load while the gui is already up and running and it keeps checking for the RenderEngineGuiPlugin component on every update (and stops checking only when the it gets the world entity form the sever). In simple worlds, this does not happen because the server is usually loaded quite fast and have the world entity ready for gui on initalization.

This PR adds a check on the Scene3d plugin side so that it does not attempt to find the world's `RenderEngineGuiPlugin` component if the world entity is not ready yet.

# To test

run ign gazebo with a complex world, e.g.

```
ign gazebo -v tunnel.sdf
```

You should see a few `RenderEngineGuiPlugin component not found` msgs printed in the console.

## Checklist
- [ ] Added tests

### DIFF
--- a/src/gui/plugins/scene3d/Scene3D.cc
+++ b/src/gui/plugins/scene3d/Scene3D.cc
@@ -2721,21 +2721,20 @@ void Scene3D::Update(const UpdateInfo &_info,
   if (this->dataPtr->worldName.empty())
   {
     // TODO(anyone) Only one scene is supported for now
+    Entity worldEntity;
     _ecm.Each<components::World, components::Name>(
-        [&](const Entity &/*_entity*/,
+        [&](const Entity &_entity,
           const components::World * /* _world */ ,
           const components::Name *_name)->bool
         {
           this->dataPtr->worldName = _name->Data();
+          worldEntity = _entity;
           return true;
         });
 
     if (!this->dataPtr->worldName.empty())
     {
       renderWindow->SetWorldName(this->dataPtr->worldName);
-      auto worldEntity =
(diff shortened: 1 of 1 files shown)


ahcorde (approved):

src/gui/plugins/scene3d/Scene3D.cc
```diff
@@ -2730,20 +2730,23 @@ void Scene3D::Update(const UpdateInfo &_info,
...
-      this->dataPtr->renderUtil->SetEngineName(renderEngineGuiComp->Data());
-    }
-    else
-    {
-      igndbg << "RenderEngineGuiPlugin component not found, "
-        "render engine won't be set from the ECM" << std::endl;
+      renderWindow->SetWorldName(this->dataPtr->worldName);
+      auto worldEntity =
+        _ecm.EntityByComponents(components::Name(this->dataPtr->worldName),
+          components::World());
```
chapulina, line 2738: This is not introduced by this PR, but the `worldEntity` could be got from the `Each` call above on line 2724.

chapulina (approved):

src/gui/plugins/scene3d/Scene3D.cc
iche033: updated. iche033/scene3d_world_name

PR #669, opened 2021-03-04
Cache link poses to improve performance

adlarkin:

# 🦟 Bug fix

Related to https://github.com/ignitionrobotics/ign-physics/issues/219.

## Summary
This PR is motivated by #656 and https://github.com/ignitionrobotics/ign-physics/issues/219.

https://github.com/ignitionrobotics/ign-physics/issues/219 points out that the `UpdateSim` method in the physics system [goes over all links](https://github.com/ignitionrobotics/ign-gazebo/blob/86e6ad93581362eba4e89a58dc1fef058806a081/src/systems/physics/Physics.cc#L1702) since the physics system doesn't know which links changed. For non-static links, we are [always performing pose updates](https://github.com/ignitionrobotics/ign-gazebo/blob/86e6ad93581362eba4e89a58dc1fef058806a081/src/systems/physics/Physics.cc#L1726-L1774), even if the pose hasn't changed. I'm proposing that we cache each non-static link's pose from the previous iteration and only perform a pose update if the link's pose is different from the previous iteration.

Testing this change by profiling a world as done in #656 (3000 non-static simple shapes, unbounded RTF, headless, and TPE for the physics engine) seems to result in a noticeable performance improvement.

Before link caching, `UpdateSim` takes around `8.25ms`. `Local pose`, which contains the pose updates, takes around `5.2ms`, which is more than half of the `UpdateSim` time:

before_link_caching

After link caching, `UpdateSim` takes around `5.7ms`, and `Local pose` now takes around `2.4ms`:

after_link_caching

Taking a look at the some of the other code in `UpdateSim`, I am wondering if we can do something similar for non-static link velocities and accelerations. @chapulina what do you think about this?

## Checklist
- [x] Added tests

### DIFF
--- a/src/systems/physics/Physics.cc
+++ b/src/systems/physics/Physics.cc
@@ -175,10 +175,10 @@ class ignition::gazebo::systems::PhysicsPrivate
   public: void RemovePhysicsEntities(const EntityComponentManager &_ecm);
 
   /// \brief Update physics from components
-  /// \param[in] _ecm Constant reference to ECM.
+  /// \param[in] _ecm Mutable reference to ECM.
   public: void UpdatePhysics(EntityComponentManager &_ecm);
 
-  /// \brief Step the simulationrfor each world
+  /// \brief Step the simulation for each world
   /// \param[in] _dt Duration
   public: void Step(const std::chrono::steady_clock::duration &_dt);
 
@@ -211,6 +211,11 @@ class ignition::gazebo::systems::PhysicsPrivate
 
   /// \brief Keep track of what entities are static (models and links).
   public: std::unordered_set<Entity> staticEntities;
+
+  /// \brief Keep track of poses for links attached to non-static models.
+  /// This allows for skipping pose updates if a link's pose didn't change
+  /// after a physics step.
+  public: std::unordered_map<Entity, ignition::math::Pose3d> linkWorldPoses;
 
   /// \brief A map between model entity ids in the ECM to whether its battery
   /// has drained.
@@ -1142,6 +1147,7 @@ void PhysicsPrivate::RemovePhysicsEntities(const EntityComponentManager &_ecm)
             this->entityLinkMap.Remove(childLink);
             this->topLevelModelMap.erase(childLink);
             this->staticEntities.erase(childLink);
+            this->linkWorldPoses.erase(childLink);
           }
 
           for (const auto &childJoint :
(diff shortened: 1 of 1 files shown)


src/systems/physics/Physics.cc
```diff
@@ -212,6 +213,12 @@ class ignition::gazebo::systems::PhysicsPrivate
   /// \brief Keep track of what entities are static (models and links).
   public: std::unordered_set<Entity> staticEntities;
 
+  /// \brief Keep track of non-static link poses.
+  /// This allows for skipping pose updates if a link's pose didn't change
+  /// after a physics step.
+  public: std::unordered_map<Entity, ignition::physics::Pose<double, 3>>
```
adlarkin, line 219: I hardcoded `double` and `3` because these are the values that seem to match the template values of the link's world pose in `UpdateSim` (I ran `gdb` with a breakpoint on this line and then ran `whatis worldPose`, which returned `type = const Eigen::Transform<double, 3, 1, 0> &`):
https://github.com/ignitionrobotics/ign-gazebo/blob/86e6ad93581362eba4e89a58dc1fef058806a081/src/systems/physics/Physics.cc#L1727

Is there a better way to define these template values, especially if `double` and `3` aren't always guaranteed to work? From my understanding, `double` is the pose's `Scalar` value, and `3` is the pose's `dimension` value (more information about these template parameters can be found [here](https://github.com/ignitionrobotics/ign-physics/blob/6aa968042659ad4ee60e9126a7ff81fe68d0d05c/include/ignition/physics/Geometry.hh#L43-L44)).

src/systems/physics/Physics.cc
```diff
@@ -1726,51 +1734,62 @@ void PhysicsPrivate::UpdateSim(EntityComponentManager &_ecm)
...
-
-          _ecm.SetChanged(topLevelModelEnt, components::Pose::typeId,
-              ComponentState::PeriodicChange);
-        }
-        else
+        // update the link or top level model pose if this is the first update,
+        // or if the link pose has changed since the last update
+        // (if the link pose hasn't changed, there's no need for a pose update)
+        if ((this->linkWorldPoses.find(_entity) == this->linkWorldPoses.end())
+            || (this->linkWorldPoses[_entity].matrix() != worldPose.matrix()))
```
adlarkin, line 1741: I took a look at the documentation for [`Eigen::Transform`](https://eigen.tuxfamily.org/dox/classEigen_1_1Transform.html) (`ignition::Physics::Pose` is really `Eigen::Transform` [under the hood](https://github.com/ignitionrobotics/ign-physics/blob/6aa968042659ad4ee60e9126a7ff81fe68d0d05c/include/ignition/physics/Geometry.hh#L44)), and it doesn't seem to have an `==` operator defined. So, in order to see if the transforms are equal, I had to call the transform's [`matrix()` method](https://eigen.tuxfamily.org/dox/classEigen_1_1Transform.html#a758fbaf6aa41a5493659aa0c9bfe0dcf) and then compare the matrices for equality. Is this okay, or is there a better way of checking if the two transforms are equal?

adlarkin (commented):
A few questions for the PR reviewers (cc @azeey).

src/systems/physics/Physics.cc
azeey: This physics system is specifically for `double` scalar and `3` dimension (i.e, `FeaturePolicy3d`), so I think it's okay to hard code them.

src/systems/physics/Physics.cc
azeey: Comparing the matrices directly is likely doing comparison on doubles, which means the matrices have to be exactly the same for equality. This might be okay for the purpose of this code because it only causes an update if the matrices aren't equal. A different way would be to compare the norm of the difference of the translations with some tolerance and then convert the relative rotation to angle axis representation and check the angle (https://github.com/ignitionrobotics/ign-physics/blob/9ba272864ec261ffe972365c82a808b476628659/test/Utils.hh#L181)

Since we have comparison functions for `ignition::math::Pose3d` in Physics.cc already, it might be easier to store them as `ignition::math::Pose3d`. You can do the conversion with `ignition::math::eigen3::convert`.

adlarkin: > Comparing the matrices directly is likely doing comparison on doubles, which means the matrices have to be exactly the same for equality.

You're right; I looked into this some more, and the [MatrixBase == operator](https://eigen.tuxfamily.org/dox/classEigen_1_1MatrixBase.html#a80e3e1e83fdf43f9f7fb6ff51836b24d) has no tolerance for floats. While we could use [`isApprox()`](https://eigen.tuxfamily.org/dox/classEigen_1_1DenseBase.html#ae8443357b808cd393be1b51974213f9c), it looks like there could be some issues with zero vectors/matrices.

> Since we have comparison functions for `ignition::math::Pose3d` in `Physics.cc` already, it might be easier to store them as `ignition::math::Pose3d`. You can do the conversion with `ignition::math::eigen3::convert`.

I went ahead and used `ignition::math::Pose3d` and the corresponding comparison functions in `Physics.cc` in https://github.com/ignitionrobotics/ign-gazebo/pull/669/commits/6bf05727f38212c16f9b4ac3e26b6d5058f005d8. Another nice thing about using `ignition::math::Pose3d` is that it removes the hardcoded values from `physics::Pose` that [I mentioned above](https://github.com/ignitionrobotics/ign-gazebo/pull/669#discussion_r587785042). If this looks good to you, I think we can mark this as resolved.

src/systems/physics/Physics.cc
adlarkin: https://github.com/ignitionrobotics/ign-gazebo/pull/669/commits/6bf05727f38212c16f9b4ac3e26b6d5058f005d8 replaces `ignition::physics::Pose<double, 3>` with `ignition::math::Pose3d`, so I'm marking this as resolved.

src/systems/physics/Physics.cc
```diff
@@ -212,6 +212,11 @@ class ignition::gazebo::systems::PhysicsPrivate
   /// \brief Keep track of what entities are static (models and links).
   public: std::unordered_set<Entity> staticEntities;
 
+  /// \brief Keep track of non-static link poses.
+  /// This allows for skipping pose updates if a link's pose didn't change
+  /// after a physics step.
```
iche033, line 217: In a model, there will be links that never move over time (i.e. ones not connected by joints). Looking at the code below, does `non-static link poses` here refer to its parent not being static?

iche033 (commented):
nice, changes look good to me. Just one question about the non-static link pose comment

src/systems/physics/Physics.cc
adlarkin: > Looking at the code below, does non-static link poses here refer to its parent not being static?

Yes, that is the correct interpretation. I've updated the documentation in https://github.com/ignitionrobotics/ign-gazebo/pull/669/commits/c88b4747c9a1b1d6f77fb61a4e797d417a71d195 to make this clear. If it looks good to you, we can mark this as resolved.

iche033: looks good to me.

iche033 (approved):

PR #678, opened 2021-03-10
iterate through changed links only in UpdateSim

adlarkin:

# 🎉 New feature

requires:
* https://github.com/ignitionrobotics/ign-physics/pull/223
* https://github.com/ignitionrobotics/ign-physics/pull/238
* https://github.com/ignitionrobotics/ign-math/pull/196

related to https://github.com/ignitionrobotics/ign-physics/issues/219

## Summary

As mentioned in https://github.com/ignitionrobotics/ign-physics/issues/219, `UpdateSim` in the `ign-gazebo` physics system iterates over all links, because it has no way of knowing which links have moved. https://github.com/ignitionrobotics/ign-physics/pull/223 and https://github.com/ignitionrobotics/ign-physics/pull/238 should resolve this issue, so this PR uses each of these `ign-physics` PRs (and https://github.com/ignitionrobotics/ign-math/pull/196) to have `UpdateSim` iterate over just the links that have moved.

As a result of this change, the link pose caching that was done in https://github.com/ignitionrobotics/ign-gazebo/pull/669 is no longer needed if a physics engine is being used that writes output pose data (we still need the caching done in https://github.com/ignitionrobotics/ign-gazebo/pull/656 regardless of whether the physics engine being used writes output pose data or not). I also had to add some functionality to the `EntityFeatureMap` class (introduced in https://github.com/ignitionrobotics/ign-gazebo/pull/586) that maps physics link entity IDs (`std::size_t`) to the corresponding entity in `ign-gazebo` (the [link data](https://github.com/ignitionrobotics/ign-physics/blob/ignition-physics3_3.1.0/include/ignition/physics/ForwardStep.hh#L39-L44) that `ign-gazebo` gets from `ign-physics` contains a link's pose and link ID).

This new behavior is **optional**. In other words, if a physics engine is being used that writes output pose data, then the output pose data will be used to update the simulation state. If a physics engine is being used that does not write output pose data, then the approach for updating the simulation state remains unchanged (getting all link poses from the latest step via `ECM`).

## Test it

Testing this PR can follow the testing approach taken in #656 and #669, which involves profiling a world as follows:
* [3000 non-static (or static) simple shapes](https://gist.github.com/adlarkin/c0a97f50bfaa04131ec7275eaa3f7c0f) sitting on a static ground plane
* running with unbounded RTF (-z 1000000000)
* running headless (with -s and only the Physics system)

Profiling the world above can be done with both DART and TPE physics engines. In the comparisons below, the metrics being compared to (i.e., before this PR) are being done on commit https://github.com/ignitionrobotics/ign-gazebo/commit/2bf29c214bc81d5b75a087f995582d7a82908445 of `ign-gazebo`, and commit https://github.com/ignitionrobotics/ign-physics/commit/54d2a4433cf231deb3c8ace085c35cea5c9c06c6 of `ign-physics`.

#### Static

Before this PR:
* RTF: 21%
* `UpdateSim` time: 3ms

static_tpe_pre_PR

After this PR:
* RTF: 40%
* `UpdateSim` time: 0.67ms

static_tpe_post_PR

#### Non-Static

Before this PR:
* RTF: 4.8%
* `UpdateSim` time: 13.6ms

non_static_tpe_pre_PR

After this PR:
* RTF: 13.8%
* `UpdateSim` time: 0.8ms

non_static_tpe_post_PR

#### Static

Before this PR:
* RTF: 8.5%
* `UpdateSim` time: 4.1ms

static_dart_pre_PR

After this PR:
* RTF: 12.5%
* `UpdateSim` time: 1.1ms

static_dart_post_PR

#### Non-Static

Before this PR:
* RTF: 0.95%
* `UpdateSim` time: 15.6ms

non_static_dart_pre_PR

After this PR:
* RTF: 1.1%
* `UpdateSim` time: 2.8ms

non_static_dart_post_PR

## Takeaways

The TPE performance improvement is noticeable (2x RTF increase for static shapes, and 3x RTF increase for non-static shapes). The performance improvement for DART is not as noticeable (1.5x RTF increase for static shapes, and no real change in RTF for non-static shapes), probably because `UpdateSim` is a small part of the `Update` in the physics system - for DART, most of the work in the physics system `Update` comes from the physics step itself, which we probably don't have a ton of control over when it comes to optimization.

### Future Work

Although there are noticeable performance improvements here, there are a few things that I'd like to do next to create even more performance improvements:
1. address https://github.com/ignitionrobotics/ign-gazebo/issues/706 - if you take a look at the profiler images above, you'll notice that most of the time in `UpdateSim` is taken up by `Models`, which is a loop that updates model poses if necessary. Since we now only have the links that have changed after each physics step, it would be nice to make a bi-directional mapping between models and their canonical links (currently, we only have a mapping from a model to its canonical link - see #685). That way, we can loop through all of the changed links and update the models that have this link as its canonical link instead of looping through all of the models and seeing if a model's canonical link has changed pose. I believe that this would decrease the `Model` block time significantly.
1. optimize the `EntityComponentManager::Each` call. As I mentioned in https://github.com/ignitionrobotics/ign-gazebo/pull/678#issuecomment-806317054, reducing the number of components used in the `ecm.Each` call sped up the time for the `ecm.Each` call to complete by approximately 3x (wow!). Considering how often we use `ecm.Each` in `ign-gazebo`, I imagine that optimizing this function call would result in a massive performance benefit. I have created an issue for this so that it can be tracked: https://github.com/ignitionrobotics/ign-gazebo/issues/711

## Checklist
- [x] Added tests

### DIFF
--- a/src/systems/physics/EntityFeatureMap.hh
+++ b/src/systems/physics/EntityFeatureMap.hh
@@ -137,6 +137,7 @@ namespace systems::physics_system
         return castEntity;
       }
     }
+
     /// \brief Helper function to cast from an entity type with minimum features
     /// to an entity with a different set of features. This overload takes a
     /// physics entity as input
@@ -187,6 +188,21 @@ namespace systems::physics_system
       return kNullEntity;
     }
 
+    /// \brief Get the physics entity with required features that has a
+    /// particular ID
+    /// \param[in] _id The ID of the desired physics entity
+    /// \return If found, returns the corresponding physics entity. Otherwise,
+    /// nullptr
+    public: RequiredEntityPtr GetPhysicsEntityPtr(std::size_t _id) const
+    {
+      auto it = this->physEntityById.find(_id);
+      if (it != this->physEntityById.end())
+      {
+        return it->second;
+      }
+      return nullptr;
+    }
+
     /// \brief Check whether there is a physics entity associated with the given
     /// Gazebo entity
     /// \param[in] _entity Gazebo entity.
@@ -215,6 +231,7 @@ namespace systems::physics_system
     {
       this->entityMap[_entity] = _physicsEntity;
       this->reverseMap[_physicsEntity] = _entity;
+      this->physEntityById[_physicsEntity->EntityID()] = _physicsEntity;
     }
 
     /// \brief Remove entity from all associated maps
@@ -226,8 +243,9 @@ namespace systems::physics_system
       if (it != this->entityMap.end())
       {
         this->reverseMap.erase(it->second);
-        this->entityMap.erase(it);
+        this->physEntityById.erase(it->second->EntityID());
         this->castCache.erase(_entity);
+        this->entityMap.erase(it);
         return true;
       }
       return false;
@@ -242,8 +260,9 @@ namespace systems::physics_system
       if (it != this->reverseMap.end())
       {
         this->entityMap.erase(it->second);
-        this->reverseMap.erase(it);
+        this->physEntityById.erase(it->first->EntityID());
         this->castCache.erase(it->second);
+        this->reverseMap.erase(it);
         return true;
       }
       return false;
@@ -257,13 +276,13 @@ namespace systems::physics_system
       return this->entityMap;
     }
 
-    /// \brief Get the total number of entries in the three maps. Only used for
+    /// \brief Get the total number of entries in the maps. Only used for
     /// testing.
     /// \return Number of entries in all the maps.
     public: std::size_t TotalMapEntryCount() const
     {
       return this->entityMap.size() + this->reverseMap.size() +
-             this->castCache.size();
+             this->castCache.size() + this->physEntityById.size();
     }
 
     /// \brief Map from Gazebo entity to physics entities with required features
@@ -271,6 +290,10 @@ namespace systems::physics_system
 
     /// \brief Reverse map of entityMap
     private: std::unordered_map<RequiredEntityPtr, Entity> reverseMap;
+
+    /// \brief Map of physics entity IDs to the corresponding physics entity
+    /// with required features
+    private: std::unordered_map<std::size_t, RequiredEntityPtr> physEntityById;
 
     /// \brief Cache map from Gazebo entity to physics entities with optional
     /// features
--- a/src/systems/physics/EntityFeatureMap_TEST.cc
+++ b/src/systems/physics/EntityFeatureMap_TEST.cc
@@ -118,8 +118,8 @@ TEST_F(EntityFeatureMapFixture, AddCastRemoveEntity)
 
   testMap.AddEntity(gazeboWorld1Entity, testWorld1);
 
-  // After adding the entity, there should be one entry each in two maps
-  EXPECT_EQ(2u, testMap.TotalMapEntryCount());
+  // After adding the entity, there should be one entry each in three maps
+  EXPECT_EQ(3u, testMap.TotalMapEntryCount());
   EXPECT_EQ(testWorld1, testMap.Get(gazeboWorld1Entity));
   EXPECT_EQ(gazeboWorld1Entity, testMap.Get(testWorld1));
 
@@ -128,7 +128,7 @@ TEST_F(EntityFeatureMapFixture, AddCastRemoveEntity)
       testMap.EntityCast<TestOptionalFeatures1>(gazeboWorld1Entity);
   ASSERT_NE(nullptr, testWorld1Feature1);
   // After the cast, there should be one more entry in the cache map.
-  EXPECT_EQ(3u, testMap.TotalMapEntryCount());
+  EXPECT_EQ(4u, testMap.TotalMapEntryCount());
 
   // Cast to optional feature2
   auto testWorld1Feature2 =
@@ -136,12 +136,12 @@ TEST_F(EntityFeatureMapFixture, AddCastRemoveEntity)
   ASSERT_NE(nullptr, testWorld1Feature2);
   // After the cast, the number of entries should remain the same because we
   // have not added an entity.
-  EXPECT_EQ(3u, testMap.TotalMapEntryCount());
+  EXPECT_EQ(4u, testMap.TotalMapEntryCount());
 
   // Add another entity
   WorldPtrType testWorld2 = this->engine->ConstructEmptyWorld("world2");
   testMap.AddEntity(gazeboWorld2Entity, testWorld2);
-  EXPECT_EQ(5u, testMap.TotalMapEntryCount());
+  EXPECT_EQ(7u, testMap.TotalMapEntryCount());
   EXPECT_EQ(testWorld2, testMap.Get(gazeboWorld2Entity));
   EXPECT_EQ(gazeboWorld2Entity, testMap.Get(testWorld2));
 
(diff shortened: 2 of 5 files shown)


adlarkin:
Note to self: some of the caching that was done in #656 may no longer be needed (I will look into this). #669 has already been removed in this PR since https://github.com/ignitionrobotics/ign-physics/pull/223 gives us only the links that have moved in a physics step

adlarkin:
Note to self: remove all instances of the pose link map cache that was introduced in #669 since it's no longer needed

azeey (commented):
At a high level, my main concern is that this changes the requirement on the ign-physics plugin such that the `ForwardStep::World::Step` function now has to return the world poses. We have been trying not to require features in a released version and this feels like we're breaking that. I'd like to hear @chapulina's thoughts on this. 

A possible workaround to maintain backward compatibility would be to populate `stepOutput` in ign-gazebo if the return of `ForwardStep::World::Step` doesn't contain any world poses.

chapulina:
> We have been trying not to require features in a released version

+1, we should be mindful of not introducing new required APIs that may break existing 3rd party physics engines.

> possible workaround to maintain backward compatibility

That sounds like a good idea. 

Another thing to consider is working on this feature on top of Edifice so that we're free to change requirements and implement the feature in the ideal way. Once that's settled, we can think of how to backport it.

adlarkin:
Thanks for the comments, @azeey and @chapulina. I am going to update this PR and https://github.com/ignitionrobotics/ign-physics/pull/223 to target Edifice, and then we can determine the next best steps.

adlarkin:
This is now ready for review, and targets Edifice (`main` branch).

As I mentioned in the PR description, this change eliminates the need for #669, which is something to keep in mind when we make the next forward-port.

We'll need to make a minor release for `ign-math6` so that ignitionrobotics/ign-math#196 can be used.

adlarkin:
> As I mentioned in the PR description, this change eliminates the need for #669, which is something to keep in mind when we make the next forward-port.

I just updated this branch to include the changes in #695, and resolved conflicts to use the changes in this PR instead of the changes in #669: https://github.com/ignitionrobotics/ign-gazebo/pull/678/commits/1074b5a5cfddc2c8edb3f2975174c8f1b3105f8a

azeey (approved):
LGTM!

azeey:

adlarkin:
> LGTM!

I'm making one last change that will make this new physics output optional (if poses aren't a part of `ignition::physics::ForwardStep::Output`, then the physics system will iterate through all of the links in `UpdateSim` instead of through the links in `ignition::physics::ForwardStep::Output`). Once that change is in, I believe this can be merged.

adlarkin:
> I'm making one last change that will make this new physics output optional

@azeey this new behavior is now optional: https://github.com/ignitionrobotics/ign-gazebo/pull/678/commits/040b4a16fe162c167852c519895fac01543c1876

In order to achieve optional behavior, this PR now requires https://github.com/ignitionrobotics/ign-gazebo/pull/678/commits/040b4a16fe162c167852c519895fac01543c1876.

chapulina:
Once conflicts are fixed, I think this may be ready to go with the latest nightlies / bottles

adlarkin:
> Once conflicts are fixed

In the process of resolving this now; I'll let you know when I am done.

adlarkin:
Conflicts have been resolved in https://github.com/ignitionrobotics/ign-gazebo/pull/678/commits/dca0f6686cf6e7b7c0173a9f54bd58a414f04dc8, but it looks like there's a new test failure that I need to resolve (`INTEGRATION_wind_effects`). Using physics engine output to determine which links changed is still optional behavior (if a physics engine doesn't write output, the `ECM` is used like we've been doing up until this point).

@azeey, I had to make a few changes to the physics system when resolving merge conflicts in order to support the work you did in #685. The main difference that was made was including a `modelWorldPoses` cache since the frame data that is used in `UpdateSim` only contains links that have changed instead of all links (which means that we don't update every model's pose every time, so we should cache model poses to avoid losing pose information for models that aren't updated). The merge commit (https://github.com/ignitionrobotics/ign-gazebo/pull/678/commits/dca0f6686cf6e7b7c0173a9f54bd58a414f04dc8) shows how I resolved the conflicts and how I made use of this `modelWorldPoses` cache, and as we discussed offline, I made some optimizations to the model pose update `ecm.each` call in https://github.com/ignitionrobotics/ign-gazebo/pull/678/commits/25719238fee10560800cbbe81a97ef3db11f0669.

To give more insight into the benefits of https://github.com/ignitionrobotics/ign-gazebo/pull/678/commits/25719238fee10560800cbbe81a97ef3db11f0669, here are the profiler results before this commit using the same testing methodology described in the PR description. As we can see, the `ecm.Each` model loop takes `2.5ms`:

ecm_each_original

After applying the changes made in https://github.com/ignitionrobotics/ign-gazebo/pull/678/commits/25719238fee10560800cbbe81a97ef3db11f0669, the `ecm.Each` model loop takes `0.79ms`:

modified_ecm_each

It's interesting to see how much faster the `ecm.Each` call can be when it is used with fewer components (in this case, we only use the components we absolutely need with `ecm.Each`, and then query other components later on if we need them). Perhaps optimizing `ecm.Each` is something I can look into as I continue to work on simulation performance updates.

---

**Testing the optional behavior**

To revert to the old behavior of using the `ECM` to go through all links for updates instead of using physics output for updating links, make the following changes in `ign-physics` (this is assuming you're on commit `b10a4c3` in `ign-physics`):
```diff
diff --git a/dartsim/src/SimulationFeatures.cc b/dartsim/src/SimulationFeatures.cc
index 2c4ef89..e7aa863 100644
--- a/dartsim/src/SimulationFeatures.cc
+++ b/dartsim/src/SimulationFeatures.cc
@@ -59,7 +59,7 @@ void SimulationFeatures::WorldForwardStep(
 
   // TODO(MXG): Parse input
   world->step();
-  this->Write(_h.Get<ChangedWorldPoses>());
+  //this->Write(_h.Get<ChangedWorldPoses>());
   // TODO(MXG): Fill in state
 }
 
diff --git a/tpe/plugin/src/SimulationFeatures.cc b/tpe/plugin/src/SimulationFeatures.cc
index 30db7b4..10992c3 100644
--- a/tpe/plugin/src/SimulationFeatures.cc
+++ b/tpe/plugin/src/SimulationFeatures.cc
@@ -62,7 +62,7 @@ void SimulationFeatures::WorldForwardStep(
     }
   }
   world->Step();
-  this->Write(_h.Get<ChangedWorldPoses>());
+  //this->Write(_h.Get<ChangedWorldPoses>());
 }
```

You can also add some print statements inside the `if`...`else` block of the `PhysicsPrivate::ChangedLinks` method in the physics system to get further validation on which method is being used for recording link updates.

chapulina:
Removing from **beta**, this can come in a minor release.

adlarkin:
> it looks like there's a new test failure that I need to resolve (INTEGRATION_wind_effects)

I fixed the failure in https://github.com/ignitionrobotics/ign-gazebo/pull/678/commits/4f44bfca6b6d958a7c8bf25e7502b2df52fccf03. The issue was that since we are now only updating the simulation if a pose change greater than `1e-6` took place ([check in `ign-physics` if the engine writes output](https://github.com/ignitionrobotics/ign-physics/blob/ignition-physics4_4.0.0-pre1/dartsim/src/SimulationFeatures.cc#L84-L96) and [check in `ign-gazebo` if the physics engine doesn't write output](https://github.com/ignitionrobotics/ign-gazebo/blob/dc2f521013b3d0f478497b5d7098d93840ffa0c1/src/systems/physics/Physics.cc#L1783-L1796)), the object wasn't moving fast enough after applying the wind force in the test. As a result, it took several iterations after applying the wind force for the object to undergo a pose change greater than `1e-6`, so an update wasn't registered until this point, and [this check](https://github.com/ignitionrobotics/ign-gazebo/blob/4f44bfca6b6d958a7c8bf25e7502b2df52fccf03/test/integration/wind_effects.cc#L325) failed:
```
ign-gazebo/test/integration/wind_effects.cc:325: Failure
Expected: (std::fabs(lastVelMagnitude - velMagnitude)) < (1e-6), actual: 6.34714e-05 vs 1e-06

[  FAILED  ] WindEffectsTest.TopicsAndServices (90 ms)
[----------] 4 tests from WindEffectsTest (1445 ms total)

[----------] Global test environment tear-down
[==========] 4 tests from 1 test suite ran. (1445 ms total)
[  PASSED  ] 3 tests.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] WindEffectsTest.TopicsAndServices

 1 FAILED TEST
```

I increased the amount of wind force applied so that the object has a higher velocity, which results in more frequent physics updates after the wind force is no longer applied. This ensures that we don't see any velocity jumps after applying the wind force in the `INTEGRATION_wind_effects` test.

adlarkin:
I have addressed all review feedback and updated this PR with the latest updates from the `main` branch. All tests and codecheck pass for me locally. I believe this is ready for final review!

azeey (approved):
Great work!!

adlarkin:
Now that Edifice has been released, should I change the base branch to `ign-gazebo5` before merging?

PR #645, opened 2021-02-22
some world files added

addy1997:
@chapulina, I am continuing this pull-up request from #644. I have included a few world files like camera, sensors, cars, tri-copter. Note: I have taken the mesh files from Google warehouse and constructed the .sdf file by referring to the tutorials on gazebosim.org. 

Thanks

PR #680, opened 2021-03-11
Code for actor walking in circular path

addy1997:
I have added the code for the Actor to move in the circle.


chapulina (changes requested):
Thank you for the pull request, @addy1997 . Did you mean to include all the other worlds too? I recommend creating separate branches for each feature so you can open pull requests for them separately.

Similar to #645 , how do you feel about uploading `Actor_circle_path.sdf` to Fuel? Once you've uploaded a lot of these worlds, I suggest you post about them on https://community.gazebosim.org/ so more people take a look at them!

addy1997:
@chapulina Nope, I didn't mean to include all the worlds. This was just for the actor's circular path. Would uploading a single actor world file make the entire repo bulky?

chapulina:
> Would uploading a single actor world file make the entire repo bulky?

Got it. One world file is not that large, but in general, we're trying not to add new worlds for now unless they're demonstrating some new functionality. And eventually we'll move all of them to Fuel.

addy1997:
> > Would uploading a single actor world file make the entire repo bulky?
> 
> Got it. One world file is not that large, but in general, we're trying not to add new worlds for now unless they're demonstrating some new functionality. And eventually, we'll move all of them to Fuel.

Thanks for that. I have uploaded the world model with the proper format and thumbnails. [Link](https://app.ignitionrobotics.org/Gambit/fuel/worlds/actor%20walk%20circle)

PR #684, opened 2021-03-16
Set collision detector and solver from SDF

chapulina:
# 🎉 New feature

Requires https://github.com/ignitionrobotics/ign-physics/pull/225

## Summary

This makes it possible to choose a different collision detector and solver when using DART. This way, users can choose the combination that suits their simulation best.

## Test it

Try setting different options in SDF, for example:

```
      <dart>
        <collision_detector>bullet</collision_detector>
        <solver>
          <solver_type>pgs</solver_type>
        </solver>
      </dart>
```

Try the example world:

`ign gazebo physics_options.sdf`

See that it behaves different from `rolling_shapes.sdf`:

`rolling_shapes.sdf` | `physics_options.sdf`
-- | -- 
default_rolling | bullet_pgs

I also made the options visible on the GUI:

image

## Checklist
- [x] Added tests
- [ ] ~~Updated migration guide (as needed)~~


chapulina:
Removed from **beta**, this can come in a minor release.

chapulina:
This is ready for review, @scpeters

src/LevelManager.cc
```diff
@@ -111,6 +111,34 @@ void LevelManager::ReadLevelPerformerInfo()
   this->runner->entityCompMgr.CreateComponent(this->worldEntity,
       components::Physics(*physics));
 
+  // Populate physics options that aren't accessible outside the Element()
+  // See https://github.com/osrf/sdformat/issues/508
+  if (physics->Element())
+  {
+    if (physics->Element()->HasElement("dart"))
```
adlarkin, line 118:
```suggestion
  if (physics->Element() && physics->Element()->HasElement("dart"))
```
Nit: I believe this can be condensed into a single `if`.

src/SdfEntityCreator.cc
```diff
@@ -279,6 +279,34 @@ Entity SdfEntityCreator::CreateEntities(const sdf::World *_world)
   this->dataPtr->ecm->CreateComponent(worldEntity,
       components::Physics(*physics));
 
+  // Populate physics options that aren't accessible outside the Element()
+  // See https://github.com/osrf/sdformat/issues/508
+  if (physics->Element())
+  {
+    if (physics->Element()->HasElement("dart"))
+    {
```
adlarkin, line 287: Same comment here about condensing into a single `if`

src/systems/physics/Physics.hh
```diff
@@ -39,6 +39,7 @@
 #include <ignition/physics/RemoveEntities.hh>
 #include <ignition/physics/Shape.hh>
 #include <ignition/physics/SphereShape.hh>
+#include <ignition/physics/World.hh>
```
adlarkin, line 42: Is there a reason for including this header (and other headers) in the header file, if they're really only used in the source file? Does it have something to do with making builds green for macOS and/or Windows?

test/integration/physics_system.cc
```diff
@@ -1333,3 +1334,46 @@ TEST_F(PhysicsSystemFixture, NestedModelIndividualCanonicalLinks)
   EXPECT_NE(postUpModelPoses.end(), modelIt);
   EXPECT_NEAR(zExpected, modelIt->second.Z(), 1e-2);
 }
+
+/////////////////////////////////////////////////
+TEST_F(PhysicsSystemFixture, PhysicsOptions)
```
adlarkin, line 1339: Would it be worth adding another test that checks the default values for `collisionDetector` (`fcl`) and `solverType` (`dantzig`)? That way, we can confirm that both default and user-defined input behave as expected.

src/systems/physics/Physics.hh
chapulina: There's a little comment on line 26, these were originally here to satisfy the initial `entityCast` implementation. There's a chance that this isn't needed anymore since #586, I haven't checked.

test/integration/physics_system.cc
chapulina: The components were not being populated by default, only when the user set custom options. I think it's useful for users to know the default values, if available, so I added that in 144adfbfb132003179fedc527e590bf67d4e6617

test/integration/physics_system.cc
```diff
@@ -1375,12 +1375,53 @@ TEST_F(PhysicsSystemFixture, NestedModelIndividualCanonicalLinks)
...
+      _ecm.Each<components::World, components::PhysicsCollisionDetector,
+                components::PhysicsSolver>(
+        [&](const ignition::gazebo::Entity &, const components::World *,
+            const components::PhysicsCollisionDetector *_collisionDetector,
+            const components::PhysicsSolver *_solver)->bool
+        {
+          EXPECT_NE(nullptr, _collisionDetector);
+          if (_collisionDetector)
+          {
+            EXPECT_EQ("ode", _collisionDetector->Data());
```
adlarkin, line 1400: The default for `<collision_detector>` listed in http://sdformat.org/spec?ver=1.8&elem=physics#physics_dart is `fcl`. Is the SDF spec out of date?

adlarkin (approved):
One minor comment, but otherwise, LGTM assuming that CI checks come back green :crossed_fingers:

test/integration/physics_system.cc
chapulina: Yeah that's the default on Gazebo classic, but on Ignition, we started using Bullet (https://github.com/ignitionrobotics/ign-physics/commit/f298fb42eb000a89872bb1bc71b65cfd5e4e0179), and then ODE (https://github.com/ignitionrobotics/ign-physics/commit/8b00f01d4e3afe9af0f740e4f429ef3cb877b120).

Ultimately, I think the SDF spec shouldn't be opinionated with this kind of element, because each simulator may choose a different default value. This is slightly related to https://github.com/osrf/sdformat/issues/508.

adlarkin: Ah, I see - thanks for sharing those links. I just took a look at the [Gazebo classic dependencies](http://gazebosim.org/tutorials?tut=install_dependencies_from_source&cat=install#SDFormat), and it looks like Gazebo 11 supports SDF specification 1.7, so should we at least update the SDF website starting at specification 1.8? I also wonder if for the time being (i.e., until https://github.com/osrf/sdformat/issues/508 is addressed), we should document somewhere about the different physics engine default parameters between Gazebo classic and ign-gazebo. I can also leave a comment on https://github.com/osrf/sdformat/issues/508 if it would be helpful to have these thoughts tracked there.

chapulina: > it looks like Gazebo 11 supports SDF specification 1.7, so should we at least update the SDF website starting at specification 1.8? 

I'm not sure this solves the problem. Ignition Citadel uses 1.7 and uses a different default from Gazebo classic. In the spirit of being simulator-agnostic, I think the SDF spec shouldn't dictate the default values for these options.

> we should document somewhere about the different physics engine default parameters between Gazebo classic and ign-gazebo.

Documented the options supported by Ignition on 16b1af4. And added the high-level comparison to Gazebo classic on https://github.com/ignitionrobotics/docs/pull/186.

PR #690, opened 2021-03-17
[BULLET] Making GetContactsFromLastStepFeature optional in Collision Features

Lobotuerk:
# 🎉 New feature

Makes GetContactFromLastStepFeture not a requirement for Collisions in IGN-Physics. Fixes https://github.com/ignitionrobotics/ign-physics/pull/208#discussion_r594540971

## Checklist
- [x] Added tests

### DIFF
--- a/src/systems/physics/Physics.cc
+++ b/src/systems/physics/Physics.cc
@@ -304,21 +304,23 @@ class ignition::gazebo::systems::PhysicsPrivate
             MinimumFeatureList,
             ignition::physics::sdf::ConstructSdfCollision>{};
 
-  /// \brief Collision type with collision features.
-  public: using ShapePtrType = ignition::physics::ShapePtr<
-            ignition::physics::FeaturePolicy3d, CollisionFeatureList>;
-
-  /// \brief World type with just the minimum features. Non-pointer.
-  public: using WorldShapeType = ignition::physics::World<
-            ignition::physics::FeaturePolicy3d, CollisionFeatureList>;
-
-  //////////////////////////////////////////////////
-  // Contacts
-
   /// \brief Feature list to handle contacts information.
   public: struct ContactFeatureList : ignition::physics::FeatureList<
             CollisionFeatureList,
             ignition::physics::GetContactsFromLastStepFeature>{};
+
+  /// \brief Collision type with collision features.
+  public: using ShapePtrType = ignition::physics::ShapePtr<
+            ignition::physics::FeaturePolicy3d, ContactFeatureList>;
+
+  /// \brief World type with just the minimum features. Non-pointer.
+  public: using WorldShapeType = ignition::physics::World<
+            ignition::physics::FeaturePolicy3d, ContactFeatureList>;
+
+  //////////////////////////////////////////////////
+  // Contacts
+
+
 
   //////////////////////////////////////////////////
   // Collision filtering with bitmasks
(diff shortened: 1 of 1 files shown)


azeey:
This is not compiling mainly because of https://github.com/ignitionrobotics/ign-gazebo/blob/f5a362cf579ef23d2cd9d5ec1cd2f856410db3c3/src/systems/physics/Physics.cc#L2202-L2203

The `EntityFeatureMap::Get` expects an `EntityPtr` with required/minimum FeatureList. In this snippet of code `contact.collision1` has `ContactFeatureList`, which is not the required feature in `EntityCollisionMap`. Ideally, this should downcast cast to the entity with the required FeatureList, but there's an implicit conversion issue that causes an ambiguity in the compiler: the compiler is unable to decide whether to call `EntityFeatureMap::Get(const Entity &)` or `EntityFeatureMap::Get(const RequiredEntityPtr &_physEntity)`. The temporary solution is to manually create an ShapePtr with the required features from `contact.collision1` before calling `EntityFeatureMap::Get`. 

```diff
diff --git a/src/systems/physics/Physics.cc b/src/systems/physics/Physics.cc
index 9b419b8c..5ce52614 100644
--- a/src/systems/physics/Physics.cc
+++ b/src/systems/physics/Physics.cc
@@ -313,3 +313,3 @@ class ignition::gazebo::systems::PhysicsPrivate
   public: using ShapePtrType = ignition::physics::ShapePtr<
-            ignition::physics::FeaturePolicy3d, ContactFeatureList>;
+            ignition::physics::FeaturePolicy3d, CollisionFeatureList>;
 
@@ -432,2 +432,3 @@ class ignition::gazebo::systems::PhysicsPrivate
             CollisionFeatureList,
+            ContactFeatureList,
             CollisionMaskFeatureList,
@@ -2201,5 +2202,6 @@ void PhysicsPrivate::UpdateCollisions(EntityComponentManager &_ecm)
     const auto &contact = contactComposite.Get<WorldShapeType::ContactPoint>();
-    auto coll1Entity = this->entityCollisionMap.Get(contact.collision1);
-    auto coll2Entity = this->entityCollisionMap.Get(contact.collision2);
-
+    auto coll1Entity =
+        this->entityCollisionMap.Get(ShapePtrType(contact.collision1));
+    auto coll2Entity =
+        this->entityCollisionMap.Get(ShapePtrType(contact.collision2));
```

I'll continue to think about a way to fix the implicit conversion problem, but I don't think it's needed for this PR.

src/systems/physics/Physics.cc
```diff
@@ -373,6 +382,7 @@ class ignition::gazebo::systems::PhysicsPrivate
           physics::World,
           MinimumFeatureList,
           CollisionFeatureList,
+          ContactFeatureList,
```
Blast545, line 385: If adding this, `CollisionFeatureList,` can be removed

Lobotuerk: Following that logic, `MinimumFeatureList` should have been removed before, I'm leaving them all just in case

azeey: I agree with @Lobotuerk, I think we want to support physics engines that support `CollisionFeatureList`, but not `ContactFeatureList`.

azeey (approved):
The failing test on Ubuntu is also failing on our ign-gazebo4 CI https://build.osrfoundation.org/job/ignition_gazebo-ci-ign-gazebo4-bionic-amd64/54/, so it's not caused by this PR.

Blast545 (approved):
LGTM

gonzodepedro (approved):
LGTM

PR #694, opened 2021-03-18
Standalone Ignition Gazebo executables

mjcarroll:
# 🎉 Standalone Executables 

## Summary

Adds `ign-gazebo-server` and `ign-gazebo-gui` to be used as standalone executables in `valgrind`, `gdb`, `heaptrack`, etc.

## Checklist
- [ ] Added tests


chapulina:
This can come in a minor, removed from **beta**

scpeters:
We just merged the proof-of-concept for this in https://github.com/ignitionrobotics/ign-transport/pull/216, but it was a bit different since `ign-transport` uses separate commands for each executable and put most of the command-line parsing into the c++ executables:

* `ign topic` -> `ign-transport-topic`
* `ign service` -> `ign-transport-service`

while `ign gazebo` uses arguments to the `gazebo` command to determine whether to open the server or gui or both:

* `ign gazebo -s` -> `ign-gazebo-server`
* `ign gazebo -g` -> `ign-gazebo-gui`
* `ign gazebo` -> both `ign-gazebo-server` and `ign-gazebo-gui`

Furthermore, the [cmdgazebo.rb](https://github.com/ignitionrobotics/ign-gazebo/blob/ignition-gazebo5_5.0.0/src/cmd/cmdgazebo.rb.in#L392-L427) ruby command has complex logic for finding the world `file` to load (specified as a positional argument):

* check if `file` is a path to an existing file
* if not, check for `file` as a relative path in all the paths listed in `IGN_GAZEBO_RESOURCE_PATH`
* if still not, check for `file` as a relative path to the `worldInstallDir` function
* if still not, call `findFuelResource` to try to download the file
* then after all that, parse the file through embedded ruby (ERB)

it would be helpful to have a plan for what will be done in C++ and what will be done in ruby in order to proceed

mjcarroll:
```
ign gazebo -s -> ign-gazebo-server
ign gazebo -g -> ign-gazebo-gui
ign gazebo -> both ign-gazebo-server and ign-gazebo-gui
```

This seems reasonable, but I think that @ahcorde 's current work may allow us to not run them as separate processes.  Part of this was an artifact of OGRE.

> command has complex logic for finding the world file to load

I think it would also be helpful to have this logic implemented in C++, so it could be accessible via API.

chapulina:
> I think it would also be helpful to have this logic implemented in C++, so it could be accessible via API.

All of that should be available from C++, except for the ERB parsing. That's why we had to duplicate a lot of the file locating logic from C++ into the Ruby script. It would be great to have that consistent regardless of whether we use the C++ API or the CLI.

ahcorde:
> This seems reasonable, but I think that @ahcorde 's current work may allow us to not run them as separate processes. Part of this was an artifact of OGRE.

PR https://github.com/ignitionrobotics/ign-gazebo/pull/793

peci1:
@mjcarroll why is this PR closed? Is there any ongoing effort in this direction for ign-gazebo?

mjcarroll:
This is out-of-date enough that I'm going to close this PR and re-open with a new implementation.

PR #702, opened 2021-03-24
Fixed collision visual bounding boxes

jennuine:
Signed-off-by: Jenn Nguyen <jenn@openrobotics.org>

# 🦟 Bug fix

## Summary

Encountered a new bug with collision visualizations. When collisions are turned on for a model, then any collision entity that was selected from the entity tree (which "highlights" the collision entity by putting a bounding box around it), will have bounding boxes around them once you turn off, then turn back on collisions. The only way to make the bounding boxes disappear is to select the collision entity again then deselect but once you turn collisions off then back on again for the model the bounding boxes reappear. 

**Steps to reproduce bug before PR**:
1. `ign gazebo -v 4 "https://fuel.ignitionrobotics.org/1.0/OpenRobotics/worlds/Edifice demo"`
2. Turn on collisions for Mecanum lift
3. In entity tree, open Mecanum lift > chassis
4. Click on collision entities for bar_1 and bar_2
5. Right-click Mecanum lift and toggle off collisions
6. Right-click Mecanum lift and toggle back on collisions

bar_1 and bar_2 should have bounding boxes around them.

If you run the above steps above with this PR, then the bounding boxes will not be shown around the collisions unless it is the most recent highlighted one from the entity tree. For example, if collisions are turned off and you select bar_2 from the entity tree then when you turn collisions back on (with bar_2 highlighted), bar_2 will have a bounding box around it.

This bug fix will need to be backported to `ign-gazebo4` for Dome.

## Checklist
- [ ] Added tests

### DIFF
--- a/src/rendering/RenderUtil.cc
+++ b/src/rendering/RenderUtil.cc
@@ -2164,11 +2164,9 @@ void RenderUtil::ViewCollisions(const Entity &_entity)
 
     if (showCol)
     {
-      // turn off wireboxes for collision entity except for most recent
-      // selected entity (it is still highlighted in entity tree)
+      // turn off wireboxes for collision entity
       if (this->dataPtr->wireBoxes.find(colEntity)
-            != this->dataPtr->wireBoxes.end()
-          && colEntity != this->dataPtr->selectedEntities.back())
+            != this->dataPtr->wireBoxes.end())
       {
         ignition::rendering::WireBoxPtr wireBox =
           this->dataPtr->wireBoxes[colEntity];
--- a/src/rendering/RenderUtil.cc
+++ b/src/rendering/RenderUtil.cc
@@ -2161,5 +2161,21 @@ void RenderUtil::ViewCollisions(const Entity &_entity)
 
     this->dataPtr->viewingCollisions[colEntity] = showCol;
     colVisual->SetVisible(showCol);
+
+    if (showCol)
+    {
+      // turn off wireboxes for collision entity except for most recent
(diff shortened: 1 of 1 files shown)


src/rendering/RenderUtil.cc
```diff
@@ -2161,5 +2161,21 @@ void RenderUtil::ViewCollisions(const Entity &_entity)
 
     this->dataPtr->viewingCollisions[colEntity] = showCol;
     colVisual->SetVisible(showCol);
+
+    if (showCol)
+    {
+      // turn off wireboxes for collision entity except for most recent
+      // selected entity (it is still highlighted in entity tree)
```
chapulina, line 2168: How about also showing collision for all selected entities? There may still be more than one selected. I tried with `log_record_dbl_pendulum.sdf`:

coll

jennuine: Since collision entities aren't created until after the first "View > Collisions", `RenderUtil::SetSelectedEntities` does nothing (does not add the collision entity to the list since it's not created yet) so there is no way of getting selected entities until after collisions are first created (toggled on). To avoid user confusion, I went ahead and turned off the bounding box for all selected collisions when toggling on: 4dfe6d32614474aad1819a532d11b0978294e594

chapulina: Nice! This behaviour is more consistent! :+1:

chapulina (approved):

chapulina:
Also cancelled the Windows build for this PR due to the long queue. I think it's ready to go in! Merging!

PR #717, opened 2021-03-28
Respect spotlight direction

peci1:
# 🦟 Bug fix

## Summary
According to SDF and ign-rendering, spot lights have a direction parameter. However, this parameter was ignored when setting the light in the scene. This PR fixes it.

Of course, this might break some scenes which were set up with the direction ignored. However, I still feel that the change is worth it (but I'd suggest the SubT team to thoroughly check if some worlds aren't affected :) ). If merging in Dome is not possible, merge to Edifice, please.

## Checklist
- [ ] Added tests

### DIFF
--- a/src/rendering/SceneManager.cc
+++ b/src/rendering/SceneManager.cc
@@ -939,6 +939,7 @@ rendering::LightPtr SceneManager::CreateLight(Entity _id,
       spotLight->SetInnerAngle(_light.SpotInnerAngle());
       spotLight->SetOuterAngle(_light.SpotOuterAngle());
       spotLight->SetFalloff(_light.SpotFalloff());
+      spotLight->SetDirection(_light.Direction());
       break;
     }
     case sdf::LightType::DIRECTIONAL:


peci1:
Here's a picture of the current behavior. The link the light is attached to is on the robot's left side with X axis pointing towards the box. I tried many values of the `<direction>` tag, but it always looks the same. If you think it through and connect with the fact that the default value for `<direction>` is `0 0 -1`, this is exactly it. The light is always shining downwards.

Rotating the link itself rotates the light correctly.

2021-03-28T21:30:52 869741306

ahcorde (approved):

chapulina (changes requested):
> If merging in Dome is not possible, merge to Edifice, please.

Yeah fixing this bug in Dome could break current users, targeting Edifice feels safer. Would you be able to retarget the PR to `main`? We can get it in before the stable release as a critical bug fix. We should also add a note to the migration guide. Thanks!

peci1:
Okay, replaced by #718. I'm not sure if the light visuals do not need some adjustment too, but that can be fixed anytime later without breaking anything.

PR #728, opened 2021-03-30
Adding HaltMotion (Or new suggested name) to physics plugin

Lobotuerk:
Signed-off-by: Tomas Lorente <jtlorente@ekumenlabs.com>

# 🎉 New feature

## Summary
Adds the functionality to stop a model's joint's movement in other cases than the battery being 0. By adding a HaltMotion component to the model, its joints movement can be turned on/off. Also added some QoL to kineticEnergyMonitor plugin with profilers and some checking.

(Updates)
19acc8d ~~adds the capability to turn movement halting off if you only need the advice (and making it backward compatible).~~
4a9f551 Moved everything into GameLogicPlugin, just left here the halt component evaluation in physics and the QoL changes in kineticEnergyMonitor

## Test it
TO-DO tested it with SubT worlds for now

## Checklist
- [ ] Added tests


src/systems/kinetic_energy_monitor/KineticEnergyMonitor.hh
```diff
@@ -97,9 +97,10 @@ namespace systems
       </plugin>
     </model>
   \endverbatim */
-  class IGNITION_GAZEBO_VISIBLE KineticEnergyMonitor:
-    public System,
+  class IGNITION_GAZEBO_VISIBLE KineticEnergyMonitor
+  : public System,
```
gonzodepedro, line 101: Unrelated change

Lobotuerk: I made some format changes, following the rest of the IGN plugins. Should that be another PR?

gonzodepedro: I'm ok with it, if it was intended.

include/ignition/gazebo/components/HaltMotion.hh
```diff
@@ -0,0 +1,42 @@
...
+
+namespace ignition
+{
+namespace gazebo
+{
+// Inline bracket to help doxygen filtering.
+inline namespace IGNITION_GAZEBO_VERSION_NAMESPACE {
+namespace components
+{
+  /// \brief A component used to indicate that an entity casts shadows
```
azeey, line 32: nit: Can you update these docs?

src/systems/kinetic_energy_monitor/KineticEnergyMonitor.hh
```diff
@@ -114,6 +115,10 @@ namespace systems
                            EntityComponentManager &_ecm,
                            EventManager &_eventMgr) final;
 
+    /// Documentation inherited
+    public: void Update(const UpdateInfo &_info,
```
azeey, line 119: Use of `Update` is discouraged. Only the physics system should use it. Can the functionality be added `PreUpdate`?

src/systems/kinetic_energy_monitor/KineticEnergyMonitor.cc
```diff
@@ -161,12 +171,54 @@ void KineticEnergyMonitor::Configure(const Entity &_entity,
...
+
+  if (_info.paused)
+    return;
+
+  // Update Halt Motion component
+  auto *haltMotionComp =
+    _ecm.Component<components::HaltMotion>(
+      _ecm.ParentEntity(this->dataPtr->linkEntity));
+
+  if (this->dataPtr->haltMode &&
```
azeey, line 206: Should we check if `haltMotionComp` is nullptr before dereferencing it?

src/systems/kinetic_energy_monitor/KineticEnergyMonitor.cc
```diff
@@ -124,6 +132,8 @@ void KineticEnergyMonitor::Configure(const Entity &_entity,
   ignmsg << "KineticEnergyMonitor publishing messages on "
     << "[" << topic << "]" << std::endl;
 
+  this->dataPtr->haltMode = sdfClone->Get<bool>("halt_mode", false).first;
```
azeey, line 135: Can you add documentation about `halt_mode` in `KineticEnergyMonitor.hh`?

azeey (commented):
IMO, the kinetic energy monitor should not be concerned with halting a vehicle. If we adhere to the separation of concerns principle, I would think another System in SubT (eg. GameLogicPlugin) would monitor the output of the kinetic energy monitor and set the state of the `HaltMotion` component.

I also left some comments in case we decide to proceed with keeping that logic in the kinetic energy monitor.

Lobotuerk:
> IMO, the kinetic energy monitor should not be concerned with halting a vehicle. If we adhere to the separation of concerns principle, I would think another System in SubT (eg. GameLogicPlugin) would monitor the output of the kinetic energy monitor and set the state of the `HaltMotion` component.

I do agree with moving the HaltMotion logic into the GameLogicPlugin, but that would require adding A PreUpdate there. @nkoenig Do you have a strong feeling for either of them?

nkoenig:
> IMO, the kinetic energy monitor should not be concerned with halting a vehicle. If we adhere to the separation of concerns principle, I would think another System in SubT (eg. GameLogicPlugin) would monitor the output of the kinetic energy monitor and set the state of the `HaltMotion` component.
> 
> I also left some comments in case we decide to proceed with keeping that logic in the kinetic energy monitor.

Agreed. @Lobotuerk , can you remove the logic for halting motion from the Kinetic Energy Monitor and place it in the SubT Game Logic Plugin?

nkoenig:
> 
> I do agree with moving the HaltMotion logic into the GameLogicPlugin, but that would require adding A PreUpdate there. @nkoenig Do you have a strong feeling for either of them?

Adding a PreUpdate to the GameLogicPlugin sounds good.

Lobotuerk:
@azeey When moving the halting logic to GameLogic, should I leave the HaltMotion component creation inside KineticEnergyMonitor?

src/systems/kinetic_energy_monitor/KineticEnergyMonitor.hh
Lobotuerk: Moved to https://github.com/osrf/subt/pull/858

src/systems/kinetic_energy_monitor/KineticEnergyMonitor.cc
Lobotuerk: Moved to https://github.com/osrf/subt/pull/858 and fixed

src/systems/kinetic_energy_monitor/KineticEnergyMonitor.cc
Lobotuerk: Added on  f5c472e

include/ignition/gazebo/components/HaltMotion.hh
Lobotuerk: Updated in 67c262b

azeey:
> @azeey When moving the halting logic to GameLogic, should I leave the HaltMotion component creation inside KineticEnergyMonitor?

IMO, We don't need to modify `KineticEnergyMonitor` at all. All the logic can be in `GameLogic` and `Physics`.

azeey (approved):
LGTM! It would be great if you can add a test showing that a vehicle stops once the `HaltMotion` component is set.

chapulina (commented):
I'm wondering if this functionality couldn't be achieved using `components::JointVelocityReset` / `components::JointPositionReset`?

nkoenig (approved):

Lobotuerk:
> I'm wondering if this functionality couldn't be achieved using `components::JointVelocityReset` / `components::JointPositionReset`?

It's better for a long time halt to simply stop the physics layer before applying forces, than trying to reset the joint every single timestep

PR #742, opened 2021-04-06
ECM's ChangedState gets message with modified components

jennuine:
# 🎉 New feature

## Summary
This PR update's ECM's `ChangedState` functions to include added, removed, or modified components in the serialized state message.

Previously, ECM's [ChangedState](https://ignitionrobotics.org/api/gazebo/4.0/classignition_1_1gazebo_1_1EntityComponentManager.html#a7b4d4e3c44ffc8d2b429bafb0a256070) functions returned a message with the serialized state of entities that were changing (new or removed) in the current iteration but not when components were modified (new, removed, or modified).

## Test it

A new test was added to `UNIT_EntityComponentManager_TEST`, to run all tests:
```bash
./build/ignition-gazebo4/bin/UNIT_EntityComponentManager_TEST
```

To run only the new test (for 1st iteration):
```bash
./build/ignition-gazebo4/bin/UNIT_EntityComponentManager_TEST --gtest_filter=EntityComponentManagerRepeat/EntityComponentManagerFixture.ChangedStateComponents/0
```

Also, `INTEGRATION_log_system` was updated to check for pose components in the state message from the `/world/log_pendulum/changed_state` topic:
```bash
./build/ignition-gazebo4/bin/INTEGRATION_log_system --gtest_filter=LogSystemTest.RecordAndPlayback
```

## Checklist
- [X] Added tests
- [X] Updated documentation (as needed)

### DIFF
--- a/src/EntityComponentManager.cc
+++ b/src/EntityComponentManager.cc
@@ -211,7 +211,6 @@ void EntityComponentManager::ClearNewlyCreatedEntities()
 {
   std::lock_guard<std::mutex> lock(this->dataPtr->entityCreatedMutex);
   this->dataPtr->newlyCreatedEntities.clear();
-  this->dataPtr->modifiedComponents.clear();
 
   for (auto &view : this->dataPtr->views)
   {
@@ -1282,7 +1281,10 @@ void EntityComponentManager::SetState(
         // values.
         // igndbg << *comp << "  " << *newComp.get() << std::endl;
         // *comp = *newComp.get();
-        this->dataPtr->AddModifiedComponent(entity);
+
+        // When above TODO is addressed, uncomment AddModifiedComponent below
+        // unless calling SetChanged (which already calls AddModifiedComponent)
+        // this->dataPtr->AddModifiedComponent(entity);
       }
     }
   }
@@ -1393,7 +1395,6 @@ void EntityComponentManager::SetState(
           }
         }
         this->SetChanged(entity, compIter.first, flag);
-        this->dataPtr->AddModifiedComponent(entity);
       }
     }
   }
@@ -1429,6 +1430,7 @@ void EntityComponentManager::SetAllComponentsUnchanged()
 {
   this->dataPtr->periodicChangedComponents.clear();
   this->dataPtr->oneTimeChangedComponents.clear();
+  this->dataPtr->modifiedComponents.clear();
 }
 
 /////////////////////////////////////////////////
--- a/src/EntityComponentManager.cc
+++ b/src/EntityComponentManager.cc
@@ -72,6 +72,12 @@ class ignition::gazebo::EntityComponentManagerPrivate
       msgs::SerializedStateMap &_msg,
       const std::unordered_set<ComponentTypeId> &_types = {});
 
+  /// \brief Add newly modified (created/modified/removed) components to
+  /// modifiedComponents list. The entity is added to the list when it is not
+  /// a newly created entity or is not an entity to be removed
+  /// \param[in] _entity Entity that has component newly modified
+  public: void AddModifiedComponent(const Entity &_entity);
+
   /// \brief Map of component storage classes. The key is a component
   /// type id, and the value is a pointer to the component storage.
   public: std::unordered_map<ComponentTypeId,
@@ -92,6 +98,12 @@ class ignition::gazebo::EntityComponentManagerPrivate
 
   /// \brief Entities that need to be removed.
   public: std::unordered_set<Entity> toRemoveEntities;
+
+  /// \brief Entities that have components newly modified
+  /// (created/modified/removed) but are not entities that have been
+  /// newly created or removed (ie. newlyCreatedEntities or toRemoveEntities).
+  /// This is used for the ChangedState functions
+  public: std::unordered_set<Entity> modifiedComponents;
 
   /// \brief Flag that indicates if all entities should be removed.
   public: bool removeAllEntities{false};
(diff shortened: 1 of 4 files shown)


src/EntityComponentManager.cc
```diff
@@ -1257,6 +1282,7 @@ void EntityComponentManager::SetState(
         // values.
         // igndbg << *comp << "  " << *newComp.get() << std::endl;
         // *comp = *newComp.get();
+        this->dataPtr->AddModifiedComponent(entity);
```
iche033, line 1285: is this call needed here? Given that the above msg says internal error, I expect that something is wrong is it gets to this else block. Looking at the code, the component value is just being updated by removing the old component and creating a new one?

src/EntityComponentManager.cc
```diff
@@ -1367,6 +1393,7 @@ void EntityComponentManager::SetState(
           }
         }
         this->SetChanged(entity, compIter.first, flag);
+        this->dataPtr->AddModifiedComponent(entity);
```
iche033, line 1396: I think the `SetChanged` call above will also call `AddModifiedComponent`

src/EntityComponentManager.cc
```diff
@@ -199,6 +211,7 @@ void EntityComponentManager::ClearNewlyCreatedEntities()
 {
   std::lock_guard<std::mutex> lock(this->dataPtr->entityCreatedMutex);
   this->dataPtr->newlyCreatedEntities.clear();
+  this->dataPtr->modifiedComponents.clear();
```
iche033, line 214: should this call be in `ClearNewlyCreatedEntities` or maybe it fits better in `SetAllComponentsUnchanged`?

iche033:
I wonder if we can safely remove [this call](https://github.com/ignitionrobotics/ign-gazebo/blob/ign-gazebo4/src/systems/log/LogRecord.cc#L328) to record the `dynamic_pose/info` topic now that the state msg also captures the pose changes - mainly so that we don't record the pose twice in the log file.

src/EntityComponentManager.cc
jennuine: It's not needed currently but I put the call there in case the above `TODO` is addressed. Should I comment it out with a note to add it in when the `TODO` is addressed and if `SetChanged` is not called?

iche033: I see, yeah you can comment it out and add a note here

jennuine: Done 70c432050ff7c18d5cb3f79de5cb697ce950ff14

src/EntityComponentManager.cc
jennuine: Oops, thanks 70c432050ff7c18d5cb3f79de5cb697ce950ff14

src/EntityComponentManager.cc
jennuine: Moved to `SetAllComponentsUnchanged` 70c432050ff7c18d5cb3f79de5cb697ce950ff14

jennuine:
Here is a plot of a test I did running a `subt` simulation before this PR and another after. In both tests, I manually drove the robot to the same location. The `state.tlog` file size for the sim test before the PR is `6.5 MB` and after is `15 MB`.

Figure_4

In the legend, where it says `before`/`after` is the raw data and where it says `smooth` is just a Gaussian filter applied to the `before`/`after` data to make it easier to visualize. Although it's hard to get a real representation since I'm running the sim for only ~1.5 mins and I'm not averaging multiple runs, it's just to give a general idea.

---

> I wonder if we can safely remove [this call](https://github.com/ignitionrobotics/ign-gazebo/blob/ign-gazebo4/src/systems/log/LogRecord.cc#L328) to record the `dynamic_pose/info` topic now that the state msg also captures the pose changes - mainly so that we don't record the pose twice in the log file.

I did a quick test by commenting out the call, recording data, then playing it back and everything seems to be working properly. Should this be addressed in another PR or this one? I believe there will need to be some clean up needed after removing the call (e.g., in test and possibly in `LogPlayback`)

iche033:
> I did a quick test by commenting out the call, recording data, then playing it back and everything seems to be working properly. Should this be addressed in another PR or this one?

yeah let's do that in a separate PR. I think it'll require thorough testing to make sure we don't break anything.

iche033:
looks like there is a related test failure on [homebrew](https://build.osrfoundation.org/job/ignition_gazebo-ci-pr_any-homebrew-amd64/5411/). The `INTEGRATION_log_system` test failed with a segfault

jennuine:
> looks like there is a related test failure on [homebrew](https://build.osrfoundation.org/job/ignition_gazebo-ci-pr_any-homebrew-amd64/5411/). The `INTEGRATION_log_system` test failed with a segfault. Not sure if it's flaky

I think this may be a flaky test, I just re-ran homebrew CI and the test passed. It's hard to pinpoint exactly where since I can't reproduce this but I believe the seg fault happened somewhere towards the end of this function
https://github.com/ignitionrobotics/ign-gazebo/blob/8829840d05c45db1599bb0fca7ef9c4914b807af/test/integration/log_system.cc#L1038

because this error message printed 100 times: `[Err] [Physics.cc:1726] Internal error: link [29] not in entity map`

iche033:
I do see a lot of the `[Err] [Physics.cc:1726] Internal error: link [29] not in entity map` msgs when running the `integration_log_system test`, which seems to be happening on `ign-rendering4` branch  as well (cc @chapulina)

But I can't reproduce the test failure.

iche033 (approved):

PR #751, opened 2021-04-09
Don't store duplicate ComponentTypeId in ECM

chapulina:
# 🦟 Bug fix

## Summary

While looking at the ECM earlier with @adlarkin and @iche033 , I noticed that it was weird that the ECM is storing a map from `ComponentTypeId` to `ComponentKey`, because `ComponentKey` is `std::pair<ComponentTypeId, ComponentId>`, which means that `ComponentTypeId` is duplicated.

I tried changing that to `ComponentId` and it only needs these few changes to compile fine.

The map was introduced back at https://github.com/ignitionrobotics/ign-gazebo/pull/416. I don't remember this detail coming up then, maybe it was overlooked?

I haven't profiled this yet, so I'll leave it as draft until I do just to make sure this isn't introducing any regressions.

## Checklist
- [x] Added tests
- [ ] ~~Updated documentation (as needed)~~
- [ ] ~~Updated migration guide (as needed)~~

### DIFF
--- a/src/EntityComponentManager_TEST.cc
+++ b/src/EntityComponentManager_TEST.cc
@@ -2158,6 +2158,11 @@ TEST_P(EntityComponentManagerFixture, SetEntityCreateOffset)
   manager.SetEntityCreateOffset(1000);
   Entity entity2 = manager.CreateEntity();
   EXPECT_EQ(1001u, entity2);
+
+  // Apply a lower offset, prints warning but goes through.
+  manager.SetEntityCreateOffset(500);
+  Entity entity3 = manager.CreateEntity();
+  EXPECT_EQ(501u, entity3);
 }
 
 // Run multiple times. We want to make sure that static globals don't cause
--- a/src/EntityComponentManager.cc
+++ b/src/EntityComponentManager.cc
@@ -1310,7 +1310,7 @@ void EntityComponentManager::SetChanged(
   }
   else if (_c == ComponentState::OneTimeChange)
   {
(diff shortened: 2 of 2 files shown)


src/EntityComponentManager_TEST.cc
```diff
@@ -2158,6 +2158,11 @@ TEST_P(EntityComponentManagerFixture, SetEntityCreateOffset)
   manager.SetEntityCreateOffset(1000);
   Entity entity2 = manager.CreateEntity();
   EXPECT_EQ(1001u, entity2);
+
+  // Apply a lower offset, prints warning but goes through.
+  manager.SetEntityCreateOffset(500);
+  Entity entity3 = manager.CreateEntity();
+  EXPECT_EQ(501u, entity3);
```
adlarkin, line 2165: Did you intentionally mean to apply a lower offset to test the warning? If not, we can move this code block above the [first offset call](https://github.com/ignitionrobotics/ign-gazebo/blob/8d35373b44dbe8fa7469d949bd48b8dfd6c17d52/src/EntityComponentManager_TEST.cc#L2157-L2158), which should remove the warning.

adlarkin (approved):
LGTM :+1: I don't think this will improve performance at all since we are updating a container's value, but it's definitely good to remove this duplicate data type.

src/EntityComponentManager_TEST.cc
chapulina: Yup, it was an easy addition to test coverage

Issue #711, opened 2021-03-26
Improve the performance of EntityComponentManager::Each


adlarkin:
## Desired behavior
As discussed in https://github.com/ignitionrobotics/ign-gazebo/pull/678#issuecomment-806317054, it seems like adding more components to an `ecm.Each` call reduces performance. Considering how often we call this method in `ign-gazebo`, it would be nice to optimize `ecm.Each` so that performance is the same regardless of the number of components used.

## Alternatives considered
TBD

## Implementation suggestion
TBD - I plan to look into this further and will post an update here based on my findings.

## Additional context
Could be related to https://github.com/ignitionrobotics/ign-gazebo/issues/429

adlarkin:
After some further investigation, I believe that the views are responsible for most of the performance in `ecm.Each`. The issue is that the views aren't really working as a true "cache". Once we have the view, we have to find the components in the view that correspond to the current entity, every single time we want to use an entity (`view.Component<ComponentTypeTs>(entity, this)...`): https://github.com/ignitionrobotics/ign-gazebo/blob/ignition-gazebo5_5.0.0/include/ignition/gazebo/detail/EntityComponentManager.hh#L386-L394

In reality, we shouldn't have to do this, because the view should already have these components cached for every entity that has them. I think it could be good to change how components are saved in a view. Currently, we store components in a view using a `std::map`, with a `std::pair` being the key of this map: https://github.com/ignitionrobotics/ign-gazebo/blob/ignition-gazebo3_3.8.0/include/ignition/gazebo/detail/View.hh#L129-L131

I tried to put a band-aid on the drawbacks of using a `std::map` with a key of `std::pair` in #752, but in reality, I believe that a better solution would be to store something like a table of component data in a view. This table would have the columns be the components, and the rows be entities - then, when it's time to use this view with an entity, the entity would simply index the proper row of data in the view's table, instead of having to find each entity's component individually.

If we were to take this approach, I believe that it would be crucial to create the following mappings (these mappings need to be fast):
* Entity ID to its row index in a view's table (we can't just assume that the Entity ID is the row ID in a view's table)
* A set of components that make up a view to the ID of the view. Right now, it looks like we have no mapping or hash function for this, which means that comparing two sets of components (when searching for views) could be slow:
    * https://github.com/ignitionrobotics/ign-gazebo/blob/ignition-gazebo3_3.8.0/src/EntityComponentManager.cc#L110-L111
[...]

PR #752, opened 2021-04-09
Use unordered maps for component look-ups in views

adlarkin:

## Summary
Related to #711 

Currently, when calling `EntityComponentManager::Each`, the following methods get called:
1. [EntityComponentManager::Each](https://github.com/ignitionrobotics/ign-gazebo/blob/ignition-gazebo3_3.8.0/include/ignition/gazebo/detail/EntityComponentManager.hh#L362-L380)
2. [View::Component](https://github.com/ignitionrobotics/ign-gazebo/blob/ignition-gazebo3_3.8.0/include/ignition/gazebo/detail/View.hh#L63-L75)
3. [View::ComponentImplementation](https://github.com/ignitionrobotics/ign-gazebo/blob/ignition-gazebo3_3.8.0/src/View.cc#L61-L69)
4. [EntityComponentManager::ComponentImplementation](https://github.com/ignitionrobotics/ign-gazebo/blob/ignition-gazebo3_3.8.0/src/EntityComponentManager.cc#L598-L608)
5. [ComponentStorageBase::Component](https://github.com/ignitionrobotics/ign-gazebo/blob/ignition-gazebo3_3.8.0/include/ignition/gazebo/detail/ComponentStorageBase.hh#L187-L199)

There are a few `std::map` lookups in this process:
* https://github.com/ignitionrobotics/ign-gazebo/blob/ignition-gazebo3_3.8.0/src/View.cc#L68
* https://github.com/ignitionrobotics/ign-gazebo/blob/ignition-gazebo3_3.8.0/include/ignition/gazebo/detail/ComponentStorageBase.hh#L191

Since lookup time in a `std::map` is `O(log(n))`, it's better to use `std::unordered_map` which has a lookup time of `O(1)` (assuming no collisions). This PR makes these changes, with the hope of improving the performance of `ecm.Each`.

### Testing

I tested these changes applied to `ign-gazebo` at commit https://github.com/ignitionrobotics/ign-gazebo/commit/a41cf80049c07977079d0b215398dc4225b63038. I used the testing methodology that was followed in #678 (take a look at that PR for more details about how to run the tests). The results are as follows (the model `ecm.Each` call I am referring to in the test results is a part of the `UpdateSim` in the physics system that can be found [here](https://github.com/ignitionrobotics/ign-gazebo/blob/a41cf80049c07977079d0b215398dc4225b63038/src/systems/physics/Physics.cc#L1810-L1812)):

#### 3000 non-static simple shapes
Using `std::map`: 15.75% RTF, .803ms for the model `ecm.Each` call

map_views_non_static

Using `std::unordered_map`: 16.5% RTF, .306ms for the model `ecm.Each` call

unordered_map_views_non_static

#### 3000 static simple shapes
Using `std::map`: 50% RTF, .754ms for the model `ecm.Each` call

map_views_static

Using `std::unordered_map`: 70% RTF, .289ms for the model `ecm.Each` call

unordered_map_views_static

### Takeaways
After running some tests, it appears that the performance gains from this change are minimal, and in some cases, not noticeable at all. This change seems to help the use case where `ecm.Each` is being used on a large(r) set of components that match a lot of entities.

We could also change [EntityComponentManagerPrivate::views](https://github.com/ignitionrobotics/ign-gazebo/blob/ignition-gazebo3_3.8.0/src/EntityComponentManager.cc#L110-L111) to an unordered map, but this would require some more work/thought because [the key of this map is actually a `std::set`](https://github.com/ignitionrobotics/ign-gazebo/blob/ignition-gazebo3_3.8.0/include/ignition/gazebo/detail/View.hh#L37-L38). I don't think we should be using a `std::set` as the key of a map, but this seems to be necessary due to some limitations in our current implementation of views (this leads me to my next point that we need to fix the current design of views - see the next paragraph). 

The real reason why this change does not help the performance of `ecm.Each` much is because of how views are currently being used. See https://github.com/ignitionrobotics/ign-gazebo/issues/711#issuecomment-816890851 for more information - I plan to address what is mentioned in this comment next.

_Also, on a related note: I believe that changing https://github.com/ignitionrobotics/ign-gazebo/blob/ignition-gazebo3_3.8.0/include/ignition/gazebo/detail/View.hh#L37-L38 to a `std::unordered_set` may help a bit with performance, but I don't believe this can be done in an existing release because there are methods in the ECM (like [FindView](https://github.com/ignitionrobotics/ign-gazebo/blob/ignition-gazebo3_3.8.0/include/ignition/gazebo/EntityComponentManager.hh#L681-L689)) that use this `std::set`._

**Edit:**
I'd like to give one other perspective as to why the changes here don't result in much of an improvement. If we look at the [callback loop](https://github.com/ignitionrobotics/ign-gazebo/blob/ignition-gazebo3_3.8.0/include/ignition/gazebo/detail/EntityComponentManager.hh#L371-L379) in `ecm.Each` again, the time complexity can be calculated as:
```
O(E * (C * L))
```
The variables are defined as follows:
* `E`: the number of entities (or number of `for` loops)
* `C`: the number of components
* `L`: the lookup time for a component in a view
* `C * L`: the time complexity for `view.Component<ComponentTypeTs>(entity, this)...`

Before this PR, `L` was `O(log(n))` since `std::map`s were used. After this PR, `L` is now `O(1)`. It's worth analyzing the time complexity for two cases:
* We have a lot of entities and few components in an `ecm.Each` call. The complexity is now:
```
O(E * (C * L)) = O(E * (C * 1)) = O(C * E)
```
Before this PR, the time complexity for this case would be `O(C * E * log(E))`.

* The number of components is large (i.e., approaching the number of entities), the time complexity is now:
```
O(E * (E * 1)) = O(E^2)
```
Before this PR, the time complexity for this case would be `O(E^2 * log(E))`. 

As we can see in this analysis, the real issue is that we still have to spend time finding each component in the view (as mentioned in https://github.com/ignitionrobotics/ign-gazebo/issues/711#issuecomment-816890851) - the time complexity for the `ecm.Each` callback loop should really be `O(E)` (all components should be cached at the time of view creation, which would mean that `C * L` is no longer part of the equation for time complexity). If anyone believes that my analysis here may be incorrect, feel free to mention what I am missing, and I will update this accordingly.

## Checklist
- [x] Added tests

### DIFF
--- a/include/ignition/gazebo/detail/ComponentStorageBase.hh
+++ b/include/ignition/gazebo/detail/ComponentStorageBase.hh
@@ -17,7 +17,7 @@
 #ifndef IGNITION_GAZEBO_DETAIL_COMPONENTSTORAGEBASE_HH_
 #define IGNITION_GAZEBO_DETAIL_COMPONENTSTORAGEBASE_HH_
 
-#include <map>
+#include <unordered_map>
 #include <utility>
 #include <vector>
 #include "ignition/gazebo/components/Component.hh"
@@ -212,7 +212,7 @@ namespace ignition
       private: ComponentId idCounter = 0;
 
       /// \brief Map of ComponentId to Components (see the components vector).
-      private: std::map<ComponentId, int> idMap;
+      private: std::unordered_map<ComponentId, int> idMap;
 
       /// \brief Sequential storage of components.
       public: std::vector<ComponentTypeT> components;
--- a/include/ignition/gazebo/detail/View.hh
+++ b/include/ignition/gazebo/detail/View.hh
@@ -17,9 +17,9 @@
 #ifndef IGNITION_GAZEBO_DETAIL_VIEW_HH_
 #define IGNITION_GAZEBO_DETAIL_VIEW_HH_
 
-#include <map>
 #include <set>
 #include <string>
+#include <unordered_map>
 #include <utility>
 #include "ignition/gazebo/components/Component.hh"
 #include "ignition/gazebo/Entity.hh"
@@ -126,9 +126,32 @@ class IGNITION_GAZEBO_VISIBLE View
   /// \brief List of entities about to be removed
   public: std::set<Entity> toRemoveEntities;
 
+  /// \brief Hash functor for std::pair<Entity, ComponentTypeId>
+  public: struct Hasher
+          {
+            std::size_t operator()(
+                std::pair<Entity, ComponentTypeId> _pair) const
+            {
+              _pair.first ^= _pair.second + 0x9e3779b9 + (_pair.second << 6)
+                 + (_pair.second >> 2);
+              return _pair.first;
+            }
+          };
+
+  /// \brief Equality functor for std::pair<Entity, ComponentTypeId>
+  public: struct EqualTo
+          {
+            bool operator()(const std::pair<Entity, ComponentTypeId> &_lhs,
+                const std::pair<Entity, ComponentTypeId> &_rhs) const
+            {
+              return _lhs.first == _rhs.first &&
+                _lhs.second == _rhs.second;
+            }
+          };
+
   /// \brief All of the components for each entity.
-  public: std::map<std::pair<Entity, ComponentTypeId>,
-          ComponentId> components;
+  public: std::unordered_map<std::pair<Entity, ComponentTypeId>, ComponentId,
+            Hasher, EqualTo> components;
 };
 /// \endcond
 }


adlarkin:
When running tests locally, I noticed that the [`Unit_Gui_TEST`](https://github.com/ignitionrobotics/ign-gazebo/blob/13ac7ff493583f11b9a0c4dfea9e8ae8220e2585/src/gui/Gui_TEST.cc) segfaults on shutdown. It appears to be related to the [unordered map that I introduced in `View`](https://github.com/ignitionrobotics/ign-gazebo/blob/13ac7ff493583f11b9a0c4dfea9e8ae8220e2585/include/ignition/gazebo/detail/View.hh#L152-L154) (if I revert this back to a `std::map`, the segfault disappears). I'm not sure why there's a crash on shutdown though, because I did some debugging and verified that all checks in the GUI unit test pass. Any thoughts?

adlarkin:
ABI checker is failing - I'm not sure if this is a false positive or not, since the files I am modifying are in the `include/.../detail` directory.

iche033:
hmm I know we don't promise ABI/API compatibility for the headers in `detail` directory and users should be using the functions in these classes but I wonder if this would cause any undefined behavior for users linking to our libraries? cc @j-rivero

iche033:
> After running some tests, it appears that the performance gains from this change are minimal, and in some cases, not noticeable at all. This change seems to help the use case where ecm.Each is being used on a large(r) set of components that match a lot of entities.

Do you have some rough numbers from profiling? I'm mainly gauging to see if it's worth risking breaking ABI to get these performance improvements.

adlarkin:
> Do you have some rough numbers from profiling? I'm mainly gauging to see if it's worth risking breaking ABI to get these performance improvements.

I tested 3000 non-static simple shapes, iterating through all of the models and extracting some components from them. The profiling time for the `ecm.Each` call before this PR was 4.2ms, and after this PR was 2.9ms. So again, not that much - just a little more than 1ms gain for a world with A LOT of shapes. For worlds with fewer shapes, the improvements will be less noticeable I believe.

> hmm I know we don't promise ABI/API compatibility for the headers in detail directory and users shouldn't be using the functions in these classes but I wonder if this would cause any undefined behavior for users linking to our libraries?

I'm not sure about this either. I originally thought that these changes would have to target fortress, but @chapulina believes that they could land in citadel for the reasons you stated (not promising ABI/API for headers in the detail directory). We can wait to see what @j-rivero says about the linking.

j-rivero:
> We can wait to see what @j-rivero says about the linking.

Umm, downloaded the latest version of `ignition-gazebo5` and run:
```bash
include/ignition/gazebo on  main via 喝 v3.10.2 ❯ nm -DC /usr/lib/x86_64-linux-gnu/libignition-gazebo5.so.5 | grep View::AddComponent
00000000001e42f0 T ignition::gazebo::v5::detail::View::AddComponent(unsigned long, unsigned long, int)
```

Somehow the binary symbol for the detail class is there. Might be wrong but looks like an valid ABI breakage to me.

iche033:
> Somehow the binary symbol for the detail class is there. Might be wrong but looks like an valid ABI breakage to me.

hmm sounds like we won't be able to land this in a released version and should probably also ticket an issue about hiding symbols of classes in `detail`. 

the perf improvement from 4.2s to 2.9s is actually good in my opinion. We could target this to F or wait till you find a better solution with caching data in Views.

adlarkin:
> ticket an issue about hiding symbols of classes in detail

Done: #759

> We could target this to F or wait till you find a better solution with caching data in Views

I went ahead and re-targeted this PR to Fortress, and also added some actual performance numbers/comparisons in the PR description. Based on the numbers I'm seeing in my tests/comparisons, I think that this change could be worthwhile for worlds with a lot of entities. If you wouldn't mind reading the updated PR description and letting me know what you think, that would be great!

adlarkin:
This PR is probably no longer needed since #856 is now open. Once #856 is merged, we can close this.

adlarkin:
I'm closing this since #856 has been merged.

PR #758, opened 2021-04-12
4 to 5 20210412

nkoenig:
# ➡️ Forward port

Port ign-gazebo4 to ign-gazebo5

Branch comparison: https://github.com/ignitionrobotics/ign-gazebo/compare/ign-gazebo5...ign-gazebo4

Requires release of https://github.com/osrf/sdformat/pull/539


src/rendering/SceneManager.cc
```diff
@@ -1298,6 +1298,24 @@ rendering::ParticleEmitterPtr SceneManager::UpdateParticleEmitter(Entity _id,
...
+    for (int i = 0; i < _emitter.header().data_size(); ++i)
+    {
+      const auto &data = _emitter.header().data(i);
+      const std::string key = "particle_scatter_ratio";
+      if (data.key() == "particle_scatter_ratio" && data.value_size() > 0)
+      {
+        // \todo(anyone) switch to use the follow API when merging forward to
+        // edifice
+        // emitter->SetParticleScatterRatio(math::parseFloat(data.value(0)));
+        emitter->SetUserData(key, math::parseFloat(data.value(0)));
```
adlarkin, line 1313: Can we go ahead and make this change in this PR before merging the forward port? This API should now be supported thanks to https://github.com/ignitionrobotics/ign-rendering/pull/275

cc @iche033

chapulina: Please take a look at the current version and let me know if this is how you peeps were planning to handle this.

adlarkin: Looking at https://github.com/ignitionrobotics/ign-rendering/pull/275, I'm fairly confident that this change needs to be made in this forward port in order to ensure proper scatter ratio functionality in `ign-gazebo5`. If you look at the changes made in https://github.com/ignitionrobotics/ign-rendering/pull/275, you'll notice that the `Ogre2ParticleNoiseListener` now gets the particle scatter ratio by calling `emitterPtr->ParticleScatterRatio()` instead of `ogreVisual->UserData("particle_scatter_ratio")`.

So, making the change that is commented in this TODO note should set the particle scatter ratio correctly for when the `Ogre2ParticleNoiseListener` retrieves it via `emitterPtr->ParticleScatterRatio()` later on.

@chapulina, it looks like you went ahead and made this change a few hours ago, right? Thanks :bow: I did notice that you left the `SetUserData` call:
https://github.com/ignitionrobotics/ign-gazebo/blob/2fc9b6f07d9ae16237a142189a8902dab2aab390/src/rendering/SceneManager.cc#L1310-L1312

Can we completely remove the `SetUserData` call, or is it needed for backwards compatibility of some sort? I'm pretty sure the `SetUserData` call is no longer needed.

chapulina: > Can we completely remove the SetUserData call, or is it needed for backwards compatibility of some sort?

Yup, the reasoning is that this is data that any plugin could be looking at, not only our plugins. So we keep the data around during one more major release to provide a tick-tock transition to users.

But admittedly, I haven't looked closely into the particle implementation.

adlarkin: > the reasoning is that this is data that any plugin could be looking at, not only our plugins

That makes sense :+1: Let's just make sure we don't forget to remove it when it's time to forward-port to `ign-gazebo6` (it can be easy to glance over `TODO` note(s) when checking forward ports, especially if they're very large :grimacing:)

> I haven't looked closely into the particle implementation.

I haven't been super involved with the particle emitter implementation, but have reviewed a few PRs for it. I understand it well enough to confidently say that for usage with `ign-rendering` at least, calling `SetUserData` here is no longer needed thanks to the new API in `ign-rendering`. So, I think that what you've done here (make use of the new API and leave `SetUserData` in for now, but deprecate it in case we have downstream users elsewhere using this) is the correct fix, but @iche033 would you mind confirming as a sanity check?

chapulina:
Needs `ign-physics` 4.1.0 for link velocity tests to pass: https://github.com/ignitionrobotics/ign-physics/pull/255

chapulina:
@osrf-jenkins run tests please :crossed_fingers:

chapulina:
No more upstream releases needed. Now I'm just fixing the Windows build

src/rendering/SceneManager.cc
iche033: oh calling `SetUserData` with `particle_scatter_ratio` will no longer have any effect. It was removed as soon as we added the new `SetParticleScatterRatio` API:
https://github.com/ignitionrobotics/ign-rendering/pull/275/files#diff-fd78319e07767eb0a94e57f99e5460cd1dea217968dbe4191768feaf19d8e1adL108

I don't think users should be relying on custom key-value pairs stored in UserData. We didn't document it or put it in the tutorials as it's more for internal use. They will probably only find out if they look at the source code.

chapulina: Thanks for the context, I imagined it wouldn't have any effect internally anymore.

> I don't think users should be relying on custom key-value pairs stored in UserData. 

We have been using some user data as sort of a public API, like `gazebo-entity` and `pause-update`. These are being used across plugins, and I don't think we'll move them to `ign-rendering` APIs because they're Gazebo-speciifc.

Do you have a suggestion on how to work around this? I can imagine creating a `GazeboNode` class that extends `rendering::Node` and adds Gazebo-specific APIs.

> They will probably only find out if they look at the source code.

We should keep in mind that people use the official plugins as examples to develop their own plugins. I wouldn't be surprised if there are people relying on `gazebo-entity` for example.

---

In summary, I think we should be mindful about this, but I agree that this sounds like a corner case that most likely no one is using, so I'll just remove it :+1:

chapulina: Related, are there plans to add a `scatter_ratio` field to the particle message so this isn't stored in the header anymore?

chapulina (approved):
Windows is building, I haven't spotted any test regressions and all comments have been addressed.

PR #777, opened 2021-04-20
Patch particle emitter2 service

nkoenig:
# 🦟 Bug fix

## Summary

Requesting the particle emitter service from a websocket server which is also managing a gazebo instance causes the service handler to block simulation. As a result, the condition variable times out and particle emitter information is not returned. 

This PR caches particle emitter information during the `EachNew`, removing the blocking `wait_for`. 

## Checklist
- [ ] Added tests
- [X] All tests passed (See [test coverage](https://ignitionrobotics.org/docs/all/contributing#test-coverage))


src/systems/particle_emitter2/ParticleEmitter2.cc
```diff
@@ -100,20 +93,9 @@ bool ParticleEmitter2Private::EmittersService(
   _res.Clear();
 
   // Lock and wait for an iteration to be run and fill the state
```
iche033, line 95: I think this comment no longer applies?

src/systems/particle_emitter2/ParticleEmitter2.cc
```diff
@@ -160,71 +142,71 @@ void ParticleEmitter2::PreUpdate(const ignition::gazebo::UpdateInfo &_info,
...
+          // correctly.
+          this->dataPtr->emitterTopicMap[topic] = _entity;
+
+          // Store the emitter information in the service message, which
+          // can then be used in the particle_emitters service.
           msgs::ParticleEmitter *emitterMsg =
             this->dataPtr->serviceMsg.add_particle_emitter();
           emitterMsg->CopyFrom(_emitter->Data());
+          math::Pose3d pose = worldPose(_entity, _ecm);
+          msgs::Set(emitterMsg->mutable_pose(), pose);
```
iche033, line 193: initially I thought since particle emitter is now s on the same level as sensor, visual and collision, the pose field would be in parent frame so that they are consistent. Is it in world frame here because the particle emitter is missing a parent id field (and so the subscribers can't compute the pose themselves in world frame)?

german-e-mas:
I used it alongside https://gitlab.com/ignitionrobotics/web/app/-/merge_requests/117 and https://gitlab.com/ignitionrobotics/web/gz3d/-/merge_requests/48 and it works as expected.

src/systems/particle_emitter2/ParticleEmitter2.cc
nkoenig: Removed in d57cfa204f4f713031e844edda0a6f003940f60b

src/systems/particle_emitter2/ParticleEmitter2.cc
nkoenig: Using relative pose and added frame information in d57cfa204f4f713031e844edda0a6f003940f60b

iche033 (approved):

PR #787, opened 2021-04-26
Recommend usage of SdfEntityCreator in Fortress

adlarkin:

## Summary
As discussed offline, the usage of the `SdfEntityCreator` should be ~~_required_~~ _strongly encouraged_ beginning in `Fortress` in order to properly handle new features related to nested models.

## Checklist
- [ ] Added tests

### DIFF
--- a/Migration.md
+++ b/Migration.md
@@ -5,6 +5,12 @@ Deprecated code produces compile-time warnings. These warning serve as
 notification to users that their code should be upgraded. The next major
 release will remove the deprecated code.
 
+## Ignition Gazebo 5.x to 6.x
+
+* Entity creation is now required to be done through the
+  `ignition::gazebo::SdfEntityCreator`. In earlier versions of Ignition Gazebo,
+  usage of this class for entity creation is optional, but strongly encouraged.
+
 ## Ignition Gazebo 4.x to 5.x
 
 * Use `cli` component of `ignition-utils1`.


Migration.md
```diff
@@ -5,6 +5,12 @@ Deprecated code produces compile-time warnings. These warning serve as
 notification to users that their code should be upgraded. The next major
 release will remove the deprecated code.
 
+## Ignition Gazebo 5.x to 6.x
+
+* Entity creation is now required to be done through the
+  `ignition::gazebo::SdfEntityCreator`. In earlier versions of Ignition Gazebo,
+  usage of this class for entity creation is optional, but strongly encouraged.
```
chapulina, line 12: I think it's tough to make this statement without providing users the full context. Someone may come and ask what is the `EntityComponentManager::CreateEntity` function for then; should it be made private and a friend of the `SdfEntityCreator` so no system can use it, but the `SdfEntityCreator` can?

I think the message we'd like to convey is that there's functionality in the `Physics` plugin that relies on entities being created by the `SdfEntityCreator`, specifically because of their order of creation, and that the plugin may misbehave for entities created directly through the ECM's API.

We may even want to document that on the `Physics` system's header.

adlarkin: > Someone may come and ask what is the `EntityComponentManager::CreateEntity` function for then; should it be made private and a friend of the `SdfEntityCreator` so no system can use it, but the `SdfEntityCreator` can?

As I look closer at the `SdfEntityCreator` API, I think that we would need to modify/expand its API a bit if we want to push users towards using it. There appear to be `CreateEntities` methods for various `sdf` tags (sensor, model, etc.), but I also think that we should have a method that creates an "empty" entity: something like this:

```c++
/// \brief Create an entity
/// \brief _parent The entity's parent
/// \return The newly created entity
Entity CreateEntity(const Entity _parent = kNullEntity);
```

That way, for things like system plugins, users can create additional entities as needed (I had to do something similar to this in the [logical audio sensor plugin](https://github.com/ignitionrobotics/ign-gazebo/blob/ignition-gazebo5_5.0.0/src/systems/logical_audio_sensor_plugin/LogicalAudioSensorPlugin.cc#L480-L501) - I'm not sure if what I did in this plugin is actually "best practices" or not). In terms of how this method is implemented, perhaps we can do something like have it call `EntityComponentManager::CreateEntity` if it's a friend of the `SdfEntityCreator`. This method would also take care of setting the entity's parent, so users can use it like this:

```c++
// this is the way users may currently be creating entities
auto entity = ecm.CreateEntity();
sdfEntityCreator.SetParent(entity, parent);

// users should now do this instead
auto entity = sdfEntityCreator.CreateEntity(parent);

// they can also do something like this if the entity has no parent
auto entity = sdfEntityCreator.CreateEntity();
```

Thoughts?

chapulina: > we should have a method that creates an "empty" entity

As the name says, the `SdfEntityCreator` is concerned with creating entities from SDF structures :wink: Empty entities should be created directly with the ECM. It's important to keep the separation of concerns among the classes, I'm not sure what's would be the advantage of creating empty entities from the `SdfEntityCreator`.

adlarkin: All good points, and you're right (again :slightly_smiling_face:). It seems to me like for _advanced_ use cases, users may need to create entities and components with the ECM (i.e., they can't just rely on the `SdfEntityCreator`). So, with that being said, I think that we can't _require_ the `SdfEntityCreator` to be used since some scenarios require making entities that don't come from `SDF`. I think that the best thing to do is to encourage the usage of `SdfEntityCreator` whenever possible, including how to use the `SdfEntityCreator` (this is related to #628).

adlarkin: Also, I have updated the PR title and description to indicate that `SdfEntityCreator` usage is _recommended_, but not _required_.

chapulina: >  Empty entities should be created directly with the ECM.

As discussed during the meeting, we could also provide APIs to create entities from the `Model`, `Link` and `World` classes. We should be trying to add more functions to those classes so that beginners can use them, while advanced users can use the ECM for custom entities.

> Does simply re-wording the migration guide to discuss these items sound good enough to you? Or do we need to add more documentation elsewhere (and potentially modify/document source code) as well?

I think that this, combined with the documentation that will come out of #628, should provide lots of pointers to users.

adlarkin: It seems to me like updating the migration guide here and merging this PR may be difficult to do before addressing #628 and #805. Would it be best to mark this PR as a draft, address these other issues first, and then update the migration guide accordingly based on the documentation created by these other issues?

adlarkin:
I'm marking this as a draft for now (and moving this from `in review` to `to do` for the Ignition core development board) based on [this conversation](https://github.com/ignitionrobotics/ign-gazebo/pull/787#discussion_r630270558). I'm planning to re-visit this once #628 and #805 are addressed.

chapulina:
I talked to @adlarkin offline, and we decided to close this PR in favor of:

* Documenting on the `Physics` system that it won't function correctly unless entities are created in the proper order. That is, parents must be created before their children.
* Write a tutorial with examples of how to create entities using the create service (recommended way) or the `SdfEntityCreator` / `EntityComponentManager` APIs (advanced usage, and users should be careful about the order in which entities are created).

PR #788, opened 2021-04-26
explain why detail::View symbols are visible

### LINKED ISSUE #759: Symbols of detail::View class are public
When working on #752, it became apparent that [the `detail::View` class has public visibility](https://github.com/ignitionrobotics/ign-gazebo/blob/ignition-gazebo5_5.0.0/include/ignition/gazebo/detail/View.hh#L48). This makes it difficult to change things in this class since it will break ABI. Considering the fact that `detail` classes were never intended to be used by external users, we should probably make the visibility of this class hidden. I believe that this could be as simple as switching `detail::View`'s visibility to `IGNITION_GAZEBO_HIDDEN` for Fortress, but as @iche033 pointed out, we will need to test this to ensure that other platforms and libraries aren't broken by this change.

@chapulina, what do you think about this? Are there any reasons why we should keep the visibility of `detail::View` (and anything else in `detail`) public, or are we okay to make the change to hidden?

adlarkin:

# 🦟 Bug fix

Fixes #759

## Summary
~~This hides the symbols of the `detail::View` class so that we can make changes to it as needed without breaking ABI. Since this is an ABI-breaking change, this has to be done in fortress.~~

**Edit:** the symbol visibility for `detail::View` cannot be hidden. See the comments below for an explanation.

## Checklist
- [ ] Added tests

### DIFF
--- a/include/ignition/gazebo/detail/View.hh
+++ b/include/ignition/gazebo/detail/View.hh
@@ -45,6 +45,11 @@ using ComponentTypeKey = std::set<ComponentTypeId>;
 /// use a cache to improve performance. The assumption is that entities
 /// and the types of components assigned to entities change infrequently
 /// compared to the frequency of queries performed by systems.
+///
+/// Note that symbols for this class are visible because methods from this class
+/// are used in templated Ignition::Gazebo::EntityComponentManager methods.
+/// However, users should not use this class (or anything else in namespace
+/// ignition::gazebo::detail) directly.
 class IGNITION_GAZEBO_VISIBLE View
 {
   /// Get a pointer to a component for an entity based on a component type.


scpeters:
linking error in Ubuntu:

~~~
undefined reference to `ignition::gazebo::v6::detail::View::AddComponent(unsigned long, unsigned long, int)'
~~~

[Build Status](https://build.osrfoundation.org/job/ignition_gazebo-ci-pr_any-ubuntu_auto-amd64/6028/) https://build.osrfoundation.org/job/ignition_gazebo-ci-pr_any-ubuntu_auto-amd64/6028/

maybe the symbols need to be visible?

adlarkin:
> maybe the symbols need to be visible?

So, after looking into the linking errors and chatting with @mjcarroll offline, I believe that the symbols for `detail::View` need to be visible. The reason for this is because when users call [`EntityComponentManager::Each`](https://github.com/ignitionrobotics/ign-gazebo/blob/ignition-gazebo5_5.0.0/include/ignition/gazebo/detail/EntityComponentManager.hh#L357-L395), [`View::Component`](https://github.com/ignitionrobotics/ign-gazebo/blob/ignition-gazebo5_5.0.0/include/ignition/gazebo/detail/View.hh#L50-L75) is called, which then calls [`View::ComponentImplementation`](https://github.com/ignitionrobotics/ign-gazebo/blob/ignition-gazebo5_5.0.0/include/ignition/gazebo/detail/View.hh#L107-L115). Since we are using templates for things like `Each`, this means that the templates are expanded where it's being used, so the `detail::View` class needs to be `visible` in order for downstream users to be able to call `ecm.Each`. There's also at least one other method that has this same issue (template method that uses a view in the definition): [`EntityComponentManager::AddComponentsToView`](https://github.com/ignitionrobotics/ign-gazebo/blob/ignition-gazebo5_5.0.0/include/ignition/gazebo/detail/EntityComponentManager.hh#L468-L516)

So, unless we can hide the templates somehow, then I believe we will need to keep `detail::View` symbols as visible, which is probably why they were never hidden in the first place.

@chapulina what do you think? I think we should probably close this PR and also close #759, with a note explaining why we can't hide the symbol visibility. Perhaps it would also be worth adding some documentation to the `detail::View` class, explaining why it's visible and not hidden.

chapulina:
> @chapulina what do you think? 

Ouch, ok, it makes sense. I thought that we shouldn't need to make the symbols visible because they were headers. Gotta learn more about symbol generation.

I guess then that we can't ever make ABI breaking changes to `detail`. So even though users shouldn't be using them directly, we'll be keeping them stable. This is something good to know, and it means that our message to users can be simpler: "_Anything that's installed will be kept ABI / API stable, regardless of whether they're in a `detail` folder._"

adlarkin:
> I guess then that we can't ever make ABI breaking changes to `detail`

I don't think this statement is 100% true. If you look at [`ComponentStorageBase.hh`](https://github.com/ignitionrobotics/ign-gazebo/blob/ignition-gazebo5_5.0.0/include/ignition/gazebo/detail/ComponentStorageBase.hh), both the `ComponentStorageBase` and `ComponentStorage` classes are hidden, and work just fine since these classes aren't used in template methods elsewhere (i.e., these classes are completely self-contained in their respective library). So, I think it's more accurate to say that we can't make ABI breaking changes to _some  things_ in `detail`, which means that there's still a chance of API/ABI breaking changes for some things in `detail`.

I went ahead and reverted back to `visible` symbols in `detail::View` and added a note that explains why `detail::View` needs to be visible in https://github.com/ignitionrobotics/ign-gazebo/pull/788/commits/f6606d782ad80f92df528c582ee703cf68a90ad4. I also updated the PR title. If this looks good to you, I think we can merge this PR and then close #759.

chapulina (approved):
> these classes aren't used in template methods elsewhere

Interesting, I think I'm starting to understand. It's very nuanced, something to be careful about when using these classes on our own headers too.

PR #792, opened 2021-04-26
3 ➡️ 4

chapulina:
# ➡️ Forward port

Port `ign-gazebo3` to `ign-gazebo4`

Branch comparison: https://github.com/ignitionrobotics/ign-gazebo/compare/ign-gazebo4...ign-gazebo3

### DIFF
--- a/src/systems/velocity_control/VelocityControl.cc
+++ b/src/systems/velocity_control/VelocityControl.cc
@@ -15,9 +15,11 @@
  *
  */
 
+#include <map>
 #include <mutex>
 #include <string>
 #include <vector>
+#include <unordered_map>
 
 #include <ignition/common/Profiler.hh>
 #include <ignition/math/Vector3.hh>
@@ -37,9 +39,14 @@ using namespace systems;
 
 class ignition::gazebo::systems::VelocityControlPrivate
 {
-  /// \brief Callback for velocity subscription
+  /// \brief Callback for model velocity subscription
   /// \param[in] _msg Velocity message
   public: void OnCmdVel(const ignition::msgs::Twist &_msg);
+
+  /// \brief Callback for link velocity subscription
+  /// \param[in] _msg Velocity message
+  public: void OnLinkCmdVel(const ignition::msgs::Twist &_msg,
+    const ignition::transport::MessageInfo &_info);
 
   /// \brief Update the linear and angular velocities.
   /// \param[in] _info System update information.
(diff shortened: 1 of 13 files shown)


scpeters (approved):
timeout of `UNIT_Server_TEST` on bionic workflow; I assume it is flaky

src/systems/velocity_control/VelocityControl.cc
```diff
@@ -88,19 +117,47 @@ void VelocityControl::Configure(const Entity &_entity,
...
+    "/model/" + this->dataPtr->model.Name(_ecm) + "/cmd_vel");
+  auto modelTopic = validTopic(modelTopics);
   this->dataPtr->node.Subscribe(
-    topic, &VelocityControlPrivate::OnCmdVel, this->dataPtr.get());
-
-  ignmsg << "VelocityControl subscribing to twist messages on [" << topic << "]"
+    modelTopic, &VelocityControlPrivate::OnCmdVel, this->dataPtr.get());
+  ignmsg << "VelocityControl subscribing to twist messages on ["
+         << modelTopic << "]"
          << std::endl;
```
adlarkin, line 133: This seems to be slightly different from the `diff` I am seeing here: https://github.com/ignitionrobotics/ign-gazebo/compare/ign-gazebo4...ign-gazebo3

It looks like multiple topics are now being supported instead of just one. Is this because of new features in `ign-gazebo4` that are to be supported, while being backwards compatible with `ign-gazebo3` functionality?

chapulina: I believe the valid topic was just implemented differently on each branch, getting the similar results.

On #427, targeted at `ign-gazebo3`, this pattern was used:

```
modelTopic = defaultTopic
if (sdfTopic)
  modelTopic = sdfTopic
modelTopic = valid(modelTopic)
```

On #522, targeted at `ign-gazebo4`, the following pattern was used. This is slightly better because it will fallback to a valid default topic if the SDF topic can't be made valid (warning as needed).

```
topics = {sdfTopic, defaultTopic}
topic = firstValidTopic(topics)
```

So as we merge forward, I kept the approach that was already on `ign-gazebo4`, while changing the topic name to `modelTopic` to differentiate from the new `linkTopic`.

adlarkin (approved):

src/systems/velocity_control/VelocityControl.cc
chapulina: That function itself can be backported, but it's of little value on `ign-gazebo3` because the `Name` component only supports spaces from Dome, see https://github.com/ignitionrobotics/ign-gazebo/pull/244

PR #793, opened 2021-04-27
Run server and client in the same process

ahcorde:

# 🎉 New feature

## Summary

Related to this issue https://github.com/ignitionrobotics/ign-gazebo/issues/556

## Test it

```bash
ign gazebo -r -v 4 --same-process shapes.sdf
ign gazebo -r -v 4 --same-process camera_sensor.sdf  # -> crash
```

## Checklist
- [ ] Added tests


ahcorde:
I made a comment in the related issue https://github.com/ignitionrobotics/ign-gazebo/issues/556#issuecomment-827510853

I want to make it work then I will clean up te code.

ahcorde:
I updated the base PR to match with the recent refactoring in the scene3D.

Current status
 - Gui and sensor thread shared the same scene. 
    - not able to visualize the objects in the scene 3D
    - When I try to load the ImageDisplay then I get a crash
 - Ogre crashes when we try to use the wheel to zoom in and zoom out. In particular in the method `OgreRayQuery::ClosestPoint`. The vertex from the meshes are not fetched properly.
 - Ogre2 crashes on start in the method `void Ogre2RenderTarget::Render()` in the call `engine->OgreRoot()->renderOneFrame();`

to reproduce:

```bash
/home/ahcorde/ignition_fortress_leak/install/bin/ign gazebo --gui-config src/ign-gazebo/src/gui/gui.config camera_sensor.sdf -v 4 --same_process --render-engine ogre --render-engine-gui ogre --render-engine-server ogre
```

If you want to try Ogre two, you should modify the gui.config file and replace `ogre` with `ogre2`

```bash
/home/ahcorde/ignition_fortress_leak/install/bin/ign gazebo --gui-config src/ign-gazebo/src/gui/gui.config camera_sensor.sdf -v 4 --same_process --render-engine ogre2 --render-engine-gui ogre2 --render-engine-server ogre2
```

## Limitations

everything should run with ogre or ogre2 but we cannot mix both renders when gui and server are running in the same process.

ahcorde:
I have two issues right now:

 - Sometimes ign-gazebo starts with some repeated entities in the Entity tree

Selección_102

 - When the entity is removed from the entity tree, it removed from the scene but not from the entity tree

same_process_remove_entity

ahcorde:
# Decisions that I have made

There are two differences main way to launch ign-gazebo: with sensors and without sensors. This have some implications. 

## When we are runninig with sensors

There are two RenderUtils available: One in the sensor system plugin and another in the GUI. This is the easiest case because there are two system plugins that remove entities: Sensors and Physics which runs at the same speed. But the Scene 3D is running just at 30Hz. This may cause some issues in the Entity tree plugin. Some of the EachRemoved cycles are lost. I have created some events to be able to remove/add entities in the Entity tree plugin, this event are emitted in the renderutils class.

 - RemoveFromECM
 - AddToECM

The GzSceneManager should not update renderutil because the sensor thread will do it.

## When running without sensors

In this case the physics system plugin will be the only one with access to the removeEach calls. In this case this plugin will send an event called UpdateGUIECM to avoid missing any removals. 

There is another event called EnableSensors this event is used `GzSceneManager`. In this case the sensor system plugin is not updating the RenderUtils because it not running, so this event allow us to check if we should update the Scene 3D from the Gui system plugin.

 - EnableSensors
 - UpdateGUIECM

# Render

On of the limitiation that we have, it is that we can only update the render scene in the same thread, otherwise we will get a crash. When running in the same process there are two Renderutils trying to update the scene3d in  different threads, it's important to load in the first time the Scene3D to set the option `useCurrentGLContext` to 1. For this reason we need to create a event called Render, this event is emitted in the `GzSceneManager` plugin every time that an event Render is received in the eventFilter method.

-----

# Notes

 - Run the world without the sceneBroadcaster when using the `--same_process` flag
 - We can use Ogre or Ogre2 at the same time, we cann't mix both render engines when same process is activated.

ahcorde:
Recommended branch of ign-gui https://github.com/ignitionrobotics/ign-gui/pull/new/ahcorde/6/new_scene

src/systems/physics/Physics.cc
```diff
@@ -1132,6 +1141,9 @@ void PhysicsPrivate::RemovePhysicsEntities(const EntityComponentManager &_ecm)
       [&](const Entity &_entity, const components::Model *
           /* _model */) -> bool
       {
+        if (this->eventManager)
```
azeey, line 1144: Can the GUI not also use `EachRemoved` instead of relying on this event being emitted?

iche033:
With the latest changes, the events for keeping GUI plugins in sync with server side seems to be working from testing. 

I also want to brainstorm ideas on how we could reuse more of the current ECM infrastructure. I think the problem all came down to the GuiRunner running the gui plugins at 30 Hz in a separate thread. What do you think of the following proposal:

* Emit an event when SimulationRunner updates the systems
https://github.com/ignitionrobotics/ign-gazebo/blob/cc4cf71f45836ad6478ab8c69af5dc4a19201d5d/src/SimulationRunner.cc#L523-L527
* In GuiRunner same process mode, we can connect to this event instead of running its [own loop](https://github.com/ignitionrobotics/ign-gazebo/blob/cc4cf71f45836ad6478ab8c69af5dc4a19201d5d/src/gui/GuiRunner.cc#L114-L123).
* In call back we can call `UpdatePlugins`. We can also throttle it to some rate, e.g. 30Hz. by checking time since last update call. 
*  In order not to miss important events like entities removed / added, we can use ECM's `HasNewEntities, HasEntitiesMarkedForRemoval, HasOneTimeComponentChanges` functions in the callback and check if we need bypass throttling and call `UpdatePlugins` 
https://github.com/ignitionrobotics/ign-gazebo/blob/cc4cf71f45836ad6478ab8c69af5dc4a19201d5d/include/ignition/gazebo/EntityComponentManager.hh#L493-L502
 
The idea above is so that the gui plugins can keep using the `EachRemoved` call without having to connect to new events. This reduces the effort required to port other gui plugins (or user's own custom gui plugins) to work in the same process mode.

ahcorde:
@iche033, your proposal it's much more simpler https://github.com/ignitionrobotics/ign-gazebo/pull/793/commits/5521b407e41376643d2064a79643f52673c73f4b

src/systems/sensors/Sensors.cc
```diff
@@ -362,6 +411,19 @@ Sensors::~Sensors()
   this->dataPtr->Stop();
 }
 
+void Sensors::SetSameProcess(bool _sameProcess)
```
iche033, line 414: one idea is to add a `EntityComponentManager::SameProcess()` function so we could access this property from all class. Maybe also a bit hacky though.

src/gui/plugins/scene_manager/GzSceneManager.cc
```diff
@@ -81,9 +122,12 @@ void GzSceneManager::Update(const UpdateInfo &_info,
     EntityComponentManager &_ecm)
 {
   IGN_PROFILE("GzSceneManager::Update");
-
-  this->dataPtr->renderUtil.UpdateECM(_info, _ecm);
-  this->dataPtr->renderUtil.UpdateFromECM(_info, _ecm);
+  if (this->dataPtr->emitFirstRender &&
+    (!this->dataPtr->sameProcess || !this->dataPtr->enableSensors))
```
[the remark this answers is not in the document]
ahcorde: Yes, you are right

src/systems/sensors/Sensors.cc
ahcorde: This is a new method of the class `ISystemConfigure`. All system plugins that override this method should get access to this data.

https://github.com/ignitionrobotics/ign-gazebo/blob/5521b407e41376643d2064a79643f52673c73f4b/include/ignition/gazebo/System.hh#L102-L105

ahcorde:
I made a quick test with the `3k_shapes.world`.

This is the data:

| | Two processes  | same process |
| ----- | ------ | ------- |
|mean | 0.008285111 |0.009956856|
|median | 0.0083 | 0.01 |
|min	 | 0.0075 | 0.0097 |
|max | 0.0085 |  0.0101 |

20% of improvement

Orage line is with ignition running in the same process, blue in two processes:

Figure_1

ahcorde:
@chapulina do you mind to take a look ?

adlarkin:
> I made a quick test with the `3k_shapes.world`.
>
> 20% of improvement
> 
> Orage line is with ignition running in the same process, blue in two processes:
> 
> Figure_1

Is the y-axis RTF (so, for example, I assume `.01` would be `1%` RTF), or something else?

src/ign.hh
```diff
@@ -76,4 +76,35 @@ extern "C" int runGui(const char *_guiConfig);
...
+/// \param[in] _logCompress --log-compress option
+/// \param[in] _playback --playback option
+/// \param[in] _physicsEngine --physics-engine option
+/// \param[in] _renderEngineServer --render-engine-server option
+/// \param[in] _renderEngineGui --render-engine-gui option
+/// \param[in] _file Path to file being loaded
+/// \param[in] _recordTopics Colon separated list of topics to record. Leave
+/// null to record the default topics.
+/// \param[in] _guiConfig Path to Ignition GUI configuration file.
+/// \return 0 if successful, 1 if not.
```
adlarkin, line 100: The implementation returns `-1` if not successful, should the docs be updated?

ahcorde:
@adlarkin Yes, you are right

src/rendering/RenderUtil.cc
```diff
@@ -1824,12 +2018,14 @@ void RenderUtilPrivate::RemoveRenderingEntities(
       });
 }
 
+static std::mutex mutexInit;
```
adlarkin, line 2021: Should this be a part of the `RenderUtil` class instead of being declared globally?

src/rendering/SceneManager.cc
```diff
@@ -175,8 +175,9 @@ rendering::VisualPtr SceneManager::CreateModel(Entity _id,
 
   if (this->dataPtr->scene->HasVisualName(name))
   {
-    ignerr << "Visual: [" << name << "] already exists" << std::endl;
```
adlarkin, line 178: Should we keep this error message, or at least turn it into a warning?

src/systems/sensors/Sensors.cc
```diff
@@ -135,6 +139,8 @@ class ignition::gazebo::systems::SensorsPrivate
   /// \brief Pointer to the event manager
   public: EventManager *eventManager{nullptr};
 
+  public: bool sameProcess = false;
```
adlarkin, line 142: nit: missing documentation

src/systems/sensors/Sensors.cc
```diff
@@ -174,6 +180,9 @@ class ignition::gazebo::systems::SensorsPrivate
 
   /// \brief Stop the rendering thread
   public: void Stop();
+
+  /// \brief Stop the rendering thread
+  public: void Render();
```
adlarkin, line 185:
```suggestion
  /// \brief Start the rendering thread
  public: void Render();
```

src/Server.cc
```diff
@@ -206,6 +206,22 @@ Server::Server(const ServerConfig &_config)
...
+  }
+}
+
+EventManager &Server::SharedEventManager()
+{
+  for (std::unique_ptr<SimulationRunner> &runner : this->dataPtr->simRunners)
+  {
+    return runner->EventMgr();
+  }
+}
```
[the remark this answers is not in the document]
ahcorde: I didn't check this yet, 

simulation runners are used for levels? should all runners share the same ecm and event manager when running in the same process ?

@chapulina @iche033  ?

include/ignition/gazebo/System.hh
```diff
@@ -98,6 +98,12 @@ namespace ignition
                   const std::shared_ptr<const sdf::Element> &_sdf,
                   EntityComponentManager &_ecm,
                   EventManager &_eventMgr) = 0;
+
+      /// \brief Set the system plugins if the simulation is running in the
+      /// same process
+      /// \param[in] _sameProcess True if the server and client are running in
+      /// the same process, False otherwise.
+      public: virtual void SetSameProcess(bool /*_sameProcess*/){}
```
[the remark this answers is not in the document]
ahcorde: I can't make it pure virtual, anyhow if this method make sense I can digg more into it

ahcorde:
@chapulina what do you think about this new method in `GuiSystem.hh`

```
    public: virtual void Configure(
      EventManager &/*_event*/, bool /*sameProcess*/){}
```

src/Server.cc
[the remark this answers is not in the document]
diegoferigo: AFAIK a multi-world simulation would have a different simulation runner for each world operating on a standalone ECM. A possible pattern for returning the ECM and the event manager is maybe changing the signature to:

```cpp
EventManager &Server::SharedEventManager(const unsigned int _worldIndex = 0)
```

src/gui/plugins/scene_manager/GzSceneManager.cc
```diff
@@ -92,25 +147,39 @@ bool GzSceneManager::eventFilter(QObject *_obj, QEvent *_event)
   if (_event->type() == gui::events::Render::kType)
   {
     this->dataPtr->OnRender();
+    if (this->dataPtr->sameProcess)
+    {
+      this->dataPtr->eventManager->Emit<ignition::gazebo::events::Render>();
```
chapulina, line 152: If we rely on the GUI to drive the sensor's render loop, I think it will be difficult to lockstep sensors and physics.

I'm not sure what's the best way to go about this. I think that the sensors system should be the one rendering the scene at every iteration, and the GUI should run in a separate process and paint that rendered scene to the window at a lower frame rate. I think that moving the paint code to `ign-gui`'s `MinimalScene` may complicate this even further :confused: 

I don't know if you have any good ideas here, @iche033 , you've been doing a lot of work with the rendering thread.

chapulina (commented):
Did another pass! The rendering sensors and GUI systems seem to be working properly when running in the `--same-process`!

> @chapulina what do you think about this new method in GuiSystem.hh ?
>     public: virtual void Configure(EventManager &/*_event*/, bool /*sameProcess*/){}

The GUI plugins have been communicating with each other through Qt's event system. I understand that this is needed for the current approach of emitting a render event from a GUI system to a server system. If we end up going with this approach, I think this function is ok.

src/gui/plugins/scene_manager/GzSceneManager.cc
iche033: theoretically speaking, it may be possible to have one main rendering loop in a thread shared by sensors and gui. 

so before these changes, we have two threads for rendering, one on server side and one on the GUI side:
* server:  This [RenderThread](https://github.com/ignitionrobotics/ign-gazebo/blob/84956cf51f01414b16527aa6c8b0b0a98535feb8/src/systems/sensors/Sensors.cc#L292) in sensors system (separate thread from the main ECS thread)
* gui: This [RenderThread](https://github.com/ignitionrobotics/ign-gui/blob/6d3690db7cb072664a7382752654763c7cee5a34/src/plugins/minimal_scene/MinimalScene.hh#L172) in Scene3D

What if we make them (merge?) the same thread? That's easy to say but probably difficult to implement...and we may run into other unexpected issues

If you search for `moveToThread` in MinimalScene.cc, e.g. [here](https://github.com/ignitionrobotics/ign-gui/blob/6d3690db7cb072664a7382752654763c7cee5a34/src/plugins/minimal_scene/MinimalScene.cc#L521), you can see that we create the opengl objects then move them to the rendering thread. So we'll just need to move them to the same thread that the sensors system in doing rendering and we end up with one thread shared by GUI and sensors.

iche033: This is all very high level idea. Maybe that should be experimented separately and come as a follow-up PR?

chapulina: > one main rendering loop in a thread shared by sensors and gui.

That would be ideal :sparkles: 

> experimented separately and come as a follow-up PR?

That could work, we're working against `main`, so it should be ok to merge one approach and revert it later for another one.

src/SimulationRunner.cc
```diff
@@ -536,6 +536,11 @@ void SimulationRunner::UpdateSystems()
     IGN_PROFILE("Update");
     for (auto& system : this->systemsUpdate)
       system->Update(this->currentInfo, this->entityCompMgr);
+
+    if (serverConfig.SameProcessAsGUI())
+    {
+      this->eventMgr.Emit<events::UpdateSystems>();
```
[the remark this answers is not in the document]
ahcorde: I think this is not that easy because of the ECM.  we can't lose any step in the loop that include any change in the ECM, otherwise gui plugins are not aware of these changes. For these reason I emitted the event UpdateSystem, which internally will be only executed if there is any change in the ECM

https://github.com/ignitionrobotics/ign-gazebo/blob/47a87eb14d204a7590a218497aac1d88dab594d6/src/gui/GuiRunner.cc#L253-L261


I have added a pool of workers to update the GUI https://github.com/ignitionrobotics/ign-gazebo/pull/793/commits/dc02d76d13359ad5097d74c0da0d2017e3fb6dbf . What do you think about this @chapulina @iche033  ?

src/gui/GuiRunner.cc
```diff
@@ -161,6 +200,14 @@ void GuiRunner::OnPluginAdded(const QString &_objectName)
     return;
   }
 
+  // The call above always returns the same plugin, which is the first plugin
```
chapulina, line 203: Ouch, that's broken. Fix in https://github.com/ignitionrobotics/ign-gui/pull/249

ahcorde: @chapulina, We can open an issue to fix this in the future when these two PR are in
 - https://github.com/ignitionrobotics/ign-gazebo/pull/904
 - https://github.com/ignitionrobotics/ign-gui/pull/249

or wait for the forward port.

chapulina: Yeah I wouldn't block this PR on this.

But note that the current implementation is calling `Configure` on the same plugin multiple times, which may go against developers' expectations.

src/gui/plugins/scene_manager/GzSceneManager.cc
```diff
@@ -44,9 +45,31 @@ inline namespace IGNITION_GAZEBO_VERSION_NAMESPACE {
...
     //// \brief Pointer to the rendering scene
     public: rendering::ScenePtr scene;
 
+    /// \brief Event Manager
+    public: EventManager *eventManager{nullptr};
+
+    /// \brief Is the simulation running the GUI and server in the same process
+    public: bool sameProcess{false};
+
+    /// \brief is the sensors system plugin running ?
```
adlarkin, line 64:
```suggestion
    /// \brief is the sensors system plugin running?
```

src/rendering/SceneManager.cc
```diff
@@ -1070,7 +1088,7 @@ rendering::LightPtr SceneManager::CreateLight(Entity _id,
   {
     ignerr << "Light with Id: [" << _id << "] already exists in the scene"
            << std::endl;
-    return rendering::LightPtr();
+    return this->dataPtr->lights.find(_id)->second;
```
adlarkin, line 1091: nit: it looks like `this->dataPtr->lights.find(_id)` is called in the body of the enclosing `if` block. We could cache the resulting iterator from the first call to avoid calling `find` a second time

src/systems/sensors/Sensors.cc
```diff
@@ -135,6 +140,9 @@ class ignition::gazebo::systems::SensorsPrivate
   /// \brief Pointer to the event manager
   public: EventManager *eventManager{nullptr};
 
+  /// \brief Is the GUI and server running in the same process?
```
adlarkin, line 143:
```suggestion
  /// \brief Are the GUI and server running in the same process?
```

src/systems/sensors/Sensors.cc
```diff
@@ -303,6 +318,42 @@ void SensorsPrivate::Run()
...
+void SensorsPrivate::Render()
+{
+  if (!this->sameProcess)
+    return;
+
+  if (!this->running)
+    return;
+
+  if (!this->scene)
+    return;
```
adlarkin, line 331:
```suggestion
  if (!this->sameProcess || !this->running || !this->scene)
    return;
```

src/systems/sensors/Sensors.cc
```diff
@@ -303,6 +318,42 @@ void SensorsPrivate::Run()
...
+  {
+    IGN_PROFILE("Update");
+    this->renderUtil.Update();
+  }
+
+  {
+    IGN_PROFILE("PreRender");
+    this->eventManager->Emit<events::PreRender>();
+    // Update the scene graph manually to improve performance
+    // We only need to do this once per frame It is important to call
```
adlarkin, line 343:
```suggestion
    // We only need to do this once per frame. It is important to call
```

src/rendering/RenderUtil.cc
```diff
@@ -1019,12 +1039,8 @@ void RenderUtil::Update()
 
   this->dataPtr->UpdateThermalCamera(thermalCameraData);
 }
-
-//////////////////////////////////////////////////
-void RenderUtilPrivate::CreateRenderingEntities(
-    const EntityComponentManager &_ecm, const UpdateInfo &_info)
+void RenderUtilPrivate::AddSensors(const EntityComponentManager &_ecm)
```
adlarkin, line 1042:
```suggestion

//////////////////////////////////////////////////
void RenderUtilPrivate::AddSensors(const EntityComponentManager &_ecm)
```

include/ignition/gazebo/gui/GuiSystem.hh
```diff
@@ -53,6 +54,14 @@ namespace gazebo
...
     public: virtual void Update(const UpdateInfo &/*_info*/,
                                 EntityComponentManager &/*_ecm*/){}
+
+    /// \brief When running in the same process GUI and server we need to
+    /// configure the gui system plugins with the event manager to listen
+    /// ecm updates and set if we are running in the same process
+    /// \param[in] _ecm Mutable event manager
+    /// \param[in] _sameProcess is server and gui running in the same process ?
+    public: virtual void Configure(
+      EventManager &/*_event*/, bool /*sameProcess*/){}
```
adlarkin, line 64: As discussed in https://github.com/ignitionrobotics/ign-gazebo/pull/793#discussion_r654643454, would it be possible to use a component instead of having this method being used in all GUI systems?

src/SimulationRunner.cc
```diff
@@ -538,6 +538,11 @@ void SimulationRunner::UpdateSystems()
       system->Update(this->currentInfo, this->entityCompMgr);
   }
 
+  if (serverConfig.SameProcessAsGUI())
+  {
+    this->eventMgr.Emit<events::UpdateSystems>();
+  }
```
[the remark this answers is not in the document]
ahcorde: this will emit a signal to update `GuiSystem` plugins, the update call will update only system plugins, right ?

include/ignition/gazebo/Events.hh
```diff
@@ -58,6 +58,31 @@ namespace ignition
...
+      using EnableSensors =
+        ignition::common::EventT<void(bool), struct EnableSensorsTag>;
+
+      /// \brief Event used to emit a render event when running in one process.
+      /// Some remove events are lost because of the rate when running in the
+      /// same process without sensors. This event is launched in the Simulation
+      /// runner when there is any new entity or entity marked to be removed
+      /// to remove/add entities in the renderUtil
+      using UpdateSystems =
+        ignition::common::EventT<void(void), struct UpdateSystemsTag>;
```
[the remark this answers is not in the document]
ahcorde: maybe is not explained properly. 

When running in the same process without sensors if there is an entity that we is marked to be removed, we need to call this event otherwise we will potentially lost the step where the entity is removed, because the gui loop rate is much more slower than the physics rate loop.

include/ignition/gazebo/gui/GuiSystem.hh
ahcorde: I made this for `sameProcess` in this commit https://github.com/ignitionrobotics/ign-gazebo/pull/793/commits/9962dbf826132a6d777792ecf2715d51b09e0786

But I don't know if we can do it with the `EventManager` because we need the reference and I believe components are just a copy, isn't it ?

include/ignition/gazebo/Events.hh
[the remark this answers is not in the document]
adlarkin: That makes sense :+1: I think that part that is confusing to me is that both of these events say that they emit render events, but the variable names don't seem to have anything to do with rendering. Should the part about emitting rendering events be updated?

src/SimulationRunner.cc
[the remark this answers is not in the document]
adlarkin: Ah yes, I believe you are right :+1: this code should be okay then. Maybe we can add a comment to the code here that indicates that the event is for `GuiSystem`s so that others don't get confused with system plugins?

include/ignition/gazebo/gui/GuiSystem.hh
adlarkin: > components are just a copy, isn't it ?

I believe so - usually, the data in a component is instantiated via the `=` operator, which results in a copy unless the type of data being assigned has the `=` operator overwritten to avoid copies. But, we haven't done this for all component data types, so I believe it's a copy.

src/gui/plugins/entity_tree/EntityTree.cc
```diff
@@ -132,6 +132,11 @@ void TreeModel::AddEntity(unsigned int _entity, const QString &_entityName,
     return;
   }
 
+  if (this->entityItems.find(_entity) != this->entityItems.end())
+  {
+    return;
+  }
```
chapulina, line 138: Broke this off into https://github.com/ignitionrobotics/ign-gazebo/pull/974

src/gui/GuiRunner.cc
chapulina: Addressed in d6d1f8a

src/SimulationRunner.cc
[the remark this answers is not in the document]
chapulina: I think I understand why it's best to trigger the GUI updates before `PostUpdate`. I documented it on #975

src/SimulationRunner.cc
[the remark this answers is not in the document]
chapulina: I'm proposing renaming the event and added some docs in #975

chapulina (commented):
I took a look at the sensors - 3D scene synchronization, and it concerns me that we're instantiating 2 `RenderUtil` classes that compete to create and update entities. I'm also wondering what impact the GUI running the render loop will have on the sensors' update rate. See more detailed comments below.

chapulina:
Didn't make it into `beta`, can come in a minor release.

azeey:
Since this is a really old PR and there has been big changes in the codebase since then, I'll go ahead and close it.

PR #799, opened 2021-04-30
Scale mode - part1

caguero:
This pull request should update the visuals of simple shapes.

Main questions for GUI experts:

* Not sure if this is the right way to traverse the node looking for the visual leaf? It looks a bit hacky.
* Not sure if the approach to scale the WireBox is the good one.

sphere_scale


chapulina (commented):
To have changes reflected on the server, we should make a request to the `UserCommands` system when the user releases the mouse, like it's done to translate and rotate here:

https://github.com/ignitionrobotics/ign-gazebo/blob/4e40b36943ebdf02ed73874eb9b9630ec4771974/src/gui/plugins/scene3d/Scene3D.cc#L1482

Are you planning to support changing the scale on physics too? ( https://github.com/ignitionrobotics/ign-physics/issues/65 )

src/gui/plugins/scene3d/Scene3D.cc
```diff
@@ -1607,10 +1607,53 @@ void IgnRenderer::HandleMouseTransformControl()
     else if (this->dataPtr->transformControl.Mode() ==
         rendering::TransformMode::TM_SCALE)
     {
+      // Check if the model that we're trying to scale looks like a simple shape
+      // It will need to have only two children nodes: one for the visual gizmo
+      // and another one for the link.
```
iche033, line 1612: How about just traversing down the tree (e.g. depth first or breadth first?) until you find a leaf node with the `geometry-type` UserData? Because we attach other objects to nodes, e.g. After doing `View` -> `Collisions`, new collision visuals are attached and the logic below will no longer work.

src/rendering/RenderUtil.cc
```diff
@@ -2058,6 +2058,27 @@ void RenderUtilPrivate::LowlightNode(const rendering::NodePtr &_node)
...
+
+  ignition::rendering::WireBoxPtr wireBox = this->dataPtr->wireBoxes[_entityId];
+  if (!wireBox)
+    return;
+
+  auto visParent = wireBox->Parent();
+  if (!visParent)
+    return;
+
+  visParent->SetLocalScale(_scale);
```
iche033, line 2079: I notice that the the wirebox is scaled correctly the first time but subsequent scales are not correct. It's likely that you'll need to multiply by the new scale instead of setting it directly.

src/gui/plugins/scene3d/Scene3D.cc
caguero: Are you envisioning something like this? 00b7d93

caguero:
> To have changes reflected on the server, we should make a request to the `UserCommands` system when the user releases the mouse, like it's done to translate and rotate here:
> 
> https://github.com/ignitionrobotics/ign-gazebo/blob/4e40b36943ebdf02ed73874eb9b9630ec4771974/src/gui/plugins/scene3d/Scene3D.cc#L1482
> 
> Are you planning to support changing the scale on physics too? ( [ignitionrobotics/ign-physics#65](https://github.com/ignitionrobotics/ign-physics/issues/65) )

It's still unclear but definitely not in this pull request.

src/gui/plugins/transform_control/TransformControl.qml
```diff
@@ -222,7 +227,6 @@ ToolBar {
     // TODO(anyone) enable scale snap values below when support is added in ign-physics
     // Also be sure to replace the placeholder 0's in all of the `OnSnapUpdate` calls in
     // this file to xScaleEntry.value, yScaleEntry.value, and zScaleEntry.value, respectively
```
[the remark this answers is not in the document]
caguero: Removed in 00b7d93.

src/rendering/RenderUtil.cc
```diff
@@ -2058,6 +2058,27 @@ void RenderUtilPrivate::LowlightNode(const rendering::NodePtr &_node)
   }
 }
 
+/////////////////////////////////////////////////
+void RenderUtil::ScaleWireBox(const Entity &_entityId,
+    const math::Vector3d &_scale)
+{
+  if (this->dataPtr->wireBoxes.find(_entityId) ==
+          this->dataPtr->wireBoxes.end())
```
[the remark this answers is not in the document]
caguero: I think the issue is that we're not tracking wireboxes in the `sceneManager`. I added a note to do it in the future if you think it's OK.

src/rendering/RenderUtil.cc
caguero: Fixed in 2410668 (by you :)

caguero:
Thanks for the comments. I think this is good to go. I removed the draft.

chapulina (changes requested):
> > Are you planning to support changing the scale on physics too?
> It's still unclear but definitely not in this pull request.

I don't think this PR should be merged in its current state. The current implementation is only on the client, so it doesn't affect sensors or physics, so I don't think this is useful to users and may cause confusion.

How about working on this feature in a separate branch, and only merging into a release branch after it's complete?

caguero:
> > > Are you planning to support changing the scale on physics too?
> > > It's still unclear but definitely not in this pull request.
> 
> I don't think this PR should be merged in its current state. The current implementation is only on the client, so it doesn't affect sensors or physics, so I don't think this is useful to users and may cause confusion.
> 
> How about working on this feature in a separate branch, and only merging into a release branch after it's complete?

I'd prefer to do incremental merges, as opposed to a big merge but I totally agree. Is disabling the scale icon from the scene an option?

chapulina:
> I'd prefer to do incremental merges

Agreed that this would make it easier to review.

> Is disabling the scale icon from the scene an option?

My main concern is that the final implementation may end up looking completely different from the current one. Once the commands are sent to the server, the way that the scene is updated will be very different.

We probably won't need the `SetWireBoxScale` function anymore for example, and that's being added to the public API. Note how `RenderUtil` doesn't have any public functions for setting pose, for example, because translation and rotation are being set via the ECM.

So, I think it's ok for this to be implemented incrementally, but I think it's best if we don't make any changes to the public API or the UI until we have the final implementation working.

caguero:
> > I'd prefer to do incremental merges
> 
> Agreed that this would make it easier to review.
> 
> > Is disabling the scale icon from the scene an option?
> 
> My main concern is that the final implementation may end up looking completely different from the current one. Once the commands are sent to the server, the way that the scene is updated will be very different.
> 
> We probably won't need the `SetWireBoxScale` function anymore for example, and that's being added to the public API. Note how `RenderUtil` doesn't have any public functions for setting pose, for example, because translation and rotation are being set via the ECM.
> 
> So, I think it's ok for this to be implemented incrementally, but I think it's best if we don't make any changes to the public API or the UI until we have the final implementation working.

I see. It sounds reasonable, thanks for the explanation. I'll retarget it then.

caguero:
> I see. It sounds reasonable, thanks for the explanation. I'll retarget it then.

Retargeted to `scale` branch.

src/rendering/RenderUtil.cc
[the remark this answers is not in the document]
chapulina: If I understand correctly, this function is getting `entityId`  -> `wirebox` -> `parent visual`, and setting the scale on the parent visual.

The `NodeById` function should give you the parent visual directly from the entity ID. So I'm not sure why we need to go through the wirebox.

The wirebox is created at runtime when the visual is highlighted, so it may be a less reliable way of getting to the visual.

src/rendering/RenderUtil.cc
```diff
@@ -2058,6 +2058,29 @@ void RenderUtilPrivate::LowlightNode(const rendering::NodePtr &_node)
...
+  /// when WireBoxes are tracked in the scene manager.
+  if (this->dataPtr->wireBoxes.find(_entityId) ==
+          this->dataPtr->wireBoxes.end())
+  {
+    return;
+  }
+
+  ignition::rendering::WireBoxPtr wireBox = this->dataPtr->wireBoxes[_entityId];
+  if (!wireBox)
+    return;
```
chapulina, line 2075: It would be helpful to print warnings for these returns, so we can debug situations when the scaling fails.

src/gui/plugins/scene3d/Scene3D.cc
```diff
@@ -1341,6 +1341,51 @@ void IgnRenderer::XYZConstraint(math::Vector3d &_axis)
...
+      auto v = std::dynamic_pointer_cast<ignition::rendering::Visual>(n);
+      if (v)
+      {
+        try
+        {
+          int userData = std::get<int>(v->UserData("geometry-type"));
+
+          if (userData == static_cast<int>(sdf::GeometryType::BOX)      ||
+              userData == static_cast<int>(sdf::GeometryType::CYLINDER) ||
+              userData == static_cast<int>(sdf::GeometryType::SPHERE))
```
chapulina, line 1369: Can you add a TODO to add ellipsoid when we forward-port to Edifice?

Scaling capsules is trickier, so I think we don't need  to do it in the first pass.

src/rendering/SceneManager.cc
```diff
@@ -265,6 +265,11 @@ rendering::VisualPtr SceneManager::CreateVisual(Entity _id,
   rendering::VisualPtr visualVis = this->dataPtr->scene->CreateVisual(name);
   visualVis->SetUserData("gazebo-entity", static_cast<int>(_id));
   visualVis->SetUserData("pause-update", static_cast<int>(0));
+
+  auto geometry = _visual.Geom();
+  if (geometry)
+    visualVis->SetUserData("geometry-type", static_cast<int>(geometry->Type()));
```
chapulina, line 271: How about adding a function to `ign-rendering` that returns this information? I think this can only be done from Fortress, so we'll need to keep using user data here.

caguero: Sounds good to me.

src/gui/plugins/scene3d/Scene3D.cc
caguero: Done in 7d0728a.

src/rendering/RenderUtil.cc
caguero: Done in 7d0728a.

caguero: I might be wrong but Ian recommended this method because the scene manager isn't tracking wireboxes so far.

caguero:
@mjcarroll , this should be ready for another pass when you have some time. Thanks!

src/gui/plugins/scene3d/Scene3D.cc
iche033: expand todo note to include `sdf::GeometryType::ELLIPSOID`?

caguero: Done in 3b857a9.

mjcarroll:

src/gui/plugins/scene3d/Scene3D.hh
```diff
@@ -439,6 +439,16 @@ inline namespace IGNITION_GAZEBO_VERSION_NAMESPACE {
...
     private: void DeselectAllEntities(bool _sendEvent);
 
+    /// \brief Helper function that traverses the node tree looking for any
+    /// child with the "geometry-type" user data. If found, the function also
+    /// checks if the value is a simple shape (box, cylinder or sphere).
+    /// \param[in] _node Root node.
+    /// \param[out] _visual The visual pointer to the simple shape if found.
+    /// \return True if any simple shape was found in the tree or false
+    /// otherwise.
+    private: bool ContainsSimpleShape(const rendering::NodePtr &_node,
```
mjcarroll, line 449: Could you return the VisualPtr as `nullptr` if there was no shape, rather than as an arg and as a bool?

caguero: Thanks! Changed in d449365.

jennuine (commented):
Currently, if a model is made up of simple shapes, it will scale the entire model. Would it be possible to update the scale for certain parts of the model? For example, in `diff_drive.sdf` would it be possible to set the scale for the wheels only and not the entire vehicle?

src/gui/plugins/scene3d/Scene3D.cc
```diff
@@ -1607,10 +1654,20 @@ void IgnRenderer::HandleMouseTransformControl()
     else if (this->dataPtr->transformControl.Mode() ==
         rendering::TransformMode::TM_SCALE)
     {
+      // Check if the model that we're trying to scale looks like a simple shape
+      auto node = this->dataPtr->transformControl.Node();
+      auto v = std::dynamic_pointer_cast<ignition::rendering::Visual>(node);
+      rendering::VisualPtr visualSimple = this->ContainsSimpleShape(node);
+      if (!visualSimple)
+        return;
```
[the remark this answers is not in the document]
caguero: Thanks! I printed a warning in de79f9a. The scope if this initial sprint for scaling is to support simple meshes only. The goal is to write all the boilerplate and initial infrastructure. Ideally, we'll support scaling meshes or parts of the model in the future.

jennuine (approved):
LGTM with minor comment.

src/gui/plugins/transform_control/TransformControl.qml
```diff
@@ -29,10 +29,10 @@ ToolBar {
   Layout.minimumHeight: 100
 
   // TODO(anyone) enable scale button when support is added in ign-physics
```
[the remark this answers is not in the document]
caguero: Removed in 8670c8f

Issue #810, opened 2021-05-06
Use math::SpeedLimiter inside controllers


chapulina:
Controllers like `diff_drive` have a [SpeedLimiter](https://github.com/ignitionrobotics/ign-gazebo/blob/ign-gazebo5/src/systems/diff_drive/SpeedLimiter.hh) class inside them. This file is currently repeated inside these plugins:

* `diff_drive`
* `ackermann_steering`

Now that we have the [ignition::math::SpeedLimiter](https://ignitionrobotics.org/api/math/6.8/classignition_1_1math_1_1SpeedLimiter.html) class, we should use that and remove those custom limiters. The [mecanum_drive](https://github.com/ignitionrobotics/ign-gazebo/blob/ign-gazebo5/src/systems/mecanum_drive/MecanumDrive.cc) plugin is already using that.

## Desired behavior

* all `SpeedLimiter` files are removed from `ign-gazebo`
* `diff_drive` and `ackermann_steering` use `math::SpeedLimiter`

All tests should continue to pass.

## Alternatives considered

We moved the `SpeedLimiter` to `ign-math` so we could use it in multiple places. An alternative is to keep duplicating the file, but that doesn't scale well. Plus, we made improvements and fixes to the limiter when porting it to math.

## Implementation suggestion

Take a look at how the mecanum drive is using it.

scpeters:
For reference, 9bbccebb80aae51dd645ed55c8d6c25687b8f9dc is the commit in #683 that changed the mecanum drive to use the `ignition::math::SpeedLimiter`

LolaSegura:
Hi I have been working on this issue and I noticed it has a 'citadel' tag. I was wondering, wouldn't it make sense to push this on the latest version (gazebo6) as I saw it hasn't been changed in either of the previous versions.

scpeters:
> Hi I have been working on this issue and I noticed it has a 'citadel' tag. I was wondering, wouldn't it make sense to push this on the latest version (gazebo6) as I saw it hasn't been changed in either of the previous versions.

we have added some "guiding principles" to our ignitionrobotics.org/about page, but they haven't been deployed yet, so I'll excerpt one here:

~~~
<b>Updates</b>:
            New backwards compatible features are added to stable releases
            whenever possible. This way, users can take advantage of new features
            without the need to upgrade to a later release, and can postpone
            upgrade efforts until breaking changes are needed.
~~~

[...]

PR #833, opened 2021-05-26
Using math::SpeedLimiter on the diff_drive controller.

LolaSegura:
Signed-off-by: LolaSegura <lsegura@ekumenlabs.com>

## Summary
This replace the diff_drive controller speed limiter class for the one implemented in ignition math. Part of #810. 

## Checklist
- [ ] Added tests


src/systems/diff_drive/DiffDrive.cc
```diff
@@ -197,56 +196,48 @@ void DiffDrive::Configure(const Entity &_entity,
...
+  // Instantiate the speed limiters.
+  this->dataPtr->limiterLin = std::make_unique<ignition::math::SpeedLimiter>();
+  this->dataPtr->limiterAng = std::make_unique<ignition::math::SpeedLimiter>();
 
+  // Parse speed limiter parameters.
   if (_sdf->HasElement("min_velocity"))
   {
-    minVel = _sdf->Get<double>("min_velocity");
-    hasVelocityLimits = true;
+    double minVel = _sdf->Get<double>("min_velocity");
```
francocipollone, line 206: nit: it can be `const`

src/systems/diff_drive/DiffDrive.cc
```diff
@@ -197,56 +196,48 @@ void DiffDrive::Configure(const Entity &_entity,
...
+    this->dataPtr->limiterLin->SetMinVelocity(minVel);
+    this->dataPtr->limiterAng->SetMinVelocity(minVel);
   }
   if (_sdf->HasElement("max_velocity"))
   {
-    maxVel = _sdf->Get<double>("max_velocity");
-    hasVelocityLimits = true;
+    double maxVel = _sdf->Get<double>("max_velocity");
```
francocipollone, line 212: nit: it can be `const`

src/systems/diff_drive/DiffDrive.cc
```diff
@@ -197,56 +196,48 @@ void DiffDrive::Configure(const Entity &_entity,
...
+    this->dataPtr->limiterLin->SetMaxVelocity(maxVel);
+    this->dataPtr->limiterAng->SetMaxVelocity(maxVel);
   }
   if (_sdf->HasElement("min_acceleration"))
   {
-    minAccel = _sdf->Get<double>("min_acceleration");
-    hasAccelerationLimits = true;
+    double minAccel = _sdf->Get<double>("min_acceleration");
```
francocipollone, line 218: nit: it can be `const`

src/systems/diff_drive/DiffDrive.cc
```diff
@@ -197,56 +196,48 @@ void DiffDrive::Configure(const Entity &_entity,
...
+    this->dataPtr->limiterLin->SetMinAcceleration(minAccel);
+    this->dataPtr->limiterAng->SetMinAcceleration(minAccel);
   }
   if (_sdf->HasElement("max_acceleration"))
   {
-    maxAccel = _sdf->Get<double>("max_acceleration");
-    hasAccelerationLimits = true;
+    double maxAccel = _sdf->Get<double>("max_acceleration");
```
francocipollone, line 224: nit: it can be `const`

src/systems/diff_drive/DiffDrive.cc
```diff
@@ -197,56 +196,48 @@ void DiffDrive::Configure(const Entity &_entity,
...
+    this->dataPtr->limiterLin->SetMaxAcceleration(maxAccel);
+    this->dataPtr->limiterAng->SetMaxAcceleration(maxAccel);
   }
   if (_sdf->HasElement("min_jerk"))
   {
-    minJerk = _sdf->Get<double>("min_jerk");
-    hasJerkLimits = true;
+    double minJerk = _sdf->Get<double>("min_jerk");
```
francocipollone, line 230: nit: it can be `const`

src/systems/diff_drive/DiffDrive.cc
```diff
@@ -197,56 +196,48 @@ void DiffDrive::Configure(const Entity &_entity,
...
+    this->dataPtr->limiterLin->SetMinJerk(minJerk);
+    this->dataPtr->limiterAng->SetMinJerk(minJerk);
   }
   if (_sdf->HasElement("max_jerk"))
   {
-    maxJerk = _sdf->Get<double>("max_jerk");
-    hasJerkLimits = true;
+    double maxJerk = _sdf->Get<double>("max_jerk");
```
francocipollone, line 236: nit: it can be `const`

francocipollone (commented):
LGTM! I just left a couple of nits.

I will add @scpeters as a reviewer as well

scpeters:
this is looking good!

it's helpful to cross-reference issues from pull requests, so I would add that this is "Part of https://github.com/ignitionrobotics/ign-gazebo/issues/810" in the pull request description

src/systems/diff_drive/DiffDrive.cc
LolaSegura: Addressed all.

scpeters:
I noticed that codecov is complaining about test coverage. We are not strict about requiring code coverage, but I think we can easily increase this for this test. I believe the `diff_drive` system has the following integration test:

* https://github.com/ignitionrobotics/ign-gazebo/blob/ign-gazebo3/test/integration/diff_drive_system.cc

The test uses several different world files (search for `test/worlds` to find [lines like this](https://github.com/ignitionrobotics/ign-gazebo/blob/ignition-gazebo3_3.8.0/test/integration/diff_drive_system.cc#L203)) such as [test/worlds/diff_drive.sdf](https://github.com/ignitionrobotics/ign-gazebo/blob/ignition-gazebo3_3.8.0/test/worlds/diff_drive.sdf). This world has a plugin that [specifies max_acceleration and max_velocity](https://github.com/ignitionrobotics/ign-gazebo/blob/ignition-gazebo3_3.8.0/test/worlds/diff_drive.sdf#L230-L231), but doesn't specify `min_acceleration` or `min_velocity`. If we added `min_acceleration` of `-1` and `min_velocity` of `0.5`, that will improve the coverage and add symmetry to the test. To get even more coverage and without changing the dynamics of the system we could add a `min_jerk` value of `-Inf` and a `max_jerk` value of `Inf` (I think that will work).

Since these are easy changes to improve code coverage, do you mind giving them a try? Thanks!

LolaSegura:
> If we added `min_acceleration` of -1 and `min_velocity` of 0.5, that will improve the coverage and add symmetry to the test.

@scpeters  here when you mentioned the `min_velocity` did you meant to write -0.5? Because otherwise it would take the same value as the `max_velocity`.

scpeters:
> > If we added `min_acceleration` of -1 and `min_velocity` of 0.5, that will improve the coverage and add symmetry to the test.
> 
> @scpeters here when you mentioned the `min_velocity` did you meant to write -0.5? Because otherwise it would take the same value as the `max_velocity`.

thanks for catching that; I meant to write `-0.5`

francocipollone (changes requested):
ptal at the comments. Also, run `codecheck` to match style.

francocipollone:
@LolaSegura . I've tried it locally.
 - It seems that just adding the `min_jerk` and `max_jerk` properties are causing a miss behavior in the diff drive when moving forward(probably backward too).
 - Leaving those properties commented out the test moving forward works fine.
 - It is failing only one test when moving backward:
 ```
 Expected: (expectedLowerPosition.Y()) < (expectedGreaterPosition.Y()), actual: 0.710173 vs 1.40412e-12
 ```
 Re-check the desire angle you are publishing.

LolaSegura:
>If we added min_acceleration of -1 and min_velocity of 0.5 -0.5, that will improve the coverage and add symmetry to the test

I added the `<min_velocity>` and `<min_acceleration>` to the `diff_drive.sdf` and the `diff_drive_custom_topics.sdf`. With this the test its working fine. 

> To get even more coverage and without changing the dynamics of the system we could add a min_jerk value of -Inf and a max_jerk value of Inf (I think that will work).

I try adding this limits as well, but I encounter a problem because when being parsed the limits where interpreted as zero, causing the test to fail. An option would be setting the `<max_jerk>` with a big value and the `<min_jerk>` with a small one to get more test coverage.

francocipollone:
It is ready for review @scpeters

scpeters (commented):
> > To get even more coverage and without changing the dynamics of the system we could add a min_jerk value of -Inf and a max_jerk value of Inf (I think that will work).
> 
> I try adding this limits as well, but I encounter a problem because when being parsed the limits where interpreted as zero, causing the test to fail. An option would be setting the `<max_jerk>` with a big value and the `<min_jerk>` with a small one to get more test coverage.

I'll look into this briefly

test/integration/diff_drive_system.cc
```diff
@@ -54,145 +54,170 @@ class DiffDriveTest : public ::testing::TestWithParam<int>
...
+
+      if(forward)
+      {
+        EXPECT_LT(v, 0.5);
+        EXPECT_LT(a, 1);
+      }
+      else
+      {
+        EXPECT_GT(v, -0.5);
+        EXPECT_GT(a, -1);
```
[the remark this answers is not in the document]
LolaSegura: Addressed both comments.

scpeters:
> > To get even more coverage and without changing the dynamics of the system we could add a min_jerk value of -Inf and a max_jerk value of Inf (I think that will work).
> 
> I try adding this limits as well, but I encounter a problem because when being parsed the limits where interpreted as zero, causing the test to fail. An option would be setting the `<max_jerk>` with a big value and the `<min_jerk>` with a small one to get more test coverage.

thanks for testing this; I just looked into it a bit deeper and found that it is a parsing issue with `libsdformat` (https://github.com/osrf/sdformat/issues/261). We can leave out the jerk tests for now

scpeters (approved):

PR #848, opened 2021-06-03
Fix potentially flaky integration component test case

adlarkin:
# 🦟 Bug fix

## Summary
I noticed that the components integration test fails at times for me locally. I found that for the `LogicalMicrophone` component test I wrote last year, I forgot to initialize one of the streams with the relevant data :upside_down_face: I believe that this should make the test pass consistently now.

## Checklist
- [x] Added tests

### DIFF
--- a/test/integration/components.cc
+++ b/test/integration/components.cc
@@ -903,7 +903,7 @@ TEST_F(ComponentsTest, LogicalMicrophone)
   comp1.Serialize(ostr);
   EXPECT_EQ("0 0.5", ostr.str());
 
-  std::istringstream istr;
+  std::istringstream istr(ostr.str());
   components::LogicalMicrophone comp3;
   comp3.Deserialize(istr);
   EXPECT_EQ(comp1, comp3);
--- a/test/integration/components.cc
+++ b/test/integration/components.cc
@@ -882,31 +882,33 @@ TEST_F(ComponentsTest, LogicalAudioSourcePlayInfo)
 TEST_F(ComponentsTest, LogicalMicrophone)
 {
   logical_audio::Microphone mic1;
-  mic1.id = 0;
+  mic1.id = 4;
   mic1.volumeDetectionThreshold = 0.5;
 
   logical_audio::Microphone mic2;
-  mic2.id = 1;
-  mic2.volumeDetectionThreshold = mic1.volumeDetectionThreshold;
+  mic2.id = 8;
+  mic2.volumeDetectionThreshold = 0.8;
 
   // create components
   auto comp1 = components::LogicalMicrophone(mic1);
   auto comp2 = components::LogicalMicrophone(mic2);
 
   // equality operators
-  EXPECT_NE(mic1, mic2);
-  EXPECT_FALSE(mic1 == mic2);
-  EXPECT_TRUE(mic1 != mic2);
+  EXPECT_NE(comp1, comp2);
+  EXPECT_FALSE(comp1 == comp2);
+  EXPECT_TRUE(comp1 != comp2);
 
   // stream operators
   std::ostringstream ostr;
   comp1.Serialize(ostr);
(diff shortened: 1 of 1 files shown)


scpeters:
out of curiosity, what console output do you see from a failure of this test?

adlarkin:
> out of curiosity, what console output do you see from a failure of this test?

Sometimes, I will see this:
```
  comp1
    Which is: 32-byte object <E0-D7 33-3B AC-55 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 E0-3F>
  comp3
    Which is: 32-byte object <E0-D7 33-3B AC-55 00-00 00-00 00-00 00-00 00-00 D0-2B 80-3B AC-55 00-00 60-D0 14-84 D3-7F 00-00>
```

If you look at the code I am changing, before this PR, I am just comparing a deserialized default-constructed `std::istringstream` with `comp1`, which is a `logical_audio::Microphone` object that is not default-constructed. I'm honestly amazed that this comparison passes most of the time, because it's an incorrect comparison.

scpeters:
your change looks right to me, but I wanted to understand why it wasn't always failing. I think it's because the [logical_audio::Microphone == operator](https://github.com/ignitionrobotics/ign-gazebo/blob/ignition-gazebo3_3.8.0/include/ignition/gazebo/components/LogicalAudio.hh#L130) only compares the `id` values and not the `volumeDetectionThreshold`. I think it would be worth adding some additional `EXPECT_NEAR` or `EXPECT_DOUBLE_EQ` statements for the `volumeDetectionThreshold` to confirm the equality

ahcorde (approved):

adlarkin:
> your change looks right to me, but I wanted to understand why it wasn't always failing. I think it's because the [logical_audio::Microphone == operator](https://github.com/ignitionrobotics/ign-gazebo/blob/ignition-gazebo3_3.8.0/include/ignition/gazebo/components/LogicalAudio.hh#L130) only compares the `id` values and not the `volumeDetectionThreshold`. I think it would be worth adding some additional `EXPECT_NEAR` or `EXPECT_DOUBLE_EQ` statements for the `volumeDetectionThreshold` to confirm the equality

Yeah, that's a good point! I also think an issue here was that `mic1.id` was set to `0`, which I believe is (usually) the default value given to unsigned ints. So, I am guessing that before the changes in this PR, the default-constructed `comp3` was right _most of the time_ because `comp3.id` was set to `0`. However, since there's no guarantee regarding what a data type will be initialized to unless explicitly assigned, there are probably times where `comp3.id` doesn't default to `0`, which results in the test failure.

Regarding the comment about the `==` operator only comparing IDs: I believe the reason for this is because for use cases like the [LogicalAudioSensorPlugin](https://github.com/ignitionrobotics/ign-gazebo/tree/ignition-gazebo3_3.8.0/src/systems/logical_audio_sensor_plugin), the responsibility is on the plugin/user to ensure that no two microphones are created with the same ID. So, when comparing for equality in the component context, IDs should be enough since IDs are what determine uniqueness for a logical microphone..

Anyways, I think that it's still good to check the values of the `volumeDetectionThreshold` after de-serialization to make sure that those are still what we expect them to be. So, I went ahead and made a few changes in https://github.com/ignitionrobotics/ign-gazebo/pull/848/commits/481e413c646dd06d2caf6a31de6220570f21e7d0:
* set the IDs of the two test microphone components to something other than `0` so that we don't have the default-initialized value issues we've been seeing
* compare the components when checking equality operators, not the data structs that make up the component (I realized the equality checks were being done between `mic1` and `mic2` instead of `comp1` and `comp2`)
* checking the `volumeDetectionThreshold` values after de-serialization to make sure they haven't changed

@scpeters, does this look good to you?

mjcarroll:

chapulina (approved):
sneaky

PR #857, opened 2021-06-10
Remove unneeded camera follow offset checks

iche033:

# 🦟 Bug fix

## Summary

When handling mouse movements in camera follow mode, , there is a check to see if camera has actually moved or not. If it has moved, the camera offset is marked as dirty. However, I found that this check always passes because a variable (`camWorldPos`) is never set. It also looks like the check is just unnecessary and we are just doing additional calculations so I just removed the check.

## Checklist
- [ ] Added tests

### DIFF
--- a/src/gui/plugins/scene3d/Scene3D.cc
+++ b/src/gui/plugins/scene3d/Scene3D.cc
@@ -1644,10 +1644,6 @@ void IgnRenderer::HandleMouseViewControl()
             << std::endl;
   }
 
-  math::Vector3d camWorldPos;
-  if (!this->dataPtr->followTarget.empty())
-    this->dataPtr->camera->WorldPosition();
-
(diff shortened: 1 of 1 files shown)


src/gui/plugins/scene3d/Scene3D.cc
```diff
@@ -1644,10 +1644,6 @@ void IgnRenderer::HandleMouseViewControl()
             << std::endl;
   }
 
-  math::Vector3d camWorldPos;
-  if (!this->dataPtr->followTarget.empty())
-    this->dataPtr->camera->WorldPosition();
```
iche033, line 1649: the bug is likely here. This should probably have been:

```
math::Vector3d camWorldPos;
if (!this->dataPtr->followTarget.empty())
  camWorldPos = this->dataPtr->camera->WorldPosition();
```

jennuine (approved):

PR #862, opened 2021-06-16
Support locked entities, and headless video recording using sim time

nkoenig:

# 🎉 New feature

Related PR: https://github.com/ignitionrobotics/ign-sensors/pull/137

## Summary

The goal of this PR is to support generating videos from log files using a headless simulation instance. In order to accomplish this, I need to spawn a camera into simulation and keep that camera around during log seeks. This lead to the need for `locked` entities within the ECM.

Recording a video using simulation time is required when you want a video to playback at real-time speeds. I copied the sim time video recording code from the Scene3d plugin. 

## Test it

There are ECM tests for the locking feature.
The easiest way to test the sim time video recording is to use the  `camera_video_record_dbl_pendulum.sdf` example world file, and add `<use_sim_time>true</use_sim_time>` to the plugin. Depending on your machine, you may have to adjust the physics step to make sure simulation is running slower than real-time. Record a video, and note the simulation time duration for the video. The resulting video should have matching duration, which can be inspected using:

`mplayer -identify VIDEO.MP4 2>/dev/null | grep ID_LENGTH`

## Checklist
- [x] Added tests


include/ignition/gazebo/EntityComponentManager.hh
```diff
@@ -91,6 +91,35 @@ namespace ignition
       public: void RequestRemoveEntity(const Entity _entity,
           bool _recursive = true);
 
+      /// \brief Prevent an entity and optionally its children from
+      /// being removed.
```
chapulina, line 95: The word "lock" implies to me that more is being prevented than just removal. Without reading these docs, I assumed the function would prevent the entity's components from being modified.

How about something like `MarkAsUnremovable`?

iche033:
The sim time video recording feature is working for me. I also tried running faster than real time and that's working as well and produced the correct video length. (RTF just drops a little when video recording is on). 

another nice video recorder param that we could port over later from Scene3D is <bitrate> so users can record high quality videos.

iche033:
> another nice video recorder param that we could port over later from Scene3D is so users can record high quality videos.

#864

include/ignition/gazebo/EntityComponentManager.hh
nkoenig: Changed the name in 1e908d310e46dfd94c1529c1a658ac8ec09397df.

mjcarroll (approved):
LGTM with Green CI (minus tests we know are flaky)

mjcarroll (approved):

PR #863, opened 2021-06-17
Set gui camera pose

### LINKED ISSUE #101: Get/set user camera pose

jennuine:
Signed-off-by: Jenn Nguyen <jenn@openrobotics.org>

# 🎉 New feature

Closes #101 

## Summary
Extended the [view_angle](https://github.com/ignitionrobotics/ign-gazebo/tree/ign-gazebo2/src/gui/plugins/view_angle) to allow the user to set the exact camera pose from the GUI.

## Test it
- `ign gazebo shapes.sdf`
- From the plugins menu, select `View Angle`
- Enter different values

Peek 2021-06-16 18-38

## Checklist
- [ ] Added tests


src/gui/plugins/view_angle/ViewAngle.cc
```diff
@@ -61,7 +76,17 @@ void ViewAngle::LoadConfig(const tinyxml2::XMLElement *)
...
+
+  // Subscribe to camera pose
+  std::string topic = "/gui/camera/pose";
+  this->dataPtr->node.Subscribe(
+    topic, &ViewAnglePrivate::CamPoseCb, this->dataPtr.get());
+
+  // Move to pose service
+  this->dataPtr->moveToPoseService = "/gui/move_to/pose";
+
+  this->dataPtr->viewAngle = this;
```
chapulina, line 89: This is adding a circular dependency which could bring problems down the road. How about adding the `CamPoseCb` function to `ViewAngle` instead of `ViewAnglePrivate` so that you can do `this->CamPoseChanged()`?

src/gui/plugins/view_angle/ViewAngle.qml
```diff
@@ -182,5 +183,130 @@ ToolBar {
...
+      Layout.row: 1
+      Layout.column: 5
+    }
+    IgnSpinBox {
+      id: x
+      Layout.row: 1
+      Layout.column: 6
+      value: ViewAngle.camPose[0]
+      maximumValue: 1000.00
+      minimumValue: -1000.00
```
chapulina, line 210: I think we can increase these limits. 1 km can be too small for large worlds

src/gui/plugins/view_angle/ViewAngle.hh
```diff
@@ -36,6 +36,13 @@ namespace gazebo
   {
     Q_OBJECT
 
+    /// \brief gui camera pose (QList order is x, y, z, roll, pitch, yaw)
+    Q_PROPERTY(
+      QList<QString> camPose
```
chapulina, line 41: Could this store `double`s instead of `QString`s?

src/gui/plugins/view_angle/ViewAngle.qml
```diff
@@ -182,5 +183,130 @@ ToolBar {
...
+      Layout.row: 1
+      Layout.column: 7
+    }
+    IgnSpinBox {
+      id: roll
+      Layout.row: 1
+      Layout.column: 8
+      value: ViewAngle.camPose[3]
+      maximumValue: 6.28
+      minimumValue: 0.00
```
chapulina, line 272: I think we can accept negative rotation values

src/gui/plugins/view_angle/ViewAngle.qml
```diff
@@ -182,5 +183,130 @@ ToolBar {
         ViewAngle.OnAngleMode(0, 0, 1)
       }
     }
+
+    // set camera pose
```
chapulina, line 187: How about making this look a bit more like the pose fields in the component inspector? I think we should start being a bit more consistent with the UI across various widgets.

campose

That one also has the benefit of expanding, which is useful in case the numbers are very big.

chapulina (commented):
This is super helpful, I'm tired of echoing the pose topic, then having to convert quaternions to Euler :sweat_smile:

src/gui/plugins/view_angle/ViewAngle.cc
jennuine: ahhh I had it that way because I misremembered what broke abi :sweat_smile: fixed 2b4c28a5b5178962e6721685c2171f5709629c52

src/gui/plugins/view_angle/ViewAngle.qml
jennuine: How's 1000km? 0cdb38b3fdf9d3bebd4c2cc4a3d362103453de07

src/gui/plugins/view_angle/ViewAngle.hh
jennuine: 0cdb38b3fdf9d3bebd4c2cc4a3d362103453de07

src/gui/plugins/view_angle/ViewAngle.qml
jennuine: 0cdb38b3fdf9d3bebd4c2cc4a3d362103453de07

src/gui/plugins/view_angle/ViewAngle.qml
jennuine: How's this? 0cdb38b3fdf9d3bebd4c2cc4a3d362103453de07

Peek 2021-06-17 13-56

src/gui/plugins/view_angle/ViewAngle.cc
chapulina: Ah gotcha. For future reference, we don't need to worry about ABI for Gazebo-GUI plugins, because their headers are not being installed. We're using the PIMPL pattern on some of them just in case we ever need to install the headers.

src/gui/plugins/view_angle/ViewAngle.qml
chapulina: Nice!

chapulina (approved):
Works great :+1:

nkoenig:
I added a header in f08cd03baacf4686dd975d38238db8106005b279. Let me know if you have objections or suggestions.

nkoenig (approved):

Issue #865, opened 2021-06-17
Resource Spawner plugin showing owner's name twice


jennuine:
## Environment
* OS Version: Ubuntu 20.04 Focal
* Source or binary build?
    both

## Description
* Expected behavior: Under "Fuel resources" expected to see owner's name appear in list once.
* Actual behavior: Owner's name appears twice but when one name is clicked it highlights both (the same owner name).

This was tested in both source and binary builds and does not happen on Ubuntu Bionic.

## Steps to reproduce

1. `ign gazebo`
2. Add "Resource Spawner" from plugin menu
3. Wait for fuel resources to be finished loading

## Output

![image](https://user-images.githubusercontent.com/8602001/122484744-1cc45f00-cf8a-11eb-842f-bcbb633b1fd0.png)

azeey:
Fixed by #1962

PR #873, opened 2021-06-22
Proposal for #856: Deprecations instead of removals

chapulina:
Some proposals to #856 , focused on providing a smoother migration for users with tick-tock.

* Keep `ComponentKey` around, it's still a unique identifier, but we use `Entity` as the `ComponentId`
* Tick-tock the `StorageDescriptor` classes, because they are outside of `detail`.
    * This also allows deprecating `Factory::Register` without loss of functionality.
    * Kept the `details/ComponentStorageBase.hh` header with minimal code not to break public API.
* Renamed the new `ComponentStorage` class to `EntityStorage`, so it's not confused with the old class.

I considered changing the `CreateComponent` function to `NewComponent`, for example, so we could deprecate the old one that returns `ComponentKey`. But that would mean many systems would need to be updated to use the new function. On the other hand, keeping the function name while changing the return type requires zero changes to our systems because they're all ignoring that return type anyway. So I think it will be easier for most users if we just change the return type.

Also removed some runtime warnings from deprecated functions, the compiler warnings should be enough.


include/ignition/gazebo/detail/ComponentStorageBase.hh
```diff
@@ -0,0 +1,55 @@
+/*
+ * Copyright (C) 2018 Open Source Robotics Foundation
```
ahcorde, line 2:
```suggestion
 * Copyright (C) 2021 Open Source Robotics Foundation
```

Migration.md
```diff
@@ -5,6 +5,10 @@ Deprecated code produces compile-time warnings. These warning serve as
 notification to users that their code should be upgraded. The next major
 release will remove the deprecated code.
 
+## Ignition Gazebo 5.x to 6.x
+
+*
```
ahcorde, line 10: ?

include/ignition/gazebo/components/Factory.hh
```diff
@@ -77,10 +78,57 @@ namespace components
...
+    /// of the component and its storage based on an ID.
+    /// \param[in] _type Type of component to register.
+    /// \param[in] _compDesc Object to manage the creation of ComponentTypeT
+    ///  objects.
+    /// \param[in] _storageDesc Ignored.
+    /// \tparam ComponentTypeT Type of component to register.
+    /// \deprecated See function that doesn't accept a storage
+    public: template<typename ComponentTypeT>
+    void IGN_DEPRECATED(6) Register(const std::string &_type,
+        ComponentDescriptorBase *_compDesc, StorageDescriptorBase *)
```
ahcorde, line 127:
```suggestion
        ComponentDescriptorBase *_compDesc, StorageDescriptorBase * /*_storageDesc*/)
```

include/ignition/gazebo/components/Factory.hh
```diff
@@ -236,6 +284,16 @@ namespace components
       return comp;
     }
 
+    /// \brief Create a new instance of a component storage.
+    /// \param[in] _typeId Type of component which the storage will hold.
+    /// \return Always returns nullptr.
+    /// \deprecated Storages aren't necessary anymore.
+    public: std::unique_ptr<ComponentStorageBase> IGN_DEPRECATED(6) NewStorage(
+        const ComponentTypeId &)
```
ahcorde, line 292:
```suggestion
        const ComponentTypeId & /*_typeId*/)
```

ahcorde (changes requested):

adlarkin:
The PR and title mention #865, but I believe that you meant #856?

chapulina:
> The PR title and description mention #865, but I believe that you meant #856?

Ahhh numbers!

include/ignition/gazebo/detail/ComponentStorageBase.hh
chapulina: This file is being restored because #856 was deleting it, it's not new :wink:

Migration.md
chapulina: Left there as a reminder for us to fill it once we settle on the API

adlarkin:
> Keep `ComponentKey` around, it's still a unique identifier, but we use `Entity` as the `ComponentId`

I thought about doing something like this as well, but I believe the issue here is that `Entity` is of type `uint64_t`, while `ComponentId` is of type `int`. So, is it safe/non-breaking to use `Entity` in place of `ComponentId` for the sake of preserving `ComponentKey`? While it's probably unlikely that we would have any issues casting between the two types, it is worth noting that one type is a signed integer while the other type is unsigned. So, we would need to make sure that we don't have any strange integer overflow bugs.

chapulina:
> What if we re-named it to EntityComponentStorage?

That works for me :+1: As I mentioned on the other PR, I'm still in the process of understanding what this class is for :smile: 

> is it safe/non-breaking to use Entity in place of ComponentId

Yeah I thought of that too. It's unfortunate that `ComponentId` is `int`, because I'm pretty sure it's not meant to store negative values. There's a real possibility of overflow.

Some options are changing `ComponentId`'s type to match `Entity`'s, or change `ComponentKey` to hold an `Entity` instead. I think both options should be seamless for users who are being strict about using the aliases. What do you think?

adlarkin:
> As I mentioned on the other PR, I'm still in the process of understanding what this class is for

I left a comment that explains things in greater detail: https://github.com/ignitionrobotics/ign-gazebo/pull/856#discussion_r659100121

Hopefully that helps!

> It's unfortunate that `ComponentId` is `int`, because I'm pretty sure it's not meant to store negative values. There's a real possibility of overflow.

I would think that `ComponentId` is `int` due to the intention of returning something like `-1` as an invalid ID. In fact, it looks like there's an alias that does exactly this: https://github.com/ignitionrobotics/ign-gazebo/blob/1117e5da9b71e6e2f05d7ceef47a6815c91cbec4/include/ignition/gazebo/Types.hh#L99-L100

However, looking at the code for `ComponentStorageBase`, it looks like it's not being used :thinking:: https://github.com/ignitionrobotics/ign-gazebo/blob/97935d96eac49e76bab5c28603eb1db5d77cfb20/include/ignition/gazebo/detail/ComponentStorageBase.hh#L158

> Some options are changing `ComponentId`'s type to match `Entity`'s, or change `ComponentKey` to hold an `Entity` instead. I think both options should be seamless for users who are being strict about using the aliases. What do you think?

Since either one of these approaches will result in a "breaking" change, I'd prefer changing `ComponentKey` to hold an `Entity` so that the code is easier to read/understand. If we're going to make a change, we might as well make the one that produces the clearest results :nerd_face: But if anyone else has a good reason for changing `ComponentId`'s type to match `Entity`'s type, I'd be interested to hear it.

chapulina:
> What if we re-named it to EntityComponentStorage? 

Done 5bb0efc

> I'd prefer changing ComponentKey to hold an Entity

Done 5bb0efc

adlarkin (approved):
Thanks for the suggestions! I think this makes the transition a smoother process.

As we discussed offline, I went ahead and renamed `ComponentStorage` to `EntityComponentStorage` in https://github.com/ignitionrobotics/ign-gazebo/pull/856/commits/d3e5a25e232210782a8c61319365f1639fbde0fb to keep authorship of those files - so, once you merge in the latest changes from #856 and address the comment I left below, I think we can merge this :+1:

src/EntityComponentManager_TEST.cc
```diff
@@ -2561,6 +2562,77 @@ TEST_P(EntityComponentManagerFixture, RemovedComponentsSyncBetweenServerAndGUI)
...
+      DoubleComponent(0.123)));
+  ComponentKey cDoubleEDouble = {DoubleComponent::typeId, eDouble};
+
+  EXPECT_NE(nullptr, manager.CreateComponent<IntComponent>(eIntDouble,
+      IntComponent(456)));
+  ComponentKey cIntEIntDouble = {IntComponent::typeId, eIntDouble};
+
+  EXPECT_NE(nullptr, manager.CreateComponent<DoubleComponent>(eIntDouble,
+      DoubleComponent(0.456)));
+  ComponentKey cDoubleEIntDouble = {DoubleComponent::typeId, eIntDouble};
```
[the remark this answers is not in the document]
chapulina: Added in b5b03d4

Issue #871, opened 2021-06-21
Pose updates for nested models are incorrect


shameekganguly:
## Environment
* OS Version: Ubuntu
* Source: Built from latest bazel branches

## Description
A common model composition paradigm used in our org is to model robot arms and end-effectors separately and connect them 
together with fixed joints, as such:
```
<model name = "parent_model">
    <link name="parent_link">...</link>
    <include><uri>model://child</uri> <name> child_model </name></include>
    <joint name="parent_joint">
        <parent> parent_model </parent>
        <child> parent_model </child>
    </joint>
</model>
```
However, pose updates seem to be wrong for the included model since updating to Ignition Edifice. I have a simple test sdf file with two double pendulums (video attached). Both pendulums are planar and have the base link fixed to the world, followed by revolute joint - link - revolute joint - link. The left pendulum is modeled with a nested model, i.e.  model{revolute joint - link} - revolute joint - link, whereas the right pendulum is modeled as a flat list.

The link poses are correct for the right pendulum, whereas they are wrong for some of the links for the left nested pendulum.

## Steps to reproduce
```
blaze run -c opt ign_tools:ign -- gazebo -r -v 4 test_pendulums.txt
```

## Output
[test_pendulums.txt](https://github.com/ignitionrobotics/ign-gazebo/files/6689063/test_pendulums.txt)
https://user-images.githubusercontent.com/2412842/122818055-92cc0d00-d28d-11eb-926c-ccdc9a169f4c.mov
<img width="1363" alt="Screen Shot 2021-06-21 at 12 39 20 PM" src="https://user-images.githubusercontent.com/2412842/122818208-c4dd6f00-d28d-11eb-82e9-0903cc3af9e6.png">

PR #876, opened 2021-06-23
Physics system: update link poses if the canonical link pose has been updated

adlarkin:
Signed-off-by: Ashton Larkin <ashton@openorobotics.org>

# 🦟 Bug fix

* Related to #871
* Fixes https://github.com/ignitionrobotics/ign-physics/issues/264#issuecomment-865123602
* Fixes https://github.com/ignitionrobotics/ign-gazebo/pull/258#discussion_r468990490 in Dome (this has already been fixed in Edifice; see https://github.com/ignitionrobotics/ign-gazebo/pull/685)

## Summary
As pointed out in https://github.com/ignitionrobotics/ign-physics/issues/264#issuecomment-865123602, if the pose of a model's canonical link changes, then the poses for all other links in this model must be updated (even if these links haven't moved) since these other link poses are given w.r.t. the model whose pose was just updated. I appeared to have removed this functionality in #669 (on accident, of course :slightly_smiling_face:) when working on physics system performance updates, so I am adding it back in this PR. I've also added logic (and test coverage) to support pose updates for canonical links that are **not** the first link of a model (this has already been fixed in Edifice, but wasn't fixed for Dome yet).

A note about nested models: since nested model support is limited in Dome, I wasn't able to write tests that check for pose updates of nested models in this PR. I have left a [TODO note](https://github.com/ignitionrobotics/ign-gazebo/blob/3e8a08b61c6d06d4ef104403986b0b1a9d86f097/test/worlds/only_canonical_link_moves.sdf#L134-L137) for myself to add this test coverage when we forward port these changes to Edifice. Since we don't have test coverage for nested models yet, we should probably leave #871 open until this PR is forward ported to Edifice and nested model test coverage is included.

It's also worth noting that the physics system in Edifice/Dome has a few differences that were made for performance. This means that we may have some conflicts to resolve when porting this forward - I'm expecting that the implementation for this bug fix in Edifice/Fortress will differ somewhat from the Dome implementation (I can handle merging this bug fix forward once this PR is merged).

## Checklist
- [x] Added tests


test/integration/physics_system.cc
```diff
@@ -987,3 +1070,25 @@ TEST_F(PhysicsSystemFixture, NestedModel)
...
   EXPECT_EQ("model_01", parentIt->second);
 }
+
+/////////////////////////////////////////////////
+// This tests whether pose updates are correct for a model whose canonical link
+// changes, but other links do not
+TEST_F(PhysicsSystemFixture, MovingCanonicalLinkOnly)
+{
+  TestStaticNonCanonicalLinkUpdates(
+      "/test/worlds/static_non_canonical_link.sdf");
```
ahcorde, line 1080: ignition::common::JoinPaths

test/integration/physics_system.cc
```diff
@@ -987,3 +1070,25 @@ TEST_F(PhysicsSystemFixture, NestedModel)
...
+// This tests whether pose updates are correct for a model whose canonical link
+// changes, but other links do not.
+//
+// This is similar to the MovingCanonicalLinkOnly test, with the difference in
+// this test being that the canonical link for the model in this test was
+// manually specified by the user instead of defaulting to the first link in the
+// model.
+TEST_F(PhysicsSystemFixture, MovingCustomCanonicalLinkOnly)
+{
+  TestStaticNonCanonicalLinkUpdates("/test/worlds/custom_canonical_links.sdf");
```
ahcorde, line 1093: ignition::common::JoinPaths

test/integration/physics_system.cc
```diff
@@ -71,6 +73,87 @@ class PhysicsSystemFixture : public ::testing::Test
...
+  /// means that the model's pose changed).
+  /// \param[in] _fileName The name of the file to test.
+  /// \note This method is to avoid code duplication so that the same models can
+  /// be tested with different links as the canonical link.
+  protected: void TestStaticNonCanonicalLinkUpdates(
+                 const std::string &_fileName)
+  {
+    ignition::gazebo::ServerConfig serverConfig;
+
+    const auto sdfFile = std::string(PROJECT_SOURCE_PATH) + _fileName;
```
ahcorde, line 90: ignition::common::JoinPaths

ahcorde (changes requested):

azeey:
FYI, https://github.com/ignitionrobotics/ign-physics/issues/264 should be partially fixed by https://github.com/ignitionrobotics/ign-physics/pull/268

test/integration/physics_system.cc
adlarkin: https://github.com/ignitionrobotics/ign-gazebo/pull/876/commits/db748c0c700e7153f736e632234ab707a29bf1d7

adlarkin:
@azeey, this is ready for review now. I also left a note in the PR description about not being able to test nested models for this in Dome, as we discussed offline.

include/ignition/gazebo/Util.hh
```diff
@@ -140,6 +141,14 @@ namespace ignition
         const Entity &_entity,
         const EntityComponentManager &_ecm);
 
+    /// \brief Get all of the links that belong to a model. If the model has
+    /// nested models, the links for the nested models will also be retrieved.
+    /// \param[in] _model The model to retrieve links from.
+    /// \param[in] _ecm The entity component manager.
+    /// \return All of the links that belong to _model.
+    std::unordered_set<Entity> allModelLinks(const Entity &_model,
+        const EntityComponentManager &_ecm);
```
diegoferigo, line 150: What about adding a new method in `Model` that complements the following but works with nested models? It could be a useful helper there, I fear that adding it to `Util.hh` would not get enough visibility and could get lost.

https://github.com/ignitionrobotics/ign-gazebo/blob/5d2c076dfd16886146d62df2ac0bc8e8dea5fdc8/include/ignition/gazebo/Model.hh#L148-L152

adlarkin: Thanks for pointing out this method in `Model` - I didn't even think to check this class when working on the PR originally :upside_down_face: so, if I understand correctly, you're recommending that I should move the `allModelLinks` method from `Util.hh` to `Model.hh`? If so, I think that makes sense.

diegoferigo: Yes, as far as I know, all these classes (`World`, `Model`, `Link`, ...) are abstractions of the ECM. From my point of view, they are useful to 1) act as an intermediate layer detaching the low-level logic that interfaces directly with the ECM so that this logic can be updated / optimized in upstream and directly propagated downstream, and 2) expose high-level functionalities that need getting and processing raw component data (e.g. the [`Link::AddWorldWrench`](https://github.com/ignitionrobotics/ign-gazebo/blob/5d2c076dfd16886146d62df2ac0bc8e8dea5fdc8/include/ignition/gazebo/Link.hh#L245-L253)).

Maybe also this new functionality to get all the links considering a nested topology could be useful to complement the existing `Model::Links`. If you agree with it, maybe moving the function from `Util.hh` to the `Model` class might makes sense.

src/systems/physics/Physics.cc
```diff
@@ -1923,34 +1978,47 @@ void PhysicsPrivate::UpdateSim(EntityComponentManager &_ecm)
...
-            // Unlike canonical links, pose of regular links can move relative
-            // to the parent. Same for links inside nested models.
-            *_pose = components::Pose(math::eigen3::convert(worldPose) +
-                                      parentWorldPose.Inverse());
-            _ecm.SetChanged(_entity, components::Pose::typeId,
-                ComponentState::PeriodicChange);
+          }
+          // don't update the link's pose if it was already updated due to a
+          // model pose upate
+          else if (linkPoseChanged.find(_entity) == linkPoseChanged.end())
```
azeey, line 2018: If the canonical link of the model is not the first link in the model (which can happen if the `canonical_link` attribute is used), the non-canonical links get updated first and then when it's time for the canonical link to get updated, all the non-canonical links get updated again. This fine in terms of correctness, but it might be a performance hit. If we want to fix that, we should update just the canonical links first and then the rest of the links. 

In fact, if we are looping through all the links of a model above, I don't know if there's even a need for this `UpdateRelativeLinkPose` here.

adlarkin: I went ahead and updated the canonical links first in https://github.com/ignitionrobotics/ign-gazebo/pull/876/commits/862f0b7dc93b332b8890621e2b4d982fe51688ce. It's not the prettiest solution since we don't have the `ModelCanonicalLink` component that was introduced in #685, but this should be easier to do once we forward-port to Edifice since this component is available starting in Edifice.

As a result of this change, the `UpdateRelativeLinkPose` method is no longer needed, so I removed it.

src/systems/physics/Physics.cc
```diff
@@ -1861,13 +1875,49 @@ ignition::math::Pose3d PhysicsPrivate::RelativePose(const Entity &_from,
...
+    ignerr << "The parent component of " << _topLevelModel
+           << " could not be found. This should never happen!\n";
+    return;
+  }
+  math::Pose3d parentWorldPose =
+      this->RelativePose(worldComp->Data(), _linkParent->Data(), _ecm);
+
+  // Unlike canonical links, pose of regular links can move relative
+  // to the parent. Same for links inside nested models.
+  *_linkPose = components::Pose(_linkWorldPose +
```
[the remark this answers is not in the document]
adlarkin: Switched to `operator*` in https://github.com/ignitionrobotics/ign-gazebo/commit/862f0b7dc93b332b8890621e2b4d982fe51688ce

include/ignition/gazebo/Util.hh
adlarkin: Thanks @diegoferigo and @azeey for your inputs. I made some changes that address https://github.com/ignitionrobotics/ign-gazebo/pull/876#discussion_r660714580 in https://github.com/ignitionrobotics/ign-gazebo/commit/862f0b7dc93b332b8890621e2b4d982fe51688ce, and as a result, I actually don't need the `allModelLinks` method anymore. So, I believe we can mark this as resolved, unless either of you think it would still be worth adding a method like this to the `Model` class (but since this would now be functionality that's unrelated to this PR, perhaps we should open a separate PR to do this if we decide to add it in).

azeey: If it's not needed here, opening a separate PR sounds good.

azeey (approved):
LGTM! I tested it using the attached `shapes.sdf` where I modified all the links to be in one model , set the canonical link, and changed the poses of the links.

[shapes.sdf.txt](https://github.com/ignitionrobotics/ign-gazebo/files/6785721/shapes.sdf.txt)

test/integration/physics_system.cc
```diff
@@ -33,6 +34,7 @@
 
 #include "ignition/gazebo/Server.hh"
 #include "ignition/gazebo/SystemLoader.hh"
+#include "ignition/gazebo/Util.hh"
```
ahcorde, line 37: alphabetize ?

ahcorde (approved):
just a small comment

test/integration/physics_system.cc
adlarkin: From my understanding, I thought that upper case (in this case, `Util`) comes before lowercase (`test_config`) because of how case ordering works in the [ASCII table](https://www.asciitable.com/) (upper case comes before lower case). However, I still haven't found a definitive guide/resource that confirms this. Does anyone else know how alphabetization should be handled for different-case letters?

diegoferigo: Maybe checking how popular tools like clang-format are implemented could give a direction? My usual configuration keeps the order as it is right now, i.e. with the lower case headers below the upper case one.

adlarkin: > Maybe checking how popular tools like clang-format are implemented could give a direction?

It looks like there are [several options](https://clang.llvm.org/docs/ClangFormatStyleOptions.html) in clang-format (see the section titled `SortIncludes`), and I'm not sure which one is default. I'll stick to `CaseSensitive` since that is what I have seen elsewhere.

> Btw, having a project-wise style configuration would be a dream becoming true :smile: 

We actually had a discussion about this today in our internal simulation meeting. We use [`cppcheck`](https://github.com/danmar/cppcheck/) which catches some styling, but not all. We discussed using something like clang, but @chapulina mentioned that she's tried this before, and that it took quite some time to run (I believe it took a few hours to complete :exploding_head:). So, if you have any thoughts about style configurations we could use, we're open to suggestions :smile: we can open a separate issue to discuss this.

PR #881, opened 2021-06-24
Scale mode - Part2

caguero:
# 🎉 New feature

## Summary

Adds a serializer to the `ModelSDF` component.

## Test it

It depends on pull request #799. 

## Checklist
- [X] Added tests


include/ignition/gazebo/components/Model.hh
```diff
@@ -29,14 +32,64 @@ namespace gazebo
...
+    /// \param[out] _model Model to populate
+    /// \return The stream.
+    public: static std::istream &Deserialize(std::istream &_in,
+                sdf::Model &_model)
+    {
+      sdf::Root root;
+      std::string sdf(std::istreambuf_iterator<char>(_in), {});
+
+      sdf::Errors errors = root.LoadSdfString(sdf);
+      if (!errors.empty())
```
jennuine, line 71: It may be better to check `root.Element()->GetElement("model") == nullptr` instead since errors can happen but depending on the error the model could still be loaded anyway. For example, in parameter passing if a user tries to update a parameter from an original model incorrectly then `errors` will be nonempty but those updates are skipped and the original model is still loaded (assuming no other errors occurred)

include/ignition/gazebo/components/Model.hh
```diff
@@ -29,14 +32,64 @@ namespace gazebo
 {
 // Inline bracket to help doxygen filtering.
 inline namespace IGNITION_GAZEBO_VERSION_NAMESPACE {
+namespace serializers
+{
+  class SdfModelSerializer
```
jennuine, line 37: Is this better suited in [Serialization.hh](https://github.com/ignitionrobotics/ign-gazebo/blob/ign-gazebo4/include/ignition/gazebo/components/Serialization.hh)?

caguero: I don't have a strong opinion about it because I see examples in both places, `Serialization.hh` and directly in the component. I think I'd go for an approach where each component is self-contained and includes its serialization if possible. `Serialization.hh` could contain any generic serialization that can be reused across many components, such as `MsgSerializer` for example.

include/ignition/gazebo/components/Model.hh
caguero: Thanks, good point! Changed in f9e2626.

include/ignition/gazebo/components/Model.hh
jennuine: Sounds good to me!

jennuine (approved):
LGTM, just need to resolve conflicts

include/ignition/gazebo/components/Model.hh
```diff
@@ -29,14 +32,64 @@ namespace gazebo
...
+    /// \param[out] _model Model to populate
+    /// \return The stream.
+    public: static std::istream &Deserialize(std::istream &_in,
+                sdf::Model &_model)
+    {
+      sdf::Root root;
+      std::string sdf(std::istreambuf_iterator<char>(_in), {});
+
+      sdf::Errors errors = root.LoadSdfString(sdf);
+      if (!root.Element()->GetElement("model"))
```
azeey, line 71: `GetElement` will always return a valid pointer. Can you use `HasElement` instead?

azeey (changes requested):
One quick comment

include/ignition/gazebo/components/Model.hh
jennuine: Oh, good catch! I forgot about that :bowing_woman:

caguero: Thanks! Updated in 0dd2af3

Issue #885, opened 2021-06-25
New release for Edifice


Tobias-Fischer:
Hi there,
We are in the process of packaging Edifice in conda-forge (see https://github.com/conda-forge/libignition-gazebo-feedstock/pull/5). At the moment Windows fails, which I think is because https://github.com/ignitionrobotics/ign-gazebo/pull/758 is not contained in the 5.0.0 release. Any chance to release a new version so we don't need to apply patches?

/cc @traversaro

chapulina:
Sure thing, I'll work on a release as soon as https://github.com/ignitionrobotics/ign-gazebo/pull/877 is in :+1:

chapulina:
The GitHub release is out, debs in the oven: https://build.osrfoundation.org/job/ign-gazebo5-debbuilder/

Tobias-Fischer:
Many thanks @chapulina!

PR #877, opened 2021-06-23
4 to 5

ahcorde:
# ➡️ Forward port

Port 4 to 5

Branch comparison: https://github.com/ignitionrobotics/ign-gazebo/compare/ign-gazebo5...ign-gazebo4

**Note to maintainers**: Remember to **Merge** with commit (not squash-merge or rebase)

Require https://github.com/osrf/sdformat/pull/598

PR #892, opened 2021-06-28
🎈 5.1.0~pre1

chapulina:
# 🎈 Release

Preparation for 5.1.0~pre1 release.

Comparison to 5.0.0: https://github.com/ignitionrobotics/ign-gazebo/compare/ignition-gazebo5_5.0.0...ign-gazebo5

The current debs are crashing when trying to load poses from SDF (https://github.com/ignitionrobotics/ign-gazebo/issues/891). If it's an ABI issue, this pre-release may fix it.

See #885

## Checklist
- [x] Asked team if this is a good time for a release
- [x] There are no changes to be ported from the previous major version - there's always something, a big pass was done in #877
- [x] No PRs targeted at this major version are close to getting in - this release is already large enough that I don't think we need to wait for changes to come in
- [x] Bumped minor for new features, patch for bug fixes
- [x] Updated changelog
- [x] Link to PR updating dependency versions in appropriate repository in [ignition-release](https://github.com/ignition-release) (as needed): https://github.com/ignition-release/ign-gazebo5-release/pull/11

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -38,7 +38,7 @@ endif()
 # Search for project-specific dependencies
 #============================================================================
 
-ign_find_package(sdformat11 REQUIRED VERSION 11.2)
+ign_find_package(sdformat11 REQUIRED VERSION 11.2.1)
 set(SDF_VER ${sdformat11_VERSION_MAJOR})
 
 #--------------------------------------
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,7 +14,7 @@ find_package(ignition-cmake2 2.8.0 REQUIRED)
 #============================================================================
 # Configure the project
 #============================================================================
-ign_configure_project(VERSION_SUFFIX)
+ign_configure_project(VERSION_SUFFIX pre1)
 
 #============================================================================
 # Set project-specific options
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,7 +3,7 @@ cmake_minimum_required(VERSION 3.10.2 FATAL_ERROR)
 #============================================================================
 # Initialize the project
 #============================================================================
-project(ignition-gazebo5 VERSION 5.0.0)
+project(ignition-gazebo5 VERSION 5.1.0)
 
 #============================================================================
 # Find ignition-cmake
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,6 +2,197 @@
 
 ### Ignition Gazebo 5.X.X (20XX-XX-XX)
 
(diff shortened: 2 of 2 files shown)


mjcarroll (approved):

chapulina:
Oh I just remembered another issue with the release, all the new test failures related to wheel friction.

adlarkin:
As we discussed offline, it seems like #891 is because of an ABI break in sdformat: https://github.com/osrf/sdformat/compare/sdformat11_11.1.0...sdf11

So, making a pre-release here won't resolve that, right? I think we need to update sdformat and make releases there?

scpeters:
> As we discussed offline, it seems like #891 is because of an ABI break in sdformat: [osrf/sdformat@sdformat11_11.1.0...sdf11](https://github.com/osrf/sdformat/compare/sdformat11_11.1.0...sdf11)
> 
> So, making a pre-release here won't resolve that, right? I think we need to update sdformat and make releases there?

usually when there's an ABI change, you can fix it by rebuilding the downstream packages, so I think rebuilding ign-gazebo could help

scpeters:
> > As we discussed offline, it seems like #891 is because of an ABI break in sdformat: [osrf/sdformat@sdformat11_11.1.0...sdf11](https://github.com/osrf/sdformat/compare/sdformat11_11.1.0...sdf11)
> > So, making a pre-release here won't resolve that, right? I think we need to update sdformat and make releases there?
> 
> usually when there's an ABI change, you can fix it by rebuilding the downstream packages, so I think rebuilding ign-gazebo could help

we might need to rebuild ignition-physics as well

chapulina:
We decided to make a new SDFormat release reverting the ABI breaking changes. I'll hold this release until that's out so we don't build against the broken ABI. CC @azeey .

This will also give @scpeters some time to look into the friction changes.

scpeters:
> We decided to make a new SDFormat release reverting the ABI breaking changes. I'll hold this release until that's out so we don't build against the broken ABI. CC @azeey .
> 
> This will also give @scpeters some time to look into the friction changes.

the test failures are already present on `ign-gazebo5`, and they are fixed in my local tests of https://github.com/ignitionrobotics/sdformat/pull/606. So I think they are related to the ABI break and should be fixed once we make a new libsdformat11 release

chapulina:
> So I think they are related to the ABI break

Ohhhh I didn't think of that because the ABI break has been crashing Gazebo on startup for me. But now that you said it I can imagine that it may manifest in different ways on CI.

Let's wait for the new SDF release then :pray:

chapulina:
I don't see new test failures :smile:

PR #895, opened 2021-06-29
Add blender sdf export tutorial

ddengster:
Signed-off-by: ddengster <ed.fan@osrfoundation.org>

# 🎉 New feature

## Summary
Adds a blender script to export models in the sdf format, accompanied with a simple tutorial as to how to use it.

## Checklist
- [ ] Added tests


examples/scripts/blender/sdf_exporter.py
```diff
@@ -0,0 +1,206 @@
...
+    static.text = "true"
+    link = ET.SubElement(model, "link", attrib={"name":"testlink"})
+    # for each geometry in geometry library add a <visual> tag
+    for o in mesh_objects:
+        visual = ET.SubElement(link, "visual", attrib={"name":o.name})
+
+        geometry = ET.SubElement(visual, "geometry")
+        mesh = ET.SubElement(geometry, "mesh")
+        uri = ET.SubElement(mesh, "uri")
+        uri.text = dae_filename
```
chapulina, line 47: It's common practice to place the meshes in a `meshes` directory and the textures in `materials/textures`. Would it be too troublesome to generate that directory structure?

examples/scripts/blender/sdf_exporter.py
```diff
@@ -0,0 +1,206 @@
...
+    name = ET.SubElement(model, 'name')
+    name.text = "L1"
+    version = ET.SubElement(model, 'version')
+    version.text = "1.0"
+    sdf_tag = ET.SubElement(model, "sdf", attrib={"sdf":"1.8"})
+    sdf_tag.text = "model.sdf"
+
+    author = ET.SubElement(model, 'author')
+    name = ET.SubElement(author, 'name')
+    name.text = "Generated by blender sdf tools"
```
chapulina, line 168:
```suggestion
    name.text = "Generated by blender SDF tools"
```

examples/scripts/blender/sdf_exporter.py
```diff
@@ -0,0 +1,206 @@
...
+    sdf_file = open(prefix_path+sdf_filename, "w")
+    sdf_file.write(reparsed.toprettyxml(indent="  "))
+    sdf_file.close()
+
+    ##############################
+    ### generate model.config ####
+    ##############################
+    model = ET.Element('model')
+    name = ET.SubElement(model, 'name')
+    name.text = "L1"
```
chapulina, line 160: Is there a way to get this string from within Blender? Otherwise, maybe we can call it something generic like `my_model`?

chapulina (approved):
Thanks for the script! It has a lot of hardcoded values for things like casting shadows and bitmasks, but I think that this is a nice starting point that users can adapt to their needs. It would be nice to move these values to the top of the script to call attention to the fact that they can be customized.

tutorials/blender_sdf_exporter.md
```diff
@@ -0,0 +1,21 @@
+\page blender_sdf_exporter Blender SDF Exporter
+
+Blender is a DCC tool to create 3d models. In some cases you may be using it to bake
```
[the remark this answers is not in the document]
ddengster: it stands for Digital Content Creation 3aaabae

examples/scripts/blender/sdf_exporter.py
ddengster: 3aaabae

examples/scripts/blender/sdf_exporter.py
ddengster: switch to `my_model`, float to top of function 3aaabae

examples/scripts/blender/sdf_exporter.py
ddengster: Moved the dae export to the meshes folder d6222f9

There is a problem that the textures have to be exported at the same location as the dae. At the moment the fix would entail lifting the entire collada export function and modifying it, plus the relocation of the texture files. I wouldn't do it within this script.

This is something I'll fix at the import stage, specifically for the collada world exporter, which places textures at the same location of the dae file.

chapulina: Got it, it's ok like that. It's more of a convention than a rule, Ignition supports any folder structure if the paths are declared correctly.

chapulina (approved):

PR #900, opened 2021-07-02
Make Video recorder plugin work with MinimalScene

ahcorde:

# 🎉 New feature

## Summary

This PR is part of the consolidation between Scene 3D in `ign-gazebo` and `ign-gui`. This PR adds a plugin to record the user camera.

It builds on top of #813

## Checklist
- [ ] Added tests


ahcorde:
Require: https://github.com/ignitionrobotics/ign-rendering/pull/358

ahcorde:

ahcorde:
@chapulina open for review

chapulina (approved):
@ahcorde , for expediency, I pushed some suggestions directly to this branch in 7650dda7343bad48a040dfa234eec5320904e025 instead of opening a PR against this branch. In summary:

* Revert changes to `gui.config`
* Merge `VideoRecorderLogic` into `VideoRecorder` like we're doing for `TransformControl`, with a `legacy` flag

Feel free to revert any of the changes I pushed as you see fit.

It seem to work for me with both `MinimalScene` and `GzScene3D`, but I'd appreciate it if @iche033 gives his ok when he's back. I'm mainly concerned about the extra camera update that we're doing in this plugin.

iche033:
> It seem to work for me with both MinimalScene and GzScene3D, but I'd appreciate it if @iche033 gives his ok when he's back. I'm mainly concerned about the extra camera update that we're doing in this plugin.

I think we may be able to skip the camera update in this plugin. Given that the render event is emitted after the `camera->Update()` call in the scene 3d plugin, the image should be ready when the VideoEncoder calls `AddFrame` in this plugin

chapulina:
> the render event is emitted after the camera->Update() call in the scene 3d plugin

Humm I think we may have a different issue then. Because the `MinimalScene` is not calling the render event in lockstep with simulation. This makes it hard for this plugin to record in lockstep...

ahcorde:
@chapulina @iche033 what about adding a new camera in the same position and orientation as the original one ? then we can do whatever we want with this camera, isn't it ?

chapulina:
> what about adding a new camera in the same position and orientation as the original one ?

That's an interesting idea :thinking: The new camera would need to track the original camera. And I imagine we'd have to protect against both cameras rendering at the same time, which I understand is not supported by `ign-rendering` / Ogre.

iche033:
> > what about adding a new camera in the same position and orientation as the original one ?

> That's an interesting idea thinking The new camera would need to track the original camera. And I imagine we'd have to protect against both cameras rendering at the same time, which I understand is not supported by ign-rendering / Ogre.

The other option for lockstepping is to spawn the new camera on the server side and do the recording there (instead of the recording from the gui). Maybe that's what you mean? You can probably tweak the [camera_video_recorder](https://github.com/ignitionrobotics/ign-gazebo/tree/main/src/systems/camera_video_recorder) system with a new option to record from the gui camera instead of a camera sensor. If user indicates to record from the gui camera, the system just creates a new camera with same configuration as the gui camera and tracks the gui camera's pose.

chapulina:
The new approach of synchronizing `Scene3D` and the `Sensors` system that @mjcarroll is thinking of implementing could allow lockstepping the GUI camera when running with `--same-process`.

src/gui/plugins/video_recorder/VideoRecorder.cc
```diff
@@ -195,17 +194,9 @@ void VideoRecorderPrivate::OnRender()
     if (t - this->updateTime != std::chrono::seconds(0))
     {
       this->updateTime = t;
-      update = true;
     }
   }
```
chapulina, line 198: I think we can remove this entire block, I don't think `updateTime` is being used for anything now

chapulina (approved):

PR #907, opened 2021-07-08
Toggle contact visuals from EntityContextMenu

atharva-18:
Signed-off-by: Atharva Pusalkar <atharvapusalkar18@gmail.com>

# 🎉 New feature
## Summary
Adds the ability to visualize contact sensor data from the entity context menu. The `Visualize Contacts` plugin is modified to only change the size of the markers.

contacts

## Test it
Test it by right-clicking an entity and selecting `View -> Contacts`. The `Visualize Contacts` plugin must be loaded to use this feature.

## Checklist
- [ ] Added tests


ahcorde:
I want to hear the opinion of @chapulina or @iche033 

To be able to see contacts in the scene, Ignition Gazebo requires to run the `VisualizeContacts` plugin. I would expect to see them without this plugin enabled such us Transparent, Wireframes, etc.

Maybe we can use the plugin just to change the parameters of the contact visual.

chapulina:
> I would expect to see them without this plugin enabled such us Transparent, Wireframes, etc.

Yeah me too. If the user doesn't load that plugin and tries to choose the option from the menu, it won't work and they don't know why. This could be confusing.

I like the idea of toggling the contacts from the menu, but with the current plugin I think it will be difficult to do it.

diegoferigo:
Is there any way to trigger a plugin load event from the GUI that is propagated to the server? The implementation seems fairly easy in the single-process setup that is ongoing, not sure about the current multiprocess situation.

ahcorde (changes requested):
Do you min to clean the COMMIT history of this PR ?

atharva-18:
> Do you min to clean the COMMIT history of this PR ?

Done [here](https://github.com/ignitionrobotics/ign-gazebo/compare/62de97b059600e19573c6f615d9df580bbbc5287..7df812027e75acb95726d8bf1707e8b203673252)

I am able to load the `VisualizeContacts` GUI plugin (if it is present) by doing
```
--- a/src/gui/plugins/modules/EntityContextMenu.cc
+++ b/src/gui/plugins/modules/EntityContextMenu.cc
@@ -167,6 +167,7 @@ void EntityContextMenu::OnRequest(const QString &_request, const QString &_data)
   else if (request == "view_contacts")
   {
+    gui::App()->LoadPlugin("VisualizeContacts");
     ignition::msgs::StringMsg req;
     req.set_data(_data.toStdString());
     this->dataPtr->node.Request(this->dataPtr->viewContactsService, req, cb);
   }
```
But I don't think that currently there is a way to check if the GUI plugin is already loaded or if the user has manually closed the plugin.

cc: @chapulina

src/gui/plugins/visualize_contacts/VisualizeContacts.hh
```diff
@@ -20,6 +20,9 @@
 
 #include <memory>
 
+#include <ignition/msgs/boolean.pb.h>
+#include <ignition/msgs/stringmsg.pb.h>
```
ahcorde, line 24: I think this will create some warnings on Windows

```suggestion
#ifdef _MSC_VER
#pragma warning(push, 0)
#endif
#include <ignition/msgs/boolean.pb.h>
#include <ignition/msgs/stringmsg.pb.h>
#ifdef _MSC_VER
#pragma warning(pop)
#endif
```

ahcorde (changes requested):

src/gui/plugins/visualize_contacts/VisualizeContacts.hh
atharva-18: Fixed in [`76a7154`](https://github.com/ignitionrobotics/ign-gazebo/pull/907/commits/76a715450d59016d893b0eb657b489562e8acfb7)

ahcorde (changes requested):
To try to simplify things. I think we should use `Contacts` as the other visualization features (wireframe, joints, etc) without loading any other plugin and then if the user wants open the Visualize Contact plugin to modify the contact visual. Maybe we can modify the name of the plugin to "Config Visualize Contacts", what do you think @atharva-18 ?

atharva-18:
> To try to simplify things. I think we should use `Contacts` as the other visualization features (wireframe, joints, etc) without loading any other plugin and then if the user wants open the Visualize Contact plugin to modify the contact visual. Maybe we can modify the name of the plugin to "Config Visualize Contacts", what do you think @atharva-18 ?

Sounds good, however, we still need to have the plugin to publish the marker positions - 
https://github.com/ignitionrobotics/ign-gazebo/blob/76a715450d59016d893b0eb657b489562e8acfb7/src/gui/plugins/visualize_contacts/VisualizeContacts.cc#L241-L262

And enable collision data - 
https://github.com/ignitionrobotics/ign-gazebo/blob/76a715450d59016d893b0eb657b489562e8acfb7/src/gui/plugins/visualize_contacts/VisualizeContacts.cc#L271-L299

We can move these to Scene3D, but the [`Update Period`](https://github.com/ignitionrobotics/ign-gazebo/blob/76a715450d59016d893b0eb657b489562e8acfb7/src/gui/plugins/visualize_contacts/VisualizeContacts.cc#L317) feature would probably need some additional work.

ahcorde:
> We can move these to Scene3D, but the [`Update Period`](https://github.com/ignitionrobotics/ign-gazebo/blob/76a715450d59016d893b0eb657b489562e8acfb7/src/gui/plugins/visualize_contacts/VisualizeContacts.cc#L317) feature would probably need some additional work.

Don't worry for now about the `Update Period`.

chapulina (changes requested):
I'm not convinced that it is a good idea to add this option to the entity context menu, because:

1. The contact visualization is being enabled for the whole world, not each entity.
2. `VisualizeContacts` needs to be open for it to work. So
    1. If the user doesn't have it open, nothing will happen and they won't understand why. Or they may think there are no contacts!
    2. If the user has it open anyway, the user can use this plugin and doesn't need to use the context menu.

Thanks for the PR @atharva-18 , but I think we shouldn't get this in. Are you ok closing it or do you see any advantages that I missed?

atharva-18:
Thank you @chapulina for looking into this. You have made fair points regarding the implementation. I see no particular advantages of using the context menu over the current plugin. The plugin anyway turns on contact visualization for the entire world. So using an entity-specific menu to do the same would be absurd and redundant.

I think we can close this PR.

src/gui/plugins/modules/EntityContextMenu.qml
```diff
@@ -100,6 +109,7 @@ Item {
     if (context.type == "model" || context.type == "link")
     {
       viewCollisionsMenu.enabled = true;
+      viewContactsMenu.enabled = true;
```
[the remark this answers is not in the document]
atharva-18: Addressed in https://github.com/ignitionrobotics/ign-gazebo/pull/907#issuecomment-991080660

chapulina:
Thanks for the PR anyway, maybe we can revisit it if we ever have a way of toggling contacts for each entity separately.

PR #913, opened 2021-07-12
Fix crash on GUI entity removal with levels

adlarkin:

# 🦟 Bug fix

Fixes https://github.com/ignitionrobotics/ign-gazebo/pull/856#pullrequestreview-686953945

## Summary
After debugging the segfault pointed out in the link above, it seems like the levels system still tries to add components to entities even after they're removed (I'm not 100% certain if levels is to blame for this, but when I removed an entity in the `shapes.sdf` world, I did not see this behavior). This caused a segfault in #856 because [components are not created if an entity doesn't exist](https://github.com/ignitionrobotics/ign-gazebo/blob/ee3d3d08fe81ced1edc81f48f5792f166d7a1578/src/EntityComponentManager.cc#L599-L602), but the non existing entity (and component) were still [marked as modified](https://github.com/ignitionrobotics/ign-gazebo/blob/ee3d3d08fe81ced1edc81f48f5792f166d7a1578/src/EntityComponentManager.cc#L587-L590). This "mismatch" in data is problematic when [`EntityComponentManagerPrivate::CalculateStateThreadLoad`](https://github.com/ignitionrobotics/ign-gazebo/blob/ee3d3d08fe81ced1edc81f48f5792f166d7a1578/src/EntityComponentManager.cc#L1023) is called in [`EntityComponentmanager::State`](https://github.com/ignitionrobotics/ign-gazebo/blob/ee3d3d08fe81ced1edc81f48f5792f166d7a1578/src/EntityComponentManager.cc#L1089), so, this PR adds a check when calling `EntityComponentManager::CreateComponentImplementation` to make sure that the entity exists.

While the bug fix here ended up being a simple one, I found something interesting when comparing #856 to the `main` branch. In the `main` branch, components are always created, regardless of whether an entity exists or not (I was also able to confirm this by testing the levels world that caused a segfault in #856 - as you can see, there's no checking for if the entity exists): https://github.com/ignitionrobotics/ign-gazebo/blob/c07b49c51caae6da1566c383d2b6ff91f4cb8097/src/EntityComponentManager.cc#L530-L566

So, on the `main` branch, we don't see this segfault since both 1) the component is created, and 2) the (non existent) entity and the component are marked as modified. This actually seems like incorrect behavior to me, and I believe that we should fix `main` (and other released versions as well) to prevent creating components for entities that do not exist. @iche033, what do you think?

## Checklist
- [ ] Added tests

### DIFF
--- a/src/EntityComponentManager_TEST.cc
+++ b/src/EntityComponentManager_TEST.cc
@@ -250,6 +250,27 @@ TEST_P(EntityComponentManagerFixture, EntitiesAndComponents)
   EXPECT_TRUE(manager.EntityHasComponentType(entity, IntComponent::typeId));
   EXPECT_EQ(123, intComp->Data());
 
+  // Try to create/query a component from an entity that does not exist. nullptr
+  // should be returned since a component cannot be attached to a non-existent
+  // entity
+  EXPECT_FALSE(manager.HasEntity(kNullEntity));
+  /*
+   * TODO(adlarkin) test with the following methods once the deprecation
+   * process has been sorted out:
+   *  EntityHasComponent
+   *  EntityHasComponentType
+   */
+  EXPECT_EQ(nullptr, manager.CreateComponent<IntComponent>(kNullEntity,
+        IntComponent(123)));
+  EXPECT_EQ(nullptr, manager.ComponentDefault<IntComponent>(kNullEntity, 123));
+  EXPECT_EQ(nullptr, manager.Component<IntComponent>(kNullEntity));
+  EXPECT_FALSE(manager.ComponentData<IntComponent>(kNullEntity).has_value());
+  EXPECT_EQ(ComponentState::NoChange, manager.ComponentState(kNullEntity,
+        IntComponent::typeId));
+  // (make sure the entity wasn't implicitly created during the invalid
+  // component calls)
+  EXPECT_FALSE(manager.HasEntity(kNullEntity));
+
   // Remove all entities
   manager.RequestRemoveEntities();
   EXPECT_EQ(3u, manager.EntityCount());
--- a/include/ignition/gazebo/EntityComponentManager.hh
+++ b/include/ignition/gazebo/EntityComponentManager.hh
@@ -181,7 +181,8 @@ namespace ignition
       /// the component.
       /// \param[in] _data Data used to construct the component.
       /// \return A pointer to the component that was created. nullptr is
-      /// returned if the component was not able to be created.
+      /// returned if the component was not able to be created. If _entity
+      /// does not exist, nullptr will be returned.
       public: template<typename ComponentTypeT>
               ComponentTypeT *CreateComponent(
                   const Entity _entity,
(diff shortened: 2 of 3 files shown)


src/EntityComponentManager_TEST.cc
```diff
@@ -250,6 +250,27 @@ TEST_P(EntityComponentManagerFixture, EntitiesAndComponents)
...
+  // Try to create/query a component from an entity that does not exist. nullptr
+  // should be returned since a component cannot be attached to a non-existent
+  // entity
+  EXPECT_FALSE(manager.HasEntity(kNullEntity));
+  /*
+   * TODO(adlarkin) test with the following methods once the deprecation
+   * process has been sorted out:
+   *  EntityHasComponent
+   *  EntityHasComponentType
+   */
```
adlarkin, line 262: I'll address this once #873 is merged.

iche033:
I think the diff_drive plugin is creating various vel cmd components on every [PreUpdate](https://github.com/ignitionrobotics/ign-gazebo/blob/ign-gazebo5/src/systems/diff_drive/DiffDrive.cc#L390) for entities that no longer exist hence causing the crash. We should probably fix the diff drive plugin too.

Maybe print an error in `CreateComponentImplementation` if users are trying to create components for entities that do not exist?

>  I believe that we should fix main (and other released versions as well) to prevent creating components for entities that do not exist. @iche033, what do you think?

yeah that sounds good to me.

src/EntityComponentManager_TEST.cc
adlarkin: This has been resolved in #873

adlarkin:
> Maybe print an error in CreateComponentImplementation if users are trying to create components for entities that do not exist?

Done in https://github.com/ignitionrobotics/ign-gazebo/pull/913/commits/fd46a05363eec3fd4e9290279628e11583fda43f

---

I have also updated the diff_drive plugin to skip removed entities in https://github.com/ignitionrobotics/ign-gazebo/pull/927. Once that change gets forward-ported to main, we should be able to merge this PR (I have started the forward-port process in https://github.com/ignitionrobotics/ign-gazebo/pull/933).

adlarkin:
> We should probably fix the diff drive plugin too.

@iche033, the diff drive plugin should now be fixed in all released versions (#927 has been forward-ported from Citadel to Fortress). Would you mind testing the bug you found in https://github.com/ignitionrobotics/ign-gazebo/pull/856#pullrequestreview-686953945 one last time before I merge it?

iche033 (commented):
CI caught a compilation error and I can reproduce it locally. Looks like it could be introduced due to recent changes in the target branch

```
EntityComponentManager_TEST.cc:233:3:   required from here
/home/osrf/code/ign_f_ws2/src/ign-gazebo/test/gtest/include/gtest/gtest.h:1527:11: error: no match for ‘operator==’ (operand types are ‘const std::pair<long unsigned int, long unsigned int>’ and ‘ignition::gazebo::v6::components::Component<int, ignition::gazebo::v6::components::IntComponentTag>* const’)
   if (lhs == rhs) {
```

adlarkin:
> CI caught a compilation error and I can reproduce it locally

Oops, it looks like that came from https://github.com/ignitionrobotics/ign-gazebo/pull/913/commits/158a1af2efcf6be5b466e1b1f0acd74cfd2aac79 :sweat_smile: it should be fixed in https://github.com/ignitionrobotics/ign-gazebo/pull/913/commits/713c848b53818bed1de5bd5ccf3305a47d0776ad. I just tested the levels world, and removing entities seemed to work fine for me. Just as a note, I typically just build gazebo from source and then use the binary dependencies for PR testing, but I found that I also had to build `ign-launch` from source as well. I believe it's because of the changes that were made in `include/ignition/gazebo/components/Factory.hh` in #856.

iche033 (approved):

PR #921, opened 2021-07-17
Make Gazebo aware of SetCameraPassCountPerGpuFlush

darksylinc:
# 🎉 New feature

See ignitionrobotics/ign-rendering/issues/323

## Summary

ign-rendering has a new feature aimed to improve performance but also fixes a few bugs related to particle simulations; which is controlled via `Scene::SetCameraPassCountPerGpuFlush`.

A value of 0 will use the legacy mode (i.e. previous behavior) to ease porting from older version.
A value > 0 controls performance vs RAM tradeoff (note that it is an upper limit; a very high value doesn't immediately waste RAM. See `Scene::SetCameraPassCountPerGpuFlush` documentation for details)

This PR depends on https://github.com/ignitionrobotics/ign-rendering/pull/353 **otherwise it will not compile**.

## Test it

Just run the code again. This PR changes SetCameraPassCountPerGpuFlush explicitly to 6 to take advantage of the new feature. Set it to 0 for legacy.

As for performance, there won't be _yet_ measurable differences because most sensors are currently downloading rendering data immediately after submitting rendering; which forces a flush per sensor. The final improvement will come when we make use of AsyncTickets in Ogre 2.2

However after the PR it can be immediately tested that particles FXs involving GpuRays behave differently, because pre-PR each cubemap face would be rendered at different timestamps; post-PR they are all rendered at the same timestamp (even in legacy mode).

## Checklist
- [ ] Added tests

### DIFF
--- a/src/gui/plugins/scene3d/Scene3D.cc
+++ b/src/gui/plugins/scene3d/Scene3D.cc
@@ -2019,7 +2019,7 @@ void IgnRenderer::Initialize()
 
   auto root = scene->RootVisual();
 
-  scene->SetNumCameraPassesPerGpuFlush( 6u );
+  scene->SetCameraPassCountPerGpuFlush( 6u );
 
   // Camera
   this->dataPtr->camera = scene->CreateCamera();
--- a/src/gui/plugins/scene3d/Scene3D.cc
+++ b/src/gui/plugins/scene3d/Scene3D.cc
@@ -2019,7 +2019,7 @@ void IgnRenderer::Initialize()
 
   auto root = scene->RootVisual();
 
-  scene->SetCameraPassCountPerGpuFlush( 6u );
+  scene->SetCameraPassCountPerGpuFlush(6u);
 
   // Camera
   this->dataPtr->camera = scene->CreateCamera();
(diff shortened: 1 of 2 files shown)


src/gui/plugins/scene3d/Scene3D.cc
```diff
@@ -2019,6 +2019,8 @@ void IgnRenderer::Initialize()
 
   auto root = scene->RootVisual();
 
+  scene->SetCameraPassCountPerGpuFlush( 6u );
```
ahcorde, line 2022:
```suggestion
  scene->SetCameraPassCountPerGpuFlush(6u);
```

src/systems/sensors/Sensors.cc
```diff
@@ -196,6 +196,7 @@ void SensorsPrivate::WaitForInit()
       igndbg << "Initializing render context" << std::endl;
       this->renderUtil.Init();
       this->scene = this->renderUtil.Scene();
+      this->scene->SetCameraPassCountPerGpuFlush( 6u );
```
ahcorde, line 199:
```suggestion
      this->scene->SetCameraPassCountPerGpuFlush(6u);
```

ahcorde (changes requested):
style

iche033:

ahcorde (approved):

iche033 (approved):

PR #925, opened 2021-07-21
Place a component's removed flag in the ECM, not in the component itself

adlarkin:


chapulina (approved):
Sweeeet! :lollipop: Thanks for sorting that out!

PR #935, opened 2021-07-27
Functions to enable velocity and acceleration checks on Link

chapulina:
# 🎉 New feature

## Summary

This is broken off #926 for easier review.

The `Link` class provides convenient functions to retrieve velocities and accelerations without needing to touch the ECM directly. But in order to use those functions, users need to create the necessary components. This defeats the purpose of having an API that hides away the ECM's complexity.

This PR adds the `Enable(Velocity|Acceleration)Checks` functions, which provide a way for users of the `Link` API to create those components without necessarily knowing that they're doing so.

## Test it

I used the new functions in some existing plugins to demonstrate how it comes in handy for plugin developers, see below. The tests and documentation should also be helpful to understand the intended usage.

## Checklist
- [x] Added tests


src/Link.cc
```diff
@@ -44,6 +44,23 @@ class ignition::gazebo::LinkPrivate
 using namespace ignition;
 using namespace gazebo;
 
+//////////////////////////////////////////////////
+// Helper function to create and remove components to enable and disable
+// functionality.
+template <class ComponentType>
```
arjo129, line 50: I'm just wondering if it would make sense to make this API public. There are many use cases for `enableComponent<T>` apart from just `LinearVelocity`. I'm not sure where the best place is to put it but this pattern is prevalent throughout almost all plugins. It'd be far cleaner to just expose this.

src/systems/kinetic_energy_monitor/KineticEnergyMonitor.cc
```diff
@@ -127,40 +127,13 @@ void KineticEnergyMonitor::Configure(const Entity &_entity,
...
 
-  if (!_ecm.Component<components::WorldPose>(this->dataPtr->linkEntity))
-  {
-    _ecm.CreateComponent(this->dataPtr->linkEntity,
-        components::WorldPose());
-  }
+  Link link(this->dataPtr->linkEntity);
+  link.EnableVelocityChecks(_ecm, true);
 
   if (!_ecm.Component<components::Inertial>(this->dataPtr->linkEntity))
```
arjo129, line 133: This is one such place where the templated command would be very powerful.

arjo129 (commented):
These new APIs are awesome! They'll make it so much easier for systems authors (and result in lot cleaner systems). One small suggestion is lets make the helper function `void enableComponent(EntityComponentManager &_ecm, Entity _entity, bool _enable)`  in `link.cc` public. There are many places where this pattern is used and it'd be incredibly helpful to reduce the complexity of systems code.

src/Link.cc
mjcarroll: I agree, this may make sense to expose in `utils` or the like.

mjcarroll (commented):
I agree with @arjo129 that moving this `enableComponent` function into the public API would make a lot of sense.  It's a pattern that we do a good bit throughout the codebase.

src/Link.cc
chapulina: Moved it to `Util.hh` in 9f61ff885ad48bc7c010a9ef964e2fbf31f2d062. Outside of the `Link` class, the `enable` word may not make much sense though. I tried to explain in the docs because I couldn't think of a better name.

src/systems/kinetic_energy_monitor/KineticEnergyMonitor.cc
chapulina: Using `enableComponent` here now: 9f61ff885ad48bc7c010a9ef964e2fbf31f2d062

include/ignition/gazebo/Util.hh
```diff
@@ -169,6 +169,33 @@ namespace ignition
...
+    /// \brief Helper function to "enable" a component (i.e. create it if it
+    /// doesn't exist) or "disable" a component (i.e. remove it if it exists).
+    /// \param[in] _ecm Mutable reference to the ECM
+    /// \param[in] _entity Entity whose component is being enabled
+    /// \param[in] _enable True to enable (create), false to disable (remove).
+    /// \return True if a component was created or removed, false if nothing
+    /// changed.
+    template <class ComponentType>
+    bool IGNITION_GAZEBO_VISIBLE enableComponent(EntityComponentManager &_ecm,
+        Entity _entity, bool _enable)
```
arjo129, line 181:
```suggestion
        Entity _entity, bool _enable = true)
```
Sorry I should probably have suggested this yesterday, but I think it'd be nice to have this as in most cases in systems we enable components. Feel free to refute this suggestion.

arjo129 (approved):
LGTM. I've got a small suggestion to make that might improve quality of life but feel free to reject it if you think its inappropriate.

include/ignition/gazebo/Util.hh
chapulina: Ahhh I thought the same :smile: Will do :+1:

PR #936, opened 2021-07-27
Support adding systems that don't come from a plugin

chapulina:
# 🎉 New feature

## Summary

This is broken off #926 for easier review.

When using the `Server` API directly, it's often useful to instantiate systems locally without actually loading a plugin's shared library. This PR adds a new `Server::AddSystem` that takes a raw pointer and behaves just like the existing `AddSystem` function that takes a plugin.

I changed the test `Relay` class to use this approach, which should be a bit faster than loading the shared library.

Also added the convenient `gazebo::worldEntity` function that returns the first world found.

## Test it

Take a look at the unit tests added, and see on #926 how the test fixture uses this.

## Checklist
- [x] Added tests


src/SimulationRunner.hh
```diff
@@ -160,9 +180,29 @@ namespace ignition
...
+      /// \note This actually adds system to a queue. The system is added to the
+      /// runner at the begining of the a simulation cycle (call to Run). It is
+      /// also responsible for calling `Configure` on the system.
       /// \param[in] _system System to be added
-      public: void AddSystem(const SystemPluginPtr &_system);
+      /// \param[in] _entity Entity of system to be added. Nullopt if system
+      /// doesn't connect to an entity.
+      /// \param[in] _sdf Pointer to the SDF of the entity. Nullopt defaults to
+      /// world.
+      public: void AddSystem(System *_system,
```
azeey, line 203: Have you considered using a `std::shared_ptr<System>` here?

include/ignition/gazebo/Server.hh
```diff
@@ -218,6 +219,16 @@ namespace ignition
...
                   const unsigned int _worldIndex = 0);
 
+      /// \brief Add a System to the server. The server must not be running when
+      /// calling this. The server does not take ownership of the pointer and
+      /// assumes that the pointer is valid during the server's lifetime.
+      /// \param[in] _system system to be added
+      /// \param[in] _worldIndex Index of the world to query.
+      /// \return Whether the system was added successfully, or std::nullopt
+      /// if _worldIndex is invalid.
+      public: std::optional<bool> AddSystem(System *_system,
```
azeey, line 229: Not really feedback on this PR, but I just realized that `_worldIndex` is used to index into an array of `SimulationRunner`s, which implies that there's a 1:1 relationship between worlds and `SimulationRunner`s. And since each `SimulationRunner` has its own `EntityComponentManager`, there can only be one world in an `EntityComponentManager`.

src/SimulationRunner.hh
```diff
@@ -160,9 +180,29 @@ namespace ignition
...
+          std::optional<Entity> _entity = std::nullopt,
+          std::optional<sdf::ElementPtr> _sdf = std::nullopt);
```
azeey, line 205: What do you think about using `std::shared_ptr<const Element>` (a.k.a `sdf::ElementConstPtr` after the next libsdformat release) here so that code that only has access to `const Element` would be able to call this. For example, if a system wants to add another system during `System::Configure`, it wouldn't be able to call this without a `const_cast`.

azeey (commented):
Thanks for the addition. This will be really useful. I have a couple of comments about the API.

include/ignition/gazebo/Server.hh
chapulina: Yeah this was documented a while back: https://ignitionrobotics.org/api/gazebo/5.1/terminology.html

> * Simulation runner: Runs a whole world or some levels of a world, but no more than 1 world.
>      * It has a single ECM with all the entities and components relevant to the levels / world / performer being simulated.
>           * It's still TBD how to support multiple `<worlds>` in parallel.

5 major versions later, and it's still TBD :grimacing:

src/SimulationRunner.hh
chapulina: > using `std::shared_ptr<const Element>`

Done in d8c6047

> if a system wants to add another system during System::Configure

Note that this isn't possible right now, since systems don't have access to the `SimulationRunner` or `Server` APIs. Unless they spawn a new server 😳😵

src/SimulationRunner.hh
chapulina: Good call, I don't remember why I originally thought a raw pointer would make more sense. Changed in d8c6047

mjcarroll (approved):
Great addition!  Do you think we should add some documentation/tutorial for developers so that they are aware that this option exists? (Could be done in a follow-up, not going to block this)

src/SimulationRunner.hh
azeey: > Note that this isn't possible right now, since systems don't have access to the SimulationRunner or Server APIs. Unless they spawn a new server

Not sure why I thought it was possible. I guess a contrived scenario is where a test creates its own `Server` and passes it to an internal `System`.

azeey (approved):
I see `UNIT_ModelCommandAPI_TEST` failed on macOS. Is that a new flake?

mjcarroll:
> I see UNIT_ModelCommandAPI_TEST failed on macOS. Is that a new flake?

It's a new test I believe from the `ign model` command?

chapulina:
> Do you think we should add some documentation/tutorial for developers so that they are aware that this option exists?

I could write up a simple tutorial, but I'm not sure in which context we want to encourage users to use this API directly. The `TestFixture` class added in #926 uses it internally, and I wrote a tutorial/example using that class because I can see a clear use case.

PR #943, opened 2021-07-29
Simplifications to storage

chapulina:
Targeted at #856 , I just realized there was never a PR opened for this.

I'll clean up the branch before opening for review. Opening the PR for visibility.


chapulina:
@adlarkin , this is ready!

src/EntityComponentManager.cc
```diff
@@ -166,6 +150,51 @@ class ignition::gazebo::EntityComponentManagerPrivate
...
+  /// keeps track of components that were removed in the current simulation
+  /// step, while componentsMarkedAsRemoved keeps track of components that are
+  /// currently removed based on all simulation steps.
+  public: std::unordered_map<Entity, std::unordered_set<ComponentTypeId>>
+    componentsMarkedAsRemoved;
+
+  /// \brief A map of an entity to its components
+  public: std::unordered_map<Entity,
+           std::vector<std::unique_ptr<components::BaseComponent>>>
+             storage;
```
adlarkin, line 165: nit: rename to `componentStorage`?

src/EntityComponentManager.cc
```diff
@@ -582,42 +611,77 @@ bool EntityComponentManager::CreateComponentImplementation(
...
+  if (entityCompIter == this->dataPtr->storage.end())
   {
-    case ComponentAdditionResult::FAILED_ADDITION:
-      ignwarn << "Attempt to create a component of type [" << _componentTypeId
-        << "] attached to entity [" << _entity << "] failed.\n";
+    ignerr << "Attempt to create a component of type [" << _componentTypeId
+      << "] attached to entity [" << _entity
+      << "] failed: entity not in storage." << std::endl;
+    return false;
+  }
```
adlarkin, line 636: minor optimization: should we place the `componentTypeIndex` and `storage` validity checks before the `components::Factory::Instance` call, so that we avoid creating the component if there's no entity for it in the storage?

src/EntityComponentManager.cc
```diff
@@ -582,42 +611,77 @@ bool EntityComponentManager::CreateComponentImplementation(
...
+
+  // If entity has never had a component of this type
+  const auto compIdxIter = typeMapIter->second.find(_componentTypeId);
+  if (compIdxIter == typeMapIter->second.end())
```
adlarkin, line 640:
```suggestion
  const auto compIdxIter = typeMapIter->second.find(_componentTypeId);
  // If entity has never had a component of this type
  if (compIdxIter == typeMapIter->second.end())
```
nit: place the comment right before the `if` statement it applies to so that it's clear to readers of the code

src/EntityComponentManager.cc
```diff
@@ -1021,13 +1132,13 @@ void EntityComponentManagerPrivate::CalculateStateThreadLoad()
...
     return;
 
-  this->entityComponentsDirty = false;
-  this->entityComponentIterators.clear();
-  auto startIt = this->entityComponents.begin();
-  int numComponents = this->entityComponents.size();
+  this->componentTypeIndexDirty = false;
+  this->componentTypeIndexIterators.clear();
+  auto startIt = this->componentTypeIndex.begin();
+  int numComponents = this->componentTypeIndex.size();
```
adlarkin, line 1141: `this->componentTypeIndex.size()` is actually the number of entities, not components (the key in the `componentTypeIndex` map is the entity). This makes me wonder if the documentation/variable naming for the rest of this method needs to be updated :thinking:

adlarkin (commented):
Looks good! Just a few minor comments

src/EntityComponentManager.cc
chapulina: hah good call, changed the wording in 01be05f

chapulina:
@adlarkin, thanks for the review, I addressed all your comments

adlarkin:
Great; merging! :tada:

PR #944, opened 2021-07-30
removed unneeded plugin update

jennuine:
# 🦟 Bug fix

## Summary
When a plugin was added an `Update` was called immediately. With #497 updates are also being called in a loop, which was causing deadlocks since both could be called at the same time. This PR removes the initial update (after the plugin is inserted) since updates are already happening periodically. 

## Checklist
- [ ] Added tests

### DIFF
--- a/include/ignition/gazebo/gui/GuiRunner.hh
+++ b/include/ignition/gazebo/gui/GuiRunner.hh
@@ -48,10 +48,6 @@ class IGNITION_GAZEBO_VISIBLE GuiRunner : public QObject
   /// \brief Destructor
   public: ~GuiRunner() override;
 
-  /// \brief Callback when a plugin has been added.
-  /// \param[in] _objectName Plugin's object name.
-  public slots: void OnPluginAdded(const QString &_objectName);
-
   /// \brief Make a new state request to the server.
   public slots: void RequestState();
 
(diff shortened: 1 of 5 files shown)


include/ignition/gazebo/gui/GuiRunner.hh
```diff
@@ -48,10 +48,6 @@ class IGNITION_GAZEBO_VISIBLE GuiRunner : public QObject
   /// \brief Destructor
   public: ~GuiRunner() override;
 
-  /// \brief Callback when a plugin has been added.
-  /// \param[in] _objectName Plugin's object name.
-  public slots: void OnPluginAdded(const QString &_objectName);
```
chapulina, line 53: Ahh I forgot that this header is installed, let's leave the function in, but without doing anything.

chapulina: 9ecc40977235ea31eb9ca1df3f73b7605c685a89

chapulina (approved):
Thanks for cleaning this up! Works for me. Let's just wait for CI results.

iche033 (commented):
ran into this issue on ign-gazebo4 so ported the fix over and confirmed that the issue is fixed.

JointPositionController tests are failing on CI though

chapulina:
> JointPositionController tests are failing on CI though

Should be fixed by waiting for updates in 20cdc9f

PR #950, opened 2021-07-30
Send mouse hover event only on spawn preview to improve framerate

iche033:

# 🦟 Bug fix

## Summary

Scene3D appear to be slow when there are high poly count meshes in the scene. This turns out that it's not a rendering issue. The problem is that we are making ray queries every frame through the `ScreenToScene` call in the `BroadcastHoverPos` function. I believe this is intended mainly for spawning entities. The `hoverDirty` variable is never set to `false` even when the mouse is outside the 3D window. It's only set to `false` in spawn entity mode. So it's incurring unnecessary computation every frame.

Note with this change, we will only broadcast the hover event on spawn entity mode. The gain here is a big improvement in frame rate.

To reproduce the issue:

launch ign-gazebo and drag and drop a high polycount model into the scene, e.g. [Urban 2 Story](https://app.ignitionrobotics.org/OpenRobotics/fuel/models/Urban%202%20Story) model. Try moving the camera and notice that it's laggy.

Before this change, I get ~3-4 FPS, and after this change, I get 40-50 FPS

## Checklist
- [ ] Added tests

### DIFF
--- a/src/gui/plugins/scene3d/Scene3D.cc
+++ b/src/gui/plugins/scene3d/Scene3D.cc
@@ -889,7 +889,7 @@ void IgnRenderer::HandleMouseEvent()
 /////////////////////////////////////////////////
 void IgnRenderer::BroadcastHoverPos()
 {
-  if (this->dataPtr->hoverDirty)
+  if (this->dataPtr->spawnPreview && this->dataPtr->hoverDirty)
   {
     math::Vector3d pos = this->ScreenToScene(this->dataPtr->mouseHoverPos);
 


src/gui/plugins/scene3d/Scene3D.cc
```diff
@@ -889,7 +889,7 @@ void IgnRenderer::HandleMouseEvent()
 /////////////////////////////////////////////////
 void IgnRenderer::BroadcastHoverPos()
 {
-  if (this->dataPtr->hoverDirty)
+  if (this->dataPtr->spawnPreview && this->dataPtr->hoverDirty)
```
chapulina, line 892: There are other plugins listening to the hover event, such as `TapeMeasure`:

https://github.com/ignitionrobotics/ign-gazebo/blob/24be610ac1e8889b6c538a110c84da7f91360a39/src/gui/plugins/tape_measure/TapeMeasure.cc#L235

chapulina (changes requested):
I agree it's a performance issue, but we need to find an alternative solution. Maybe throttle it?

iche033:
replaced by:
* #969
* https://github.com/ignitionrobotics/ign-rendering/pull/378
* https://github.com/ignitionrobotics/ign-rendering/pull/383

PR #951, opened 2021-07-31
5 ➡️ 6

chapulina:
# ➡️ Forward port

Port `ign-gazebo5` to `main`

Branch comparison: https://github.com/ignitionrobotics/ign-gazebo/compare/main...ign-gazebo5


src/gui/GuiRunner.cc
```diff
@@ -151,25 +151,10 @@ void GuiRunner::RequestState()
 }
 
 /////////////////////////////////////////////////
-void GuiRunner::OnPluginAdded(const QString &_objectName)
+void GuiRunner::OnPluginAdded(const QString &)
```
azeey, line 154: Can this be removed since this is targeting `main`?

azeey (approved):
LGTM! Just one minor comment

src/gui/GuiRunner.hh
```diff
@@ -49,6 +49,7 @@ class IGNITION_GAZEBO_GUI_VISIBLE GuiRunner : public QObject
   public: ~GuiRunner() override;
 
   /// \brief Callback when a plugin has been added.
+  /// This function has no effect and is left here for ABI compatibility.
```
adlarkin, line 52: Since we don't have to worry about API/ABI breaks in fortress, should this method be removed?

src/gui/GuiRunner.cc
chapulina: This function may make a comeback on Fortress, in https://github.com/ignitionrobotics/ign-gazebo/pull/793, so I'd rather leave it there until we sort that out.

adlarkin (approved):
LGTM! I have just one quesiton.

src/gui/GuiRunner.hh
adlarkin: Nevermind, I see that this was asked/answered already in https://github.com/ignitionrobotics/ign-gazebo/pull/951#discussion_r681152457

PR #958, opened 2021-08-06
Fix crash in the follow_actor example

luca-della-vedova:
This PR fixes a sneaky segfault introduced in #858, while the fix works I'm open to better solutions if anything is recommended (it makes the function arguments a bit bulky).

The issue was that the previous version of the code was relying on an unfortunate coincidence related to the variable names for the `entityPoses` and `trajectoryPoses` data structures, specifically when updating them [here](https://github.com/ignitionrobotics/ign-gazebo/blob/3e8d4e7dd97d6dccdab456625b4b17c9749ebf64/src/rendering/RenderUtil.cc#L845-L855) the variables would refer to the local variables introduced in the beginning of the `Update()` function, [here](https://github.com/ignitionrobotics/ign-gazebo/blob/3e8d4e7dd97d6dccdab456625b4b17c9749ebf64/src/rendering/RenderUtil.cc#L535-L537).
However, when during the refactor the code was moved into a [new function](https://github.com/ignitionrobotics/ign-gazebo/blob/a1ebffa973707a7273c33736c6dcbb7b2b171eb5/src/rendering/RenderUtil.cc#L2030-L2031), the Implementation class variables (i.e. [this](https://github.com/ignitionrobotics/ign-gazebo/blob/a1ebffa973707a7273c33736c6dcbb7b2b171eb5/src/rendering/RenderUtil.cc#L189-L190) and [this](https://github.com/ignitionrobotics/ign-gazebo/blob/a1ebffa973707a7273c33736c6dcbb7b2b171eb5/src/rendering/RenderUtil.cc#L219-L220)) with the same name are being silently used behind the scenes, the problem is that at the beginning of the `Update()` function there is a `std::move` [call](https://github.com/ignitionrobotics/ign-gazebo/blob/3e8d4e7dd97d6dccdab456625b4b17c9749ebf64/src/rendering/RenderUtil.cc#L535-L537) that moves from the member data structures into the local variables, hence it becomes a case of using a data structure after it has been moved, which causes a segfault for a null pointer exception in the data pointer.

To reproduce the bug / check the PR just run the `follow_actor` demo with and without the PR:
`ign gazebo follow_actor.sdf`


iche033 (commented):
nice, I can reproduce the crash and this fixes it for me too. 

Adding a sensors system and a camera to the follow_actor integration test also causes the test to crash. I've added it in the [fix/follow_actor_crash_test](https://github.com/ignitionrobotics/ign-gazebo/compare/fix/follow_actor_crash...fix/follow_actor_crash_test) branch so we can test the rendering side as part of the integration test in the future. Can you take a look and merge if that looks good to you?

adlarkin:
Thanks @luca-della-vedova for catching this! Sorry I didn't notice this when working on #858 :grimacing: 

> Adding a sensors system and a camera to the follow_actor integration test also causes the test to crash

Do we need to fix this as well?

iche033:
> Do we need to fix this as well?

oh I meant adding it to the test  causes to crash before the changes in this PR, and confirmed that the test no longer crashes after this PR.

src/rendering/RenderUtil.cc
```diff
@@ -358,7 +358,9 @@ class ignition::gazebo::RenderUtilPrivate
   /// data.
   /// \sa actorManualSkeletonUpdate
   public: void UpdateAnimation(const std::unordered_map<Entity,
-              AnimationUpdateData> &_actorAnimationData);
+              AnimationUpdateData> &_actorAnimationData,
+              const std::unordered_map<Entity, math::Pose3d> &_entityPoses,
+              const std::unordered_map<Entity, math::Pose3d> &_trajectoryPoses);
```
adlarkin, line 363: Can these new parameters be added to the method documentation as well?

luca-della-vedova: [Done](https://github.com/ignitionrobotics/ign-gazebo/pull/958/commits/0dd0970925853701625586b6b18073992248108b), I hope I did it right (position and content) since I'm not too familiar with this part of the code.

luca-della-vedova:
> nice, I can reproduce the crash and this fixes it for me too.
> 
> Adding a sensors system and a camera to the follow_actor integration test also causes the test to crash. I've added it in the [fix/follow_actor_crash_test](https://github.com/ignitionrobotics/ign-gazebo/compare/fix/follow_actor_crash...fix/follow_actor_crash_test) branch so we can test the rendering side as part of the integration test in the future. Can you take a look and merge if that looks good to you?

Done!

iche033 (approved):
looks good to me!

src/rendering/RenderUtil.cc
adlarkin: Looks great, thanks!

adlarkin (approved):

Issue #102, opened 2020-04-28
Copy / paste entities from the GUI


chapulina:
Add options to copy / cut / paste models through the GUI.

Hotkeys:
* Cut = CTRL+X
* Copy =  CTRL+C
* Paste = CTRL+V

The option should also be available from the context (right-click) menu.

chapulina:
I'll dump some more things I discussed with @adlarkin here:

# The quick way

The quickest approach at the moment would make use of the [SpawnFromDescription](https://github.com/ignitionrobotics/ign-gui/blob/cac3da702a312f81fee928f9e52cf61e533939fa/include/ignition/gui/GuiEvents.hh#L101-L103) event and pass the contents of `components::ModelSdf` (requires https://github.com/ignitionrobotics/ign-gazebo/pull/851)

Drawbacks of this approach:

* Only works for models (not lights, actors, etc)
* Requires the GUI to parse the SDF file to create the spawn preview that's attached to the mouse. This can take a non-negligible time for models with multiple links, joints and materials.

# The :sparkles:  great :sparkle:  way

My recommended implementation involves using the [clone_name](https://github.com/ignitionrobotics/ign-msgs/blob/ign-msgs7/proto/ignition/msgs/entity_factory.proto#L58-L59) field of `msgs::EntityFactory`:

### 1: UserCommands

- [ ] Make the `UserCommands` system parse the `clone_name` field here:

    https://github.com/ignitionrobotics/ign-gazebo/blob/24be610ac1e8889b6c538a110c84da7f91360a39/src/systems/user_commands/UserCommands.cc#L726-L731

    This would probably entail:
    1. Getting an entity ID from its name
    2. Copying all entities and components that compose this entity
        * Recursively include all children
        * Make sure the new entities have different IDs
    
    This could be in its own PR. One way to verify that this works would be:

    1. Start `ign gazebo shapes.sdf`
    2. On another terminal, call:
        
            ign service -s /world/shapes/create \
                --reqtype ignition.msgs.EntityFactory \
                --reptype ignition.msgs.Boolean \
                --timeout 300 \
               --req 'clone_name: "box", pose {z: 10}, name: "box_copy"'

    3. A red box is spawned and falls on top of the original one.

## 2: 3D Scene

- [ ] Create a new `SpawnFromCloneName` GUI event (or some name like that) that just takes a top-level entity's name. And listen to it from `Scene3D`.

[...]

PR #959, opened 2021-08-06
Add support for cloning entities

adlarkin:

# 🎉 New feature

Part of #102

## Summary
This PR completes "The Great Way", part 1 of https://github.com/ignitionrobotics/ign-gazebo/issues/102#issuecomment-890165166. I had to make an ABI-breaking change in `Component.hh` by adding a new virtual `Clone` method for components, so this is targeting main. If there's a way to achieve component cloning functionality without breaking ABI, let me know!

## Test it
New tests have been written in this PR (see `src/Component_TEST.cc`, src/`EntityComponentManager_TEST.cc`).

To test this manually, you can do the following:
1. start gazebo:
```
ign gazebo shapes.sdf
```
2. in a new terminal, run the following command to clone the red box. Since we set the cloned box's `position.z` is set to 3 in the service call, the cloned box should appear above the original box:
```
ign service -s /world/shapes/create --reqtype ignition.msgs.EntityFactory --reptype ignition.msgs.Boolean --timeout 300 --req 'clone_name: "box", pose: {position: {z: 3}}, name: "box_copy"'
```

## Checklist
- [ ] Added tests


adlarkin:
Currently, I am having an issue where the cloned entity cannot be moved. Here's what I am seeing when I run through the steps in the `Test It` section of the PR description (`box_copy` is the clone. Also, notice how when I move the original box, `box_copy` moves back to the origin): 

clone_wont_move

Does anyone know what might be causing this?

chapulina:
> Does anyone know what might be causing this?

I looked a bit into it and I'm not sure. I verified that:

* the `UserCommands` system receives the pose command
* all entities are created and **appear to** have the correct components, but I didn't double check each of them
* the `Physics` system **appears to** correctly creates `ign-physics` entities
* the `Physics` system receives and applies the `WorldPoseCmd`

I did notice that the code never reaches this point when moving the copied model, but it does when moving other models:

https://github.com/ignitionrobotics/ign-gazebo/blob/9fbb9b988c20711fcff706fb14d034516e027e6b/src/systems/physics/Physics.cc#L2251

My suggestion would be to trace back what's missing for the code to get there. I hope that helps

src/EntityComponentManager.cc
```diff
@@ -206,6 +209,79 @@ Entity EntityComponentManagerPrivate::CreateEntityImplementation(Entity _entity)
...
+        << "has this name." << std::endl;
+      return kNullEntity;
+    }
+    uniqueNameGenerated = true;
+  }
+
+  auto clonedEntity = this->CreateEntity();
+
+  if (_parent != kNullEntity)
+    this->CreateComponent(clonedEntity, components::ParentEntity(_parent));
```
iche033, line 242: also call `this->SetParentEntity(clonedEntity, _parent);`?

iche033 (commented):
looks like the `box` and `box_copy` have the same canonical link in the physics system. Maybe when cloning the entity, you'll need to update the cloned entity's canonical entity to the cloned child link?

src/EntityComponentManager.cc
adlarkin: Yeah, thanks for catching this :+1: I added it in https://github.com/ignitionrobotics/ign-gazebo/pull/959/commits/8b0e17c2887028a8662aea087d2905a31679ecae

adlarkin:
> looks like the `box` and `box_copy` have the same canonical link in the physics system. Maybe when cloning the entity, you'll need to update the cloned entity's canonical link to the cloned child link?

Yeah, that was the issue! I addressed this in https://github.com/ignitionrobotics/ign-gazebo/pull/959/commits/8b0e17c2887028a8662aea087d2905a31679ecae, and things seem to be working now. So, this PR should be ready for review :crossed_fingers: 

---

One other thing to note is a TODO note that I left in the code: https://github.com/ignitionrobotics/ign-gazebo/blob/8b0e17c2887028a8662aea087d2905a31679ecae/src/systems/user_commands/UserCommands.cc#L744

This is referring to the following field in the `entity_factory` message: https://github.com/ignitionrobotics/ign-msgs/blob/7a702296e6b58c7c3bfd9638c296a9e62f48810c/proto/ignition/msgs/entity_factory.proto#L72-L74

We have a method for getting the pose of an entity in the world: https://github.com/ignitionrobotics/ign-gazebo/blob/09e2ea78bf10d0bf84871e4ad3d0e120b85100eb/include/ignition/gazebo/Util.hh#L38-L43

Do we have any other similar methods for getting the pose of an entity in a particular frame?

iche033:
> Do we have any other similar methods for getting the pose of an entity in a particular frame?

There is this [RelativePose](https://github.com/ignitionrobotics/ign-gazebo/blob/ign-gazebo5/src/systems/physics/Physics.cc#L2088) function in the physics system that may be useful here?

iche033:
copying simple shapes and meshes work well. I then tried to copy lights and sensors and ran into issues:
* lights - crashes when trying to clone a light with the service call
* sensors - cloning works but the data is published to the same topic, e.g. for camera sensors, in ImageDisplay you'll see images switching between two camera views

I'm testing with `fuel_textured_mesh.sdf` world with the following cmds:

```
ign service -s /world/fuel/create --reqtype ignition.msgs.EntityFactory --reptype ignition.msgs.Boolean --timeout 300 --req 'clone_name: "tube_light", pose: {position: {z: 1}}, name: "tube_light_copy"'
ign service -s /world/fuel/create --reqtype ignition.msgs.EntityFactory --reptype ignition.msgs.Boolean --timeout 300 --req 'clone_name: "camera", pose: {position: {z: 3}}, name: "camera_copy"'
```

iche033:
> sensors - cloning works but the data is published to the same topic,

actually I think this may be ok. It's the same behavior when user tries to insert the same sensor model twice into the scene. If a user wants different camera topics, they will need to leave out the `<topic>` field and let gazebo auto generate the topic name

adlarkin:
> There's a possible race condition with the views
> For now, I've made a temporary fix for this in fe95598, but I will open a new PR with a better fix that is more performance-friendly.

I've opened a better fix in #1001 which seems to resolve the race condition issue, so I have reverted the temporary fix in https://github.com/ignitionrobotics/ign-gazebo/pull/959/commits/d9de4396020b02f320dea403863cb1af2c55f55f.

adlarkin:
#1001 has been merged, and this PR has been updated to include these changes in https://github.com/ignitionrobotics/ign-gazebo/pull/959/commits/876724b6f732efc43d8091074c964d8380fca49d. So, now, I believe that the only other issue (for lights, at least) is what I described in the second half of https://github.com/ignitionrobotics/ign-gazebo/pull/959#issuecomment-906883962 (cloning -> deleting -> re-cloning while simulation is paused), which seems to be a bug that already existed before this PR.

adlarkin:
I found that when cloning actors, the same fix with the light sdf (https://github.com/ignitionrobotics/ign-gazebo/pull/959#issuecomment-906883962) had to be applied. Cloning actors should work now with https://github.com/ignitionrobotics/ign-gazebo/pull/959/commits/2fbb3849932a04fa35d7dec60f408a334760bcf8.

I believe that cloning support now exists for simple shapes, meshes, sensors, lights, and actors. This should be ready for another round of review.

iche033:
>  So, now, I believe that the only other issue (for lights, at least) is what I described in the second half of #959 (comment) (cloning -> deleting -> re-cloning while simulation is paused), which seems to be a bug that already existed before this PR.

I think the problem is in the order which the rendering thread processes requests from the ECM. So this may be what's happening: When the user Clones -> Deletes -> Clones a `tube_light`, and hits Play, the sensors system starts processing these requests. Since sim was paused when these operations took place, these requests all come in at the same time. The rendering thread then carries out these operations in the following order as seen [here](https://github.com/ignitionrobotics/ign-gazebo/blob/ign-gazebo5/src/rendering/RenderUtil.cc#L661-L678):
1. Delete `tube_light_copy `
2. Create `tube_light_copy`
3. Create `tube_light_copy` -> crash because light with the same name already exists.

At minimum, I think we should catch duplicate lights in SceneManager e.g. using `scene->HasLightName` check, so we can avoid crashes when we roll out the clone feature. We can then ticket this issue about creating and deleting entities with the same name when sim is paused.

adlarkin:
> At minimum, I think we should catch duplicate lights in SceneManager e.g. using `scene->HasLightName` check, so we can avoid crashes when we roll out the clone feature

I went ahead and made the following changes to catch duplicate lights:
```diff
diff --git a/src/rendering/SceneManager.cc b/src/rendering/SceneManager.cc
index 283c12f1..b8f891ac 100644
--- a/src/rendering/SceneManager.cc
+++ b/src/rendering/SceneManager.cc
@@ -1101,6 +1101,12 @@ rendering::LightPtr SceneManager::CreateLight(Entity _id,
     name = parent->Name() +  "::" + name;
 
   rendering::LightPtr light;
+  if (this->dataPtr->scene->HasLightName(name))
+  {
+    ignerr << "A light named [" << name << "] already exists in the scene.\n";
+    return light;
+  }
+
```

However, I must say that I am still a little confused by the behavior. If I follow the same pattern of creating -> deleting -> creating the light copy while simulation is paused, and then press play, no crash occurs and I see this error message, as expected:
```
[Err] [SceneManager.cc:1106] A light named [tube_light_copy] already exists in the scene.
```

But, after doing this, if I press play and then delete -> create the light copy again, I see the same error message when re-creating the light, which doesn't make sense to me since the light has been deleted and re-created successfully in the GUI. So, my concern is that while things may appear to be okay on the GUI, something still isn't right on the server side.

src/EntityComponentManager.cc
```diff
@@ -270,6 +276,154 @@ Entity EntityComponentManagerPrivate::CreateEntityImplementation(Entity _entity)
   return _entity;
 }
 
+/////////////////////////////////////////////////
+Entity EntityComponentManager::Clone(const Entity _entity, const Entity _parent,
+    const std::string &_name, bool _allowRename)
+{
+  // During cloning, we populate two maps:
+  //  - map of cloned model entities to the non-cloned model's canonical link
```
chapulina, line 284: If feels strange to give the canonical link component special treatment. The ECM should be treating all components the same way as much as possible.

I believe that a more general approach would be to treat all components that have `Entity` data type the same way. This may need some template magic. At the moment, the only components that hold `Entity` are `CanonicalLink` and `ParentEntity`, but in the future we may have other components that point to other entities (I don't know, like `EndEffectorLink`), and the current approach won't scale well for them.

src/systems/user_commands/UserCommands.cc
```diff
@@ -725,9 +725,28 @@ bool CreateCommand::Execute()
     }
     case msgs::EntityFactory::kCloneName:
     {
-      // TODO(louise) Implement clone
-      ignerr << "Cloning an entity is not yet supported." << std::endl;
-      return false;
+      auto entityToClone = this->iface->ecm->EntityByComponents(
+          components::Name(createMsg->clone_name()));
```
chapulina, line 729: We should check if `kNullEntity != entityToClone` in case the user passes a bad `clone_name`

src/systems/user_commands/UserCommands.cc
```diff
@@ -725,9 +725,28 @@ bool CreateCommand::Execute()
...
+      auto parentComp =
+        this->iface->ecm->Component<components::ParentEntity>(entityToClone);
```
chapulina, line 731: For now, it's best we shortcut here if `!parentComp || parentComp->Data() != worldEntity`, because we don't have good support for creating non-top-level entities yet. i.e. we don't want to let users clone a joint just yet.

src/EntityComponentManager.cc
```diff
@@ -270,6 +276,154 @@ Entity EntityComponentManagerPrivate::CreateEntityImplementation(Entity _entity)
...
+    // we're cloning a canonical link, so we map the original canonical link
+    // to the cloned canonical link
+    _oldToClonedCanonicalLink[_entity] = clonedEntity;
+  }
+
+  for (const auto &childEntity :
+      this->EntitiesByComponents(components::ParentEntity(_entity)))
+  {
+    this->Clone(childEntity, clonedEntity, "", true, _oldModelCanonicalLink,
+        _oldToClonedCanonicalLink);
```
chapulina, line 421: I think we should check the return value of this function to handle children not being created correctly.

src/EntityComponentManager.cc
```diff
@@ -270,6 +276,154 @@ Entity EntityComponentManagerPrivate::CreateEntityImplementation(Entity _entity)
...
+    // skip the Name and ParentEntity components since those were already
+    // handled above
+    if ((type == components::Name::typeId) ||
+        (type == components::ParentEntity::typeId))
+      continue;
+
+    auto originalComp = this->ComponentImplementation(_entity, type);
+    auto clonedComp = originalComp->Clone();
+
+    if (type == components::Light::typeId)
```
chapulina, line 383: Handling each component type's special case within the ECM doesn't scale well. How about overriding the `Component::Clone()` function that you created so each component can handle its cloning special needs separately?

In any case, I'd say we don't need to worry about updating the contents of SDF components when cloning. These contents become outdated the moment the `SdfEntityCreator` generates components for them and it's not feasible to keep everything in sync. We can treat most SDF components as the "initial SDF" for an entity.

chapulina (changes requested):
I did a first pass. There are some things that concern me about the changes to the ECM as I wrote below. But I think that the addition of `Component::Clone` is good, and that's the actual ABI breaking change. So if we can't sort out the ECM until code freeze, I think we can at least get the `Component::Clone` function into Fortress and add the ECM stuff later.

src/EntityComponentManager.cc
adlarkin: > How about overriding the Component::Clone() function that you created so each component can handle its cloning special needs separately?

That sounds like a good approach :+1: 

> In any case, I'd say we don't need to worry about updating the contents of SDF components when cloning.

The issue is that if I don't update the name of the SDF for lights and actors, then they won't get created because another light/actor already exists with this name - in fact, the `SceneManager` will crash if the name isn't changed. Take a look at this code as an example:
https://github.com/ignitionrobotics/ign-gazebo/blob/1334b5cb6a123cd16c5d39b1cebb03e2f25a2bee/src/rendering/SceneManager.cc#L1111-L1117

If I try cloning a light without changing the name in the SDF, the `CreateSpotLight` call returns a `nullptr` because a light already exists with that name, and then that `nullptr` is dereferenced a few lines later, causing a crash. I can check to make sure that the name used for cloning something isn't already taken, but then I get into the issues described in https://github.com/ignitionrobotics/ign-gazebo/pull/959#issuecomment-919454527.

So, should I still modify the name in the SDF for lights and actors if I were to override `Component::Clone()` for these components?

chapulina: > The issue is that if I don't update the name of the SDF for lights and actors, then they won't get created because another light/actor already exists with this name

Argh I see :woman_facepalming: I think we shouldn't be using the `sdf::Light::Name` there, but the `Name` component.

So you can either override `Component::Clone` or update `SceneManager` to use the name component. Whatever is faster for now :+1:

src/EntityComponentManager.cc
adlarkin: https://github.com/ignitionrobotics/ign-gazebo/pull/959/commits/ba5f116ae6dbcc4602a15853cec2dae1d9d7aec6

src/systems/user_commands/UserCommands.cc
adlarkin: Sounds good. I made the change in https://github.com/ignitionrobotics/ign-gazebo/pull/959/commits/ba5f116ae6dbcc4602a15853cec2dae1d9d7aec6 and added a `TODO` note about adding future support for creating non-top level entities

src/EntityComponentManager.cc
adlarkin: > If feels strange to give the canonical link component special treatment. The ECM should be treating all components the same way as much as possible.

Agreed :+1: 

> I believe that a more general approach would be to treat all components that have Entity data type the same way. This may need some template magic.

Should I try to go ahead and update this now? Or do we want to leave a `TODO` note/ticket an issue and come back to it later? If we were to make this change now, I'm not sure off the top of my head how I'd do it, so we may need to do some brainstorming.

chapulina: I'd be ok going with this for now for expediency, but how about removing that private function from the public header?

adlarkin: > how about removing that private function from the public header?

I was hoping to move the private method to the `EntityComponentManagerPrivate` class, but I can't because the private method calls private methods from the `EntityComponentManager` class (`CreateComponentImplementation`, `ComponentImplementation`). So, unless there's another way to hide this method, it may need to stay in the public header.

src/EntityComponentManager.cc
adlarkin: The `SceneManager` has been updated to use `components::Name`: https://github.com/ignitionrobotics/ign-gazebo/pull/959/commits/cef373f156ebd36abbf87d59aa5025fd37487bb2

I believe that resolves this conversation, unless there's anything else I am missing?

src/EntityComponentManager.cc
chapulina: I pushed the idea we talked about in 96396c83c5947151ea3050bbd8234d89eccc3118.

chapulina (approved):
Great work! The usefulness of the new API and transport service goes beyond the copy/paste GUI functionality.

chapulina:
The Windows build failed with an infra issue. I'm going to merge this, hopefully it compiles on Windows :crossed_fingers: Otherwise we fix it tomorrow

PR #971, opened 2021-08-12
Don't use $HOME on most tests (InternalFixture)

chapulina:
Signed-off-by: Louise Poubel <louise@openrobotics.org>

# 🦟 Bug fix

## Summary

Tests shouldn't be relying on the user's home folder. This makes sure that most tests use the `server.config` from a custom home path.

I think I caught all tests that are loading `~/.ignition/gazebo/server.config`. This PR also fixes creating the config file in general, because now we make sure that the directories are created too.

It also backports `EnvTestFixture` from https://github.com/ignitionrobotics/ign-gazebo/pull/594, with some updates to make sure each test can keep its name, as well as supporting various test types.

## Checklist
- [x] Added tests

### DIFF
--- a/src/ServerPrivate.cc
+++ b/src/ServerPrivate.cc
@@ -229,13 +229,11 @@ void ServerPrivate::AddRecordPlugin(const ServerConfig &_config)
   bool hasRecordPath {false};
   bool hasCompressPath {false};
   bool hasRecordResources {false};
-  bool hasCompress {false};
   bool hasRecordTopics {false};
 
   std::string sdfRecordPath;
   std::string sdfCompressPath;
   bool sdfRecordResources;
-  bool sdfCompress;
   std::vector<std::string> sdfRecordTopics;
 
   if (sdfUseLogRecord)
@@ -246,8 +244,6 @@ void ServerPrivate::AddRecordPlugin(const ServerConfig &_config)
       recordPluginElem->Get<std::string>("compress_path", "");
     std::tie(sdfRecordResources, hasRecordResources) =
       recordPluginElem->Get<bool>("record_resources", false);
-    std::tie(sdfCompress, hasCompress) =
-      recordPluginElem->Get<bool>("compress", false);
 
     hasRecordTopics = recordPluginElem->HasElement("record_topic");
     if (hasRecordTopics)
--- a/src/Server_TEST.cc
+++ b/src/Server_TEST.cc
@@ -39,22 +39,11 @@
 
 #include "plugins/MockSystem.hh"
 #include "../test/helpers/Relay.hh"
+#include "../test/helpers/EnvTestFixture.hh"
 
 using namespace ignition;
 using namespace ignition::gazebo;
 using namespace std::chrono_literals;
-
-class ServerFixture : public ::testing::TestWithParam<int>
-{
-  protected: void SetUp() override
-  {
-    // Augment the system plugin path.  In SetUp to avoid test order issues.
-    ignition::common::setenv("IGN_GAZEBO_SYSTEM_PLUGIN_PATH",
-           (std::string(PROJECT_BINARY_PATH) + "/lib").c_str());
-
-    ignition::common::Console::SetVerbosity(4);
-  }
-};
 
 /////////////////////////////////////////////////
 TEST_P(ServerFixture, DefaultServerConfig)
--- a/src/Server_TEST.cc
+++ b/src/Server_TEST.cc
@@ -46,6 +46,11 @@ using namespace ignition::gazebo;
 using namespace std::chrono_literals;
 
 /////////////////////////////////////////////////
+class ServerFixture : public InternalFixture<::testing::TestWithParam<int>>
+{
+};
+
+/////////////////////////////////////////////////
 TEST_P(ServerFixture, DefaultServerConfig)
 {
   ignition::gazebo::ServerConfig serverConfig;
--- a/test/helpers/EnvTestFixture.hh
+++ b/test/helpers/EnvTestFixture.hh
@@ -1,0 +1,59 @@
+/*
+ * Copyright (C) 2020 Open Source Robotics Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+*/
+#ifndef IGNITION_GAZEBO_TEST_HELPERS_ENVTESTFIXTURE_HH_
+#define IGNITION_GAZEBO_TEST_HELPERS_ENVTESTFIXTURE_HH_
+
+#include <gtest/gtest.h>
+
+#include <ignition/common/Console.hh>
+#include <ignition/common/Util.hh>
+#include "ignition/gazebo/test_config.hh"
+
+using namespace ignition;
+
+/// \brief Common test setup for various tests
+class ServerFixture : public ::testing::TestWithParam<int>
+{
+  // Documentation inherited
+  protected: void SetUp() override
+  {
+    // Augment the system plugin path.  In SetUp to avoid test order issues.
+    common::setenv("IGN_GAZEBO_SYSTEM_PLUGIN_PATH",
+           (std::string(PROJECT_BINARY_PATH) + "/lib").c_str());
+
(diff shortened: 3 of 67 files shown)


src/Server_TEST.cc
```diff
@@ -39,23 +39,12 @@
...
-{
-  protected: void SetUp() override
-  {
-    // Augment the system plugin path.  In SetUp to avoid test order issues.
-    ignition::common::setenv("IGN_GAZEBO_SYSTEM_PLUGIN_PATH",
-           (std::string(PROJECT_BINARY_PATH) + "/lib").c_str());
-
-    ignition::common::Console::SetVerbosity(4);
-  }
-};
```
adlarkin, line 57: I believe this is breaking ABI :monocle_face: https://build.osrfoundation.org/job/ignition_gazebo-abichecker-any_to_any-ubuntu_auto-amd64/4020/API_5fABI_20report/

chapulina: I believe that build failed because of server connectivity issues, see the logs here:

https://build.osrfoundation.org/job/ignition_gazebo-abichecker-any_to_any-ubuntu_auto-amd64/4020/console

Note how the ABI report says "Compatibility 100%" and doesn't have any symbols marked with some severity. This means we're good :smile: 

It's surprising that the ABI checker is considering that the `TestFixture` symbols were added by this PR, when they have been added by #926 :woman_shrugging: 

Also, note that this PR doesn't touch any headers that are installed, so there's no possibility of breaking ABI.

test/helpers/EnvTestFixture.hh
```diff
@@ -0,0 +1,59 @@
+/*
+ * Copyright (C) 2020 Open Source Robotics Foundation
```
adlarkin, line 2:
```suggestion
 * Copyright (C) 2021 Open Source Robotics Foundation
```

test/helpers/EnvTestFixture.hh
```diff
@@ -0,0 +1,59 @@
...
+  // Documentation inherited
+  protected: void TearDown() override
+  {
+    // Restore $HOME
+    EXPECT_TRUE(common::setenv(IGN_HOMEDIR, this->realHome.c_str()));
+  }
+
+  /// \brief Directory to act as $HOME for tests
+  public: std::string fakeHome = common::joinPaths(PROJECT_BINARY_PATH, "test",
+      "fake_home");
```
adlarkin, line 54: nit: make this variable `const`?

adlarkin (approved):
> I know there are other tests still loading ~/.ignition/gazebo/server.config

Should we create an issue with all the tests that are still loading `~/.ignition/gazebo/server.config` so that we can keep track of what needs to be updated?

chapulina:
> all the tests that are still loading ~/.ignition/gazebo/server.config

Ok, I went ahead and fixed almost all of them (if not all): 886b788

test/integration/battery_plugin.cc
```diff
@@ -39,18 +39,17 @@
...
-class BatteryPluginTest : public ::testing::Test
+class BatteryPluginTest : public InternalFixture<::testing::Test>
 {
   // Documentation inherited
   protected: void SetUp() override
   {
-    common::Console::SetVerbosity(4);
-    ignition::common::setenv("IGN_GAZEBO_SYSTEM_PLUGIN_PATH",
-           (std::string(PROJECT_BINARY_PATH) + "/lib").c_str());
+    InternalFixture::SetUp();
```
adlarkin, line 52: If we call `SetUp` but don't call `TearDown`, does that mean the `IGN_HOMEDIR` environment variable doesn't get reset correctly?

test/helpers/EnvTestFixture.hh
```diff
@@ -26,7 +26,8 @@
 using namespace ignition;
 
 /// \brief Common test setup for various tests
-class ServerFixture : public ::testing::TestWithParam<int>
+template <typename TestType>
+class InternalFixture : public TestType
```
adlarkin, line 30: Would a better name for this class be `InternalEnvFixture`? I feel like the name `InternalFixture` doesn't really tell readers what the class does.

test/integration/apply_joint_force_system.cc
```diff
@@ -32,22 +32,16 @@
...
-class ApplyJointForceTestFixture : public ::testing::Test
+class ApplyJointForceTestFixture : public InternalFixture<::testing::Test>
 {
-  // Documentation inherited
-  protected: void SetUp() override
-  {
-    ignition::common::Console::SetVerbosity(4);
-    ignition::common::setenv("IGN_GAZEBO_SYSTEM_PLUGIN_PATH",
-           (std::string(PROJECT_BINARY_PATH) + "/lib").c_str());
-  }
```
adlarkin, line 50: I've noticed that this file (and many other files) have their `SetUp` method removed, instead of calling `InternalFixture::Setup`. I have also noticed that some files call `InternalFixture::SetUp`. What is the criteria for determining whether a file should have `SetUp` removed completely, or replaced with `InternalFixture::SetUp`?

adlarkin (commented):
I still need to do a deeper review pass, but before I do, there are a few higher-level questions that I have.

adlarkin:
There are also a few codecheck issues:
```
  /github/workspace/src/EntityComponentManager_TEST.cc:97:  Lines should be <= 80 characters long  [whitespace/line_length] [2]
  /github/workspace/src/SimulationRunner_TEST.cc:84:  Lines should be <= 80 characters long  [whitespace/line_length] [2]
  /github/workspace/test/integration/examples_build.cc:126:  Lines should be <= 80 characters long  [whitespace/line_length] [2]
  /github/workspace/test/integration/joint_position_controller_system.cc:43:  Lines should be <= 80 characters long  [whitespace/line_length] [2]
  /github/workspace/test/integration/joint_state_publisher_system.cc:32:  Lines should be <= 80 characters long  [whitespace/line_length] [2]
  /github/workspace/test/integration/scene_broadcaster_system.cc:35:  Line ends in whitespace.  Consider deleting these extra spaces.  [whitespace/end_of_line] [4]
  Total errors found: 6
```

ahcorde (dismissed):

test/integration/battery_plugin.cc
chapulina: We're overriding the `SetUp` function, so we explicitly call the base class's `SetUp` to make sure it runs. Since we're not overriding `TearDown`, the base class's function will be called.

test/helpers/EnvTestFixture.hh
chapulina: My concern with the word "environment" is that in GoogleTest that's a separate concept. They have both `::testing::Test` and [::testing::Environment](https://github.com/YOU-i-Labs/googletest/blob/master/googletest/docs/V1_7_AdvancedGuide.md#global-set-up-and-tear-down), and what we're handling here is the test family, not the environments.

I'm open to other name alternatives though. I actually started with `GazeboFixture` :sweat_smile:

test/integration/apply_joint_force_system.cc
chapulina: Since the tests now inherit from `InternalFixture`, they should be calling that class's `SetUp` by default. All setups that did what's now handled by the base class have been removed in favor of using the base class. 

`SetUp`s that do more than what the base class does override the function and call the base class's `SetUp` to make sure that runs as well.

test/helpers/EnvTestFixture.hh
adlarkin: > My concern with the word "environment" is that in GoogleTest that's a separate concept

Oh, I didn't realize that. I was just thinking that adding `environment` (or something similar) to the name could be good to imply to readers/users that this modifies environment variables. But, we can leave it named as-is.

test/integration/apply_joint_force_system.cc
adlarkin: That makes sense :+1: sorry, I totally forgot about the override/inheritance behavior when reviewing this yesterday :upside_down_face:

src/ServerPrivate.cc
```diff
@@ -246,8 +244,6 @@ void ServerPrivate::AddRecordPlugin(const ServerConfig &_config)
       recordPluginElem->Get<std::string>("compress_path", "");
     std::tie(sdfRecordResources, hasRecordResources) =
       recordPluginElem->Get<bool>("record_resources", false);
-    std::tie(sdfCompress, hasCompress) =
-      recordPluginElem->Get<bool>("compress", false);
```
adlarkin, line 250: Based on what I can tell, it looks like the variables `sdfCompress` and `hasCompress` aren't used. Is that why you're removing them?

adlarkin (commented):
LGTM, just one question for clarity/understanding below. Maybe we can get one more set of eyes on this to make sure we aren't introducing any new test failures.

src/ServerPrivate.cc
chapulina: Yup, it came up on codecheck while I was cleaning up the rest, so I thought it was small enough to sneak into this PR.

jennuine (approved):
LGTM

PR #975, opened 2021-08-14
WIP [Same process] Updates to Server-GUI synchronization

chapulina:
Targeted at #793 

## Server-GUI synchronization

Changes:

* `ClientUpdate`: renamed event for clarity, with accompanying documentation
* Removed `GuiRunner`'s update loop when in sameProcess and instead rely on ClientUpdate events
    * I believe these changes are in line with @iche033 's suggestion in https://github.com/ignitionrobotics/ign-gazebo/pull/793#issuecomment-857286602.
    * The problem with keeping the client's loop is that the GUI systems' `Update` may be called at any time. We don't want that because they have access to the same ECM as the server, so GUI systems could be trying to read entities and components while a server system is changing them.

Open issues:

* The client's update is performed in a separate thread that doesn't block the server's loop. This means that GUI systems may be reading from the ECM, while the server is already calling `PreUpdate` on server systems. #racecondition. Some ideas:
    * Emit events for all server updates (`Pre` / `Update` / `Post`). This way, the GUI can hold the `PreUpdate` until its updates end.
    * Treat the `GuiRunner`'s update as one of the `PostUpdates` and add it to the barrier in `SimulationRunner`

* The GUI is not getting sim time (and all `UpdateInfo`) updates when running in the same process, which will make some plugins behave badly.
    * One idea is to pass `UpdateInfo` through the `ClientUpdate` event. And if we're doing that, we might as well also pass the `ECM` and avoid the whole shared ECM logic. This pattern would look closer to our existing APIs.
        * (I know I was the one who suggested the shared ECM in the first place. I hadn't thought we'd be using events back then :innocent: )

* `GuiRunner::UpdatePlugins` calls `ECM::ClearRemovedComponents`. When running in the same process, this should be handled by `SimulationRunner`.

## Sensors - Scene3D synchronization

I'm looking at this next.


ahcorde (changes requested):

chapulina:
Didn't make it into `beta`, can come in a minor release.

chapulina:
This PR has staled, closing. We can still refer to it as needed.

Issue #942, opened 2021-07-29
Distributed simulation through distributing components


Blast545:
After working with #issue, this issue considers a different approach to implement distributed simulation as it is difficult to deal with moving performers from one secondary to the other because a lot of components have internal states that are not being serialized.

The idea could be distributing the rendering/sensors and other `postUpdate` components of the simulation through the set of secondaries. This would reduce the overhead of rendering the sensors in one PC and it may improve the previous distributed implementation if the distributed components do not need to send information back to the primary (not sure if this applies to all sensors).

In each simulation step, after running the `preUpdate` and `Update` phases, the primary will send the serialized state to all the secondaries so those will proceed with the `postUpdate` of their assigned components. I think “running ahead” doesn’t make sense in this scenario, and for a first demo I can set up a perfect synchronization similar to the [current implementation](https://github.com/ignitionrobotics/ign-gazebo/blob/24be610ac1e8889b6c538a110c84da7f91360a39/src/network/NetworkManagerPrimary.cc#L155).

So, the code implementation would be something like this:

- Primary [loads all the instances](https://github.com/ignitionrobotics/ign-gazebo/blob/24be610ac1e8889b6c538a110c84da7f91360a39/src/SimulationRunner.cc#L163) / plugins from the sdf world. 

- After everything is loaded, take the list of systems requiring `postUpdates` and distribute them evenly across the network (maybe some sensors are known in advance to consume more resources than others and might be grouped somehow). This will use a round robin system similar to the one used for the current distributed simulation to [distribute performers](https://github.com/ignitionrobotics/ign-gazebo/blob/24be610ac1e8889b6c538a110c84da7f91360a39/src/network/NetworkManagerPrimary.cc#L239), but instead with components.

- Then, run the simulation controlled by the primary. [This condition](https://github.com/ignitionrobotics/ign-gazebo/blob/24be610ac1e8889b6c538a110c84da7f91360a39/src/SimulationRunner.cc#L547) will be changed to replace a network halt waiting for the response from the sensor's `postUpdate` phase.

I'd like the primary to be able of deciding what to distribute across the secondaries, but for a first attempt to see performance improvements I might hardcode it in the secondaries.

Please let me know your thoughts @nkoenig @chapulina @mjcarroll, do you think this approach sounds reasonable?

PR #985, opened 2021-08-19
Distributed sensors first implementation

Blast545:
First iteration towards #942

## Summary
First demo to approach distributed simulation through distributing the rendering parts. The current implementation does not distribute the components itself, it reuses the existing logic defined to distribute performers to distribute the sensors/postUpdate components associated with each performer.

##Workflow of the implementation
1. Primary populates affinities distributing performers evenly.
2. Primary sets up its simulationRunner to run the systems.
3. Secondaries set up their simulationRunner to run systems with postUpdate methods. Specifically, the sensors plugin. (see secondary.sdf)
4. Primary manages the step/clock of the simulation, full synchronization.
5. Stepping starts with the primary that runs the preUpdate and update sections of the simulation.
6. Primary sends the serialized state to the secondaries.
7. Each secondary runs their assigned postUpdate tasks and replies with an acknowledge to the primary.
8. After primary receives the acknowledge from all the secondaries, it continues with the next step.

(Is it necessary to send the serialized state back from each secondary to the primary? AFAIK the postUpdate phase does not change the ecm components state. If I'm wrong, the secondaries should send the state of their assigned components for the primary to consolidate them).

## Test it

Run in three different terminals:
`./examples/scripts/distributed_rendering/primary.sh`
`./examples/scripts/distributed_rendering/secondary.sh`
`./examples/scripts/distributed_rendering/secondary.sh`

The `sensors_demo.sdf` was adjusted as demo for this PR. Half of the sensors are run in each secondary. Physics step is run in the primary,

## Checklist
- [ ] Added tests

Signed-off-by: Jorge Perez <jjperez@ekumenlabs.com>


src/SimulationRunner.cc
```diff
@@ -480,14 +483,40 @@ void SimulationRunner::AddSystemToRunner(SystemInternal _system)
...
+    }
+  }
+  else
+  {
+    if (_system.preupdate)
+      this->systemsPreupdate.push_back(_system.preupdate);
 
-  if (_system.postupdate)
-    this->systemsPostupdate.push_back(_system.postupdate);
+    if (_system.update)
```
gonzodepedro, line 514: May checking here and below that is not secondary instead of repeating lines in the if

src/SimulationRunner.cc
```diff
@@ -511,7 +540,7 @@ void SimulationRunner::ProcessSystemQueue()
   // If additional systems were added, recreate the worker threads.
   if (pending > 0)
   {
-    igndbg << "Creating PostUpdate worker threads: "
+    ignwarn << "Creating PostUpdate worker threads: "
```
gonzodepedro, line 543: Careful with this log level changes. Please revert those when promoting this from Draft to PR

Blast545:
Code from last demo is tracked up to b14ac8e. Not changing into formal PR because leaving the simulation running for some time crashes my PC. I'm investigating the cause of that.

Blast545:
Included an extra command line option `--network-render` to be able of using this sensors implementation without removing the current performers implementation.

src/SimulationRunner.cc
```diff
@@ -124,20 +124,23 @@ SimulationRunner::SimulationRunner(const sdf::World *_world,
   // If the configuration is invalid, then networkMgr will be `nullptr`.
   if (_config.UseDistributedSimulation())
   {
-    if (_config.NetworkRole().empty())
+    if (_config.NetworkRole() == "primary" ||
```
Blast545, line 127: I changed this bit because at some point there is a partial/incomplete support for ReadOnly network types. However, trying to use those interferes with the rest of the network code that closes the simulation after a node is removed.

src/network/NetworkManagerSecondary.cc
```diff
@@ -179,3 +189,76 @@ void NetworkManagerSecondary::OnStep(
...
+  auto info = convert<UpdateInfo>(_msg.stats());
+
+  // SYNC SECONDARY ECM WITH PRIMARY ECM
+  this->dataPtr->ecm->SetState(_msg.state());
+
+  // Step runner
+  this->dataPtr->stepFunction(info);
+
+  // Update state with all the performer's entities
+  std::unordered_set<Entity> entities;
```
chapulina, line 248: This set doesn't seem to be used

src/network/NetworkManagerPrimary.cc
```diff
@@ -129,19 +142,23 @@ bool NetworkManagerPrimary::Step(const UpdateInfo &_info)
...
 
+  // Different step method for render networks
+  if (this->IsRenderNetwork())
+    return this->RenderingStep(_info);
+
+  private_msgs::SimulationStep step;
+  step.mutable_stats()->CopyFrom(convert<msgs::WorldStatistics>(_info));
+
+  // Affinities that changed this step
+  this->PopulateAffinities(step);
```
chapulina, line 160: I think this needs to be done before the `SecondariesCanStep` so we don't miss it. If we leave it for later, we may not know anymore that the affinities changed (maybe, I haven't checked).

chapulina (commented):
> Is it necessary to send the serialized state back from each secondary to the primary? AFAIK the postUpdate phase does not change the ecm components state. 

I think just a boolean like you're sending now should be enough.

---

What do you think of updating the [distributed_simulation](https://github.com/ignitionrobotics/ign-gazebo/blob/ign-gazebo5/tutorials/distributed_simulation.md) tutorial in this PR? It will need to be updated anyway, and it should help reviewers understand the implementation, working like a sort of design document.

chapulina:
`ign-gazebo4` is EOL'ing with Dome in the coming days, please retarget this PR at a supported version.

FilipeAlmeida-Movai:
What is the current status? Are there any plans to integrate this into Fortress? How can I help you?

nkoenig:
> Are there any plans to integrate this into Fortress? How can I help you?

Yes, this PR is near the top of my review plate. Feel free to try it out, and offer feedback. More eyes and testing is always better.

There is also distributed simulation work being don in  #481.

Blast545:
@FilipeAlmeida-Movai Thanks for your interest in this PR! Can you tell us the setup you used for testing it and the .sdf file you have for these pictures?

At the time this PR was opened we only ran tests with the sensors available in the .sdf files uploaded with this PR, it's mainly a proof of concept to check if distributing the sensors lead to a better performance than the idea proposed with #481

FilipeAlmeida-Movai:
> @FilipeAlmeida-Movai Thanks for your interest in this PR! Can you tell us the setup you used for testing it and the .sdf file you have for these pictures?
> 
> At the time this PR was opened we only ran tests with the sensors available in the .sdf files uploaded with this PR, it's mainly a proof of concept to check if distributing the sensors lead to a better performance than the idea proposed with #481

Yes. I ran the tests scripts that are in the description of this PR,  and I got the same results. It is dividing the render scene between the secondaries, not only the sensors. So a sensor only "sees" what is rendering by the secondary that it was assigned.

You can run the tests and check if all cameras are getting all objects.

I already checked in the code where it is dividing the objects by secondaries and try a modification adapting performers and separate only the render of each performer instead of separate the entire scene and test. I run it and work! I was able to run 2 robots and separate the computation of the sensor of each one by the secondaries.

Furthermore, I am really interested in this PR. I think that dividing the sensors rendering by different GPU process is the best approach to improve the performance of the robots simulation. Please, let me know if I can help with anything!

src/network/NetworkManagerPrimary.cc
```diff
@@ -303,6 +397,33 @@ void NetworkManagerPrimary::PopulateAffinities(
...
+//////////////////////////////////////////////////
+void NetworkManagerPrimary::PopulateRenderingModels(
+    private_msgs::SimulationStateStep &_msg)
+{
+  IGN_PROFILE("NetworkManagerPrimary::PopulateDistributedModels");
+
+  auto secondaryIt = this->secondaries.begin();
+
+  // Go through models and assign a different secondary to each
+  this->dataPtr->ecm->Each<components::Model, components::Name>(
```
FilipeAlmeida-Movai, line 409: Shouldn't that separate only the sensor entities, instead of all models?

mjcarroll:
This has been open for quite some time, I don't think there is any intention to pick it up in the near future.  I'm going to close it.

PR #977, opened 2021-08-17
Joint Force-Torque Systems Plugin

Devansh16:
# 🎉 New feature

## Summary
Systems Plugin Implementation for Joint Force-Torque Sensor.

Needs https://github.com/ignitionrobotics/ign-sensors/pull/144

## Checklist
- [ ] Signed all commits for DCO
- [ ] Added tests
- [ ] Added example and/or tutorial
- [ ] Updated documentation (as needed)
- [ ] Updated migration guide (as needed)
- [ ] `codecheck` passed (See [contributing](https://ignitionrobotics.org/docs/all/contributing#contributing-code))
- [ ] All tests passed (See [test coverage](https://ignitionrobotics.org/docs/all/contributing#test-coverage))
- [ ] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Aignitionrobotics+repo%3Aosrf%2Fsdformat+archived%3Afalse+) to support the maintainers

**Note to maintainers**: Remember to use **Squash-Merge**

PR #952, opened 2021-07-31
Populate joint forces through physics

vatanaksoytezer:
# 🎉 New feature

Closes #883 

## Summary
Exports joint forces through physics.cc to enable force-torque sensors and efforts through joint state publisher.

## Test it
* Pull https://github.com/ignitionrobotics/ign-physics/pull/143 and this PR.
* Open up any world file with joint_state_publisher plugin (for eg. lift_drag.sdf)
* Listen the joint_states topic through ros2 bridge or Ignition

## Checklist
- [x ] Signed all commits for DCO
- [ ] Added tests
- [ ] Added example and/or tutorial
- [ ] Updated documentation (as needed)
- [ ] Updated migration guide (as needed)
- [ ] `codecheck` passed (See [contributing](https://ignitionrobotics.org/docs/all/contributing#contributing-code))
- [ ] All tests passed (See [test coverage](https://ignitionrobotics.org/docs/all/contributing#test-coverage))
- [ ] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Aignitionrobotics+repo%3Aosrf%2Fsdformat+archived%3Afalse+) to support the maintainers

**Note to maintainers**: Remember to use **Squash-Merge**

PR #989, opened 2021-08-20
Populate JointConstraintWrench from physics

azeey:
# 🎉 New feature

## Summary

This populates joint constraint wrenches from ign-physics. This is needed to implement Force/Torque sensors.

Needs https://github.com/ignitionrobotics/ign-physics/pull/283

## Checklist
- [x] Added tests


chapulina:
@azeey , can you comment on the relation between this PR and:

* https://github.com/ignitionrobotics/ign-gazebo/pull/952
* https://github.com/ignitionrobotics/ign-gazebo/pull/977

azeey:
#952 populates forces in the DOF of the joint whereas this PR populates the constraint (non-DOF direction) wrench on the joint. This wrench will not (hopefully) include the commanded force in the DOF of the joint. #952 need https://github.com/ignitionrobotics/ign-physics/pull/143, which we're still trying to resolve. 

#977 I believe is the ign-gazebo counter part to the ign-sensors Force/Torque sensor.

chapulina:
Thanks, so do both this PR and #952 work together, or do we have to choose one or the other?

And I see some overlap between #977 and this one. Is that a superset of this?

include/ignition/gazebo/components/JointTransmittedWrench.hh
```diff
@@ -0,0 +1,96 @@
...
+class WrenchSerializer
+{
+  /// \brief Serialization for `Wrench`.
+  /// \param[in] _out Output stream.
+  /// \param[in] _wrench Wrench to stream
+  /// \return The stream.
+  public: static std::ostream &Serialize(std::ostream &_out,
+                                         const Wrench &_wrench)
+  {
+    _out << _wrench.torque << " " << _wrench.torque;
```
chapulina, line 60:
```suggestion
    _out << _wrench.torque << " " << _wrench.force;
```

include/ignition/gazebo/components/JointTransmittedWrench.hh
```diff
@@ -0,0 +1,96 @@
...
+    return _in;
+  }
+};
+}  // namespace serializers
+
+namespace components
+{
+/// \brief Joint Transmitted wrench in SI units (Nm for revolute, N for
+/// prismatic). The first set of 3 values coorespond to the torque while the
+/// second set of 3 correspond to the force. The wrench is expressed in the
```
chapulina, line 81: The part about the set of values doesn't make much sense since we're using a struct?

include/ignition/gazebo/components/JointTransmittedWrench.hh
```diff
@@ -0,0 +1,96 @@
...
+namespace gazebo
+{
+// Inline bracket to help doxygen filtering.
+inline namespace IGNITION_GAZEBO_VERSION_NAMESPACE {
+
+/// \brief Data structure that contains torque and force.
+struct Wrench
+{
+  math::Vector3d torque;
+  math::Vector3d force;
```
chapulina, line 38: Document all members, this is a good place to mention units and reference frames

chapulina (changes requested):
I haven't tried running it but the code and test looks reasonable to me.

Can we add unit tests for the component in `test/integration/components.cc`? That would have caught the serialization typo

include/ignition/gazebo/components/JointTransmittedWrench.hh
azeey: I feel like this struct should go into a different header so it can be used by more than one component, but I'm not sure where. Would `Types.hh` be a good place?  We do use `msgs::Wrench` for the `ExternalWorldWrenchCmd` component, but I remember a discussion where we said we should avoid using messages as data types for components if possible.

chapulina: You know, I'd lean towards using `ignition::msgs::Wrench` here instead of yet another struct. That also gives us serialization for free. What do you think?

In general, I've been trying to use messages in components as much as possible. This is in relation to https://github.com/ignitionrobotics/ign-gazebo/issues/494

azeey:
@osrf-jenkins run test

azeey:
Looks like the `prerelease` repo is not enabled on CI, so it's using `ignition-physics5` from nightlies. Do we need to enable prereleases?

chapulina:
> Looks like the prerelease repo is not enabled on CI, 

Ahhhhh see https://github.com/ignition-tooling/gzdev/pull/35

chapulina:

include/ignition/gazebo/components/JointTransmittedWrench.hh
azeey: Change to force and torque components in https://github.com/ignitionrobotics/ign-gazebo/pull/989/commits/41310933ca384a67559f7b166dbabb0488b617e7

include/ignition/gazebo/components/JointTransmittedWrench.hh
azeey: Changed to `msgs::Wrench` in https://github.com/ignitionrobotics/ign-gazebo/pull/989/commits/41310933ca384a67559f7b166dbabb0488b617e7. One issue is that the message has a `offset_point` member, which we don't support. I made a note of that in the documentation.

chapulina (approved):

PR #992, opened 2021-08-24
Add linear velocity components to links that have linear velocity cmds applied

### LINKED ISSUE #966: VelocityCmd components applied to models don't have any effect
## Environment

Ubuntu 20.04, Ignition Edifice source build (latest ign-gazebo5 branch).

## Description

We relied on using `LinearVelocityCmd` and `AngularVelocityCmd` to apply velocities at the model level in our [robot simulation plugins](https://github.com/open-rmf/rmf_simulation/blob/main/rmf_robot_sim_ignition_plugins/src/slotcar.cpp#L155-L167). However, I noticed that recently the commands don't have any effect.
An example SDF that we use this plugin with is [here](https://github.com/open-rmf/rmf_demos/blob/main/rmf_demos_assets/models/TinyRobot/model.sdf).
I ran `git bisect` and found out the exact commit that introduced this behavior, #678. I didn't look into the commit diff too much but my guess is that, since we are applying the velocity to a model entity and not a link entity it is skipped by that optimization PR.
I wonder if there was a recommended way to adapt our plugin or if this is something that should be fixed in the core code. I tried to fetch the model's canonical link through the `ModelCanonicalLink` component and applying the velocities to it but didn't have much success.

## Steps to reproduce

It is arguably a bit complicated, it would require setting up the whole `rmf_demos` workspace, but let me know if you want detailed instructions.

adlarkin:

# 🦟 Bug fix

Fixes #966

## Summary
When a link has a `LinearVelocityCmd` applied to it in a given simulation step, there's no way to read what the link's linear velocity is in future simulation steps since [the `LinearVelocityCmd` is "zeroed out" after it is applied](https://github.com/ignitionrobotics/ign-gazebo/blob/ignition-gazebo3_3.9.0-pre2/src/systems/physics/Physics.cc#L2149-L2154). This PR adds a `LinearVelocity` and `WorldLinearVelocity` component to a link when a `LinearVelocityCmd` is applied to it, so that users have a way to read a link's linear velocity in future simulation steps. I've added a test case to check this in the physics system integration test, and have verified that running the test to read linear velocity data from a link with a `LinearVelocityCmd` applied to it fails without the other changes made in this PR.

We should probably make similar updates for `LinearVelocityCmd`s applied to models, but this will be easier to do starting in `ign-gazebo5` since we can make use of #736 and #685 (if a `LinearVelocityCmd` is applied to a model and we want to create `LinearVelocity` and `WorldLinearVelocity` components for the model, we could populate the data in these components with the new velocity information from the [model's canonical link](https://github.com/ignitionrobotics/ign-gazebo/blob/ignition-gazebo3_3.9.0-pre2/src/systems/physics/Physics.cc#L1799)).

On a somewhat related note to this, we could also make similar changes for `AngularVelocityCmd`, so that users who apply `AngularVelocityCmd`s can read the corresponding entity's angular velocity data in future simulation steps. We could either open a new PR for this, or add it to this one if this change seems like something worthwhile.

## Checklist
- [ ] Added tests

### DIFF
--- a/src/systems/physics/Physics.cc
+++ b/src/systems/physics/Physics.cc
@@ -1671,6 +1671,11 @@ void PhysicsPrivate::UpdatePhysics(EntityComponentManager &_ecm)
         worldLinearVelFeature->SetWorldLinearVelocity(
             math::eigen3::convert(worldLinearVel));
 
+        // create linear velocity components for the link, if they don't
+        // already exist. The component data is set in PhysicsPivate::UpdateSim
+        _ecm.ComponentDefault<components::LinearVelocity>(_entity);
+        _ecm.ComponentDefault<components::WorldLinearVelocity>(_entity);
+
         return true;
       });
 
@@ -1848,7 +1853,12 @@ void PhysicsPrivate::UpdateSim(EntityComponentManager &_ecm)
 
           // Populate world poses, velocities and accelerations of the link. For
           // now these components are updated only if another system has created
-          // the corresponding component on the entity.
+          // the corresponding component on the entity. An exception to this is
+          // if a LinearVelocityCmd component is applied to a link - if this
+          // happens, then LinearVelocity and WorldLinearVelocity components
+          // are created for this link in case the user who applied the velocity
+          // command would like to read the link's velocity in a future
+          // simulation step.
           auto worldPoseComp = _ecm.Component<components::WorldPose>(_entity);
           if (worldPoseComp)
           {
--- a/test/integration/physics_system.cc
+++ b/test/integration/physics_system.cc
@@ -33,6 +33,7 @@
 
 #include "ignition/gazebo/Server.hh"
 #include "ignition/gazebo/SystemLoader.hh"
+#include "ignition/gazebo/Util.hh"
 #include "ignition/gazebo/test_config.hh"  // NOLINT(build/include)
 
 #include "ignition/gazebo/components/AxisAlignedBox.hh"
@@ -47,6 +48,7 @@
 #include "ignition/gazebo/components/JointVelocityReset.hh"
 #include "ignition/gazebo/components/Link.hh"
 #include "ignition/gazebo/components/LinearVelocity.hh"
+#include "ignition/gazebo/components/LinearVelocityCmd.hh"
 #include "ignition/gazebo/components/Material.hh"
 #include "ignition/gazebo/components/Model.hh"
 #include "ignition/gazebo/components/Name.hh"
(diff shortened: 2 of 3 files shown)


azeey:
If the user wants the `LinearVelocity` of a link, shouldn't they create the component themselves? I say this because there are various components that are not created by default and the pattern we've been following so far is to expect users to create the components they need.

adlarkin:
> If the user wants the LinearVelocity of a link, shouldn't they create the component themselves?

I'm not opposed to this, but if we expect the user to create this component themselves, then how can they accurately populate the data for this component? They could set the data of the `LinearVelocity` component to be the data of the `LinearVelocityCmd` component, but the actual velocity vs. the commanded velocity will differ slightly due to physics forces, right?

azeey:
Maybe we don't have this documented very well, but in general, when users want a value like `LinearVelocity`, they create the component on the entity in the `PreUpdate`. The physics system will then know to populate the component from the physics engine in
https://github.com/ignitionrobotics/ign-gazebo/blob/0c401c17c1d4b748233c61921c47da5725ce9d5c/src/systems/physics/Physics.cc#L2432-L2563

adlarkin:
> Maybe we don't have this documented very well, but in general, when users want a value like LinearVelocity, they create the component on the entity in the PreUpdate. The physics system will then know to populate the component from the physics engine

Yeah, I just realized this - see my edit on https://github.com/ignitionrobotics/ign-gazebo/pull/992#issuecomment-905010859. So, should this PR be closed? If so, I can post on #992 that the solution is to create a `LinearVelocity` component for links of interest.

chapulina:
> Maybe we don't have this documented very well, but in general, when users want a value like LinearVelocity, they create the component on the entity in the PreUpdate

For beginners, my recommendation would be to encourage them to use [Link::EnableVelocityChecks](https://ignitionrobotics.org/api/gazebo/5.1/classignition_1_1gazebo_1_1Link.html#afa6f4d13e23b7046dbcf152cd078bd4d), which creates the necessary components. But I agree that it's a good idea to document that thee components should be created, for power users.

adlarkin:
I guess my only other remark is that it looks like we currently don't have code in the physics system that would take care of updating a model's velocity if a `LinearVelocity` component was attached to a model. It looks like we have implementation for handling links, sensors and collisions. However, I guess that if a user wanted to know a model's velocity/acceleration/etc., they could just look at the model's canonical link, right?

luca-della-vedova:
I agree with not adding `LinearVelocity` components automatically, building up on what @adlarkin mentioned, I wonder if you think it's a sensible default to assign to a Model the velocity of its canonical link. For reference [this](https://github.com/ignitionrobotics/ign-gazebo/blob/a4fbd0f817785ebc7e654c3621412b1dee6bafa0/src/systems/lin_vel_cmd/LinVelCmdTest.cc) was the sample system I created for the issue.
It works once I change the entity from the model's to the canonical link's. I think there is a bit of disconnect since the `LinearVelocityCmd` _can_ be applied to the model, and when it is the model itself moves, but the model's `LinearVelocity` component isn't updated and always reads back as zero.
I guess it can be made more coherent in two ways, either make it impossible to apply `LinearVelocityCmd` to models and force users to only work at the canonical link level, or make sure that when `LinearVelocityCmd` components are applied to models the model's `LinearVelocity` component is also updated.
I personally prefer the second since I think defaulting a model's velocity to that of its canonical link is a reasonable assumption, but to be fully honest I never heard of canonical links until a few days ago so I don't know if it makes sense :sweat_smile:

adlarkin:
> I think there is a bit of disconnect since the LinearVelocityCmd can be applied to the model, and when it is the model itself moves, but the model's LinearVelocity component isn't updated and always reads back as zero.

The reason why this is the current behavior is because in the physics system, we get updated [`frameData`](https://github.com/ignitionrobotics/ign-physics/blob/ignition-physics4_4.2.0/include/ignition/physics/FrameData.hh#L42) for links that have changed. But yes, I agree that this is confusing :grimacing: 

> I guess it can be made more coherent in two ways, either make it impossible to apply LinearVelocityCmd to models and force users to only work at the canonical link level, or make sure that when LinearVelocityCmd components are applied to models the model's LinearVelocity component is also updated.

I am not opposed to applying a `LinearVelocityCmd` to a model and then updating the model's `LinearVelocity` component if it exists - we'd just need to figure out how to implement this functionality (right now, nothing immediate comes to mind regarding implementation, but perhapsit wouldn't be too hard to implement). For now, I think that using the canonical link should be fine, since I understand the canonical link to serve as a "reference" for the model's velocity, pose, acceleration, etc.

@azeey what do you think about adding functionality for updating `LinearVelocityCmd` components that are attached to models? Do you envision any challenges with implementing this?

azeey:
I'm on board with setting `LinearVelocityCmd` on models and updating the `LinearVelocity` to reflect the actual velocity of the model frame. However, note that the model frame is not the same as canonical link frame since there can be a transform between the two. For example:

```xml
<model name="M">
  <link name="L">
        <pose>1 0 0   0 0 0</pose>
   </link>
</model>
```

The model frame of `M` is translated from the canonical link by `-1 0 0`.

The reason we he haven't implemented updates to the `LinearVelocity` component of models is because the dartsim plugin of ign-physics does not make the frame data readily available for models because it doesn't implement `ModelFrameSemantics`. But if we want to do it, we can compute the frame data for the model frame from the frame data of the canonical link in ign-gazebo.

PR #1001, opened 2021-08-30
Lock views during system PostUpdates

adlarkin:

# 🦟 Bug fix

Fixes a race condition discovered in https://github.com/ignitionrobotics/ign-gazebo/pull/959#issuecomment-906883962

## Summary
As a consequence of the changes in #856, new entities aren't added to a view until the view is used, because we need access to the template signature of the view when adding entities to it. This introduces a potential race condition when processing system `PostUpdate`s since system `PostUpdate`s are run in parallel (see https://github.com/ignitionrobotics/ign-gazebo/pull/959#issuecomment-906883962 for a more detailed explanation).

The fix proposed here is to assign a mutex to every view that exists. Then, when processing system `PostUpdate`s, we lock the view while entities are being added to it. Since every view has its own mutex, this means that different views can still be accessed in parallel during `PostUpdate` calls - the only time blocking may occur is when two system `PostUpdate`s are trying to read from the same view. So, there shouldn't be any major performance hits here.

## Checklist
- [ ] Added tests

### DIFF
--- a/src/EntityComponentManager.cc
+++ b/src/EntityComponentManager.cc
@@ -133,8 +133,15 @@ class ignition::gazebo::EntityComponentManagerPrivate
   public: mutable std::mutex removedComponentsMutex;
 
   /// \brief The set of all views.
+  /// The value is a pair of the view itself and a mutex that can be used for
+  /// locking the view to ensure thread safety when adding entities to the view.
   public: mutable std::unordered_map<detail::ComponentTypeKey,
-          std::unique_ptr<detail::BaseView>, detail::ComponentTypeHasher> views;
+          std::pair<std::unique_ptr<detail::BaseView>,
+            std::unique_ptr<std::mutex>>, detail::ComponentTypeHasher> views;
+
+  /// \brief A flag that indicates whether views should be locked while adding
+  /// new entities to them or not.
+  public: bool lockAddEntitiesToViews{false};
 
   /// \brief Cache of previously queried descendants. The key is the parent
   /// entity for which descendants were queried, and the value are all its
@@ -271,7 +278,7 @@ void EntityComponentManager::ClearNewlyCreatedEntities()
 
   for (auto &view : this->dataPtr->views)
   {
-    view.second->ResetNewEntityState();
+    view.second.first->ResetNewEntityState();
   }
 }
 
@@ -319,7 +326,7 @@ void EntityComponentManager::RequestRemoveEntity(Entity _entity,
   {
     for (auto &view : this->dataPtr->views)
     {
-      view.second->MarkEntityToRemove(removedEntity);
+      view.second.first->MarkEntityToRemove(removedEntity);
     }
   }
 }
@@ -377,7 +384,7 @@ void EntityComponentManager::ProcessRemoveEntityRequests()
       // Remove the entity from views.
       for (auto &view : this->dataPtr->views)
       {
-        view.second->RemoveEntity(entity);
+        view.second.first->RemoveEntity(entity);
       }
     }
     // Clear the set of entities to remove.
@@ -420,7 +427,7 @@ bool EntityComponentManager::RemoveComponent(
 
     // update views to reflect the component removal
     for (auto &viewPair : this->dataPtr->views)
-      viewPair.second->NotifyComponentRemoval(_entity, _typeId);
+      viewPair.second.first->NotifyComponentRemoval(_entity, _typeId);
   }
 
   this->dataPtr->AddModifiedComponent(_entity);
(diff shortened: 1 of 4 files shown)


src/EntityComponentManager.cc
```diff
@@ -795,27 +802,32 @@ const EntityGraph &EntityComponentManager::Entities() const
...
-  auto iter = this->dataPtr->views.insert(
-      std::make_pair(_types, std::move(_view))).first;
-  return iter->second.get();
+  auto iter = this->dataPtr->views.find(_types);
+  if (iter != this->dataPtr->views.end())
+    return iter->second.first.get();
+  this->dataPtr->views[_types] =
+    std::make_pair(std::move(_view), std::make_unique<std::mutex>());
+  return this->dataPtr->views[_types].first.get();
 }
```
adlarkin, line 831: I tried to modify the `this->dataPtr->views.insert` call to pass in a `std::pair<detail::ComponentTypeKey, std::pair<std::unique_ptr<detail::BaseView>, std::unique_ptr<std::mutex>>`, but was having trouble getting it to compile. So, I resorted to the changes made here (searching and then manually adding if the search didn't find anything). I don't think this is a bad change though since the `AddView` method is only called [here](https://github.com/ignitionrobotics/ign-gazebo/blob/166c06580036943c62d49422c3ef235b4eb7ce5d/include/ignition/gazebo/detail/EntityComponentManager.hh#L549), and the changes made here introduce only an extra lookup or two in a `std::unordered_map`, which already has `O(1)` lookup time. Since the `AddView` method is only called when a requested view doesn't already exist, this means that this code should rarely be called anyways, so I don't think the extra lookups are a problem.

src/SimulationRunner.cc
```diff
@@ -572,13 +572,15 @@ void SimulationRunner::UpdateSystems()
...
+    this->entityCompMgr.LockAddingEntitiesToViews(true);
     // If no systems implementing PostUpdate have been added, then
     // the barriers will be uninitialized, so guard against that condition.
     if (this->postUpdateStartBarrier && this->postUpdateStopBarrier)
     {
       this->postUpdateStartBarrier->Wait();
       this->postUpdateStopBarrier->Wait();
     }
+    this->entityCompMgr.LockAddingEntitiesToViews(false);
   }
```
adlarkin, line 584: I know that server system `PostUpdate`s are ran in parallel (i.e., across multiple threads), but what about GUI systems? Do any GUI system `*Update` methods run in parallel as well?

adlarkin (commented):
I've got a few comments/questions for PR reviewers.

src/SimulationRunner.cc
iche033: I think the gui system updates do not run in parallel:
https://github.com/ignitionrobotics/ign-gazebo/blob/ign-gazebo5/src/gui/GuiRunner.cc#L199-L202

adlarkin: Great, thanks for the link! I will go ahead and mark this as resolved, since it looks like we only need to do this for server systems.

src/EntityComponentManager.cc
iche033: is this something that you're thinking of doing?

```
  auto iter = this->dataPtr->views.insert(
    std::make_pair(_types,
    std::make_pair(std::move(_view), std::make_unique<std::mutex>()))).first;
  return iter->second.first.get();
```

iche033:
what's a good way to test this PR? Should I merge these changes with #959 and try cloning lights?

adlarkin:
> what's a good way to test this PR? Should I merge these changes with #959 and try cloning lights?

Yeah, that's what I did! Cloning lights will probably segfault without these changes. I'd try cloning a light several times with/without these changes to make sure that the race condition is fixed.

src/EntityComponentManager.cc
adlarkin: Yeah, that's it! I tried what you suggested above, and it works. I think the issue I was having before was that I made the view/mutex pair outside of the `insert` call, and then the compiler was complaining to me about trying to copy a `unique_ptr`. I went ahead and made the update in https://github.com/ignitionrobotics/ign-gazebo/pull/1001/commits/395495816478b6aa102dc807b6e8b2d09f291ea3.

iche033:
hmm I merged these changes into  the `adlarkin/clone_entities` branch but I'm still getting crashes. I see this error msg in the console:

```
[Err] [BaseStorage.hh:927] Another item already exists with name: tube_light_copy
```

I followed the steps you mentioned in #959:

1. Pause sim
2. clone tube light
3. remove tube light clone through Entity Tree
4. clone tube light again
5. Run sim -> crash

I don't get the crash if sim is not paused though.

adlarkin:
> I don't get the crash if sim is not paused though.

Oh yeah, so the steps I mentioned about multiple clones/removals while simulation is paused in https://github.com/ignitionrobotics/ign-gazebo/pull/959 is actually a separate bug that I don't believe is related to this PR (sorry about the confusion). That seems to be something separate that needs to be fixed, which is why I brought it up, but it has nothing to do with the race condition (at least, I don't think). With this PR, you should have no issues cloning/removing lights while simulation is running, and should have no issues cloning an entity while simulation is paused, as long as you don't remove it and try to re-clone it again while simulation is paused.

iche033:
ok I see. I tried the `adlarkin/clone_entities` again and I get a crash when cloning the light while sim is running. With the changes in this PR, I no longer get the crash.

PR #1003, opened 2021-08-31
Adds velocity control to JointPositionController.

arjo129:

# 🎉 New feature

Closes #<NUMBER>

## Summary
There are times when we don't nessecarily want to tune a PID for each joint. Often we
may assume that the physical robot already has a well tuned controller. This
was true in the case of the mbari LRAUV.
Hence, there should be a way to set position without the use of a PID
controller.  This commit adds a `use_velocity_commands` option which when set to true bypasses the force
based PID giving a very predictable position. The `use_velocity_command` mode has a `cmd_max` which in turn
can be used to set a maximum velocity which the joint can move at.

## TODO:
- ~~Add unit test.~~
- The current implementation was developed for a prismatic joint. We need to add angle wrapping for a prismatic joint otherwise the controller may take the long way around. For now we maynot actually need this.

## Checklist
- [x] Added tests


src/systems/joint_position_controller/JointPositionController.cc
```diff
@@ -63,6 +64,14 @@ class ignition::gazebo::systems::JointPositionControllerPrivate
 
   /// \brief Joint index to be used.
   public: unsigned int jointIndex = 0u;
+
+  /// \brief Operation modes
+  enum OperationMode {
+    PID, ABSOLUTE
```
chapulina, line 70: Style nit: put each enum in its own line and give it a `\brief`

src/systems/joint_position_controller/JointPositionController.cc
```diff
@@ -240,14 +258,56 @@ void JointPositionController::PreUpdate(
...
+    JointPositionControllerPrivate::OperationMode::ABSOLUTE)
+  {
+    // Calculate target velcity
+    double targetVel = 0;
+
+    // Get time in seconds
+    auto dt = std::chrono::duration<double>(_info.dt).count();
+
+    // Get the maximum amount in m that this joint may move
+    auto maxMovement = this->dataPtr->posPid.CmdMax() * dt;
```
chapulina, line 280: Should we also use `CmdMin` for negative velocities?

chapulina (commented):
LGTM, I just have a question about using `cmd_min`.

src/systems/joint_position_controller/JointPositionController.cc
arjo129: Resolved in 8ceaed2

src/systems/joint_position_controller/JointPositionController.cc
arjo129: I don't think its absolutely essential. We could add this in a future PR/release. I guess it would simulate actuators which have asymmetric forward and backward speeds. That being said, I do not feel it is necessary immediately. To be honest, this mode will be used usually when we don't really care about motor characteristics.

arjo129:
@chapulina This is ready for another round of review.

chapulina (approved):

PR #1012, opened 2021-09-07
Config file loading options

### LINKED ISSUE #796: Gazebo hangs when loading a model with a plugin
## Environment
* OS Version: Ubuntu 18.04
* Source or binary build? Binary, Ignition Dome

## Description

Running this world will cause Ignition Gazebo to hang at start because the included model has a plugin.

```
<?xml version="1.0" ?>
<sdf version="1.6">
  <world name="default">
    <include>
      <uri>https://fuel.ignitionrobotics.org/1.0/openrobotics/models/finals staging area</uri>
    </include>
  </world>
</sdf>
```

I believe the problem is that Ignition Gazebo sees the plugin, which in this case is a performer detector plugin, and doesn't load any additional plugins. Adding the following plugins to the world  will make gazebo behave correctly.

```
    <plugin
      filename="ignition-gazebo-physics-system"
      name="ignition::gazebo::systems::Physics">
    </plugin>
    <plugin
      filename="ignition-gazebo-sensors-system"
      name="ignition::gazebo::systems::Sensors">
      <render_engine>ogre2</render_engine>
    </plugin>
    <plugin
      filename="ignition-gazebo-user-commands-system"
      name="ignition::gazebo::systems::UserCommands">
    </plugin>
    <plugin
    </plugin>
```

I think we have an opt-in approach with plugins, which can cause strange and silent failures like this. An opt-out approach might be nicer, where an expert user can explicitly say that they don't want a plugin.

chapulina:
# 🎉 New feature

Closes #796

## Summary

Trying some documentation-driven-development :writing_hand: 

I'd like some feedback on the intended behavior before working on the implementation. See also the conversation on #796.

I'm not planning on changing the default behavior for Fortress because I couldn't come up with a migration plan that would have minimal impact on users, but I'm open to ideas.

Whatever is mentioned here for the GUI config will also apply to the server config.

## Test it

TODO

## Checklist
- [ ] Added tests


tutorials/gui_config.md
```diff
@@ -15,10 +15,34 @@ There are a few places where the GUI configuration can come from:
...
+The table below summarizes how all the options behave together:
+
+--gui-config     | SDF `<gui>`  | --gui-config-option      | Result
+---------------- | ------------ | ------------------------ | ------
+empty            | no plugins   | empty / force / prepend  | Load plugins from `$HOME/.ignition/gazebo/gui.config`
+empty / provided | no plugins   | ignore                   | Load no plugins
+provided         | no plugins   | empty / force / prepend  | Load plugins from `--gui-config`
+empty            | has plugins  | empty / ignore           | Load plugins from SDF
+empty            | has plugins  | force                    | `$HOME/.ignition/gazebo/gui.config`
+empty            | has plugins  | prepend                  | Load plugins from `$HOME/.ignition/gazebo/gui.config` and SDF
```
mjcarroll, line 42:
```suggestion
empty            | has plugins  | prepend                  | Load plugins from `$HOME/.ignition/gazebo/gui.config` then SDF
```

tutorials/gui_config.md
```diff
@@ -15,10 +15,34 @@ There are a few places where the GUI configuration can come from:
...
+provided         | has plugins  | empty / force            | Load plugins from `--gui-config`
+provided         | has plugins  | ignore                   | Load plugins from SDF
+provided         | has plugins  | prepend                  | Load plugins from `--gui-config` and SDF
```
mjcarroll, line 45:
```suggestion
provided         | has plugins  | prepend                  | Load plugins from `--gui-config` then SDF
```

tutorials/gui_config.md
```diff
@@ -15,10 +15,34 @@ There are a few places where the GUI configuration can come from:
 2. A `<gui>` element inside an SDF file
 3. The default configuration file at `$HOME/.ignition/gazebo/gui.config` \*
 
-Each of the items above takes precedence over the ones below it. For example,
-if a user chooses a `--gui-config`, the SDF's `<gui>` element is ignored. And
-the default configuration file is only loaded if no configuration is passed
-through the command line or the SDF file.
+Each of the items above takes precedence over the ones below it by default.
+For example, if a user chooses a `--gui-config`, the SDF's `<gui>` element is
+ignored. And the default configuration file is only loaded if no configuration
```
mjcarroll, line 20:
```suggestion
ignored.  The default configuration file is only loaded if no configuration
```

tutorials/gui_config.md
```diff
@@ -15,10 +15,34 @@ There are a few places where the GUI configuration can come from:
...
+is passed through the command line or the SDF file.
+
+The default behaviour can be overridden using the `--gui-config-options`
+command line argument, which offers options:
+
+* `force`: Force the use of the GUI config file, ignoring the SDF file. If a
+  `--gui-config` is provided explicitly, that is used, otherwise, the default
+  config file is used.
+* `ignore`: Ignore the config file completely and load only plugins from SDF.
+* `prepend`: Load plugins from both the config file and the SDF file. The config
```
mjcarroll, line 30: The only thing that I think could be troublesome here is if you had duplicate plugins.  The scenario that I'm thinking of would be  having "specific" configuration in the SDF file and "generic" configuration in the `gui.config` file.

In that case, I think that you would want to prefer the "specific" configuration from the SDF file.

The implementation should take into account this scenario as well as the possibility of having duplicated plugins from the two configurations available.

mjcarroll (commented):
A few grammar nits as well as a more general comment about loading order resolution.

tutorials/gui_config.md
```diff
@@ -15,10 +15,34 @@ There are a few places where the GUI configuration can come from:
...
-through the command line or the SDF file.
+Each of the items above takes precedence over the ones below it by default.
+For example, if a user chooses a `--gui-config`, the SDF's `<gui>` element is
+ignored. And the default configuration file is only loaded if no configuration
+is passed through the command line or the SDF file.
+
+The default behaviour can be overridden using the `--gui-config-options`
+command line argument, which offers options:
+
+* `force`: Force the use of the GUI config file, ignoring the SDF file. If a
```
nkoenig, line 26: How about making `force` override `--gui-config` as well? Specifying both `--gui-config-options force` and `--gui-config` is a bit odd, and probably will never be done. However, it seems easier and cleaner to  remember that `force` is a complete `force` without a caveat.

chapulina: I'm not sure I understand the proposal. Are you saying that `--gui-config <path_to_file>` should imply `force`? I believe that's already in the current proposal. If `--gui-config` is provided and `--gui-config-option` is empty, it works like a `force`.

See these lines on the table below:

```
--gui-config     | SDF `<gui>`  | --gui-config-option      | Result
---------------- | ------------ | ------------------------ | ------
provided         | no plugins   | empty / force / prepend  | Load plugins from `--gui-config`
provided         | has plugins  | empty / force            | Load plugins from `--gui-config`
```

tutorials/gui_config.md
chapulina: Yeah that's complicated to handle, because there are valid use cases for duplicate plugins, on both the GUI and the server. You may want multiple `ImageDisplay`s or multiple `WindEffects`, for example.

I don't have a good idea on how to handle that. Maybe each plugin can be responsible for shortcutting multiple instantiation?

chapulina:
I'll remove this from the `beta` for now to give it some more thought. It would be nice to handle plugins repeated not on purpose nicely.

tutorials/gui_config.md
chapulina: I added a `merge` option to cover this use case, with an example. Let me know what you think 4ab37d6dddc2d9feb64a2baaed21719a7b95ae88. Coming back to this after so many months, now I think this will be my favorite option.

chapulina: I reduced the number of options to 2 in d7a7677c3791890d5116705ecbc1f06a682b7044 and will work on the implementation now

chapulina:
I won't have time to go back to this soon

PR #1034, opened 2021-09-17
Fix minimal scene render engine gui name from command line arg

ahcorde:

# 🦟 Bug fix

## Summary

The render engine GUI name was not passed to the GUI side, the minimal scene was not aware of the render engine GUI name when it was defined in the command line arguments. This PR fixes the issue and allows to change the render engine of the minimal scene using the command line. If the command line is defined will overwrite the value defined in the sdf. 

Require:
 - https://github.com/ignitionrobotics/ign-gui/pull/286

## Checklist
- [ ] Added tests


chapulina (approved):
Great idea to set a property on the window, works for me! :+1: 

I pushed some changes in c6cfb44a02fdd55dc60ff265a2788d825c1edb0e so we don't need to deprecated the public functions. Adding an argument that has a default value on a new major release is fine, because it breaks ABI, but doesn't break API.

chapulina:
@osrf-jenkins run tests pleaseee

PR #1037, opened 2021-09-19
Optimize performance by skipping scene update when possible

darksylinc:
## Summary

> **IMPORTANT:**
>
> This PR depends on ignitionrobotics/ign-rendering#415 and should not be merged until that PR is.

Ray queries (performed every time the mouse moves) had a negative
performance impact caused by
https://github.com/ignitionrobotics/ign-rendering/pull/415

This commit prevents repetead calls to IgnRenderer::ScreenToScene to
perform unnecessary performance degradation since calling
ClosestPoint(true) the first time during the frame is enough

Code will not compile until PR ignitionrobotics/ign-rendering#415 is
merged

## Checklist
- [ ] Added tests

### DIFF
--- a/src/gui/plugins/scene3d/Scene3D.cc
+++ b/src/gui/plugins/scene3d/Scene3D.cc
@@ -107,6 +107,10 @@ inline namespace IGNITION_GAZEBO_VERSION_NAMESPACE {
     /// \brief Flag to indicate if hover event is dirty
     public: bool hoverDirty = false;
 
+    /// \brief Flag to indicate if we have to call
+    /// SceneManager::updateAllTransforms before rayQuery
+    public: bool rayQuerySceneDirty = true;
+
     /// \brief Mouse event
     public: common::MouseEvent mouseEvent;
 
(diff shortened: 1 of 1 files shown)


darksylinc:
**Note:** As I'm studying a few changes I noticed we may be able to optimize it back to pre ignitionrobotics/ign-rendering#415 levels by calling `this->HandleMouseEvent();` either before `this->dataPtr->renderUtil.Update();` or after `this->dataPtr->camera->Update();`

This would cause mouse events to lag behind by 1 frame; but no update of scene would be required (thus no extra CPU needed). Without ignitionrobotics/ign-rendering#415; gazebo's GUI is currently operating on data that is behind by 1 frame anyway.

chapulina:
Note that from Fortress, `GzScene3D` is being deprecated in favor of [`MinimalScene`](https://github.com/ignitionrobotics/ign-gui/blob/main/src/plugins/minimal_scene/MinimalScene.cc) (still WIP), so we'll need to port this fix there). CC @ahcorde

chapulina:
@darksylinc , are you still working on this?

darksylinc:
I forgot about this but because I thought it was already merged.

If I understand correctly this code should be ported to [MinimalScene](https://github.com/ignitionrobotics/ign-gui/blob/main/src/plugins/minimal_scene/MinimalScene.cc) now?

chapulina:
> If I understand correctly this code should be ported to [MinimalScene](https://github.com/ignitionrobotics/ign-gui/blob/main/src/plugins/minimal_scene/MinimalScene.cc) now?

Yes, please 🙇🏽‍♀️  You're welcome to leave the code in `GzScene3D` too because Fortress users may still be relying on it.

chapulina:
`Scene3D` has been removed from the `gz-sim7` branch. So this needs to target `MinimalScene` on gz-gui. Closing.

PR #1050, opened 2021-09-24
Place config files in a versioned directory

### LINKED ISSUE #765: Separate config files according to version
## Desired behavior

We place GUI and server configuration files under `~/.ignition/gazebo`, as explained on these tutorials:

* https://ignitionrobotics.org/api/gazebo/5.0/gui_config.html
* https://ignitionrobotics.org/api/gazebo/5.0/server_config.html

The problem is that these files may differ according to Gazebo's version, and if the user has more than one version installed, they may try to load an unsupported config. Ideally, the files should be installed in versioned folders, i.e.:

* `~/.ignition/gazebo/3/gui.config`
* `~/.ignition/gazebo/4/server.config`

And Gazebo should only save and load files from the current version's folder.

## Alternatives considered

Instead of version numbers, we could have collection names, i.e. `citadel`, `dome`.

## Implementation suggestion

It would be good to do this for stable releases while keeping backwards compatibility. For example, Gazebo could always look for `~/.ignition/gazebo/gui.config` first, and use that if it exists.

chapulina:
Signed-off-by: Louise Poubel <louise@openrobotics.org>

# 🦟 Bug fix

Fixes #765

## Summary

From Fortress, place the config files in a versioned directory inside `~/.ignition/gazebo`. This will make sure that Fortress users don't reuse config files from older versions unless they explicitly ask to.

I targeted this at Fortress because I still haven't thought of a nice way to make this backwards compatible. The suggestion I put on the issue:

> For example, Gazebo could always look for ~/.ignition/gazebo/gui.config first, and use that if it exists.

Wouldn't be a good idea for existing users who are migrating to Fortress, for example.

## Checklist
- [x] Added tests - existing tests should pass

### DIFF
--- a/tutorials/resources.md
+++ b/tutorials/resources.md
@@ -61,7 +61,7 @@ Ignition will look for GUI plugins on the following paths, in order:
 1. All paths set on the `IGN_GUI_PLUGIN_PATH` environment variable
 2. [GUI plugins that are installed with Ignition Gazebo](https://github.com/ignitionrobotics/ign-gazebo/tree/main/src/gui/plugins)
 3. Other paths added by calling `ignition::gui::App()->AddPluginPath`
-4. `~/.ignition/gui/plugins`
+4. `~/.ignition/gui/<#>/plugins`
 5. [Plugins which are installed with Ignition GUI](https://ignitionrobotics.org/api/gui/4.2/namespaceignition_1_1gui_1_1plugins.html)
 
 ### Physics engines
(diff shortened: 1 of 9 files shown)


tutorials/resources.md
```diff
@@ -61,7 +61,7 @@ Ignition will look for GUI plugins on the following paths, in order:
 1. All paths set on the `IGN_GUI_PLUGIN_PATH` environment variable
 2. [GUI plugins that are installed with Ignition Gazebo](https://github.com/ignitionrobotics/ign-gazebo/tree/main/src/gui/plugins)
 3. Other paths added by calling `ignition::gui::App()->AddPluginPath`
-4. `~/.ignition/gui/plugins`
+4. `~/.ignition/gui/<#>/plugins`
```
jennuine, line 64: I think this was changed by accident? The PR doesn't change the gui plugins path

jennuine (approved):
Looks good, one minor comment

Issue #997, opened 2021-08-26
While using contact sensor, the simulation stuck when contact happens


0smile:
## Environment
* OS Version: Ubuntu20.04
* Source or binary build?
  Binary, Fortress

  
## Description
I'm just testing the example world contact_sensor.sdf, I found out that when the ball hit the ground, the simulation stuck. The simulation time and real time are both barely moving, but the simulation is still working, because the iterations is still increasing slowly. 
After waiting for some time, the simulation shutdown itself and the terminus output some log, which I've added below.
* Expected behavior: The simulation should continue in regular rate while the ball hit the ground
* Actual behavior: The simulation stuck when contact happens

## Steps to reproduce
I did no other action except input "ign gazebo contact_sensor.sdf", and click the start button.

## Output
ODE INTERNAL ERROR 1: assertion "aabbBound >= dMinIntExact && aabbBound < dMaxIntExact" failed in collide() [collision_space.cpp:460]
/usr/lib/ruby/ignition/cmdgazebo6.rb:432: [BUG] Segmentation fault at 0x0000000000000000
ruby 2.7.0p0 (2019-12-25 revision 647ee6f091) [x86_64-linux-gnu]

-- Control frame information -----------------------------------------------
c:0007 p:---- s:0059 e:000058 CFUNC  :call
c:0006 p:0016 s:0054 e:000053 METHOD /usr/lib/ruby/ignition/cmdgazebo6.rb:432
c:0005 p:0060 s:0048 e:000047 BLOCK  /usr/lib/ruby/ignition/cmdgazebo6.rb:468 [FINISH]
c:0004 p:---- s:0045 e:000044 CFUNC  :fork
c:0003 p:0674 s:0041 e:000040 METHOD /usr/lib/ruby/ignition/cmdgazebo6.rb:464
c:0002 p:0683 s:0020 E:000c40 EVAL   /usr/bin/ign:275 [FINISH]
c:0001 p:0000 s:0003 E:001f70 (none) [FINISH]

-- Ruby level backtrace information ----------------------------------------
/usr/bin/ign:275:in `<main>'
/usr/lib/ruby/ignition/cmdgazebo6.rb:464:in `execute'
/usr/lib/ruby/ignition/cmdgazebo6.rb:464:in `fork'
/usr/lib/ruby/ignition/cmdgazebo6.rb:468:in `block in execute'
/usr/lib/ruby/ignition/cmdgazebo6.rb:432:in `runGui'
[...]

PR #1062, opened 2021-09-28
5 ➡️ main

azeey:
# ➡️ Forward port

Port `ign-gazebo5` to `main`

Branch comparison: https://github.com/ignitionrobotics/ign-gazebo/compare/main...ign-gazebo5

### DIFF
--- a/src/EntityComponentManager.cc
+++ b/src/EntityComponentManager.cc
@@ -53,6 +53,13 @@ class ignition::gazebo::EntityComponentManagerPrivate
   public: void InsertEntityRecursive(Entity _entity,
       std::unordered_set<Entity> &_set);
 
+  /// \brief Recursively erase an entity and all its descendants from a given
+  /// set.
+  /// \param[in] _entity Entity to be erased.
+  /// \param[in, out] _set Set to erase from.
+  public: void EraseEntityRecursive(Entity _entity,
+      std::unordered_set<Entity> &_set);
+
   /// \brief Allots the work for multiple threads prior to running
   /// `AddEntityToMessage`.
   public: void CalculateStateThreadLoad();
@@ -221,10 +228,13 @@ class ignition::gazebo::EntityComponentManagerPrivate
   /// \TODO(anyone) We shouldn't be giving canonical links special treatment.
   /// This may happen to any component that holds an Entity, so we should figure
   /// out a way to generalize this for any such component.
-  std::unordered_map<Entity, Entity> oldModelCanonicalLink;
+  public: std::unordered_map<Entity, Entity> oldModelCanonicalLink;
 
   /// \brief See above
-  std::unordered_map<Entity, Entity> oldToClonedCanonicalLink;
+  public: std::unordered_map<Entity, Entity> oldToClonedCanonicalLink;
+
+  /// \brief Set of entities that are prevented from removal.
+  public: std::unordered_set<Entity> pinnedEntities;
 };
 
 //////////////////////////////////////////////////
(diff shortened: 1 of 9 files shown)


src/EntityComponentManager.cc
```diff
@@ -490,11 +527,44 @@ void EntityComponentManager::RequestRemoveEntity(Entity _entity,
...
+      std::lock_guard<std::mutex> lock(this->dataPtr->entityRemoveMutex);
+      this->dataPtr->toRemoveEntities.insert(tmpToRemoveEntities.begin(),
+          tmpToRemoveEntities.end());
+    }
+
+    for (const auto &removedEntity : tmpToRemoveEntities)
+    {
+      for (auto &view : this->dataPtr->views)
+      {
+        view.second.first->MarkEntityToRemove(removedEntity);
```
azeey, line 564: @adlarkin Can you check if I did this right? The code in Dome/Edifice called `UpdateViews` for each removed entity, but that function has been removed in Fortress, but looking at other places where `UpdateViews` was used, I saw that it was replaced with `MarkEntityToRemove`.

src/EntityComponentManager.cc
```diff
@@ -225,6 +232,8 @@ class ignition::gazebo::EntityComponentManagerPrivate
 
   /// \brief See above
   std::unordered_map<Entity, Entity> oldToClonedCanonicalLink;
+  /// \brief Set of entities that are prevented from removal.
+  public: std::unordered_set<Entity> pinnedEntities;
```
adlarkin, line 236:
```suggestion
  std::unordered_map<Entity, Entity> oldToClonedCanonicalLink;
  
  /// \brief Set of entities that are prevented from removal.
  public: std::unordered_set<Entity> pinnedEntities;
```

src/EntityComponentManager.cc
```diff
@@ -225,6 +232,8 @@ class ignition::gazebo::EntityComponentManagerPrivate
 
   /// \brief See above
   std::unordered_map<Entity, Entity> oldToClonedCanonicalLink;
```
adlarkin, line 234: Can we add `public` to this variable and the `oldModelCanonicalLink` variable above? It looks like I forgot to add those visibility flags in when working on #959.

src/EntityComponentManager.cc
adlarkin: This looks correct to me! The following tests would fail if the removal didn't work:

https://github.com/ignitionrobotics/ign-gazebo/blob/3478e80eef05d7e23cd63a8ea3eb9eebe9d3bda0/src/EntityComponentManager_TEST.cc#L776

https://github.com/ignitionrobotics/ign-gazebo/blob/3478e80eef05d7e23cd63a8ea3eb9eebe9d3bda0/src/EntityComponentManager_TEST.cc#L1252

Looking at CI, those tests are passing. So, we should be okay :slightly_smiling_face:

azeey: Thanks!

src/EntityComponentManager.cc
azeey: Done in https://github.com/ignitionrobotics/ign-gazebo/pull/1062/commits/7fa0aa2b95d00f2dff94611c75932d455a4eafa3

src/EntityComponentManager.cc
azeey: https://github.com/ignitionrobotics/ign-gazebo/pull/1062/commits/7fa0aa2b95d00f2dff94611c75932d455a4eafa3

chapulina (approved):

adlarkin:
As a side note, I just read through #997, and it looks like we can close that issue once this PR is merged.

PR #1075, opened 2021-10-01
Set a cloned joint's parent/child link names to the cloned parent/child link names

adlarkin:

# 🦟 Bug fix

## Summary
When a joint is cloned, the cloned joint's parent and child links need to be set to the cloned parent and child links (currently, the cloned joint's parent/child links are the original joint's parent/child links). Otherwise, links will not be connected to joints when cloning things like models that have links and joints.

The approach taken here is similar to the approach taken for updating a cloned canonical link in #959, where internal maps are used to keep track of a joint's original links and cloned links. As discussed in https://github.com/ignitionrobotics/ign-gazebo/pull/959#discussion_r711385128, we shouldn't give special treatment to certain components. This is just a temporary workaround until we find a better solution.

To test this, do the following:
1. Start gazebo with a double pendulum, since this model has links and joints:
```
ign gazebo -r pendulum_links.sdf
```
2. In another terminal window, clone the double pendulum:
```
ign service -s /world/double_pendulum/create --reqtype ignition.msgs.EntityFactory --reptype ignition.msgs.Boolean --timeout 300 --req 'clone_name: "double_pendulum_with_base", pose: {position: {y: 3}}, name: "pendulum_copy"'
```
3. You should now see two pendulums operating as expected:

functioning_joints

If you repeat the steps above without this PR, you'll notice that the links on the cloned pendulum fall to the ground because the cloned pendulum does not have proper joints:

bad_joints_clone

## Checklist
- [x] Added tests


src/EntityComponentManager.cc
```diff
@@ -233,6 +253,24 @@ class ignition::gazebo::EntityComponentManagerPrivate
...
 
+  /// \brief During cloning, we populate two maps:
+  ///  - map of link entities to their cloned link
+  ///  - map of cloned joint entities to the original joint entity's parent and
+  ///    child links
+  /// After cloning is done, these maps can be used to update the cloned joint
+  /// entity's parent and child links to the cloned parent and child links.
+  /// \TODO(anyone) We shouldn't be giving joints special treatment.
+  /// We should figure out a way to update a joint's parent/child links without
+  /// having to explicitly search/track for the cloned links.
```
adlarkin, line 264: There's a similar `TODO` note here: https://github.com/ignitionrobotics/ign-gazebo/blob/8f083d6d212c5867cc18cd45162ba3e4807643c4/src/EntityComponentManager.cc#L248-L250

I don't think we have an issue for tracking the idea of "generalizable component cloning" yet, do we? Should we create one?

azeey:
Detachable joints also contain information about parent and child links. Should we add those as well or state that they're not supported?

adlarkin:
> Detachable joints also contain information about parent and child links. Should we add those as well or state that they're not supported?

As we discussed offline, adding support for cloning detachable joints would require cloning the `DetachableJoint` system. So, for now, I have made a note in the documentation that cloning detachable joints is not supported: https://github.com/ignitionrobotics/ign-gazebo/pull/1075/commits/b557328353efdbdd9869e7c4e1a414245ba91590

I've also created an issue for adding detachable joint cloning support so that we can re-visit this in the future: #1080

src/EntityComponentManager.cc
```diff
@@ -97,6 +101,22 @@ class ignition::gazebo::EntityComponentManagerPrivate
   public: bool ComponentMarkedAsRemoved(const Entity _entity,
               const ComponentTypeId _typeId) const;
 
+  /// \brief Set a cloned joint's parent or child link name.
+  /// \param[in] _joint The cloned joint.
+  /// \param[in] _link The original joint's parent or child link.
```
nkoenig, line 106:
```suggestion
  /// \param[in] _originalLink The original joint's parent or child link.
```

nkoenig (approved):
Just one minor comment.

src/EntityComponentManager.cc
adlarkin: https://github.com/ignitionrobotics/ign-gazebo/pull/1075/commits/2a23906d75186f1651e4753257028280c52c7f77

PR #1076, opened 2021-10-01
Configurable joint state publisher's topic

### LINKED ISSUE #1057: Make joint state publisher's topic configurable
## Desired behavior

The [JointStatePublisher](https://ignitionrobotics.org/api/gazebo/6.0/classignition_1_1gazebo_1_1systems_1_1JointStatePublisher.html#details) plugin always publishes on the `/world/<world_name>/model/<model_name>/state` topic. It would be convenient to let users change it, especially so that models can be more easily reused across worlds of different names.

## Alternatives considered

N/A

## Implementation suggestion

The topic should be configurable from SDF.

The same way that we accept a `joint_name` element here:

https://github.com/ignitionrobotics/ign-gazebo/blob/59ff8532d0699f18cf9d0e34aa1bc0afecb73671/src/systems/joint_state_publisher/JointStatePublisher.cc#L61

We could accept a `topic` element, like how this other plugin does it for example:

https://github.com/ignitionrobotics/ign-gazebo/blob/59ff8532d0699f18cf9d0e34aa1bc0afecb73671/src/systems/detachable_joint/DetachableJoint.cc#L99-L106

jasmeet0915:
# 🎉 New feature

Closes #1057 

## Summary

Makes the topic for JointStatePublisher configurable

#### Expected behavior: 
A user should be able to provide a topic for JointStatePublisher using `<topic>` in the sdf

configurable_joint_state_pub_topic

## Test it

 * Choose an example world (I tested with [diff_drive.sdf](https://github.com/ignitionrobotics/ign-gazebo/blob/ign-gazebo5/examples/worlds/diff_drive.sdf)) and add JointStatePublisher to it while also specifying a topic with `<topic>` tags.
 * Then check if this topic gets listed with `ign topic -l` and you are able to echo it using `ign topic -t $your_topic_name -e`. 

## Checklist
- [ ] Added tests

### DIFF
--- a/src/systems/joint_state_publisher/JointStatePublisher.cc
+++ b/src/systems/joint_state_publisher/JointStatePublisher.cc
@@ -31,6 +31,7 @@
 #include "ignition/gazebo/components/JointVelocity.hh"
 #include "ignition/gazebo/components/ParentEntity.hh"
 #include "ignition/gazebo/components/Pose.hh"
+#include "ignition/gazebo/Util.hh"
 
 using namespace ignition;
 using namespace gazebo;
@@ -89,6 +90,14 @@ void JointStatePublisher::Configure(
       this->CreateComponents(_ecm, joint);
     }
   }
+
+  // Advertise the state topic
+  // Sets to provided topic if available
+  if (_sdf->HasElement("topic"))
+  {
+    this->topic = _sdf->Get<std::string>("topic");
+  }
+
 }
 
 //////////////////////////////////////////////////
@@ -142,11 +151,15 @@ void JointStatePublisher::PostUpdate(const UpdateInfo &_info,
       worldName = _ecm.Component<components::Name>(
           parentEntity->Data())->Data();
 
-      // Advertise the state topic
-      std::string topic = std::string("/world/") + worldName + "/model/"
-        + this->model.Name(_ecm) + "/joint_state";
+      // if topic not set it will be empty, hence invalid
+      std::vector<std::string> topics;
+      topics.push_back(this->topic);
+      topics.push_back(std::string("/world/") + worldName + "/model/"
+        + this->model.Name(_ecm) + "/joint_state");
+      this->topic = validTopic(topics);
+
       this->modelPub = std::make_unique<transport::Node::Publisher>(
-          this->node.Advertise<msgs::Model>(topic));
+          this->node.Advertise<msgs::Model>(this->topic));
     }
   }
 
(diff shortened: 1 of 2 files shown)


src/systems/joint_state_publisher/JointStatePublisher.cc
```diff
@@ -142,11 +151,19 @@ void JointStatePublisher::PostUpdate(const UpdateInfo &_info,
...
+      // if topic not set it will be empty
+      std::vector<std::string> topics;
+      // this helps avoid unecesarry invalid topic error
+      if (!this->topic.empty())
+      {
+        topics.push_back(this->topic);
+      }
+      topics.push_back(std::string("/world/") + worldName + "/model/"
+        + this->model.Name(_ecm) + "/joint_state");
+      this->topic = validTopic(topics);
```
ahcorde, line 163: you should check if the topic is empty

```suggestion
      this->topic = validTopic(topics);
      if (this->topic.empty())
      {
      show and error and maybe return ?
      }
```

ahcorde (changes requested):

src/systems/joint_state_publisher/JointStatePublisher.cc
jasmeet0915: Hi, correct me if I am wrong but I am not sure I understand why this check might be required. 
There are two possible conditions: 
 * The user provides a topic in the sdf
 * The user doesn't provide a topic and goes with default one

In the first one, `this->topic` will be initially set to the value entered by user and the `topics` vector will have [2 elements](https://github.com/jasmeet0915/ign-gazebo/blob/5105afb2f4275987239c2455e964db4e336b46dd/src/systems/joint_state_publisher/JointStatePublisher.cc#L159): The topic entered by the user and the default topic (`/world/worldName/model/modelName/joint_state`). If the user entered topic is invalid, `this->topic` automatically gets set to default one in `this->topic = validTopic(topics)`.

In the second case, `this->topic` will be empty initially and won't be added to the `topics` vector according to [this](https://github.com/jasmeet0915/ign-gazebo/blob/5105afb2f4275987239c2455e964db4e336b46dd/src/systems/joint_state_publisher/JointStatePublisher.cc#L157) check. Therefore, the `topics` vector will only have the default topic which will be returned as it is from `validTopic()`.

The only condition where the suggested check might be required will be if the default topic is invalid which I don't think will be possible (is it possible?).

chapulina: > The only condition where the suggested check might be required will be if the default topic is invalid which I don't think will be possible (is it possible?).

Yup, correct. I think it may be possible that the model name contains characters that can't be handled by the function that generates valid topics.

jasmeet0915: >Yup, correct. I think it may be possible that the model name contains characters that can't be handled by the function that generates valid topics.

Oh yeah that makes sense. 
I have made the changes in the [latest commit](https://github.com/ignitionrobotics/ign-gazebo/pull/1076/commits/281f1c3ecd57e54bc985acc781651fea5da98206)!

chapulina (approved):
Thanks for iterating!

osrf-triage:

PR #1094, opened 2021-10-07
Bump main to 7.0.0~pre1

chapulina:
Part of https://github.com/ignition-tooling/release-tools/issues/459


chapulina:
`INTEGRATION_examples_build` is timing out on Focal actions. I think it's not related to this PR, we probably have too many examples and have to bump the timeout. I think that can come in a separate PR, and will probably need to target earlier versions too.

PR #1097, opened 2021-10-08
Detachable joint: support for nested models of the same name

arjo129:
# 🦟 Bug fix
Fixes #<NUMBER>

## Summary

Not sure if this is a bugfix or a feature tbh. Currently, if we are loading multiple models with child models that have the  same name, the detachable joint will pick the wrong one. This is because given the current logic we only check for a child model by name. My understanding is that we wrote this plugin so we can detach sibling models. 

The proposed fix in this PR is that if there are multiple models with a similar name, the system will look for the model which is a child of the parent model. This is by no means the best solution, as I do not handle grandchildren or the case where there are multiple children of the same name but it is a solution. Would be happy to have other suggestions as to how I should handle this.

## Checklist
- [x] Added tests


src/systems/detachable_joint/DetachableJoint.cc
```diff
@@ -129,8 +129,33 @@ void DetachableJoint::PreUpdate(
...
+        // If there is one entity select that entity itself
+        modelEntity = candidateEntities[0];
+      }
+      else
+      {
+        // If there is more than one entity with the same name, look for the
+        // entity with a parent component.
+        for(auto entity : candidateEntities)
+        {
+          // TODO(arjo): do we want to support grand children?
```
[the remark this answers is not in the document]
chapulina: @arjo129 , are you still interested in following up with this PR?

azeey:
Removing `beta` label. Let me know if you're planning to get this in before Harmonic is released.

arjo129:
I'll address this after the code freeze.

azeey:
@arjo129 Do you want to keep this PR?

arjo129:
Yep. Lets try to get this in. I'll address the feedback.

src/systems/detachable_joint/DetachableJoint.cc
```diff
@@ -234,8 +234,33 @@ void DetachableJoint::PreUpdate(
...
+          auto childEntityScope = scopedName(entity, _ecm);
+          if (childEntityScope.size() < parentEntityScopedPath.size())
+          {
+            continue;
+          }
+          if (childEntityScope.find(parentEntityScopedPath) != 0)
+          {
+            continue;
+          }
+          modelEntity = entity;
```
shameekganguly, line 261: I suggest that we also check that the child link entity is present under the selected model entity here. This way, in case of multiple candidate models, we can resolve both the model as well as the child link at once.

shameekganguly, line 261: If there are multiple possible child models even after this name resolution, I'd suggest we log a warning to the user and mention the scoped path of the selected child model.

test/integration/detachable_joint.cc
```diff
@@ -214,6 +216,77 @@ TEST_F(DetachableJointTest, IGN_UTILS_TEST_DISABLED_ON_WIN32(LinksInSameModel))
...
+    childM5Poses.push_back(poseM5->Data());
+  }
+  );
+
+  this->server->AddSystem(testSystem1.systemPtr);
+
+  const std::size_t nIters{20};
+  this->server->Run(true, nIters, false);
+
+  // Both children of model1 and model4 should not move as they are held
```
shameekganguly, line 274: nit: Should be "children of model4 and model5", right?

test/integration/detachable_joint.cc
```diff
@@ -214,6 +216,77 @@ TEST_F(DetachableJointTest, IGN_UTILS_TEST_DISABLED_ON_WIN32(LinksInSameModel))
...
+  EXPECT_EQ(childM4Poses.front(), childM4Poses.back());
+  EXPECT_EQ(childM5Poses.front(), childM5Poses.back());
+
+  // Release M5's child only
+  transport::Node node;
+  auto pub = node.Advertise<msgs::Empty>("/model/M5/detachable_joint/detach");
+  pub.Publish(msgs::Empty());
+  std::this_thread::sleep_for(250ms);
+
+  // Release M5's
```
shameekganguly, line 285: nit: remove this comment

test/integration/detachable_joint.cc
```diff
@@ -214,6 +216,77 @@ TEST_F(DetachableJointTest, IGN_UTILS_TEST_DISABLED_ON_WIN32(LinksInSameModel))
...
+  this->server->Run(true, nIters, false);
+  EXPECT_LT(childM5Poses.back().Z(), childM4Poses.front().Z());
```
shameekganguly, line 287: nit: Would it make more sense to check
```
EXPECT_LT(childM5Poses.back().Z(), childM5Poses[childM5Poses.size()-2].Z())
```
to establish that the model is falling?

src/systems/detachable_joint/DetachableJoint.cc
```diff
@@ -234,8 +234,33 @@ void DetachableJoint::PreUpdate(
     }
     else
     {
-      modelEntity = _ecm.EntityByComponents(
-          components::Model(), components::Name(this->childModelName));
+      auto candidateEntities = entitiesFromScopedName(
+        this->childModelName, _ecm);
+
```
shameekganguly, line 239: nit: Please consider filtering this list down by entities which have the Model component. This way, if the entity is found but is not a model, we can indicate that to the user. Otherwise they will end up with a slightly more confusing "Child link ... could not be found" message on line 289.

arjo129: [e8e1b16](https://github.com/gazebosim/gz-sim/pull/1097/commits/e8e1b16976b2774f3f9cbade1eb8c0978aa3279a)

src/systems/detachable_joint/DetachableJoint.cc
arjo129: [e8e1b16](https://github.com/gazebosim/gz-sim/pull/1097/commits/e8e1b16976b2774f3f9cbade1eb8c0978aa3279a)

arjo129: [e8e1b16](https://github.com/gazebosim/gz-sim/pull/1097/commits/e8e1b16976b2774f3f9cbade1eb8c0978aa3279a)

test/integration/detachable_joint.cc
arjo129: Thanks for spotting this. [19bdc85](https://github.com/gazebosim/gz-sim/pull/1097/commits/19bdc858b0f28bd1403ece5e1bb95d763a8ac65e)

test/integration/detachable_joint.cc
arjo129: [19bdc85](https://github.com/gazebosim/gz-sim/pull/1097/commits/19bdc858b0f28bd1403ece5e1bb95d763a8ac65e)

test/integration/detachable_joint.cc
arjo129: Added in [19bdc85](https://github.com/gazebosim/gz-sim/pull/1097/commits/19bdc858b0f28bd1403ece5e1bb95d763a8ac65e)

src/systems/detachable_joint/DetachableJoint.cc
```diff
@@ -234,16 +234,29 @@ void DetachableJoint::PreUpdate(
...
     {
-      auto candidateEntities = entitiesFromScopedName(
+      auto entitiesMatchingName = entitiesFromScopedName(
         this->childModelName, _ecm);
 
+      // TODO(arjoc): There is probably a more efficient way of combining entitiesFromScopedName
+      // With filtering.
+      // Filter for entities with only models
+      std::vector<Entity> candidateEntities;
+      for (const auto entity: entitiesMatchingName)
```
shameekganguly, line 244:
```suggestion
      std::copy_if(entitiesMatchingName.begin(), entitiesMatchingName.end(),
                 std::back_inserter(candidateEntities),
                 [](Entity e) { return _ecm.EntityHasComponentType(e, components::Model::typeId); });
```

src/systems/detachable_joint/DetachableJoint.cc
```diff
@@ -234,16 +234,29 @@ void DetachableJoint::PreUpdate(
...
+      {
+        if (_ecm.EntityHasComponentType(entity, components::Model::typeId))
+        {
+          candidateEntities.push_back(entity);
+        }
+      }
```
shameekganguly:
```suggestion
      std::copy_if(entitiesMatchingName.begin(), entitiesMatchingName.end(),
                 std::back_inserter(candidateEntities),
                 [](Entity e) { return _ecm.EntityHasComponentType(e, components::Model::typeId); });
```

shameekganguly: Sorry, I made this suggestion against an older revision. Will make a fresh suggestion on latest revision.

src/systems/detachable_joint/DetachableJoint.cc
```diff
@@ -234,8 +234,75 @@ void DetachableJoint::PreUpdate(
...
+        if (_ecm.EntityHasComponentType(entity, components::Model::typeId))
+        {
+          candidateEntities.push_back(entity);
+        }
+        else
+        {
+          ignwarn << "Found entity " << scopedName(entity, _ecm)
+            << " but it is not a model." << std::endl;
+        }
+      }
```
shameekganguly, line 255:
```suggestion
      std::copy_if(entitiesMatchingName.begin(), entitiesMatchingName.end(),
                 std::back_inserter(candidateEntities),
                 [&_ecm](Entity e) { return _ecm.EntityHasComponentType(e, components::Model::typeId); });
```

I'd suggest to remove the log message as it can be spammy if there is another entity with the same name. We'll log in any case that the child model could not be found on line 335.

src/systems/detachable_joint/DetachableJoint.cc
```diff
@@ -234,8 +234,75 @@ void DetachableJoint::PreUpdate(
...
+            {
+                  // Only select child entity if tbe entity has a link
+                  modelEntity = entity;
+                  selectedModelName = childEntityScope;
+                  igndbg << "Selecting " << childEntityScope << " as model to be detached" << std::endl;
+            }
+            else
+            {
+              ignwarn << "Found " << childEntityScope << " with no link named " << this->childLinkName <<std::endl;
+              this->suppressChildWarning = true;
```
shameekganguly, line 296: I don't think we should set `this->suppressChildWarning` as it seems like an input parameter for the plugin? See line 208

src/systems/detachable_joint/DetachableJoint.cc
```diff
@@ -234,8 +234,75 @@ void DetachableJoint::PreUpdate(
...
+      {
+        // If there is one entity select that entity itself
+        modelEntity = *candidateEntities.begin();
+      }
+      else
+      {
+        std::string selectedModelName;
+        auto parentEntityScopedPath = scopedName(this->model.Entity(), _ecm);
+        // If there is more than one entity with the same name, look for the
+        // entity with a parent component.
```
shameekganguly, line 267:
```suggestion
        // If there is more than one model with the given child model name, the plugin looks for a model which is 
        // - a descendant of the plugin's parent model with that name, and
        // - has a child link with the given child link name
```

src/systems/detachable_joint/DetachableJoint.cc
```diff
@@ -234,8 +234,75 @@ void DetachableJoint::PreUpdate(
...
+        for (auto entity : candidateEntities)
+        {
+          auto childEntityScope = scopedName(entity, _ecm);
+          if (childEntityScope.size() < parentEntityScopedPath.size())
+          {
+            continue;
+          }
+          if (childEntityScope.find(parentEntityScopedPath) != 0)
```
shameekganguly, line 275:
```suggestion
          if (childEntityScope.rfind(parentEntityScopedPath, 0) != 0)
```
This version is more efficient as it restricts the search to a prefix in childEntityScope.

src/systems/detachable_joint/DetachableJoint.cc
```diff
@@ -234,8 +234,75 @@ void DetachableJoint::PreUpdate(
...
+            if (kNullEntity != this->childLinkEntity)
+            {
+                  // Only select child entity if tbe entity has a link
+                  modelEntity = entity;
+                  selectedModelName = childEntityScope;
+                  igndbg << "Selecting " << childEntityScope << " as model to be detached" << std::endl;
+            }
+            else
+            {
+              ignwarn << "Found " << childEntityScope << " with no link named " << this->childLinkName <<std::endl;
```
shameekganguly, line 295:
```suggestion
              ignwarn << "Found " << childEntityScope << " with no link named " << this->childLinkName << std::endl;
```

src/systems/detachable_joint/DetachableJoint.cc
```diff
@@ -234,8 +234,75 @@ void DetachableJoint::PreUpdate(
     }
     else
     {
-      modelEntity = _ecm.EntityByComponents(
-          components::Model(), components::Name(this->childModelName));
+      auto entitiesMatchingName = entitiesFromScopedName(
```
shameekganguly, line 237: EntitiesFromScopedName seems like a very expensive function call, with complexity O(P*N*R) where P is the length of the scoped path, N is the number of named entities in the ECM and R is the size of the function output. I'd recommend that we do not use it in PreUpdate/PostUpdate as it could affect performance for large scenes.

I'd prefer if we moved the entire logic for selecting the child model and child model link to the Configure call since if the plugin sdf contains a reference to a model, presumably it should already be present in the world at the time when Configure is called. 

However, that may break existing apps which depend on run-time addition and discovery of the child model, which the existing implementation allows, if only in a restrictive way.

One alternative is to move this logic to a helper function `GetChildModelAndLinkEntities(...)` and call it in Configure first. Then in PreUpdate, we can call `GetChildModelAndLinkEntities(...)` conditionally if `childLinkEntity == kNullEntity || !_ecm.HasEntity(childLinkEntity)` to update the cached entities.

shameekganguly (changes requested):
Thanks for addressing the previous comments @arjo129! I have a few more suggestions with which we can merge these changes in.

src/systems/detachable_joint/DetachableJoint.cc
arjo129: Thats a good point. We only run this when we call attach, but reading the implementation of `entitiesFromScopedName` I can see how this can be a problem. Might be worth exploring ways to optimize this in future releases, 

Fixed in [1a7083d](https://github.com/gazebosim/gz-sim/pull/1097/commits/1a7083dfeb05bd50e6358bbbd3377f6a13e0bb4e)

src/systems/detachable_joint/DetachableJoint.cc
arjo129: [1a7083d](https://github.com/gazebosim/gz-sim/pull/1097/commits/1a7083dfeb05bd50e6358bbbd3377f6a13e0bb4e)

src/systems/detachable_joint/DetachableJoint.cc
arjo129: [1a7083d](https://github.com/gazebosim/gz-sim/pull/1097/commits/1a7083dfeb05bd50e6358bbbd3377f6a13e0bb4e)

src/systems/detachable_joint/DetachableJoint.cc
arjo129: [1a7083d](https://github.com/gazebosim/gz-sim/pull/1097/commits/1a7083dfeb05bd50e6358bbbd3377f6a13e0bb4e)

src/systems/detachable_joint/DetachableJoint.cc
arjo129: [1a7083d](https://github.com/gazebosim/gz-sim/pull/1097/commits/1a7083dfeb05bd50e6358bbbd3377f6a13e0bb4e)

src/systems/detachable_joint/DetachableJoint.cc
```diff
@@ -225,54 +314,35 @@ void DetachableJoint::PreUpdate(
     if (!this->attachRequested){
       return;
     }
-    // Look for the child model and link
-    Entity modelEntity{kNullEntity};
 
-    if ("__model__" == this->childModelName)
+    if (childLinkEntity == kNullEntity || !_ecm.HasEntity(childLinkEntity))
```
shameekganguly, line 318:
```suggestion
    if (this->childLinkEntity == kNullEntity || !_ecm.HasEntity(this->childLinkEntity))
```
to be consistent with style (sorry should have suggested this in the first place)

arjo129: Nah it's fine. I should be more careful also :stuck_out_tongue: [6fa7157](https://github.com/gazebosim/gz-sim/pull/1097/commits/6fa7157a7beaa8735ecbde4dcc037934e3ca389c)

shameekganguly (approved):

azeey:
Looks like there's a windows test failure (https://build.osrfoundation.org/job/gz_sim-pr-win/741/testReport/(root)/DetachableJointTest/NestedModelsWithSameName/). @arjo129 would you be able to take a look?

arjo129:
Will do in the morning.

shameekganguly:
Looks like the other test cases in the detachable_joint integration test are actually disabled on Windows:
`TEST_F(DetachableJointTest, IGN_UTILS_TEST_DISABLED_ON_WIN32(LinksInSameModel))`

Looking at the test log, the physics engine is not correctly loading in windows, which could cause the test failure:
[Err] [C:\J\workspace\gz_sim-pr-win\ws\gz-sim\src\systems\physics\Physics.cc:761] Failed to find plugin [libignition-physics-dartsim-plugin.so]. Have you checked the IGN_GAZEBO_PHYSICS_ENGINE_PATHenvironment variable?

@azeey I think we can disable the new test case on Windows for now and figure out the underlying issue with the physics engine not loading post-release?

shameekganguly:
https://github.com/gazebosim/gz-sim/pull/2523 disables the added test case on Windows

PR #1109, opened 2021-10-14
Get updated GUI ECM info when a user presses 'play'

adlarkin:

# 🎉 New feature

Part of https://github.com/ignitionrobotics/ign-gazebo/issues/1106

Requires:
* https://github.com/ignitionrobotics/ign-msgs/pull/190
* https://github.com/ignitionrobotics/ign-gui/pull/306
* https://github.com/ignitionrobotics/ign-gazebo/pull/1122
* https://github.com/ignitionrobotics/ign-gazebo/pull/1131

## Summary
When simulation is paused, users might want to update the GUI's ECM directly. When simulation is resumed, the server's ECM needs to be updated to reflect any changes made in the GUI's ECM. This PR implements the server's ECM updating based on changes made in the GUI.

For a higher-level design discussion of how this functionality is implemented, see https://github.com/ignitionrobotics/ign-gazebo/pull/1109#issuecomment-953472995.

## Test it
Manual testing may need to be done in order to run a gui plugin that updates the GUI's ECM. Does anyone know how automated testing can be done for this, since it relies on GUI/Server interaction?

There are also some notes about testing the PRs new service from the command line in the PR description of https://github.com/ignitionrobotics/ign-gui/pull/306.

## Checklist
- [x] Added tests


src/gui/GuiRunner.cc
```diff
@@ -110,6 +111,26 @@ GuiRunner::GuiRunner(const std::string &_worldName)
   QPointer<QTimer> timer = new QTimer(this);
   connect(timer, &QTimer::timeout, this, &GuiRunner::UpdatePlugins);
   timer->start(33);
+
+  // Advertise a service that shares the changed state of the GUI's ECM.
+  // "Changed" in this context means the changes that took place since the last
+  // simulation step
+  const std::string changedGuiEcmService = "/changedGuiEcm";
```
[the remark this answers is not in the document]
adlarkin: no longer needed as of https://github.com/ignitionrobotics/ign-gazebo/pull/1109/commits/d726c70ed82bf27aa32ddf318b364996326a7817

adlarkin:
@nkoenig I have updated this PR to reflect the design we discussed. I like the idea of keeping the ECM out of `ign-gui`, and I think that the best way to do that is have events emitted from `ign-gui` that are processed by `ign-gazebo`'s `GuiRunner`.

Regarding how this can be tested, we will probably need to incorporate @iche033's GUI model editor work and then run all of this together to see if the server ECM is updated correctly. I did some local/manual testing to confirm that the server receives/processes the GUI's ECM on play, so I think that this should all "just work" since updating the server's ECM state is done via [`ECM::SetState`](https://github.com/ignitionrobotics/ign-gazebo/blob/ignition-gazebo6_6.1.0/include/ignition/gazebo/EntityComponentManager.hh#L599-L607), which is a method that has already been tested in the ECM unit tests.

src/SimulationRunner.cc
```diff
@@ -194,7 +194,8 @@ SimulationRunner::SimulationRunner(const sdf::World *_world,
   this->node = std::make_unique<transport::Node>(opts);
 
   // TODO(louise) Combine both messages into one.
-  this->node->Advertise("control", &SimulationRunner::OnWorldControl, this);
+  this->node->Advertise("control", &SimulationRunner::OnWorldControlState,
```
nkoenig, line 197: How about keeping the old `OnWorldControl` for one development cycle? You  can add a warning message to the `OnWorldControl` function indicating that it will be replaced by `OnWorldControlState`.

src/gui/GuiRunner.cc
```diff
@@ -115,11 +120,48 @@ GuiRunner::GuiRunner(const std::string &_worldName)
...
+      reinterpret_cast<gui::events::WorldControl *>(_event);
+    if (worldControlEvent)
+    {
+      msgs::WorldControlState req;
+      req.mutable_world_control()->CopyFrom(
+          worldControlEvent->WorldControlInfo());
+
+      // If a user presses play, send the GUI's ECM state to the server
+      // in case the GUI's ECM was modified while paused
+      if (worldControlEvent->Play())
```
nkoenig, line 148: Can you check that the GUI `step` button also triggers an ECM update? I think that is missing.

src/SimulationRunner.cc
adlarkin: I assume this means we would need two different services then, right? The current service (`control`) that takes a `msgs::WorldControl`, and a new service (should it be called `control/state`?) that takes a `msgs::WorldControlState`.

src/gui/GuiRunner.cc
adlarkin: Yeah, I missed that case - thanks for pointing it out. I think the easiest way to check for play/step is to update the GUI event in https://github.com/ignitionrobotics/ign-gui/pull/306 to have a `Pause` method instead of `Play`. Then, if calling `Pause` returns false, we know that the ECM needs to be updated. Giving the new `WorldControl` GUI event a `Pause` method instead of `Play` also makes it align better with the `WorldControl` message type (`msgs::WorldControl` has a `paused` field). I'll go ahead and make this update, unless you're opposed to this change in any way.

src/SimulationRunner.cc
nkoenig: Yeah, that's correct.

adlarkin: Both services are supported as of https://github.com/ignitionrobotics/ign-gazebo/pull/1109/commits/3ca5d9dd72626b72209c422464cfb65e92d1eafc and https://github.com/ignitionrobotics/ign-gui/pull/306/commits/7ff16ce47ffbb1e0dc37d005333a4111a13e505d (using the new `control/state` service is the default), so I'll mark this as resolved. If you want to test using the old service (which doesn't take ECM state information), see the discussion in https://github.com/ignitionrobotics/ign-gui/pull/306#discussion_r739588172

src/gui/GuiRunner.cc
adlarkin: The `step` case has been handled in https://github.com/ignitionrobotics/ign-gazebo/pull/1109/commits/3ca5d9dd72626b72209c422464cfb65e92d1eafc. It looks like there's some "interesting" behavior when pressing step now. Whenever I press step while paused, the GUI usually shows that it's playing for a moment, and then goes back to paused (see below). I believe this is due to the extra time it takes for the event approach to be processed, compared to the original service approach (I don't see this GUI behavior when using the service). I'm not sure if this is necessarily a "problem", but I wanted to point it out.

step

nkoenig: Yes, I'm seeing this as well. I was able to get the play button into an odd state, where simulation was running but the play button showed the play symbol (>).

nkoenig: Clicking the step button rapidly can also cause simulation to start running.

adlarkin: Okay, so after debugging, I found why the behaviors above were occurring. Here's the program flow when you press step while paused:
1. When step is pressed while simulation is paused, the server has to set its paused state to `false` so that the step can be processed: https://github.com/ignitionrobotics/ign-gazebo/blob/35a5772ece24cf394cf4d03d95bf0502af31f90e/src/SimulationRunner.cc#L1218
2. Once the paused state is set to false, the step occurs, and world stats are published. So, in the world stats message that's published, paused is false: https://github.com/ignitionrobotics/ign-gazebo/blob/35a5772ece24cf394cf4d03d95bf0502af31f90e/src/SimulationRunner.cc#L415-L418
3. On the `ign-gui` side, the `WorldControl` plugin receives this message and sees that `paused` is false. But, the `WorldControl` plugin is currently paused since step occurred while paused, so the plugin thinks that it needs to be updated to start playing: https://github.com/ignitionrobotics/ign-gui/blob/054d02c87b72db9a80998899d45b372a0cfdac9c/src/plugins/world_control/WorldControl.cc#L245-L264
4. The server goes through its next update: this time, [it resets its paused state to `true`](https://github.com/ignitionrobotics/ign-gazebo/blob/35a5772ece24cf394cf4d03d95bf0502af31f90e/src/SimulationRunner.cc#L844), and has no `WorldControl` msgs to process. So, when the server publishes its next stats message, `paused` is now true.
5. `ign-gui`'s `WorldControl` (which is now in a playing state because of the previous stats message) processes this message, and sees that paused is true. So, now the opposite of step 3 happens, and the `WorldControl` GUI plugin goes back to a paused state.

To work around the issue outlined above, I have added a `step` key to the stats message whenever a message is published that corresponds to a step (see https://github.com/ignitionrobotics/ign-gazebo/pull/1109/commits/cabdc966701dc9fbf72444115d118f52c0d8f6e2). The `WorldControl` GUI plugin has also been updated to ignore any incoming stats messages that have this header/key combination (see https://github.com/ignitionrobotics/ign-gui/pull/306/commits/d7e2c4f31360608f14fe1f1b40fdcec16f7f90fd). This now resolves the issues related to pressing step while paused.

Once strange thing that I noticed is that when using the service approach instead of the new event approach (so, using `<use_event>false</use_event` in the gui config), if step is pressed while paused, the server will still publish a stats message with pause set to false, but when `ign-gui` processes this message, it seems to think that pause in the message is true. When using the event approach, `ign-gui` thinks pause is false when processing the message. So, I am not sure why there's a difference between the two approaches, but the fix I mentioned above ensures proper behavior for both the event and service approach.

adlarkin:
Following the conversation in https://github.com/ignitionrobotics/ign-gui/pull/306#discussion_r741531329, I have removed deprecation warnings/notes since we will be supporting both the event and service options moving forward.

adlarkin:
@nkoenig since the changes made in https://github.com/ignitionrobotics/ign-gui/pull/306 make the GUI listen to the world stats topic to get pause/play updates from the server, we might want to increase how frequently world stats messages are published to prevent slow response times when users press the play/pause GUI button. We could change the messages published per second to something like 10 or 15 (it's currently only 5): https://github.com/ignitionrobotics/ign-gazebo/blob/e61eef90a392896b85253ee971098f1eafe56162/src/SimulationRunner.cc#L672

Let me know what you think about this.

nkoenig:
> @nkoenig since the changes made in [ignitionrobotics/ign-gui#306](https://github.com/ignitionrobotics/ign-gui/pull/306) make the GUI listen to the world stats topic to get pause/play updates from the server, we might want to increase how frequently world stats messages are published to prevent slow response times when users press the play/pause GUI button. We could change the messages published per second to something like 10 or 15 (it's currently only 5):
> 
> https://github.com/ignitionrobotics/ign-gazebo/blob/e61eef90a392896b85253ee971098f1eafe56162/src/SimulationRunner.cc#L672
> 
> Let me know what you think about this.

Can you open a separate PR with this change?

adlarkin:
> Can you open a separate PR with this change?

:+1: see #1163

nkoenig (approved):

nkoenig:
@scpeters , it looks like there are problems with homebrew that you're aware of. Is it okay to merge this, or should I wait for a homebrew fix?

scpeters:
> @scpeters , it looks like there are problems with homebrew that you're aware of. Is it okay to merge this, or should I wait for a homebrew fix?

I just merged the bottle for now; we can fix it later

I'll restart the homebrew job, which should have a better chance now

nkoenig:
> I just merged the bottle for now; we can fix it later
> 
> I'll restart the homebrew job, which should have a better chance now

Thanks.

src/SimulationRunner.cc
```diff
@@ -1097,36 +1117,54 @@ void SimulationRunner::SetRunToSimTime(
...
+
+/////////////////////////////////////////////////
+bool SimulationRunner::OnWorldControlState(const msgs::WorldControlState &_req,
+    msgs::Boolean &_res)
 {
   std::lock_guard<std::mutex> lock(this->msgBufferMutex);
 
+  // update the server ECM if the request contains SerializedState information
+  if (_req.has_state())
+    this->entityCompMgr.SetState(_req.state());
```
chapulina, line 1135: Sorry to join the party late, but is this thread-safe? It's dangerous to update the ECM in a transport callback. I think we should define a clear moment in the update loop when the ECM is updated with the new state; maybe before `PreUpdate`?

adlarkin: Thanks for pointing thread safety out - I totally missed this the first time around :facepalm: @nkoenig is going to open another PR with a fix.

PR #1119, opened 2021-10-18
Better protect this->dataPtr->initialized with renderMutex.

mahiuchun:
Tested with running ThreadSanitizer on some code involves camera sensor rendering.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -38,6 +38,9 @@ endif()
 # Search for project-specific dependencies
 #============================================================================
 
+# Setting this policy enables using the protobuf_MODULE_COMPATIBLE
+# set command in CMake versions older than 13.13
+set(CMAKE_POLICY_DEFAULT_CMP0077 NEW)
 # This option is needed to use the PROTOBUF_GENERATE_CPP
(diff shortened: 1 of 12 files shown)


src/systems/sensors/Sensors.cc
```diff
@@ -417,6 +417,7 @@ void Sensors::Update(const UpdateInfo &_info,
                      EntityComponentManager &_ecm)
 {
   IGN_PROFILE("Sensors::Update");
+  std::unique_lock<std::mutex> lock(this->dataPtr->renderMutex);
```
azeey, line 420: I know this is merged, but I just noticed this. If I understand correctly, having this mutex lock here could prevent sensors from updating in parallel with physics. This would happen if the `Sensors` system comes before the `Physics` system in the SDFormat file. Here's the sequence of events that could happen:
1. In the `Sensors::PostUpdate`, a new render is triggered that takes a long time to complete. The render thread locks `renderMutex` and starts the render.
2. In the next phase, `Sensors::Update` tries to acquire `renderMutex` and blocks because the render is not complete. This also blocks `Physics::Update` from being called because it comes after `Sensors::Update`.

Can some one verify this?

Issue #1106, opened 2021-10-14
Synchronize ECM between GUI/Server


adlarkin:
## Desired behavior
Add functionality so that when simulation is paused, users can perform GUI actions that update the GUI's ECM directly. Then, when simulation is unpaused, the GUI shares its updated ECM state with the server, and the server does whatever is needed to reflect the changes made in the GUI.

## Alternatives considered
N/A (new feature). For more context, we currently update the server from the GUI via gui events.

## Implementation suggestion
See the design proposed in https://github.com/ignitionrobotics/design/pull/8

## Additional context
Here is a checklist of things that need to be done to add this functionality. The design document linked above can be referenced for more context about each checklist item. I will add relevant PRs by each item as they're created so that we can track the progress of this new feature:
- [x] Create a new message type that contains updated ECM information for when users press "play" on the GUI (see the conversation in https://github.com/ignitionrobotics/design/pull/8#discussion_r728391135 for more context)
    - https://github.com/ignitionrobotics/ign-msgs/pull/190
- [x] Update the interaction between GUI/server to make use of the new message type above when "play" is clicked in the GUI. Consider the case of multiple clients attached to the same server (see https://github.com/ignitionrobotics/design/pull/8#discussion_r729123976 for a solution)
    - https://github.com/ignitionrobotics/ign-gui/pull/306
    - https://github.com/ignitionrobotics/ign-gazebo/pull/1109
- [x] Create a GUI system plugin that tracks ECM changes from GUI user actions when simulation is paused (this is the `GuiUserCommands` plugin in the design document) - I may also be able to add this functionality to the existing `ign-gui` WorldControl plugin, since this is the plugin that interacts with the server when "play" is pressed
    - https://github.com/ignitionrobotics/ign-gui/pull/306 (I took the approach of adding to the existing `ign-gui`'s `WorldControl` plugin instead of creating a new one)
- [ ] Update the server to not only update its ECM based on the ECM-changing GUI actions, but to notify things like server systems of changes as well (notifying server systems of updates may require a lot of work). For the case of server/client running in the same process, the server's ECM wouldn't need to be updated since the server/client share the same ECM, but server systems would still need to be notified of changes
    - `TODO` note left in https://github.com/ignitionrobotics/ign-gazebo/pull/1109

adlarkin:
I'm closing this, since #1231 has been merged.

PR #1122, opened 2021-10-19
Add support for GUI entity creation

adlarkin:

# 🎉 New feature

Related to #1106

## Summary
In order to create entities in the GUI (while paused) and then share these new entities with the server, we need to:
1. Make sure that entity IDs in the GUI are unique (i.e., they don't collide with server Entity IDs)
2. Update the "new entity state" in the GUI's ECM after new GUI entity additions are processed. Otherwise, on consecutive iterations while paused, the `RenderUtil` will try to create entities that were already created.

Item 1 above was resolved through a hack that sets the GUI ECM's entity creation ID offset to one that is (probably) larger than the server ECM's entity creation ID offset (I left a `TODO` note in the code about improving this). Item 2 above actually appears to be a bug that was never fixed. The fix for this is to move the GUI ECM's `ClearNewlyCreatedEntities` and `ProcessRemoveEntityRequests` calls to the `GuiRunner::Implementation::UpdatePlugins` method. This allows the GUI ECM's state to be properly updated when simulation is both running and paused (the previous approach of having these ECM method calls in `GuiRunner::Implementation::ProcessState` meant that these ECM method calls were only made when simulation is running).

## Test it
@iche033 can you try to test this with your GUI model editor work in https://github.com/ignitionrobotics/ign-gazebo/compare/ign-gazebo6...model_editor_add_link? You should be able to remove [these lines](https://github.com/ignitionrobotics/ign-gazebo/blob/cdd9e5096ebaf57cd603a877d926a149e58dbc1c/src/gui/plugins/model_editor/ModelEditor.cc#L129-L134) from your branch because this PR sets the entity creation offset for the ECM that's used by all GUI plugins.

## Checklist
- [ ] Added tests

### DIFF
--- a/src/gui/GuiRunner.cc
+++ b/src/gui/GuiRunner.cc
@@ -217,7 +217,7 @@ void GuiRunner::UpdatePlugins()
   {
     plugin->Update(this->dataPtr->updateInfo, this->dataPtr->ecm);
   }
-  this->ecm.ClearRemovedComponents();
-  this->ecm.ClearNewlyCreatedEntities();
-  this->ecm.ProcessRemoveEntityRequests();
+  this->dataPtr->ecm.ClearRemovedComponents();
+  this->dataPtr->ecm.ClearNewlyCreatedEntities();
+  this->dataPtr->ecm.ProcessRemoveEntityRequests();
 }


adlarkin:
> we still need a way to propagate new ECM changes to other GUI plugins but that's outside the scope of this PR

I'm concerned that the changes I proposed in this PR aren't actually a proper "fix". As we discussed offline, we run the chance of other GUI plugins not being aware of new entities that were added in a GUI plugin's `Update` if we clear the GUI ECM's new entity state at the end of a GUI `Update`. For example, consider the following example:

* GUI's `Update` begins. Plugin updates occur in the following order:
    - The `EntityTree` plugin processes its update, looking for new/removed entities to update the tree with
    - Another GUI plugin processes its update, which involves adding a new entity
* GUI's `Update` ends, and the ECM clears its new entity state

The problem with the scenario above is that `EntityTree` will never be aware of the new entity that was created by the other GUI plugin, since the new entities are cleared by the time `EntityTree` processes its next update. So, although the new entity exists, we will never see it in the entity tree.

I think the proper solution is to follow the pattern of server systems. We should give GUI plugins a `PostUpdate` so that all GUI plugins have the chance to see ECM changes that took place during `Update`. Then, at the end of `PostUpdate`, we can do things like clear the ECM's new entity state.

The solution I've proposed would involve updating other GUI plugin logic as needed (for example, having the `EntityTree` check for new entities in `PostUpdate`), and would also break ABI (so, we'd need to target Garden).

Another solution proposed by @iche033 is to send any changes made on the GUI back to the server, and then have the server broadcast the changes out to all GUI plugins. While this solution would not break ABI, I worry that this would be difficult code to maintain, and also wonder how much performance overhead would be introduced by going from GUI -> Server -> GUI for every GUI change. I think that implementing this approach would be just as much (or possibly even more) work than giving GUI systems a `PostUpdate`. I also prefer the approach of adding a `PostUpdate` because we can ensure thread safety by making the ECM read-only in `PostUpdate`.

I'm open to other suggestions/thoughts if anyone has them!

PR #1148, opened 2021-10-27
Use deserialize to remove CreateComponentImplementation's template dependency

adlarkin:

# 🦟 Bug fix

## Summary
When I worked on #856, I came across an issue where I could not update derived component data when dealing with `BaseComponent` pointers in `EntityComponentManager::CreateComponentImplementation`: https://github.com/ignitionrobotics/ign-gazebo/pull/856#discussion_r684511937

The workaround at the time was to have `CreateComponentImplementation` return a flag that indicated whether the derived component data needed to be updated or not. If the data needed to be updated, it was taken care of in the template method that called `CreateComponentImplementation`, since the template method knew what the derived component type is.

However, after working on #1131, I realized that we can use component's `Serialize`/`Deserialize` methods in `EntityComponentManager::CreateComponentImplementation` to properly update the derived component data when working with `BaseComponent` pointers. The reason why this works is because these serialization methods are virtual methods that are implemented by both the `BaseComponent` and derived component classes. A regular assignment operator wouldn't work because the component classes don't have this operator implemented as a virtual override (see https://godbolt.org/z/hezKTbeEj for an example of what happens when a default assignment operator is used with polymorphism - spoiler alert: the derived class data is lost in the assignment).

As a result of this change, `CreateComponentImplementation` can return a pointer to the component instead of a update flag, and the template method that calls `CreateComponentImplementation` can then use this component directly. I can't change the signature of this method in `ign-gazebo6` since it will break ABI, but I left a `TODO` note about fixing it in `ign-gazebo7`. Since `CreateComponentImplementation` is a private method, we don't need to tick-tock it, right? We can change a private method without having to deprecate it, if I understand correctly.

## Checklist
- [ ] Added tests

### DIFF
--- a/include/ignition/gazebo/EntityComponentManager.hh
+++ b/include/ignition/gazebo/EntityComponentManager.hh
@@ -705,7 +705,17 @@ namespace ignition
       /// \param[in] _componentTypeId Id of the component type.
       /// \param[in] _data Data used to construct the component.
       /// \return True if the component's data needs to be set externally; false
-      /// otherwise.
+      /// otherwise (this return type should be ignored - see todo note below)
+      /// TODO(adlarkin) update this method to return the actual component
+      /// (components::BaseComponent *) in ign-gazebo7. This method has
+      /// been updated in ign-gazebo6 to call BaseComponent::Deserialize in the
+      /// implementation, so the component data is now set correctly and does
+      /// not need to be set in the calling template method. Once this method
+      /// returns the actual component, the template method that calls this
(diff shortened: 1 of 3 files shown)


adlarkin:
After some discussion that was had offline, a point was brought up that not all components have `Deserialize` implemented. So, we may need to close this PR and stick with the current approach (cc @chapulina).

chapulina:
It was a good idea though. Maybe in the future we can make sure all components are serializable.

adlarkin:
Closing this for now since it's not guaranteed that all components are deserializable.

PR #1159, opened 2021-11-02
Fix the force-torque sensor update rate

mjcarroll:
# 🦟 Bug fix

Makes the force-torque sensor system properly respect the `update_time` parameter from the sensor's SDF.

## Summary

This bug is introduced by calling `ForceTorqueSensor::Update` rather that `Sensors::Update` from the base class.  This was causing all of the base class update logic to be skiped and the FT sensor would update every time.

To test, runt he example sensors world
```
ign gazebo ./src/ign-gazebo/examples/worlds/sensors.sdf    
```

And verify that the output rate of the f-t sensor is reasonable (10 hz, in this case):

```
ign topic -e -t /world/sensors/model/force_torque_demo/joint/joint_01/sensor/force_torque/forcetorque 
```

### DIFF
--- a/src/systems/force_torque/ForceTorque.cc
+++ b/src/systems/force_torque/ForceTorque.cc
@@ -136,10 +136,7 @@ void ForceTorque::PostUpdate(const UpdateInfo &_info,
 
     for (auto &it : this->dataPtr->entitySensorMap)
     {
-      // Update measurement time
-      auto time = math::durationToSecNsec(_info.simTime);
-      static_cast<sensors::Sensor *>(it.second.get())->Update(
-          math::secNsecToDuration(time.first, time.second), false);
+      it.second.get()->sensors::Sensor::Update(_info.simTime, false);
     }
   }
 
--- a/src/systems/force_torque/ForceTorque.cc
+++ b/src/systems/force_torque/ForceTorque.cc
@@ -137,7 +137,9 @@ void ForceTorque::PostUpdate(const UpdateInfo &_info,
     for (auto &it : this->dataPtr->entitySensorMap)
     {
       // Update measurement time
-      it.second->Update(_info.simTime);
+      auto time = math::durationToSecNsec(_info.simTime);
+      static_cast<sensors::Sensor *>(it.second.get())->Update(
+          math::secNsecToDuration(time.first, time.second), false);
     }
   }
 
--- a/examples/worlds/sensors.sdf
+++ b/examples/worlds/sensors.sdf
@@ -36,6 +36,10 @@
       name="ignition::gazebo::systems::Magnetometer">
     </plugin>
     <plugin
+      filename="ignition-gazebo-forcetorque-system"
+      name="ignition::gazebo::systems::ForceTorque">
+    </plugin>
+    <plugin
       filename="ignition-gazebo-user-commands-system"
       name="ignition::gazebo::systems::UserCommands">
     </plugin>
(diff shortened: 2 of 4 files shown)


src/systems/force_torque/ForceTorque.cc
```diff
@@ -137,7 +137,9 @@ void ForceTorque::PostUpdate(const UpdateInfo &_info,
     for (auto &it : this->dataPtr->entitySensorMap)
     {
       // Update measurement time
-      it.second->Update(_info.simTime);
+      auto time = math::durationToSecNsec(_info.simTime);
```
chapulina, line 140: I see the same pattern is used on other sensors, but is the conversion to sec/nsec and back necessary? Can't we just call `Update(_info.simTime, false)`?

chapulina (commented):
> This bug is introduced by calling ForceTorqueSensor::Update rather that Sensors::Update from the base class. 

I see the same pattern is used for other sensors, but with this pattern, how are we handling custom updates for each sensor type? That's why they override the parent `Sensor::Update` function, right?

Is there a problem in `ign-sensors` itself? I'd expect the overriding functions to call the parent function, i.e.

```
bool ForceTorqueSensor::Update(const std::chrono::steady_clock::duration &_now)
{
    Sensor::Update(_now);
...
```

src/systems/force_torque/ForceTorque.cc
mjcarroll: This may be vestigial from switching over how we handled time in Fortress. I can do a follow up to fix all of the locations where we are using this pattern.

mjcarroll:
Ah, so I think I see the issue.

In `Sensor`, we have:

```
public: virtual bool Update(const std::chrono::steady_clock::duration &_now) = 0;
public: bool Update(const std::chrono::steady_clock::duration &_now, const bool _force);
```

Where the argument version calls the virtual version to be implemented by downstream users.

In our systems in ign-gazebo, we generally keep the pointer to the most-derived class, so we don't have access to `Sensor::Update` that contains the logic that we are most interested in.

src/systems/force_torque/ForceTorque.cc
mjcarroll: Similar to my comment on the main thread, I think we can do away with the time logic, but we still have to cast to the base class in order to get the `Update` function that we want to use here.

Our two options would be, neither of which are particularly pretty.

```
      it.second.get()->sensors::Sensor::Update(_info.simTime, false);
      static_cast<sensors::Sensor*>(it.second.get())->Update(_info.simTime, false);
```

mjcarroll: I updated to my preferred call, but I'm open to suggestions.

mjcarroll: One other option would be to put `using Sensor::Update(..., ...)` in each of the derived classes.

chapulina: > One other option would be to put using Sensor::Update(..., ...) in each of the derived classes.

chapulina: > One other option would be to put using Sensor::Update(..., ...) in each of the derived classes.

I was just looking into that, and what's currently happening is that this function:

`public: bool Update(const std::chrono::steady_clock::duration &_now, const bool _force);`

Calls this one:

`public: virtual bool Update(const std::chrono::steady_clock::duration &_now) = 0;`

So I think this PR is correct, calling the outermost function. We should do the same for all sensors. And we should rethink whether we want the 2nd function to be public at all; it seems to me that calling it is a mistake, because it only does a subset of what the other function does.

chapulina:
The approach here LGTM, we just need to fix the test:

```
  /github/workspace/test/integration/force_torque_system.cc:142: Failure
    iters
      Which is: 1000
    wrenches.size()
      Which is: 31
```

chapulina (approved):

src/systems/force_torque/ForceTorque.cc
chapulina: Follow-up PRs:

* https://github.com/ignitionrobotics/ign-gazebo/pull/1282
* https://github.com/ignitionrobotics/ign-sensors/pull/187
* https://github.com/ignitionrobotics/ign-gazebo/pull/1283

PR #1161, opened 2021-11-03
Clarify available Ignition Versions

vatanaksoytezer:
Just a small clarification on which versions are available to install.


chapulina (approved):
LGTM. As a reference, we also have this table where we list the EOL dates per library, which can be helpful: https://github.com/ignitionrobotics/docs/blob/master/tools/versions.md

PR #1163, opened 2021-11-04
Publish 10 world stats msgs/sec instead of 5

adlarkin:

Addresses https://github.com/ignitionrobotics/ign-gazebo/pull/1109#issuecomment-961207086

## Summary
The changes made in #1109 and https://github.com/ignitionrobotics/ign-gui/pull/306 result in the GUI listening to the world stats topic to receive play/pause confirmations from the server. The publisher of this topic is currently publishing 5 msgs/second, but I think it would be good to publish at a higher rate to decrease response times between a GUI play/pause button press and server confirmation.

## Test it
Use these changes with #1109 and try pressing play/pause/step in the GUI. The GUI response times should be a little faster now.

## Checklist
- [ ] Added tests

### DIFF
--- a/src/SimulationRunner.cc
+++ b/src/SimulationRunner.cc
@@ -661,7 +661,7 @@ bool SimulationRunner::Run(const uint64_t _iterations)
   if (!this->statsPub.Valid())
   {
     transport::AdvertiseMessageOptions advertOpts;
-    advertOpts.SetMsgsPerSec(5);
+    advertOpts.SetMsgsPerSec(10);
     this->statsPub = this->node->Advertise<ignition::msgs::WorldStatistics>(
         "stats", advertOpts);
   }
--- a/src/SimulationRunner.cc
+++ b/src/SimulationRunner.cc
@@ -661,6 +661,12 @@ bool SimulationRunner::Run(const uint64_t _iterations)
   if (!this->statsPub.Valid())
   {
     transport::AdvertiseMessageOptions advertOpts;
+    // publish 10 world statistics msgs/second. A smaller number isn't used
+    // because the GUI listens to these msgs to receive confirmation that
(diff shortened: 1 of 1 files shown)


chapulina (approved):
I'm afraid we may be tweaking this hardcoded value often. We could consider exposing this as an option on `ServerConfig`, and optionally as a command line option, but that may be overkill. Defaulting to something that makes the GUI usable makes sense to me.

adlarkin:
> I'm afraid we may be tweaking this hardcoded value often. We could consider exposing this as an option on `ServerConfig`, and optionally as a command line option, but that may be overkill. Defaulting to something that makes the GUI usable makes sense to me.

It would be nice to remove hardcoded values whenever possible. We can make this adjustable on `ServerConfig`, or at least add a comment in the code that explains why `10` is used. Would it be worth making either of these additions/changes before merging this PR?

chapulina:
> Would it be worth making either of these additions/changes before merging this PR?

I'd say at least the comment would be a good idea. I'm fine exposing the parameter in a later PR. We may want to think a bit more about how to expose these "small" options, so that we don't bloat `ServerConfig` and the command line options

adlarkin:
> I'd say at least the comment would be a good idea.

I left a comment in https://github.com/ignitionrobotics/ign-gazebo/pull/1163/commits/145b177670d80a5638a503a33432ce68e1f43f9e. We should wait to merge this until https://github.com/ignitionrobotics/ign-gui/pull/306 and https://github.com/ignitionrobotics/ign-gazebo/pull/1163 are merged in order for the comment to make sense with the git history.

> I'm fine exposing the parameter in a later PR. We may want to think a bit more about how to expose these "small" options, so that we don't bloat ServerConfig and the command line options

I've opened an issue for this (#1164). We can have the discussion there.

PR #1171, opened 2021-11-06
Support editing air pressure sensor in the GUI

nkoenig:

# 🎉 New feature

## Summary

Adds support for editing the noise properties of the air pressure sensor in the component inspector.

Requires:
* https://github.com/ignitionrobotics/ign-gazebo/pull/1109
* https://github.com/ignitionrobotics/sdformat/pull/737

I've made some changes to the component inspector in an attempt to make it more modular, these include:
1. We have ability to (mostly) encapsulate an inspector element into a set of three files consisting of a `.cc`, `.hh`, and `.qml`. See #1172 for an example of adding a new inspector element.
2. There is a bit of ugliness where an inspector element has to [callout to the `componentInspector`](https://github.com/ignitionrobotics/ign-gazebo/pull/1171/files#diff-ffe1a0ce87209ac513a5b088feb12af0e52506e74e95c83702a015b1815baba5R130) in order to then call the [c++ function](https://github.com/ignitionrobotics/ign-gazebo/pull/1171/files#diff-37e11db45de8ed455c819cc8d33a7bbcafb06b3f6cd9c865bdc191e6b81baf16R93). Any help here would be appreciated.
3. The header for a element can be customized. By default the header name is based on the QML filename.

## Test it

1. `ign gazebo -v 4 sensors.sdf`
2. Take note of the output from `ign model -m sensors_box -l link -s air_pressure`
3. Select the altimeter sensor in the entity tree.
4. Make some changes to the values in the component inspector. 
5. Rerun `ign model -m sensors_box -l link -s air_pressure` and you should see the changes.

## Checklist
- [ ] Added tests


mjcarroll:
> There is a bit of ugliness where an inspector element has to callout to the componentInspector in order to then call the c++ function. Any help here would be appreciated.

Here is an attempt at a different implementation that works by overloading the function protoype.

https://github.com/ignitionrobotics/ign-gazebo/commit/37eade487000fe28a8e8f9832f61e48fddff1731

nkoenig:
> > There is a bit of ugliness where an inspector element has to callout to the componentInspector in order to then call the c++ function. Any help here would be appreciated.
> 
> Here is an attempt at a different implementation that works by overloading the function protoype.
> 
> [37eade4](https://github.com/ignitionrobotics/ign-gazebo/commit/37eade487000fe28a8e8f9832f61e48fddff1731)

Thanks, but I think the ugliness is still there. The `componentInspector` is still required in your approach. I can't figure out ho to call `AirPressure::OnAirPressureNoise` from `AirPressure.qml`. I'm only able to call the function from `ComponentInspector.qml`. Also, @chapulina mentioned in a separate conversation that slots/signals are preferred to function overloading. If it's okay with you, I'll revert this change.

mjcarroll:
> If it's okay with you, I'll revert this change.

I didn't put it on your branch, made a new branch, so should still be good.

src/gui/plugins/component_inspector/AirPressure.cc
```diff
@@ -0,0 +1,120 @@
...
+ * limitations under the License.
+ *
+*/
+#include <sdf/AirPressure.hh>
+
+#include <ignition/common/Console.hh>
+#include <ignition/gazebo/components/AirPressureSensor.hh>
+#include <ignition/gazebo/EntityComponentManager.hh>
+
+#include "ComponentInspector.hh"
```
ahcorde, line 23: alphabetize

src/gui/plugins/component_inspector/AirPressure.hh
```diff
@@ -0,0 +1,60 @@
...
+    /// \param[in] _stdDevBias Bias standard deviation value
+    /// \param[in] _dynamicBiasStdDev Dynamic bias standard deviation value
+    /// \param[in] _dynamicBiasCorrelationTime Dynamic bias correlation time
+    /// value
+    public: Q_INVOKABLE void OnAirPressureNoise(
+                double _mean, double _meanBias, double _stdDev,
+                double _stdDevBias, double _dynamicBiasStdDev,
+                double _dynamicBiasCorrelationTime);
+
+    public: Q_INVOKABLE void OnAirPressureReferenceAltitude(
```
ahcorde, line 51: docs ?

ahcorde (changes requested):

chapulina (changes requested):
> There is a bit of ugliness where an inspector element has to callout to the componentInspector in order to then call the c++ function. Any help here would be appreciated.

I think I got it, see https://github.com/ignitionrobotics/ign-gazebo/pull/1179

---

I think the new sensor noise is not taking effect. The `ign-sensor` objects are held at `system::AirPressure`:

https://github.com/ignitionrobotics/ign-gazebo/blob/ee8c3fd6e1a2cef3d999370033b9484b9d537f53/src/systems/air_pressure/AirPressure.cc#L53-L54

I checked it by plotting the sensor's readings and setting very high noise.

We should propagate the changes to the sensor there. The pattern we've been using is to create a service on `UserCommands`, which sets a component called `XXXCmd`, that's processed by the responsible system, in this case `system::AirPressure`. See an example in https://github.com/ignitionrobotics/ign-gazebo/pull/536

src/gui/plugins/component_inspector/AirPressure.qml
```diff
@@ -0,0 +1,136 @@
...
+          IgnSpinBox {
+            id: referenceAltitudeSpin
+            Layout.fillWidth: true
+            height: 40
+            property double refAltitude: model.data[0]
+            value: referenceAltitudeSpin.activeFocus ? referenceAltitudeSpin.value : refAltitude 
+
+            minimumValue: 0 
+            maximumValue: 100000 
+            decimals:4
```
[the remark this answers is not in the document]
nkoenig: Fixed an issue with zero-padding when using this function in 1a3b5ee

src/gui/plugins/component_inspector/ComponentInspector.cc
```diff
@@ -100,6 +103,16 @@ namespace ignition::gazebo
 
     /// \brief Transport node for making command requests
     public: transport::Node node;
+
+    /// \brief Air pressure sensor inspector elements
+    public: std::unique_ptr<ignition::gazebo::AirPressure> airPressure;
+
+    /// \brief Set of callbacks to execture during the Update function.
```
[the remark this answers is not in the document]
nkoenig: Fixed in 1a3b5ee

src/gui/plugins/component_inspector/AirPressure.hh
nkoenig: Fixed in 1a3b5ee

src/gui/plugins/component_inspector/AirPressure.cc
nkoenig: Fixed in 1a3b5ee

chapulina:
> Yeah, the noise values will not take effect until: #1170. You should see the changes in the ECM, just not in simulation output. I'm trying not to utilize additional services via the UserCommands pattern.

I see, so is the idea that the component inspector will add the `Recreate` component to the parent model whenever the user edits a value? Is that going to happen only while paused or also while playing?

I understand that the `*Cmd` pattern involves a lot of boilerplate and bloats the list of services, but that's efficient and works at runtime. Recreating the model just to edit a few sensor properties will be extremely inefficient if we do it every time the user clicks the spin arrows. For this particular use case, I think that a single `components::SensorCmd` and a `~/sensor_config` service that handles all sensor modifications.

nkoenig:
> > Yeah, the noise values will not take effect until: #1170. You should see the changes in the ECM, just not in simulation output. I'm trying not to utilize additional services via the UserCommands pattern.
> 
> I see, so is the idea that the component inspector will add the `Recreate` component to the parent model whenever the user edits a value? Is that going to happen only while paused or also while playing?

Yes , that's correct.

> 
> I understand that the `*Cmd` pattern involves a lot of boilerplate and bloats the list of services, but that's efficient and works at runtime. 

We are currently only trying to get parameters editable by the GUI between play/pause. The service pattern would become quite large when we add in all the sensors and joints, and we don't want users to edit these parameters while simulation running (at least for right now). The ECM sync is a more efficient way to handle all of the parameter updating that we want to do via the GUI at the moment, and has a better user experience (see below). In the long run it would be nice to add in the concept of parameters, similar to ROS parameters.

> Recreating the model just to edit a few sensor properties will be extremely inefficient if we do it every time the user clicks the spin arrows. For this particular use case, I think that a single `components::SensorCmd` and a `~/sensor_config` service that handles all sensor modifications. 

The ECM is only created/synced when a sensor is added or modified. Unless a user is making a change every step, then they shouldn't notice. An asynchronous service approach adds complexity and noticeable delays. For example, the pose update still uses the service method, and I see the following behavior:

1. Translate a model in the GUI.
2. The model bounces back to it's original position.
3. Server gets the new pose and sends back the new correct pose.
4. The model bounce to the location I set in step 1.

This is the type of user experience that I'd like to avoid.

chapulina:
> we don't want users to edit these parameters while simulation running 

Got it. I think this PR is missing the concept of a read-only mode for the new widgets then, so the user doesn't edit the sensor while playing by mistake.

> The ECM sync is a more efficient way to handle all of the parameter updating that we want to do via the GUI at the moment

I can see how that's efficient if lots of changes are queued while paused and only handled on play. This important detail is missing from this PR though.

> An asynchronous service approach adds complexity and noticeable delays

Yeah it would be nice to avoid the delay. But I'm not sure how the proposed approach will avoid the asynchronous requests since it still requires a round-trip to the server.

nkoenig:
> Got it. I think this PR is missing the concept of a read-only mode for the new widgets then, so the user doesn't edit the sensor while playing by mistake.

Yup, that is correct.

> 
> > The ECM sync is a more efficient way to handle all of the parameter updating that we want to do via the GUI at the moment
> 
> I can see how that's efficient if lots of changes are queued while paused and only handled on play. This important detail is missing from this PR though.

I don't follow, what detail are you referring to?

> > An asynchronous service approach adds complexity and noticeable delays
> 
> Yeah it would be nice to avoid the delay. But I'm not sure how the proposed approach can avoid the asynchronous requests while unpaused since it still requires a round-trip to the server.

The problem is avoided because the GUI ECM has the most up-to-date ECM state, which is then sent to the server when the play button is pressed. The GUI registers the state change from pause to play when the server sends out a a new state message. The couple ECM from the server is the same as the ECM on the GUI. There is no visible jump.

nkoenig:
Retargeting to the `model_editor` branch.

nkoenig:
Added in a simulation state aware spin box that enables/disables based on the play/pause state of simulation.

src/systems/scene_broadcaster/SceneBroadcaster.cc
```diff
@@ -199,9 +199,18 @@ class ignition::gazebo::systems::SceneBroadcasterPrivate
...
+  /// default update rate of 60Hz. The paused period has a key=true and a
+  /// default update rate of 30Hz.
+  public: std::map<bool, std::chrono::duration<int64_t, std::ratio<1, 1000>>>
+      statePublishPeriod{{false, std::chrono::milliseconds(1000/60)},
+                         {true,  std::chrono::milliseconds(1000/30)}};
+
   /// \brief Period to publish state, defaults to 60 Hz.
   public: std::chrono::duration<int64_t, std::ratio<1, 1000>>
-      statePublishPeriod{std::chrono::milliseconds(1000/60)};
+      statePublishPeriodPaused{std::chrono::milliseconds(1000/30)};
```
iche033, line 213: I think this is no longer used and safe to remove

iche033 (approved):
did some quick tests changing the air pressure values and it's working for me as advertised. Just one minor comment

src/systems/scene_broadcaster/SceneBroadcaster.cc
nkoenig: I believe this was needed for the GUI to be aware of the play/paused state so that it can enable/disable the edit boxes.

nkoenig: I confirmed this. Change `itsPubTime` in line 306 below to include ` !_info.paused &&`. Run simulation and toggle the play/pause state. The edit boxes on the air pressure sensor will remain in the disabled state.

iche033: hmm I think this variable is replaced by the `statePublishPeriod` map? I see in line 306 it's using the `statePublishPeriod` variable (as opposed to `statePublishPeriodPaused`). I also tried removing this variable and ign-gazebo still builds.

nkoenig: Argh, right. My bad. Removed.

PR #1176, opened 2021-11-08
use entity instead of entity name

nkoenig:
Suggestion to use `Entity` instead of a string when creating links.


iche033 (approved):
looks good to me

PR #1165, opened 2021-11-04
Model editor: Add links to model

iche033:
# 🎉 New feature

## Summary

This pull request adds a `+` menu button to the Component Inspector for adding links to models. Currently supports adding links consisting of simple shapes and lights.

The changes are written to the ECM and GUI side and assumes that GUI and server ECM synchronization (#1109) is available. Marked as draft until the newly added entities can be reflected on the server side.

## Test it

1. launch ign gazebo with shapes.sdf world
1. Open the Model Editor GUI plugin by going to top right 3 dots menu and select `Model Editor`
1. Click on one of the shapes, e.g. box
1. Go the Component Inspect and you should see a new `+` button appear on the top bar
1. Click on the `+` button and select a link type to add, .e.g Cylinder
1. Once selected, a black cylinder should now appear at where the original box model is.

## Checklist
- [x] Signed all commits for DCO
- [ ] Added tests
- [ ] Added example and/or tutorial
- [ ] Updated documentation (as needed)
- [ ] Updated migration guide (as needed)
- [ ] `codecheck` passed (See [contributing](https://ignitionrobotics.org/docs/all/contributing#contributing-code))
- [ ] All tests passed (See [test coverage](https://ignitionrobotics.org/docs/all/contributing#test-coverage))
- [ ] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Aignitionrobotics+repo%3Aosrf%2Fsdformat+archived%3Afalse+) to support the maintainers

**Note to maintainers**: Remember to use **Squash-Merge**

PR #1177, opened 2021-11-08
Add link menu updates

nkoenig:
An suggestion for #1165, which builds on #1176. The changes include:

1. The add link menu is now a `Menu` that is a child of the `Toolbutton`.
2. A mesh can be added to a link.


iche033 (approved):
adding mesh now works for me

PR #1178, opened 2021-11-09
Wirebox deletion: Suggestion to #1053

chapulina:
Suggestion to #1053 that avoids adding GUI as a dependency of rendering.

`RenderUtil` is part of `gazebo::rendering` and is used by both the backend scene (`systems::Sensors`) and the front-end. We shouldn't introduce a dependency on GUI to the backend. The suggestion here emits the event from `GzSceneMananger`, because that's the GUI plugin responsible for connecting the GUI scene to the ECM (most of it done leveraging `RenderUtil`).

For future reference, when I first looked at #1053 I thought we may not need an event because the wirebox should be deleted as a child of the visual. But it looks like that doesn't happen; not sure if it's kept alive because the shared pointer is kept in a map inside `SelectEntities`. In any case, we need to remove the item from that map, and I think that the event is a good way of doing that.

### DIFF
--- a/src/gui/plugins/scene_manager/GzSceneManager.cc
+++ b/src/gui/plugins/scene_manager/GzSceneManager.cc
@@ -17,6 +17,8 @@
 
 #include "GzSceneManager.hh"
 
+#include <vector>
+
 #include <ignition/common/Profiler.hh>
 #include <ignition/gui/Application.hh>
 #include <ignition/gui/GuiEvents.hh>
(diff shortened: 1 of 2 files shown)


jennuine (approved):
:bowing_woman: Thank you for the suggestion! LGTM, I added `#include <vector>` in 2eb80d3b056289a7f8b78436ea79e7f61031d87f for `codechek`

PR #1179, opened 2021-11-09
Move AirPressure functions out of ComponentInspector (suggestion to #1171)

chapulina:
It seems like there's a name collision on QML between the C++ property and the QML component itself. I suspect all Ignition GUI plugins may have been getting away using the same name due to lucky (deterministic) timing.

Giving the property a different name seems to prevent it from being overridden.

### DIFF
--- a/src/gui/plugins/component_inspector/AirPressure.cc
+++ b/src/gui/plugins/component_inspector/AirPressure.cc
@@ -30,7 +30,7 @@ using namespace gazebo;
 /////////////////////////////////////////////////
 AirPressure::AirPressure(ComponentInspector *_inspector)
 {
-  _inspector->Context()->setContextProperty("AirPressure", this);
+  _inspector->Context()->setContextProperty("AirPressureImpl", this);
   this->inspector = _inspector;
 
   ComponentCreator creator =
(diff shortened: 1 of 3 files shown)


src/gui/plugins/component_inspector/AirPressure.cc
```diff
@@ -30,7 +30,7 @@ using namespace gazebo;
 /////////////////////////////////////////////////
 AirPressure::AirPressure(ComponentInspector *_inspector)
 {
-  _inspector->Context()->setContextProperty("AirPressure", this);
+  _inspector->Context()->setContextProperty("AirPressureImpl", this);
```
mjcarroll, line 33: Does the name collision come from the fact that we are adding property (typically object instances) using `CamelCase` which is what we normally use for class names?

nkoenig:
Fantastic!! Thanks.

nkoenig (approved):

src/gui/plugins/component_inspector/AirPressure.cc
chapulina: My guess is that it's a collision between this context property, and the `AirPressure.qml` item itself

PR #1185, opened 2021-11-10
Add sdfString to copy constructor.

### LINKED ISSUE #1182: SDF String not copied as part of server config copy constructor.
In `src/ServerConfig.cc`, the [copy constructor](https://github.com/ignitionrobotics/ign-gazebo/blob/ign-gazebo6/src/ServerConfig.cc#L220) for `ServerConfigPrivate` does not copy over the `sdfString` variable. This is surprising, so my recommendation would be one of the following:

(A) If it's a bug, fix it 😃
(B) If it's not a bug, then documentation should be added to that class describing why not.

Thanks!

Mokosha:
# 🦟 Bug fix

Fixes #1182

### DIFF
--- a/src/ServerConfig.cc
+++ b/src/ServerConfig.cc
@@ -220,6 +220,7 @@ class ignition::gazebo::ServerConfigPrivate
   public: explicit ServerConfigPrivate(
               const std::unique_ptr<ServerConfigPrivate> &_cfg)
           : sdfFile(_cfg->sdfFile),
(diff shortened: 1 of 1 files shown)


ahcorde (approved):

ahcorde:
@Mokosha , Do you mind to sign the [commit](https://github.com/ignitionrobotics/ign-gazebo/pull/1185/checks?check_run_id=4163029221) ?

chapulina (approved):

PR #1187, opened 2021-11-10
Together

nkoenig:
This is part of the ModelEditor work, where I'm trying to get the recreation functionality to work. Mostly looking for some assistance.

Here is how I'm currently testing:

1. `ign gazebo -v 4 sensors.sdf`
2. Add a new link to the `sensors_box` model using the GUI.
3. Click `Step` on the GUI.
4. You will see the following console output
```
Adding entity for recreation[8]
[Wrn] [Physics.cc:1102] Collision's parent entity [1073741824] not found on link map.
[Err] [Physics.cc:2365] Internal error: entity [1073741824] not in entity map
[Dbg] [EntityComponentManager.cc:1540] Updated state thread iterators: 16 threads processing around 2 entities each.
Cloning entity[8] Name[sensors_box]
Cloning entity[9] Name[link]
Cloning entity[10] Name[visual]
Cloning entity[11] Name[collision]
Cloning entity[12] Name[altimeter]
Cloning entity[13] Name[air_pressure]
Cloning entity[14] Name[imu]
Cloning entity[15] Name[magnetometer]
Cloning entity[1073741824] Name[link_1]
Cloning entity[1073741825] Name[visual]
Cloning entity[1073741826] Name[collision]
Removing recreate component from entity[17]
Adding entity for recreation[17]
```
5. I'm ignoring the Physics system error for right now. I'm trying to understand why removing the `Recreate` component from the newly create entity (17 in this case) still causes the `View` in  `SimulationRunner::ProcessRecreateEntitiesRemove` to think the `Recreate` component exists.

There are some debug statements that need to be removed once the issues are resolved.


iche033:
I can reproduce the issue that the recreate component is not removed. Looking into it.

nkoenig:
I believe the issue is that the cloned entity is added to the view's  `toAddEntities` list, and is not removed from this list by the `View::NotifyComponentRemoval` function.

iche033:
I pushed a change that works around the issue mainly by using `EachNoCache` [here](https://github.com/ignitionrobotics/ign-gazebo/pull/1187/commits/c312514705ef1feebaf469f465fc40c60e5d36a5#diff-528565fd9b23fec7ea46e2f169d2098361634a77a74cac62166310c4624ca40dR1295) when looping through entities with Recreate component. Ign-gazebo still crash in the physics system when it is stepped though. I'll continue looking at how to properly remove components from views first

iche033:
> I'll continue looking at how to properly remove components from views first
e312919 should do it. Also updated test to verify that the Recreate components are removed

The physics system crash still exists.

nkoenig:
Thanks. I'm getting closer on the physics crash.

nkoenig:
I believe physics is fixed in 199dafe

nkoenig:
This also behaves well when adding links to models with sensors.

nkoenig:
I'll go ahead and merge this in the `recreate_entities` branch.

PR #1190, opened 2021-11-11
Thruster plugin: add tests and velocity control

chapulina:
## Summary

The `Thruster` plugin was added in #749 without tests. This PR adds tests, and while at it, I also:

* Documentation
    * Documented `examples/worlds/auv_controls.sdf`
    * Documented missing XML fields in `Thruster.hh`
    * Updated variables and docs that mentioned RPM, because the plugin uses SI for angular velocity (rad / s)
* XML params
    * Some parameters were listed as "required", but also offered a default value. I made the implementation consistent with documentation by making parameters that have default values be optional.
    * Fail loading the plugin if the user gives a joint name that doesn't exist
    * Default the namespace to the model name instead of empty; empty is weird because the topic becomes `/model/joint/...` - this can potentially break users that are currently leaving the namespace empty.
* Implementation
    * Compute the desired propeller angular velocity only when the desired thrust changes
    * I think the `cmd_pos` topic is very confusing for this plugin, because the commands are forces in Newtons. So I'm proposing changing the topic to `cmd_thrust`. I didn't remove `cmd_pos` for backwards compatibility, but I think we should tick-tock it out.
    * Take the `<joint><pose>` into account
* Velocity control
    * Ported the alternative joint velocity control from downstream (https://github.com/osrf/lrauv/commit/c20b44c3b28393ea6929ab22073009dbb7a94999).
    * The motivation for this feature was that the PID was accumulating errors that made the propellers drift over time
    * While writing tests, it became clear that using `JointVelocityCmd` makes the vehicle much more unstable than directly applying the wrench to the link. So there's a trade-off.

## Test it

Follow the instructions in `auv_controls.sdf`.

Here are some handy commands to run the test worlds too:

```
ign gazebo -r -v 4 src/ign-gazebo/test/worlds/thruster_pid.sdf
ign topic -t /model/sub/joint/propeller_joint/cmd_thrust  -m ignition.msgs.Double -p 'data: 0.3'
```

```
ign gazebo -r -v 4 src/ign-gazebo/test/worlds/thruster_vel_cmd.sdf
ign topic -t /model/custom/joint/propeller_joint/cmd_thrust  -m ignition.msgs.Double -p 'data: 0.3'
```

## Checklist
- [x] Added tests


arjo129 (approved):
We should be OK to merge this as most of the issues are experienced at low velocities and I suspect due to issues with the hydrostatics.

matosinho:
I see that @chapulina added TODO to configure max and min thrust in the sdf. This means that it will be done in another PR? I think that would be a good addition.

chapulina:
> I see that @chapulina added TODO to configure max and min thrust in the sdf. This means that it will be done in another PR? I think that would be a good addition.

I wasn't planning on tackling it anytime soon, but I agree it would be a nice addition

test/integration/thruster.cc
```diff
@@ -0,0 +1,216 @@
...
+      EXPECT_NE(propellerEntity, kNullEntity);
+
+      propeller = Link(propellerEntity);
+      propeller.EnableVelocityChecks(_ecm);
+    }).
+  OnPostUpdate([&](const gazebo::UpdateInfo &_info,
+                            const gazebo::EntityComponentManager &_ecm)
+    {
+      dt = std::chrono::duration_cast<std::chrono::nanoseconds>(
+          _info.dt).count() * 1e-9;
```
arjo129, line 93: Knit: could be done with `std::duration_cast<std::chrono::seconds, double>(_info.dt).count()` thus avoiding the use of 1e-9.

test/integration/thruster.cc
```diff
@@ -0,0 +1,216 @@
...
+
+    // The joint velocity command adds some roll to the body which the PID
+    // wrench doesn't
+    if (_namespace == "custom")
+      EXPECT_NEAR(0.0, pose.Rot().Roll(), 0.1);
+    else
+      EXPECT_NEAR(0.0, pose.Rot().Roll(), _baseTol);
+  }
+
+  // omega = sqrt(thrust /
```
arjo129, line 179: Should we add the book reference in the test for clarity?

arjo129 (approved):
Thanks for doing this! The tests and up-streaming are looking good.

test/integration/thruster.cc
chapulina: Good call 9a0642b

test/integration/thruster.cc
chapulina: 9a0642b

iche033:
> I see that @chapulina added TODO to configure max and min thrust in the sdf. This means that it will be done in another PR? I think that would be a good addition.

here's a pull request for this: https://github.com/ignitionrobotics/ign-gazebo/pull/1318

osrf-triage:

PR #1192, opened 2021-11-11
Prevent creation of spurious <plugin> elements when saving worlds

azeey:
# 🦟 Bug fix

## Summary
Spurious `<plugin>` elements were created when saving worlds because `sdf::Element::GetElement` was used to check if a `<plugin>` child element exists. However this function always creates a new element if the element does not exist instead of returning a `nullptr`. The solution is to use `sdf::Element::FindElement` instead. 

## Checklist
- [x] Added tests
- ~[ ] Updated documentation (as needed)~
- ~[ ] Updated migration guide (as needed)~

### DIFF
--- a/src/SimulationRunner.cc
+++ b/src/SimulationRunner.cc
@@ -959,7 +959,7 @@ void SimulationRunner::LoadLoggingPlugins(const ServerConfig &_config)
 void SimulationRunner::LoadPlugins(const Entity _entity,
     const sdf::ElementPtr &_sdf)
 {
-  sdf::ElementPtr pluginElem = _sdf->GetElement("plugin");
+  sdf::ElementPtr pluginElem = _sdf->FindElement("plugin");
   while (pluginElem)
   {
     auto filename = pluginElem->Get<std::string>("filename");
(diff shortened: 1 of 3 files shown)


src/SimulationRunner.cc
```diff
@@ -959,16 +959,15 @@ void SimulationRunner::LoadLoggingPlugins(const ServerConfig &_config)
...
   {
     auto filename = pluginElem->Get<std::string>("filename");
     auto name = pluginElem->Get<std::string>("name");
     // No error message for the 'else' case of the following 'if' statement
     // because SDF create a default <plugin> element even if it's not
     // specified. An error message would result in spamming
-    // the console. \todo(nkoenig) Fix SDF should so that elements are not
-    // automatically added.
+    // the console.
     if (filename != "__default__" && name != "__default__")
```
chapulina, line 971: Is this check still needed? I think it was meant to avoid the plugins created by `GetElement`?

chapulina (approved):
LGTM, I just have a question

src/SimulationRunner.cc
azeey: Yes, I think it's still needed because all saved worlds prior to this PR would have these plugin tags in SDFormat files.

PR #1196, opened 2021-11-12
Model Editor: Add Joints to model

mjcarroll:
# 🎉 New feature

## Summary

Use the model editor to insert joints into a model.

## Test it

Load a world, select a model in the entity tree and try adding a new joint from the "plus" menu.

## Checklist
- [ ] Added tests

### DIFF
--- a/include/ignition/gazebo/SdfEntityCreator.hh
+++ b/include/ignition/gazebo/SdfEntityCreator.hh
@@ -122,6 +122,13 @@ namespace ignition
       /// \return Joint entity.
       public: Entity CreateEntities(const sdf::Joint *_joint);
 
+      /// \brief Create all entities that exist in the sdf::Joint object and
+      /// load their plugins.
+      /// \param[in] _joint SDF joint object.
+      /// \param[in] _resolved True if all frames are already resolved
+      /// \return Joint entity.
+      public: Entity CreateEntities(const sdf::Joint *_joint, bool _resolved);
+
       /// \brief Create all entities that exist in the sdf::Visual object and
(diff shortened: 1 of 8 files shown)


nkoenig:
I see a problem when adding a link then a joint. For example:

1. Load `shapes.sdf`
2. Add a link to the box.
3. Add a joint to the box.
4. Step simulation. You'll see the following errors:
```
[Err] [SDFFeatures.cc:384] Could not find link link in model box
[Err] [SDFFeatures.cc:750] The link of the child frame [link] with resolved link name [link] of joint [joint] in model [box] could not be resolved. The joint will not be constructed
```

mjcarroll:
> I see a problem when adding a link then a joint. For example:

I believe it's a sequencing issue in the entity creator, I was able to reproduce and I'm looking at it.

nkoenig:
> > I see a problem when adding a link then a joint. For example:
> 
> I believe it's a sequencing issue in the entity creator, I was able to reproduce and I'm looking at it.

Do you need assistance on this?

mjcarroll:
> Do you need assistance on this?

No, should have fix shortly.

nkoenig (approved):

iche033:
when I tried adding more than one joint, I see that the first joint will have a valid name, `joint`, but subsequent joint names are empty:
joint_names

mjcarroll:
> when I tried adding more than one joint, I see that the first joint will have a valid name, joint, but subsequent joint names are empty

Investigating

mjcarroll:
Fixed in #1228

Issue #1210, opened 2021-11-16
Regression in Buoyancy Example, neutral vehicle slowly sinks


arjo129:
## Environment
* OS Version: Ubuntu 20.04
* Source or binary build?
  Source 97f2e9c8499fdeb89d45220a89b7d570c9ce6690

## Description

* Expected behavior: The center vehicle in the neutral Buoyancy example in the `buoyancy.sdf` example is supposed to remain stationary.
* Actual behavior: Instead it is sinking.

![sinking_buoyant_object](https://user-images.githubusercontent.com/542272/141979483-30f48f49-b7d2-419a-a95a-43ed923cc740.gif)

## Steps to reproduce

1. `ign gazebo buoyancy.sdf`

## More details

It seems like the cause is the way the buoyancy plugin is structured. In #1064 a change was made to prevent newly spawned objects from developing a spin. To do this, we skip one iteration of the update if the pose is not ready. Unfortunately this leads to problems as it gives an initial velocity to objects. Without damping this leads to the slow sinking we see above.

chapulina:
Let's keep it open until graded buoyancy is fixed too

PR #1211, opened 2021-11-16
Fix buoyancy not being applied for one iteration

arjo129:

# 🦟 Bug fix

Workaround #1210

## Summary
This PR currently is a work around, Essentially, it skips the check introduced in #1064 for the uniform buoyancy. This is fine as #1064 is not really position dependent. We should however leave #1210 open as no fix has been applied for layered buoyancy.

## Checklist
- [ ] Added tests


src/systems/buoyancy/Buoyancy.cc
```diff
@@ -477,6 +472,13 @@ void Buoyancy::PreUpdate(const ignition::gazebo::UpdateInfo &_info,
       else if (this->dataPtr->buoyancyType
         == BuoyancyPrivate::BuoyancyType::GRADED_BUOYANCY)
       {
+        if (newPose)
+        {
+          // Skip entity if WorldPose and inertial are not yet ready
+          // TODO(arjo): Find a way of disabling gravity effects for
+          // this first iteration.
```
chapulina, line 479: I think we may be able to avoid this skip if we use `gazebo::worldPose` instead of the `WorldPose` component. That function should provide the world pose straight away.

chapulina (approved):
Nice catch! This fix works for me. I have a suggestion below for a more permanent fix.

src/systems/buoyancy/Buoyancy.cc
arjo129: Hmm, but we need to use `link::WorldInertialPose`...

chapulina: Ahh I see, maybe the implementation of that method can be improved :thinking:

chapulina:
Ouch, we should have squash-merged this. No need to go back and fix it, but something to keep in mind for next time.

PR #1235, opened 2021-12-06
Send state message when components are removed

nkoenig:

# 🦟 Bug fix

## Summary

The `WorldCmdPose` was being transmitted to the GUI and back to the Server, resulting in bad behavior. See [here](https://github.com/ignitionrobotics/ign-gazebo/pull/1231#issuecomment-985923720).

This PR changes the SceneBroadcaster to send a new state message when components are removed.

Before I write tests, I'd like to get some feedback on this approach.

## Checklist
- [x] Added tests

### DIFF
--- a/test/integration/scene_broadcaster_system.cc
+++ b/test/integration/scene_broadcaster_system.cc
@@ -22,12 +22,16 @@
 
 #include <ignition/common/Console.hh>
 #include <ignition/common/Util.hh>
+#include "ignition/gazebo/components/Model.hh"
+#include "ignition/gazebo/components/Name.hh"
+#include "ignition/gazebo/components/Pose.hh"
 #include <ignition/transport/Node.hh>
 
 #include "ignition/gazebo/Server.hh"
 #include "ignition/gazebo/test_config.hh"
 
 #include "../helpers/EnvTestFixture.hh"
+#include "../helpers/Relay.hh"
 
 using namespace ignition;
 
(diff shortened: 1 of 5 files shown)


chapulina (commented):
Nice! We just need a unit test.

nkoenig:
> Nice! We just need a unit test.

Just checking that you don't think there are unforeseen side effects of this?

chapulina:
> Just checking that you don't think there are unforeseen side effects of this?

I can't think of any unwanted effects, the worst I can think of is that we're publishing more state messages, but that should be negligible if there aren't components being removed every iteration.

azeey:
We have been preferring to zero out components instead of removing them for performance reasons, so I don't think we have any components that get removed every iteration. So I think this approach might work as a work around. My proposal for a more long term solution is to have a scheme where components are requested for removal, just like entities, and their removal happens after all the `*Update` phases have completed. I think this would remove the need to keep `WorldPoseCmd` around for an extra iteration.

nkoenig:
Tests have been added in b2ad027

test/integration/scene_broadcaster_system.cc
```diff
@@ -612,6 +616,95 @@ TEST_P(SceneBroadcasterTest, StateStatic)
...
+                  ignition::gazebo::components::Pose>(
+          [&](const ignition::gazebo::Entity &_entity,
+              const ignition::gazebo::components::Model *,
+              const ignition::gazebo::components::Name *_name,
+              const ignition::gazebo::components::Pose *)->bool
+          {
+            if (_name->Data() == "box")
+            {
+              _ecm.RemoveComponent<ignition::gazebo::components::Pose>(_entity);
+              }
```
chapulina, line 651:
```suggestion
            }
```

test/integration/scene_broadcaster_system.cc
```diff
@@ -612,6 +616,95 @@ TEST_P(SceneBroadcasterTest, StateStatic)
...
+  sleep = 0u;
+  received = false;
+  hasState = false;
+  server.RunOnce(true);
+  // cppcheck-suppress unmatchedSuppression
+  // cppcheck-suppress knownConditionTrueFalse
+  while (!received && sleep++ < maxSleep)
+    IGN_SLEEP_MS(100);
+  EXPECT_TRUE(received);
+  EXPECT_TRUE(hasState);
```
chapulina, line 705: We could check specifically that the message contains a removed component

chapulina (approved):

test/integration/scene_broadcaster_system.cc
nkoenig: Done in 326a841

test/integration/scene_broadcaster_system.cc
nkoenig: Done in 326a841

chapulina (approved):

PR #1240, opened 2021-12-08
Implement /server_control::stop

### LINKED ISSUE #1237: /server_control stop does nothing
## Environment
* OS Version: Ubuntu 18.04
* Source or binary build? Dome binary from distribution

## Description
* Expected behavior: Calling `ign service -s /server_control --reqtype ignition.msgs.ServerControl --reptype ignition.msgs.Boolean --timeout 1000 --req 'stop: 1'` should stop the Gazebo server, i.e. exit the server binary.
* Actual behavior: Nothing happens.

## Steps to reproduce

1. `ign gazebo`
2. `ign service -s /server_control --reqtype ignition.msgs.ServerControl --reptype ignition.msgs.Boolean --timeout 1000 --req 'stop: 1'`
3. Nothing happens

## Output
This is written to the console:

```
[GUI] [Dbg] [TmpIface.cc:37] OnServerControl: request
[GUI] [Dbg] [TmpIface.cc:38] stop: true

[GUI] [Dbg] [TmpIface.cc:42] OnServerControl: response
[GUI] [Dbg] [TmpIface.cc:43] data: true
```

I saw #654, so who should be actually handling the ServerControl thing?

The implementation in Dome is actually just a dummy:

https://github.com/ignitionrobotics/ign-gazebo/blob/c473a08ea3a8cc73a9fb722efb08772abcf48dfe/src/gui/TmpIface.cc#L37-L45

However, I'm seeking for a gui-only/service-calling way to exit the server (to allow exiting the simulation in a touch-only kiosk where gui is launched from a separate console than server).

Is this a kind of supported scenario? I see Fortress does not even have TmpIface, so...

I know I could `exec()` a `killall ruby`, but that's not very nice...

peci1:
# 🎉 New feature

Closes #1237

## Summary
Implemented the `stop` part of `/server_control` service.

## Test it
1. `ign gazebo -v4`
2. `ign service -s /server_control --reqtype ignition.msgs.ServerControl --reptype ignition.msgs.Boolean --timeout 1000 --req 'stop: 1'`
3. The server should stop and exit in a while.

## Checklist
- [x] Added tests

### DIFF
--- a/include/ignition/gazebo/gui/TmpIface.hh
+++ b/include/ignition/gazebo/gui/TmpIface.hh
@@ -63,15 +63,6 @@ namespace ignition
       /// \param[in] _path Path to world file.
       public slots: void OnSaveWorldAs(const QString &_path);
 
-      /// \brief Server control service callback
-      /// This is the server-side logic which provides the world_control
-      /// service.
-      /// \param[in] _req Request
-      /// \param[out] _res Response
-      /// \return True for success
-      private: bool OnServerControl(const msgs::ServerControl &_req,
-                                          msgs::Boolean &_res);
-
       /// \brief Communication node
       private: transport::Node node;
 
--- a/include/ignition/gazebo/gui/TmpIface.hh
+++ b/include/ignition/gazebo/gui/TmpIface.hh
@@ -63,6 +63,15 @@ namespace ignition
       /// \param[in] _path Path to world file.
       public slots: void OnSaveWorldAs(const QString &_path);
 
+      /// \brief This function does nothing and is kept only for retaining ABI
+      /// compatibility. /server_control service handling was moved to
+      /// ServerPrivate.
+      /// \param[in] _req Request
+      /// \param[out] _res Response
+      /// \return False.
+      private: bool IGN_DEPRECATED(3) OnServerControl(
+        const msgs::ServerControl &_req, msgs::Boolean &_res);
+
       /// \brief Communication node
       private: transport::Node node;
 
--- a/src/ServerPrivate.cc
+++ b/src/ServerPrivate.cc
@@ -419,6 +419,19 @@ void ServerPrivate::SetupTransport()
     ignerr << "Something went wrong, failed to advertise [" << pathTopic
            << "]" << std::endl;
   }
+
+  std::string serverControlService{"/server_control"};
+  if (this->node.Advertise(serverControlService,
+                           &ServerPrivate::ServerControlService, this))
+  {
+    ignmsg << "Server control service on [" << serverControlService << "]."
+           << std::endl;
+  }
+  else
+  {
+    ignerr << "Something went wrong, failed to advertise ["
+           << serverControlService << "]" << std::endl;
+  }
 }
 
 //////////////////////////////////////////////////
@@ -431,6 +444,42 @@ bool ServerPrivate::WorldsService(ignition::msgs::StringMsg_V &_res)
   for (const auto &name : this->worldNames)
   {
     _res.add_data(name);
+  }
+
+  return true;
+}
+
+//////////////////////////////////////////////////
+bool ServerPrivate::ServerControlService(
+  const ignition::msgs::ServerControl &_req, msgs::Boolean &_res)
+{
+  _res.set_data(false);
+
+  if (_req.stop())
+  {
+    this->Stop();
+    _res.set_data(true);
+  }
+
+  // TODO(anyone): implement world cloning
+  if (_req.clone() || _req.new_port() != 0 || !_req.save_world_name().empty())
+  {
+    ignerr << "ServerControl::clone is not implemented" << std::endl;
+    _res.set_data(false);
+  }
+
+  // TODO(anyone): implement adding a new world
+  if (_req.new_world())
+  {
+    ignerr << "ServerControl::new_world is not implemented" << std::endl;
+    _res.set_data(false);
+  }
+
+  // TODO(anyone): implement loading a world
+  if (!_req.open_filename().empty())
+  {
+    ignerr << "ServerControl::open_filename is not implemented" << std::endl;
+    _res.set_data(false);
   }
 
   return true;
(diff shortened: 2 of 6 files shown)


peci1:
One question that came to my mind was what to do with the `msgs::ServerControl` message definition. It is apparently very much tied to Gazebo classic with the `new_port` member.

include/ignition/gazebo/gui/TmpIface.hh
```diff
@@ -63,15 +63,6 @@ namespace ignition
...
       public slots: void OnSaveWorldAs(const QString &_path);
 
-      /// \brief Server control service callback
-      /// This is the server-side logic which provides the world_control
-      /// service.
-      /// \param[in] _req Request
-      /// \param[out] _res Response
-      /// \return True for success
-      private: bool OnServerControl(const msgs::ServerControl &_req,
-                                          msgs::Boolean &_res);
```
chapulina, line 73: This is technically ABI breaking, even if we don't expect anyone to be using this header. Can we leave the header and just remove the implementaiton? The header is already gone in newer versions.

src/ServerPrivate.cc
```diff
@@ -436,6 +449,42 @@ bool ServerPrivate::WorldsService(ignition::msgs::StringMsg_V &_res)
...
+{
+  _res.set_data(false);
+
+  if (_req.stop())
+  {
+    this->Stop();
+    _res.set_data(true);
+  }
+
+  // TODO(anyone): implement world cloning
```
chapulina, line 464: We're trying to get rid of the `anyone`s (see the [style guide](https://google.github.io/styleguide/cppguide.html#TODO_Comments)). You can put on your name, or if you'd rather not, you can put `chapulina`.

chapulina (changes requested):
Thanks, LGTM, just some minor comments.

include/ignition/gazebo/gui/TmpIface.hh
peci1: You're right, this would be technically an ABI break. I returned the function and marked it deprecated - it's not exactly following the tick-tock deprecation rule, but as it was a private function and I removed the only call to it, it should be pretty much safe.

src/ServerPrivate.cc
peci1: Okay, I definitely do not feel like adding support for these functions, so I gave these TODOs to you :)

src/gui/TmpIface.cc
```diff
@@ -25,24 +25,16 @@ using namespace gazebo;
...
 {
-  // Server control
-  this->node.Advertise("/server_control",
-      &TmpIface::OnServerControl, this);
 }
 
+
 /////////////////////////////////////////////////
 bool TmpIface::OnServerControl(const msgs::ServerControl &_req,
                                      msgs::Boolean &_res)
```
chapulina, line 33: Now we have some compiler warnings:

```
unused parameter ‘_res’ [-Wunused-parameter] msgs::Boolean &_res) ~~~~~~~~~~~~~~~^~~~
```

peci1: Ahh, right... Fixed.

peci1:
I did a few tests locally, and it seems to me it'd be better to not call `this->Stop()` directly from the callback, but rather spin a separate thread for it. Otherwise, there could possibly be a deadlock in case the service is called from a plugin whose destructor cannot be finished until the service call finishes.

What do you think about adding this async execution?

peci1:
Here's a GUI plugin using the service: https://github.com/ignitionrobotics/ign-gui/pull/335 .

chapulina:
> there could possibly be a deadlock in case the service is called from a plugin whose destructor cannot be finished until the service call finishes.

I'd have to look a bit more closely. It looks like `SimulationRunner` is not handling the stop immediately and just sets a flag to handle the stop later.

> With the async implementation, I haven't seen any yet.

But I think you're onto something. If you've seen that it works I think it's worth it pursuing that approach.

peci1:
I've changed the callback to run in a thread.

If you look into SimulationRunner::Stop(), it emits an event, which I think can be blocking and could be the source of problems I'm talking about...

src/ServerPrivate.cc
```diff
@@ -436,6 +453,49 @@ bool ServerPrivate::WorldsService(ignition::msgs::StringMsg_V &_res)
...
+      }));
+    }
+    _res.set_data(true);
+  }
+
+  // TODO(chapulina): implement world cloning
+  if (_req.clone() || _req.new_port() != 0 || !_req.save_world_name().empty())
+  {
+    ignerr << "ServerControl::clone is not implemented" << std::endl;
+    _res.set_data(false);
```
chapulina, line 479: This is already set up-top, so I think we don't need to repeat it in every block

src/ServerPrivate.cc
```diff
@@ -436,6 +453,49 @@ bool ServerPrivate::WorldsService(ignition::msgs::StringMsg_V &_res)
...
+bool ServerPrivate::ServerControlService(
+  const ignition::msgs::ServerControl &_req, msgs::Boolean &_res)
+{
+  _res.set_data(false);
+
+  if (_req.stop())
+  {
+    if (!this->stopThread)
+    {
+      this->stopThread.reset(new std::thread([this]{
```
chapulina, line 466: Can we use `std::make_shared`?

chapulina (approved):
LGTM, just a couple of minor comments left

src/ServerPrivate.cc
peci1: The individual parts are not `else if`s, but `if`s, so theoretically, the user could ask for more functions in one call. The logic I used was to report the worst result of all, i.e. report `false` as soon as any of the requested functionalities fails. It is set to false at the beginning, but I added these to "remind" or make sure that a failing function is reported. However, it might be wiser to turn the `if`s to `else if`s, as nobody probably wants to simultaneously stop the server and clone/create a world.

src/ServerPrivate.cc
peci1: Sure.

src/ServerPrivate.cc
chapulina: Ah gotcha, I think future-proofing is ok

osrf-triage:

osrf-triage:

PR #1241, opened 2021-12-09
Add wheel slip user command

ivanpauno:
# 🎉 New feature

## Summary

Depends on https://github.com/ignitionrobotics/ign-msgs/pull/205.

Add an user command to reconfigure the wheel slip parameters.
Combined with https://github.com/ignitionrobotics/ros_ign/issues/70, it will allow reconfigurability from ROS similar to what the gazebo_ros_pkgs wheel slip plugin provided.

Ideally, we would have a concept similar to parameters in ignition instead of this.

## Test it

Manual testing:

```bash
# terminal 1, launch trisphere world
ign gazebo trisphere_cycle_wheel_slip.sdf`
# terminal 2, modify the wheel slip parameters of the rear left wheel
ign service -s /world/wheel_slip/wheel_slip --reqtype ignition.msgs.WheelSlipParametersCmd --reptype ignition.msgs.Boolean -r "
model_name: \"trisphere_cycle0\"
link_name: \"wheel_rear_left\"
slip_compliance_lateral: 2
slip_compliance_longitudinal: 2
" --timeout 1000
```

https://github.com/ignitionrobotics/ign-gazebo/pull/1241/commits/80dad6988ab53c7fcb69d45f95e6e4fdb0e9f4e5 can be reverted to see a log each time the wheel slip system detected a cmd.

## Checklist
- [ ] Added tests


ivanpauno:
cc @scpeters.

I'm targetting edifice as I commented [here](https://github.com/ignitionrobotics/ign-msgs/pull/205#issuecomment-990290710), but maybe that was an incorrect choice.
Let me know if I should be targetting a different ignition version.

> Ideally, we would have a concept similar to parameters in ignition instead of this.

I started to take a look of how to do that, but I'm finally be working on a different task for 1 or 2 weeks. I'm still planning to take a deeper look at how to implement parameters in ignition after that.

I had already had implemented this on Monday when we discussed how to implement parameters in Ignition, so I'm opening this for evaluation in case that (ab)using the user commands services to provide this functionality seems like an acceptable approach.
This is much less work that supporting parameters in ignition, and supporting to bridge services between ignition and ROS 2 is something we're going to need anyway.

src/systems/user_commands/UserCommands.cc
```diff
@@ -553,6 +588,14 @@ void UserCommands::Configure(const Entity &_entity,
       &UserCommandsPrivate::VisualService, this->dataPtr.get());
 
   ignmsg << "Material service on [" << visualService << "]" << std::endl;
+
+  // Wheel slip service
+  std::string wheelSlipService
+      {"/world/" + worldName + "/wheel_slip"};
```
ahcorde, line 594:
```suggestion
      {"/world/" + validWorldName + "/wheel_slip"};
```

include/ignition/gazebo/components/WheelSlipCmd.hh
```diff
@@ -0,0 +1,46 @@
...
+ *
+ */
+#ifndef IGNITION_GAZEBO_COMPONENTS_WHEELSLIPCMD_HH_
+#define IGNITION_GAZEBO_COMPONENTS_WHEELSLIPCMD_HH_
+
+#include <ignition/gazebo/config.hh>
+#include <ignition/gazebo/Export.hh>
+#include <ignition/gazebo/components/Component.hh>
+#include <ignition/gazebo/components/Factory.hh>
+#include <ignition/gazebo/components/Serialization.hh>
```
ahcorde, line 24: alphabetize

ahcorde (changes requested):
Check linters: https://github.com/ignitionrobotics/ign-gazebo/runs/4476556123?check_suite_focus=true

include/ignition/gazebo/components/WheelSlipCmd.hh
ivanpauno: sorry, you mean like

```suggestion
#include <ignition/gazebo/components/Component.hh>
#include <ignition/gazebo/components/Factory.hh>
#include <ignition/gazebo/components/Serialization.hh>
#include <ignition/gazebo/config.hh>
#include <ignition/gazebo/Export.hh>
```

When having nested folders, I group items in each subfolder and I only alphaorder each group.
i.e. I put first all header files in the parent folder (in alpha order, in this case the items in `ignition/gazebo` are `config.hh`, `Export.hh`), after that I put the subfolders in alpha order (only `components` here) and the items of the subfolder in alpha order as well (i.e. `Component.hh`, `Factory.hh`, `Serialization.hh`).

I'm also not sure if lowercase letters are ordered after uppercase letters (like ascii code ordering, `a` goes after `Z`), or if I should ignore casing when ordering.
In the first case it should be:

```suggestion
#include <ignition/gazebo/Export.hh>
#include <ignition/gazebo/components/Component.hh>
#include <ignition/gazebo/components/Factory.hh>
#include <ignition/gazebo/components/Serialization.hh>
#include <ignition/gazebo/config.hh>
```

src/systems/user_commands/UserCommands.cc
```diff
@@ -331,6 +333,36 @@ class VisualCommand : public UserCommandBase
...
+  // Documentation inherited
+  public: bool Execute() final;
+
+  /// \brief WheelSlip equality comparision function
+  public: std::function<bool(const msgs::WheelSlipParameters &, const msgs::WheelSlipParameters &)>
+          wheelSlipEql {
+            [](const msgs::WheelSlipParameters &_a, const msgs::WheelSlipParameters &_b) {
+              return
+                _a.name() == _b.name() &&
+                _a.id() == _b.id() &&
```
scpeters, line 355: do we need to compare the `parent_name` and `parent_id` fields also?

ivanpauno:
```suggestion
                _a.id() == _b.id() &&
                _a.parent_id() == _b.parent_id() &&
                _a.parent_name() == _b.parent_name() &&
```

ivanpauno: Fixed in https://github.com/ignitionrobotics/ign-gazebo/pull/1241/commits/906c185084e9ca8970a1dc290fef7c0cdcefd1c4

src/systems/wheel_slip/WheelSlip.cc
```diff
@@ -242,9 +245,30 @@ bool WheelSlipPrivate::Load(const EntityComponentManager &_ecm,
...
 void WheelSlipPrivate::Update(EntityComponentManager &_ecm)
 {
-  for (const auto &linkSurface : this->mapLinkSurfaceParams)
+  for (auto &linkSurface : this->mapLinkSurfaceParams)
   {
-    const auto &params = linkSurface.second;
+    auto &params = linkSurface.second;
+    const auto * wheelSlipCmdComp =
+      _ecm.Component<components::WheelSlipCmd>(linkSurface.first);
+    if (wheelSlipCmdComp) {
```
jacobperron, line 253: Style nit: I don't know about the Ignition style-guide, but it seems like the rest of the code in this file puts open-braces on a new line
```suggestion
    if (wheelSlipCmdComp)
    {
```

src/systems/wheel_slip/WheelSlip.cc
```diff
@@ -242,9 +245,30 @@ bool WheelSlipPrivate::Load(const EntityComponentManager &_ecm,
...
+      const auto & wheelSlipCmdParams = wheelSlipCmdComp->Data();
+      bool changed = math::equal(
```
jacobperron, line 255: Should this condition be negated? Right now it seems to check if the values have **not** changed.

src/systems/wheel_slip/WheelSlip.cc
```diff
@@ -242,9 +245,30 @@ bool WheelSlipPrivate::Load(const EntityComponentManager &_ecm,
...
+          params.slipComplianceLateral,
+          wheelSlipCmdParams.slip_compliance_lateral(),
+          1e-6) &&
+        math::equal(
+          params.slipComplianceLongitudinal,
+          wheelSlipCmdParams.slip_compliance_longitudinal(),
+          1e-6);
+
+      if (changed) {
```
jacobperron, line 264:
```suggestion
      if (changed)
      {
```

jacobperron (commented):
I don't know much about the user command interface, but I've left some feedback below.

jacobperron:
As we discussed offline, I think implementing this as a command (instead of a parameter) makes sense. I guess it ultimately depends on how we differentiate commands vs parameters; until then, it's difficult to say how the wheel slip configuration should be exposed.

What does it look like when users want to interact with the command (e.g. from the command-line)? An example would be nice (even if it is not functional yet). I'm wondering if we really need to bridge this via `ros_ign` or if we should just recommend users make calls to ignition directly?

src/systems/wheel_slip/WheelSlip.cc
ivanpauno: yes, I'm not sure how I didn't notice this.
Thanks for the comment!

src/systems/wheel_slip/WheelSlip.cc
ivanpauno: yes, I'm not sure how I didn't notice this.
Thanks for the comment!

src/systems/wheel_slip/WheelSlip.cc
ivanpauno: Fixed in https://github.com/ignitionrobotics/ign-gazebo/pull/1241/commits/25a11e63ce5b3e1486f62996b937f6bdf868c56a

ivanpauno:
```suggestion
      if (changed)
      {
```

src/systems/wheel_slip/WheelSlip.cc
ivanpauno: Fixed in https://github.com/ignitionrobotics/ign-gazebo/pull/1241/commits/25a11e63ce5b3e1486f62996b937f6bdf868c56a.

src/systems/wheel_slip/WheelSlip.cc
ivanpauno: Fixed in https://github.com/ignitionrobotics/ign-gazebo/pull/1241/commits/d2bce8588293a49cc6cade4cc0f29f8014ff772c.
Updated OP with manual testing instructions.

Reverting https://github.com/ignitionrobotics/ign-gazebo/pull/1241/commits/80dad6988ab53c7fcb69d45f95e6e4fdb0e9f4e5 to have more feedback when testing is useful.

adityapande-1995:
The current implementation of wheel slip plugin in ``gazebo_ros_pkgs`` exposes a pair of parameters to set wheel slip parameters for all wheels : https://github.com/ros-simulation/gazebo_ros_pkgs/blob/dcda27f7ac710bdd6a8aba2692e3b3b3ac1d9c4b/gazebo_plugins/include/gazebo_plugins/gazebo_ros_wheel_slip.hpp#L31-L41, along with the ones used to set values for individual wheels. 

Should we port that functionality here ? Basically equivalent to ``link_name : *``

adityapande-1995 (commented):
The current implementation of wheel slip plugin in gazebo_ros_pkgs exposes a pair of parameters to set wheel slip parameters for all wheels : https://github.com/ros-simulation/gazebo_ros_pkgs/blob/dcda27f7ac710bdd6a8aba2692e3b3b3ac1d9c4b/gazebo_plugins/include/gazebo_plugins/gazebo_ros_wheel_slip.hpp#L31-L41, along with the ones used to set values for individual wheels.

Should we port that functionality here ? Basically equivalent to ``link_name : *`` , which would update values for all wheels in the ecm ? 

Also, could you please add a test case for this ?

ivanpauno:
> Also, could you please add a test case for this ?

Will do.

> Should we port that functionality here ? Basically equivalent to link_name : * , which would update values for all wheels in the ecm ?

I think that's a good idea, using `link_name : *` works for me

ivanpauno:
> Should we port that functionality here ? Basically equivalent to link_name : * , which would update values for all wheels in the ecm ?

Done in https://github.com/ignitionrobotics/ign-gazebo/pull/1241/commits/76e14b7597b5f25928a383bc7fcabd273bef8b1e.

src/systems/user_commands/UserCommands.cc
```diff
@@ -1366,6 +1440,72 @@ bool VisualCommand::Execute()
...
+    auto wheelSlipCmdComp =
+      this->iface->ecm->Component<components::WheelSlipCmd>(linkEntity);
+    if (!wheelSlipCmdComp)
+    {
+      this->iface->ecm->CreateComponent(
+          linkEntity, components::WheelSlipCmd(*wheelSlipMsg));
+    }
+    else
+    {
+      auto state = wheelSlipCmdComp->SetData(*wheelSlipMsg, this->wheelSlipEql) ?
```
adityapande-1995, line 1483: This line can be split into two to make the linter happy

adityapande-1995 (commented):
Thanks for the wildcard addition to link names ! Can we add some simple tests for this PR ? Otherwise looks good.

ivanpauno:
> Thanks for the wildcard addition to link names ! Can we add some simple tests for this PR ? Otherwise looks good.

I was trying to create a test, but I couldn't find a way how.
The wheel slip parameters are internally stored by the WheelSlip plugin.
I don't think there's a way to get a pointer to a system and introspect it, but maybe there is (?).

The other thing I wanted to test is to see if I could check that the generated `WheelSlipCmd`s that the `UserCommands` plugins used to communicate with the slip plugin were correct.
I couldn't make that happen either, as in the same preUpdate() step the user commands system generates the component and the wheel slip system later consumes it.

I will only add a test that shows that the service calls work, but if anyone has a better idea of how to test this I can try it.

src/systems/user_commands/UserCommands.cc
ivanpauno: done in https://github.com/ignitionrobotics/ign-gazebo/pull/1241/commits/05cdbb950a372ff7afaa9a83ac7a65b78f691bfa

adityapande-1995:
I agree, it is not possible here to test that the WheelSlip plugin has set the the correct parameters.  Otherwise looks good to me. THe CI will require a release of ``ign-msgs`` I guess.

adityapande-1995 (approved):

ivanpauno:
@adityapande-1995 I have modified how entities are identified based on feedback [here](https://github.com/ignitionrobotics/ign-msgs/pull/227#discussion_r823947032).
Implementation is in https://github.com/ignitionrobotics/ign-gazebo/pull/1241/commits/2d98ab94a7000ca6173a930ccd2948823879c459.
Could you review again?

src/systems/user_commands/UserCommands.cc
```diff
@@ -356,8 +360,13 @@ class WheelSlipCommand : public UserCommandBase
...
             {
               return
-                _a.model_name() == _b.model_name() &&
-                _a.link_name() == _b.link_name() &&
+                (
+                  (_a.entity().id() != kNullEntity && _a.entity().id() == _b.entity().id()) ||
+                  (
+                    _a.entity().name() == _b.entity().name() &&
+                    _a.entity().type() == _b.entity().type()
+                  )
```
adityapande-1995, line 368: I'm not sure, but maybe entity ids are unique, and this "or" section is redundant; i.e. maybe having the same id implies the name and types match ?

src/systems/user_commands/UserCommands.cc
```diff
@@ -1447,6 +1456,67 @@ WheelSlipCommand::WheelSlipCommand(msgs::WheelSlipParametersCmd *_msg,
...
+        break;
+      default:
+        componentType = "unknown";
+        break;
+    }
+    if (component != nullptr) {
+      return entity;
+    }
+  }
+  ignerr << "Found entity with scoped name [" << _msg.name()
```
adityapande-1995, line 1514: IIUC, this error is thrown when a user requests a change in wheelslip value of say a link (wheel) but actually enters Entity as say ``LIGHT``, instead of ``LINK``. If so, we can just throw an error at the very start saying that WheelSlip command msg is only meant for links, and we can avoid checking all these cases.

adityapande-1995 (commented):
I had 2 minor concerns about the code, otherwise looking good !

src/systems/user_commands/UserCommands.cc
ivanpauno: `id == 0` is used when you don't know the id and you want to provide a name instead.

src/systems/user_commands/UserCommands.cc
ivanpauno: WheelSlip msg is meant for links or models.
The second is used similarly to what before was the `link_name == '*'` option for.

I think that the idea is to reuse the same `Entity` msg everywhere, based on the feedback I got in https://github.com/ignitionrobotics/ign-msgs/pull/227#discussion_r823966128.

You can also use `type == NONE` (i.e. not specified), and only pass a name.

ivanpauno:
The `ign-msgs7` release was done, so I guess now CI should pass.

src/systems/user_commands/UserCommands.cc
```diff
@@ -24,9 +24,11 @@
 #include <ignition/msgs/pose.pb.h>
 #include <ignition/msgs/physics.pb.h>
 #include <ignition/msgs/visual.pb.h>
+#include <ignition/msgs/wheel_slip_parameters.pb.h>
```
scpeters, line 27: I think this needs to be updated to `wheel_slip_parameters_cmd.pb.h`

ivanpauno: Thanks!

I didn't notice the change was needed because after rebuilding the headers with the old name were still available in the install space.

ivanpauno:
The failing jobs seem unrelated.

@chapulina @adityapande-1995 can this be merged?

chapulina:
> The failing jobs seem unrelated.

This test failure looks related:

```
157: [ RUN      ] UserCommandsTest.WheelSlip
157: [31m[Err] [C:\Jenkins\workspace\ign_gazebo-pr-win\ws\ign-common\src\SystemPaths.cc:377] [m[31mUnable to find file with URI [[m[31mC:///Jenkins/workspace/ign_gazebo-pr-win/ws/ign-gazebo/test/worlds/trisphere_cycle_wheel_slip.sdf[m[31m][m[31m
157: [m[31m[Err] [C:\Jenkins\workspace\ign_gazebo-pr-win\ws\ign-common\src\SystemPaths.cc:467] [m[31mCould not resolve file [[m[31mC:/Jenkins/workspace/ign_gazebo-pr-win/ws/ign-gazebo/test/worlds/trisphere_cycle_wheel_slip.sdf[m[31m][m[31m
157: [m[31m[Err] [C:\Jenkins\workspace\ign_gazebo-pr-win\ws\ign-gazebo\src\Server.cc:158] [m[31mFailed to find world [[m[31mC:/Jenkins/workspace/ign_gazebo-pr-win/ws/ign-gazebo/test/worlds/trisphere_cycle_wheel_slip.sdf[m[31m][m[31m
157: [mC:\Jenkins\workspace\ign_gazebo-pr-win\ws\ign-gazebo\test\integration\user_commands.cc(1045): error: Expected: (nullptr) != (ecm), actual: (nullptr) vs NULL
```

I haven't been actively reviewing this, but please feel free to merge once CI is happy and it has an approval. Feel free to ping me if you need another set of eyes after the first approval.

ivanpauno:
> This test failure looks related:

Sorry, I didn't pay attention to the windows job.
Now it's failing again, but the new problem actually seems unrelated.

https://github.com/ignitionrobotics/ign-gazebo/pull/1241/commits/19606cdbb7a1cf9ebaebb56c8e943c0dfa7cc44f should've fixed the previous issue (all the other tests in that file are being skipped on Windows).

chapulina:
@ivanpauno @jacobperron @scpeters , heads up that Edifice (`ign-gazebo5`) will EOL next week. Do you think this will get merged before then?

ivanpauno:
> @ivanpauno @jacobperron @scpeters , heads up that Edifice (ign-gazebo5) will EOL next week. Do you think this will get merged before then?

It would be good to have it merged before, so we only need to forward port to newer releases.
The PR is ready and CI was passing (I updated the PR with the base branch again, but it should still pass :crossed_fingers:), so it only needs :heavy_check_mark: and to get merged.

@adityapande-1995 could you take a look again?

scpeters:
CI looks fine to me

jacobperron (approved):

osrf-triage:

Issue #797, opened 2021-04-29
Gazebo system plugins use EntityComponentManager::EachNew in PreUpdate


adlarkin:
## Environment
* OS Version: Ubuntu Bionic
* Source build, Citadel

## Description

#795 points out that `EachNew` should not be used in system plugin's `PreUpdate` function, because entities are created during `PreUpdate`, so depending on the order in which systems are run, calling `EachNew` in a system's `PreUpdate` method may result in missing newly created entities. However, there are several system plugins in the `ign-gazebo` repository that call `EachNew` in `PreUpdate`:
* Buoyancy
* Altimeter
* Imu
* Contact
* Touch
* LogicalCamera
* AirPressure
* LogRecord
* Magnetometer

These systems should be updated so that `EachNew` is not being called in `PreUpdate`, and we should also probably add/update tests for these systems as needed that spawn entities after startup.

## Steps to reproduce

For a concrete example of how/why this is problematic, launch [sensors.sdf](https://github.com/ignitionrobotics/ign-gazebo/blob/ignition-gazebo3_3.8.0/examples/worlds/sensors.sdf), which has the `UserCommands` system come after the `Imu` system. Then, add an imu sensor from the command line after startup. Since this is processed through `UserCommands`, which has its `PreUpdate` run **after** `Imu`'s `PreUpdate`, the `Imu` will not register this sensor (thanks to @chapulina for providing this test case).

Here are the steps to test/reproduce this scenario:
1. Start a running simulation using `sensors.sdf`:
```
ign gazebo sensors.sdf -r
```
2. Create an imu sensor from the command line
```
ign service -s /world/sensors/create --reqtype ignition.msgs.EntityFactory --reptype ignition.msgs.Boolean --timeout 300 --req 'sdf: ''"<?xml version=\"1.0\" ?>''<sdf version=\"1.6\">''<model name=\"spawned_model\">''<link name=\"link\"><sensor name=\"imu\" type=\"imu\"><topic>test_imu</topic></sensor></link>''</model>''</sdf>" ''pose: {position: {z: 10}}'
```

You'll then see the following error message:
```
[Err] [Imu.cc:224] Failed to update IMU: 19. Entity not found.
```

adlarkin:
The following issue could also be related: https://github.com/ignitionrobotics/ign-gazebo/issues/83

chapulina:
* #1281 tackles these sensors:
    * `AirPressure`
    * `Altimeter`
[...]

PR #1281, opened 2022-01-11
Remove EachNew calls from sensor PreUpdates

chapulina:
# 🦟 Bug fix

* Part of https://github.com/ignitionrobotics/ign-gazebo/issues/797

## Summary

Calling `EachNew` in `PreUpdate` is unreliable because new entities may be created during another system's `PreUpdate` and then you miss it. See #797 for a detailed description of the issue.

This PR tackles only non-rendering sensors. I'm using the same approach as in #1248 . This is the new flow:

1. `SdfEntityCreator` creates the sensor entity and fills some components from SDF
2. `PreUpdate`: the first `PreUpdate` doesn't do anything
3. `Update`: `Physics` populates the components created by `SdfEntityCreator`
4. `PostUpdate`: The sensor system does a few things:
    1. Creates new `ign-sensors` objects for new sensors
    2. Updates all sensors with the data from the components that were filled by physics
    3. Removes sensors that need to be removed
1. The next `PreUpdate` creates new components for the new sensors created during the previous `PostUpdate`, which is only `SensorTopic`

There's a slight change in behaviour, which is that a sensor won't have a `SensorTopic` component in the first iteration. This means that any `Each` call that includes that component will only be called from the 2nd iteration after the component is created. Since that component is not critical to the sensor's functioning and I've only seen it being used by the GUI, I think this is a safe change. See the consequences on the updated tests.

## Checklist
- [x] Added tests

### DIFF
--- a/src/systems/air_pressure/AirPressure.cc
+++ b/src/systems/air_pressure/AirPressure.cc
@@ -21,6 +21,7 @@
 
 #include <string>
 #include <unordered_map>
+#include <unordered_set>
 #include <utility>
 
 #include <ignition/plugin/Register.hh>
--- a/src/systems/air_pressure/AirPressure.cc
+++ b/src/systems/air_pressure/AirPressure.cc
@@ -49,30 +49,35 @@ using namespace systems;
 /// \brief Private AirPressure data class.
 class ignition::gazebo::systems::AirPressurePrivate
 {
-  /// \brief A map of air pressure entity to its vertical reference
+  /// \brief A map of air pressure entity to its sensor
   public: std::unordered_map<Entity,
       std::unique_ptr<sensors::AirPressureSensor>> entitySensorMap;
 
   /// \brief Ign-sensors sensor factory for creating sensors
   public: sensors::SensorFactory sensorFactory;
 
+  /// \brief Keep list of sensors that were created during the previous
+  /// `PostUpdate`, so that components can be created during the next
+  /// `PreUpdate`.
+  public: std::unordered_set<Entity> newSensors;
+
   /// True if the rendering component is initialized
   public: bool initialized = false;
 
   /// \brief Create sensor
-  /// \param[in] _ecm Mutable reference to ECM.
+  /// \param[in] _ecm Immutable reference to ECM.
   /// \param[in] _entity Entity of the IMU
   /// \param[in] _airPressure AirPressureSensor component.
   /// \param[in] _parent Parent entity component.
   public: void AddAirPressure(
-    EntityComponentManager &_ecm,
+    const EntityComponentManager &_ecm,
     const Entity _entity,
     const components::AirPressureSensor *_airPressure,
     const components::ParentEntity *_parent);
 
   /// \brief Create air pressure sensor
-  /// \param[in] _ecm Mutable reference to ECM.
-  public: void CreateAirPressureEntities(EntityComponentManager &_ecm);
+  /// \param[in] _ecm Imutable reference to ECM.
+  public: void CreateAirPressureEntities(const EntityComponentManager &_ecm);
 
   /// \brief Update air pressure sensor data based on physics data
   /// \param[in] _ecm Immutable reference to ECM.
@@ -98,7 +103,21 @@ void AirPressure::PreUpdate(const UpdateInfo &/*_info*/,
     EntityComponentManager &_ecm)
 {
   IGN_PROFILE("AirPressure::PreUpdate");
-  this->dataPtr->CreateAirPressureEntities(_ecm);
+
+  // Create components
+  for (auto entity : this->dataPtr->newSensors)
+  {
+    auto it = this->dataPtr->entitySensorMap.find(entity);
+    if (it == this->dataPtr->entitySensorMap.end())
+    {
+      ignerr << "Entity [" << entity
+             << "] isn't in sensor map, this shouldn't happen." << std::endl;
+      continue;
+    }
+    // Set topic
+    _ecm.CreateComponent(entity, components::SensorTopic(it->second->Topic()));
+  }
+  this->dataPtr->newSensors.clear();
 }
 
 //////////////////////////////////////////////////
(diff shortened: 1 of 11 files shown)


adlarkin:
It looks like there's a [new warning](https://build.osrfoundation.org/job/ignition_gazebo-ci-pr_any-ubuntu_auto-amd64/7818/cmake/new/) in Ubuntu CI:
```
FindOpenGL.cmake:275

Policy CMP0072 is not set: FindOpenGL prefers GLVND by default when...
```

Should this be addressed in a separate PR?

chapulina:
> It looks like there's a new warning in Ubuntu CI:

That's been happening since CI has migrated to Focal, it's being tracked here: https://github.com/ignitionrobotics/ign-rendering/issues/360. I mentioned in https://github.com/ignitionrobotics/ign-gazebo/pull/1274#issue-1095880182 that I think an `ign-gui3` release will fix it, and I was waiting on https://github.com/ignitionrobotics/ign-gui/pull/335 to make an `ign-gui3` release.

src/systems/air_pressure/AirPressure.cc
```diff
@@ -49,30 +50,35 @@ using namespace systems;
...
-    EntityComponentManager &_ecm,
+    const EntityComponentManager &_ecm,
     const Entity _entity,
     const components::AirPressureSensor *_airPressure,
     const components::ParentEntity *_parent);
 
   /// \brief Create air pressure sensor
-  /// \param[in] _ecm Mutable reference to ECM.
-  public: void CreateAirPressureEntities(EntityComponentManager &_ecm);
+  /// \param[in] _ecm Imutable reference to ECM.
```
adlarkin, line 80:
```suggestion
  /// \param[in] _ecm Immutable reference to ECM.
```

src/systems/altimeter/Altimeter.cc
```diff
@@ -51,30 +52,35 @@ using namespace systems;
...
+    const EntityComponentManager &_ecm,
     const Entity _entity,
     const components::Altimeter *_altimeter,
     const components::ParentEntity *_parent);
 
   /// \brief Create altimeter sensor
-  /// \param[in] _ecm Mutable reference to ECM.
-  public: void CreateAltimeterEntities(EntityComponentManager &_ecm);
+  /// \param[in] _ecm Immutable reference to ECM.
+  public: void CreateAltimeterEntities(const EntityComponentManager &_ecm);
```
adlarkin, line 83:
```suggestion
  public: void CreateAltimeterSensors(const EntityComponentManager &_ecm);
```

It looks like this method can be renamed for clarity. Even the docs say this this method should create an altimeter sensor (no an entity), so the method name should also reflect this.

src/systems/air_pressure/AirPressure.cc
```diff
@@ -49,30 +50,35 @@ using namespace systems;
...
+  public: void CreateAirPressureEntities(const EntityComponentManager &_ecm);
```
adlarkin, line 81:
```suggestion
  public: void CreateAirPressureSensors(const EntityComponentManager &_ecm);
```

src/systems/imu/Imu.cc
```diff
@@ -64,21 +70,21 @@ class ignition::gazebo::systems::ImuPrivate
   /// True if the rendering component is initialized
   public: bool initialized = false;
 
-  /// \brief Create IMU sensor
-  /// \param[in] _ecm Mutable reference to ECM.
-  public: void CreateImuEntities(EntityComponentManager &_ecm);
+  /// \brief Create IMU sensors in ign-sensors
+  /// \param[in] _ecm Immutable reference to ECM.
+  public: void CreateImuEntities(const EntityComponentManager &_ecm);
```
adlarkin, line 75:
```suggestion
  public: void CreateImuSensors(const EntityComponentManager &_ecm);
```

src/systems/logical_camera/LogicalCamera.cc
```diff
@@ -59,23 +60,28 @@ class ignition::gazebo::systems::LogicalCameraPrivate
...
+    const EntityComponentManager &_ecm,
     const Entity _entity,
     const components::LogicalCamera *_logicalCamera,
     const components::ParentEntity *_parent);
 
   /// \brief Create logicalCamera sensor
-  /// \param[in] _ecm Mutable reference to ECM.
-  public: void CreateLogicalCameraEntities(EntityComponentManager &_ecm);
+  /// \param[in] _ecm Immutable reference to ECM.
+  public: void CreateLogicalCameraEntities(const EntityComponentManager &_ecm);
```
adlarkin, line 84:
```suggestion
  public: void CreateLogicalCameraSensors(const EntityComponentManager &_ecm);
```

src/systems/magnetometer/Magnetometer.cc
```diff
@@ -56,25 +57,30 @@ class ignition::gazebo::systems::MagnetometerPrivate
...
     const Entity _entity,
     const components::Magnetometer *_magnetometer,
     const components::MagneticField *_worldField,
     const components::ParentEntity *_parent);
 
   /// \brief Create magnetometer sensor
-  /// \param[in] _ecm Mutable reference to ECM.
-  public: void CreateMagnetometerEntities(EntityComponentManager &_ecm);
+  /// \param[in] _ecm Immutable reference to ECM.
+  public: void CreateMagnetometerEntities(const EntityComponentManager &_ecm);
```
adlarkin, line 83:
```suggestion
  public: void CreateMagnetometerSensors(const EntityComponentManager &_ecm);
```

adlarkin (approved):
LGTM! There are a few methods that can be renamed for clarity, which I highlighted below.

src/systems/altimeter/Altimeter.cc
chapulina: Agreed! I was actually meaning to change it to `CreateSensors` on all of them, so I'll do that instead. These plugins are so similar and there's so much duplication that there's value in reducing the diff so it's easier to compare them side-by-side.

I'll only address the `Create` functions in this PR not to scope creep, but I think we could do the same to `Add`, `Update` and `Remove`. In fact, I'd like to move a lot of this duplication to a templated helper class, but that's for the future :grimacing:

chapulina: 75c78ad96d50849e61d03afb9e800a7f8daee6cd

osrf-triage:

osrf-triage:

PR #1249, opened 2021-12-14
Support world reset

azeey:
# 🎉 New feature

Toward #1107 

## Summary
This adds a new phase in which systems can implement special Reset functionality. The approach taken is to make an initial copy of the ECM after the entities have been created. When reset is requested, a diff is computed between the current ECM and the initial copy. This is necessary to ensure that entities that were added later on are marked as removed after reset. Likewise, entities that were removed later on are marked as newly added after reset.

The physics system has been updated to make use of the Reset phase. The changes were nontrivial because we keep a lot of state in the physics system to help with performance. Hopefully, updating other systems will be more straightforward.

The Reset API allows systems to make changes to the ECM. This can be used to, for example, change the poses of objects in the scene at every reset without having to reload the the world. As an example, I have created a small and very trivial joint position randomizer.

There things that still need to be cleaned up and tests that need to be added, but I wanted open this as a draft PR to get feedback on the approach.
 
## Checklist
- [x] Signed all commits for DCO
- [ ] Added tests
- [ ] Added example and/or tutorial
- [ ] Updated documentation (as needed)
- [ ] Updated migration guide (as needed)
- [ ] `codecheck` passed (See [contributing](https://ignitionrobotics.org/docs/all/contributing#contributing-code))
- [ ] All tests passed (See [test coverage](https://ignitionrobotics.org/docs/all/contributing#test-coverage))
- [ ] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Aignitionrobotics+repo%3Aosrf%2Fsdformat+archived%3Afalse+) to support the maintainers

**Note to maintainers**: Remember to use **Squash-Merge**

PR #1300, opened 2022-01-18
Mark Component::Clone as const

mjcarroll:
Split from #1249.  Mark `Clone` method as `const`.

This is likely not back-portable as it breaks API.  Any components that implement a custom `Clone` will fail to compile.  There are none in the main `ign-gazebo` codebase, but there is a possibility of external users doing so.


chapulina (approved):
LGTM. It may be worth dropping a note on the migration guide.

mjcarroll:
> It may be worth dropping a note on the migration guide.

73d213b

mjcarroll:

PR #1301, opened 2022-01-19
Documentation and more examples for #1297

chapulina:
* Targeted at #1297 

While reviewing #1297 I was trying to make sense of all the reference frames used throughout the plugin, so I started documenting them as I believe they currently are. Then I added a couple of examples and I think some use cases are still not working well.

I added 2 models with different offsets and volume distributions which I believe should all be equivalent and neutrally buoyant, but see how the model with a collision split in 2 bounces around. I think there's still some offset that's not properly taken into account (assuming I didn't mess anything up setting up the models). 

collision_offset

I haven't dug deeply into the math yet to see where the issue could be coming from.

### DIFF
--- a/examples/worlds/graded_buoyancy.sdf
+++ b/examples/worlds/graded_buoyancy.sdf
@@ -43,6 +43,8 @@
       <!-- enable by top level model name -->
       <enable>balloon_lighter_than_air</enable>
       <enable>box_neutral_buoyancy</enable>
+      <enable>box_neutral_buoyancy_collision_offset</enable>
+      <enable>multi_collision_neutral_buoyancy</enable>
       <enable>box_negative_buoyancy</enable>
     </plugin>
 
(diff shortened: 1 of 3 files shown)


arjo129 (approved):
Thanks for the suggestions and spotting this. I think we aren't handling multiple geometry in collisions correctly. I think this needs to be ticketed somewhere before we forget.

chapulina:
> I think we aren't handling multiple geometry in collisions correctly. I think this needs to be ticketed somewhere before we forget.

I pushed a fix on bad001b, mind giving it another try? We were accumulating the forces when we calculated the torque. I think it should be working now.

arjo129 (approved):
Tested and looks good to me. I'll add a similar test case. Thanks for catching this. Seems like a major oversight on my part :man_facepalming: .

PR #1302, opened 2022-01-20
Buoyancy: fix center of volume's reference frame

chapulina:
# 🦟 Bug fix

## Summary

While working with a model that uses NED / FSK orientation, I noticed that the center of volume calculation on the buoyancy plugin was using the wrong reference frame. According to the documentation and the usage, the reference frame is expected to be the parent link. But the value provided was being calculated in the world frame. That's evident when using a model that's oriented with Z down.

## Test it

I added a test world with 4 models that are supposed to be neutrally buoyant. Without this fix, the model that has a CoV offset and starts with a non-zero orientation isn't stable:

cov_1

## Checklist
- [x] Added tests


arjo129 (approved):
Thanks for catching this. Tested before and after the fix and can confirm this fix indeed does solve the problem at hand. Same problem affects graded buoyancy :skull_and_crossbones: I'll go take alook and fix it.

osrf-triage:

PR #1307, opened 2022-01-24
Logs a warning if a buoyancy mode is not clearly specified.

### LINKED ISSUE #1305: Fix graded buoyancy reference frame
## Environment
* OS Version: All
* Source or binary build: Fortress and Garden

## Description
This issue was actually spotted by @chapulina in #1302 for uniform buoyancy but it effects graded buoyancy as well. Basically the calculation for graded buoyancy also does not handle rotated link poses correctly. To reproduce run this sdf file: https://gist.github.com/arjo129/5d93669a05ddc9fc79c986f29f1ae925

unstablebox

Essentially all boxes should be stable, but the one with a rotated link pose is not.

arjo129:
I had originally started this branch to fix #1305. Turns out that #1305 was not actually a problem, still I decided to take the opportunity to imporve the logging so that if a user fails to specify `<uniform_fluid_density>` or `<graded_density>` then  the user will be warned. Furthermore, I add a test for graded density mode that shows that #1305 is not an issue.


caguero (commented):
Looks good to me. Just waiting for fixing the conflicts...

caguero (approved):

azeey:
I think a test introduced in this PR is failing on windows CI. Do you mind using `IGN_UTILS_TEST_DISABLED_ON_WIN32` on the test to disable it if it can't be fixed?

arjo129:
See #1368

PR #1335, opened 2022-02-12
Extend ShaderParam system to support loading different shader languages

iche033:
# 🎉 New feature

## Summary

Implemented proposed idea from: https://github.com/ignitionrobotics/ign-gazebo/pull/1275#issuecomment-1032141148

The idea is to let users specify different shader files to use. The `<shader>` sdf element now takes an optional `language` attribute. On macOS, the system will use `metal` shaders over `glsl` if specified. 

The [deformable_sphere](https://app.ignitionrobotics.org/OpenRobotics/fuel/models/deformable_sphere) and [waves](https://app.ignitionrobotics.org/OpenRobotics/fuel/models/Waves) models on Fuel have been updated to include metal shaders.

## Test it

Remove the `deformable_sphere` and `waves` models in your fuel cache (`~/.ignition/fuel/fuel.ignitionrobotics.org/openrobotics/models/`) then run  ign-gazebo with `shaders_param.sdf` world. You should get the same output as the screenshot shown in https://github.com/ignitionrobotics/ign-gazebo/pull/1310

## Checklist
- [ ] Added tests


srmainwaring:
> I integrated your shaders into the existing models. Hope I didn't mess anything up.

Thanks @iche033! - I'm sure you haven't. I'll give them a test and let you know.

Working on porting my more general wave model to Ignition at present - not quite straightforward as it updates the `Ogre::v1::Mesh` and that route is no longer available. I looked at the `Ogre2DynamicRenderable` but don't think it quite does what I need - also no support for tangent vectors which are needed for bump mapping. I'm looking at the Ogre2 DynamicGeometry example with a view to using the approach suggested there.

A couple of things I noticed when trying to map shader materials onto a custom mesh:

- the `ShaderParams` doesn't seem to supply the normals to the vertex shader (just the positions and uv0 texture coords). I can add a VES_NORMAL param to the inputs but it is zeroed.
- it doesn't look like you can use the material defined by `ShaderParams` in another system plugin declared in the same visual.

I'll post an issue with some simplified examples to illustrate.

src/systems/shader_param/ShaderParam.cc
```diff
@@ -65,11 +66,21 @@ class ignition::gazebo::systems::ShaderParamPrivate
...
-  /// \brief Path to vertex shader
-  public: std::string vertexShaderUri;
+  /// \brief Data structure for storing shader files uri
+  public: class ShaderUri
+  {
+    /// \brief Shader language: glsl or metal
+    public: std::string language;
+
+    /// \brief Path to vertex shader
+    public: std::string vertexShaderUri;
```
nkoenig, line 76: Do you want to use the URI class from ignition common?

nkoenig: Looks like this might be a a local filename, if so maybe use `std::filename::path`?

nkoenig (approved):
Looks good to me. Not sure if [srmainwaring](https://github.com/srmainwaring) has additional comments.

iche033:
@srmainwaring  oops I missed your comment

> * the ShaderParams doesn't seem to supply the normals to the vertex shader (just the positions and uv0 texture coords). I can add a VES_NORMAL param to the inputs but it is zeroed.

yep that sounds good to me.

> * it doesn't look like you can use the material defined by ShaderParams in another system plugin declared in the same visual.

hmm like the entire ogre material or specific texture? I would have expected ogre to add the texture to its resource path so that it would be available to others but that also depends on the order in which the systems are loaded.

srmainwaring:
> it doesn't look like you can use the material defined by ShaderParams in another system plugin declared in the same visual.

On this second point it turns out that only one plugin per visual is processed at present, and that is which ever one is declared last is loaded and the preceding ones are ignored. The protobuf Visual message allows for multiple plugins, but the server dispatching the message does not, and there may also be an assumption of one plugin on the GUI side as well. 

I can work around this for the moment by replicating the ShaderParam plugin logic in my plugin, so it's probably best to submit support for multiple plugins as a feature request.

On the first point, addition of normals and tangent attributes, I have some more work to get that working for my example, so again that would best be included as a follow up.

iche033:
> On this second point it turns out that only one plugin per visual is processed at present, and that is which ever one is declared last is loaded and the preceding ones are ignored. The protobuf Visual message allows for multiple plugins, but the server dispatching the message does not, and there may also be an assumption of one plugin on the GUI side as well.

ok you're right. I've ticketed an issue for this. https://github.com/ignitionrobotics/ign-gazebo/issues/1363

iche033:
I'll merge this first. We can address other issues as they come up.

osrf-triage:

https://community.gazebosim.org/t/new-ignition-releases-2022-03-25-fortress-edifice-citadel/1343/1

PR #1349, opened 2022-02-18
Add parameter to TrajectoryFollower stop rotation when bearing is reached

iche033:
# 🎉 New feature

## Summary

Added parameter `<zero_vel_on_bearing_reached>`. When set to true, it forces angular velocity to be zero when the target bearing is reached.  This is intended to remove any residue angular velocity of the link and prevent overshooting. 

## Test it

Test with trajectory_follower.sdf world. `box_02` now has this parameter set and uses high torque for faster rotation and should not overshoot

```
ign gazebo -v 4 ./src/ign-gazebo/examples/worlds/trajectory_follower.sdf
```

## Checklist
- [x] Added tests

### DIFF
--- a/src/systems/trajectory_follower/TrajectoryFollower.cc
+++ b/src/systems/trajectory_follower/TrajectoryFollower.cc
@@ -258,7 +258,7 @@ void TrajectoryFollowerPrivate::Load(const EntityComponentManager &_ecm,
 
   // Parse the optional <zero_vel_on_bearing_reached> element.
   if (_sdf->HasElement("zero_vel_on_bearing_reached"))
-    this->forceZeroAngVel = _sdf->Get<double>("zero_vel_on_bearing_reached");
+    this->forceZeroAngVel = _sdf->Get<bool>("zero_vel_on_bearing_reached");
 
   // Parse the optional <topic> element.
   this->topic = "/model/" + this->model.Name(_ecm) +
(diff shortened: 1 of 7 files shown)


src/systems/trajectory_follower/TrajectoryFollower.cc
```diff
@@ -376,11 +387,31 @@ void TrajectoryFollower::PreUpdate(
   {
     forceWorld = (*comPose).Rot().RotateVector(
       ignition::math::Vector3d(this->dataPtr->forceToApply, 0, 0));
-  }
 
+    // force angular velocity to be zero when bearing is reached
+    if (this->dataPtr->forceZeroAngVel &&
+        math::equal(std::abs(bearing.Degree()), 0.0, 1.0))
```
caguero, line 393: Do we need the second expression `math::equal(std::abs(bearing.Degree()), 0.0, 1.0)`?

We're supposed to be within the acceptable range bearing. If we need higher bearing accuracy, I think we should change the parameter `bearingTolerance`.

iche033: The behavior I found without this check is that as soon as the vehicle moves forward 1 iteration, its bearing immediately becomes outside of `bearingTolerance`. So it constantly switched back and forth between rotation and moving forward. The problem was that removing the `AngularVelocityCmd` component is found to be very expensive and this locks up simulation. With this check, I still do notice a slight pause in simulation when the component is being removed but it only happens when occasionally when bearing needs to be corrected.

iche033: I tweaked the 2nd tol check to be a percentage of `bearingTolerance`. [aa8cf68](https://github.com/ignitionrobotics/ign-gazebo/pull/1349/commits/aa8cf68ec8894ec110fa0df9c551d6d9ba755d30)

caguero (approved):
Looks good to me.

osrf-triage:

PR #1340, opened 2022-02-15
Refactor System functionality into SystemManager

mjcarroll:
SimulationRunner is getting long/crowded, so I'm refactoring out a bit of the System-oriented functionality into it's own `Manager` class as a peer to things like `EventsManager` and `LevelManager`.  This should make it a bit easier to test, as well as reduce the clutter in `SimulationRunner`

Note that this moves the functionality into internal headers, so it should not have an impact on our public API/ABI.

PR #1383, opened 2022-03-10
Prevent hanging when world has only non-world plugins

chapulina:
# 🦟 Bug fix

* Fixes the main issue in https://github.com/ignitionrobotics/ign-gazebo/issues/796

## Summary

When loading a world without any world plugins, we load a predefined set of plugins by default, [see this tutorial](https://ignitionrobotics.org/api/gazebo/6.7/server_config.html). The problem is that if a world has some non-world plugin, such as a model plugin, then we were not loading those default world plugins. Which causes the UI to "hang", because there's no `SceneBroadcaster`, and the simulation not to do anything, because there's no `Physics`.

This PR makes sure that we load the default world plugins even if there are other plugins loaded.

### Try it out

Before this PR, loading the added test world leaves an empty screen. With the PR, the UI shows the model and it moves.

`ign gazebo test/worlds/model_plugin_only.sdf`

### Target branch

I targeted `ign-gazebo6` to take advantage of the new `SystemManager` class introduced in #1340. This bug is also present in Citadel, so I'll think about a way of backporting the fix. I looked into backporting #1340 but it may depend on other non-backportable PRs.

### Implementation detail

I noticed that the `configureEntity` variable wasn't being used. I took a quick look into the "reset" PRs, and didn't see how it was meant to be used. I changed its name for clarity and also I'm making sure it's used now. Let me know if this conflicts with any of the reset work, @mjcarroll @azeey .

## Checklist
- [x] Added tests

### DIFF
--- a/src/SimulationRunner.cc
+++ b/src/SimulationRunner.cc
@@ -174,7 +174,7 @@ SimulationRunner::SimulationRunner(const sdf::World *_world,
 
   // If we have reached this point and no world systems have been loaded, then
   // load a default set of systems.
-  if (this->systemMgr->SystemsByEntity(
+  if (this->systemMgr->TotalByEntity(
       worldEntity(this->entityCompMgr)).empty())
   {
     ignmsg << "No systems loaded from SDF, loading defaults" << std::endl;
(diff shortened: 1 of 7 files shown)


ahcorde (approved):

osrf-triage:

PR #1392, opened 2022-03-15
SceneBroadcaster: only send changed state information for change events

### LINKED ISSUE #1358: Delay and time jump in object position update when using the transform tool
## Environment
* OS Version: Ubuntu 20.04
* Source or binary build?
source: ign-gazebo6 (91a79da40) (and binary 3.12.0 for comparison)

## Description
* Expected behavior: After moving an object using the transform tool, I expect the object to assume its new position and simulation to resume immediately. This is the case in citadel (3.12.0) as shown in the video:

https://user-images.githubusercontent.com/206116/155427057-455769a2-bdf3-4da3-849b-e93f845f3acc.mp4

* Actual behavior: After using the transform tool, simulation seems to hang for a second and then jump in time so that the intermediate state of the object is not displayed on the GUI. In the video below, the box snaps to the ground after it hangs in the air for a second or two.

https://user-images.githubusercontent.com/206116/155427309-2e0a7c53-deec-4915-a32d-d176fcecc1c9.mp4

Interestingly, the behavior is different when I try it on ign-gazebo 6.5.0 binaries. The small delay is still there, but it doesn't seem to jump in time.

https://user-images.githubusercontent.com/206116/155431620-1091a8fb-89c9-4743-8380-60733572683f.mp4

adlarkin:
Signed-off-by: Ashton Larkin <42042756+adlarkin@users.noreply.github.com>

# 🦟 Bug fix

Fixes #1358

## Summary
The `SceneBroadcaster` is [sending the full ECM state whenever there is a change event](https://github.com/ignitionrobotics/ign-gazebo/blob/e0d88f7dee7ffee39160dd4f203041a4d3b13870/src/systems/scene_broadcaster/SceneBroadcaster.cc#L332-L336), resulting in a performance hit because of excessive serialization/de-serialization between the server and GUI (by sharing the full state, entities/components that have not changed are still serialized, which is unnecessary). This PR takes the approach proposed in https://github.com/ignitionrobotics/ign-gazebo/issues/1358#issuecomment-1063328739 to send only the changed state from server -> GUI whenever there is a change event.

I have added test coverage to the `SceneBroadcaster` integration test to ensure that the ECM that receives the changed state (which in theory would be the GUI ECM) is properly updated to reflect the changes without losing unchanged state information. I also took some time to look at how `EntityComponentManager::SetState` is implemented ([this is how the GUI's ECM is updated whenever it receives state information from the server ECM](https://github.com/ignitionrobotics/ign-gazebo/blob/ignition-gazebo6_6.7.0/src/gui/GuiRunner.cc#L302)), and it looks to me like there shouldn't be any issues with only sending the changed state instead of the full state for change events (`EntityComponentManager::SetState` doesn't reset the ECM's current state; instead, it "applies" the incoming state to the current ECM state). It would probably be worth testing some GUI plugins before merging this to make sure that this change doesn't produce unexpected behavior changes.

## Checklist
- [x] Added tests


test/integration/scene_broadcaster_system.cc
```diff
@@ -655,21 +658,61 @@ TEST_P(SceneBroadcasterTest, IGN_UTILS_TEST_DISABLED_ON_WIN32(RemovedComponent))
             return true;
           });
       }
+      // add a component to an entity
```
chapulina, line 661: Can we also check changing a component's value and marking it as `PeriodicChange` or `OneTimeChange`? I think this is the trickiest use case to support.

chapulina (commented):
LGTM, I'd just like to see more testing with changed component values.

It's possible that this will break current users who aren't setting the changed state correctly.

srmainwaring:
@adlarkin great fix thanks. I've tested this on macOS / Metal by cherry-picking into main and it completely solves an issue I had with running a `set_pose` user command at a high update rate (#1403).

test/integration/scene_broadcaster_system.cc
adlarkin: Yeah, that's a great point - I added this coverage in https://github.com/ignitionrobotics/ign-gazebo/pull/1392/commits/92cb2f1759d765b9608b526f5ce60e43cb0d4505. In order to process `PeriodicChange` components, [the server has to run unpaused](https://github.com/ignitionrobotics/ign-gazebo/blob/92cb2f1759d765b9608b526f5ce60e43cb0d4505/src/systems/scene_broadcaster/SceneBroadcaster.cc#L342-L343). I found that since we are adding/removing pose components, running the server unpaused results in a crash due to the following code that's part of the physics system update: https://github.com/ignitionrobotics/ign-gazebo/blob/ignition-gazebo6_6.7.0/src/systems/physics/Physics.cc#L2556-L2568

The problem is that physics always expects a pose to exist, but we are removing/re-adding it in this test. So, for the update after removing the pose component, `_ecm.Component<components::Pose>(_model)` gives a `nullptr` that is later dereferenced. I tried to get around this by adding/removing a component that doesn't seem to be used by physics (for example, [`components::Transparency`](https://github.com/ignitionrobotics/ign-gazebo/blob/ign-gazebo6/include/ignition/gazebo/components/Transparency.hh)), but I was still getting a segfault. So, I created another test file that uses the `SceneBroadcaster` plugin only and not `Physics`. This test file is used for the particular test that adds/removes components and checks for `SceneBroadcaster` updates (this was done in https://github.com/ignitionrobotics/ign-gazebo/pull/1392/commits/92cb2f1759d765b9608b526f5ce60e43cb0d4505). I'm not sure if there's a better solution than the approach I took, but I spent quite a bit of time trying to debug the segfaults and found that this was the easiest way to avoid this problem for now.

adlarkin: As a follow-up, I think that the approach I took in my previous comment of only using `SceneBroadcaster` and not `Physics` for the test is okay since the test is meant to only check update functionality from `SceneBroadcaster`. However, it seems like there may be some places in the physics system where we don't check for potential `nullptr`s because of removed components, and that's the point I was trying to get across in my previous comment. It seems like the segfault issue I was having is outside the scope of this PR, so perhaps we should ticket an issue if this is an actual problem (I think it's probably fair to assume certain components like `Pose` should never be removed from entities, but I wanted to bring this up just to make sure it's not something we have been overlooking).

adlarkin:
It looks like the last part of the new test I added is a little flaky on github actions. I believe it's due to slower machines on GitHub actions. I tried to increase the wait times for messages to be received via `ign-transport` in https://github.com/ignitionrobotics/ign-gazebo/pull/1392/commits/e3c38febff1da463021715e39507f022d6ea24d9, and it seemed to fix actions on focal, but not on bionic. Jenkins CI is passing on Ubuntu though, and the tests pass locally on my machine.

chapulina (approved):
Thanks, the tests look good to me!

> It looks like the last part of the new test I added is a little flaky on github actions.

I tried to make the scene broadcaster publish more often in e8dc431f2a6fc6709eedb830bc6c0f913e75c281. Let's see how it goes, feel free to merge if Bionic passes.

osrf-triage:

Issue #191, opened 2020-06-05
System inspector


chapulina:
Similar to how we have a [component inspector](https://ignitionrobotics.org/docs/citadel/manipulating_models#component-inspector) widget, it would also be great to have a system inspector so that:

* Users can see all systems that are currently loaded and which entities they're attached to
* Users can see metadata from the system, such as the location of its library on disk
* Users can see all parameters for each system

Once #83 is solved, the widget could even display what components a system is supposed to read/write.

---

On Gazebo-classic, basic plugin information is displayed on the left panel:

![plugins_cessna](https://user-images.githubusercontent.com/5751272/83930041-85a7b300-a74a-11ea-809a-a83c63201cf7.png)

chapulina:
Done in #1404

PR #1400, opened 2022-03-19
Component inspector: refactor Pose3d C++ code into a separate class

chapulina:
# 🎉 New feature

Part of
* #191 

## Summary

While working on the system inspector, I thought it would be good to adopt some of the refactoring done for the model editor in #1231 to keep the code organized. That work is happening on the `chapulina/6/system_inspector` branch.

This PR ports the approach used on `ComponentModelEditor`:

* All the logic relative to a given display type is moved to its own class / file. 
* I ported `Pose3d` to the new format as an example. It's a comprehensive example because:
    * It registers a callback to update the view (C++ :arrow_right: QML)
    * It has a C++ `Q_INVOKABLE` function which is called from QML  (QML :arrow_right: C++)
* Some differences from `ComponentModelEditor`
    * Renamed `RegisterComponentCreator` to `AddUpdateViewCb` because I think this reflects better what the function is used for
    * Removed the `Entity` parameter from the callback because it can be attained from the inspector
    * I had to expose the `WorldName` and `TransportNode` so `Pose3d` could use them (this isn't used by the model editor because it doesn't send commands through transport directly)
    * One display type can handle multiple component types (i.e. `inspector::Pose3d` works with `components::Pose`, `components::WorldPose` and `components::WorldPoseCmd`)

## Test it

Check that all affected components are behaving as always:

1. `ign gazebo rolling_shapes.sdf`
2. See the model poses update as they roll
3. Edit the poses and see the shapes teleported accordingly
4. See how link poses are correct, and read-only
5. `ign gazebo sensors.sdf`
6. See the `WorldPose` component for the sensor entities

## Checklist
- [ ] Added tests

### DIFF
--- a/src/gui/plugins/component_inspector/CMakeLists.txt
+++ b/src/gui/plugins/component_inspector/CMakeLists.txt
@@ -1,4 +1,8 @@
 gz_add_gui_plugin(ComponentInspector
-  SOURCES ComponentInspector.cc
-  QT_HEADERS ComponentInspector.hh
+  SOURCES
+    ComponentInspector.cc
+    Pose3d.cc
+  QT_HEADERS
+    ComponentInspector.hh
+    Pose3d.hh
 )
--- a/src/gui/plugins/component_inspector/ComponentInspector.cc
+++ b/src/gui/plugins/component_inspector/ComponentInspector.cc
@@ -22,7 +22,6 @@
 #include <ignition/gui/Application.hh>
 #include <ignition/gui/MainWindow.hh>
 #include <ignition/plugin/Register.hh>
-#include <ignition/transport/Node.hh>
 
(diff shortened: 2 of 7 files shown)


nkoenig:
I'm assuming you left out the pose3d edit toggling based on the play/pause state in order to reduce the diff, correct?

chapulina:
> I'm assuming you left out the pose3d edit toggling based on the play/pause state in order to reduce the diff, correct?

Oh no, that was because I'm not backporting the model editing functionality. I'm only backporting the APIs that allow us to split up the C++ code.

So the pose editing here should continue working the same way as it already does on Citadel, which is while either playing or paused, using a service call.

nkoenig (approved):

osrf-triage:

PR #1402, opened 2022-03-19
Preserve sign of thrust_coefficient

### LINKED ISSUE #1399: ThrusterPrivateData::ThrustToAngularVec ignores sign of thrust_coefficient
## Environment
* OS Version: Ubuntu 20.04 
* Source or binary build? Source, main, latest commit = fc9cadd680e1414b38be71b404958aa8e16ff70d
  
  

## Description
* Expected behavior: clockwise (left-hand) thrusters should spin clockwise when thrust is positive, regardless of the value of  use_angvel_cmd.
* Actual behavior: when use_angvel_cmd is false the sign of thrust_coefficient is ignored. I traced this down to a bug in ThrustToAngularVec.

## Steps to reproduce

[Example model with balanced cw & ccw thrusters.](https://github.com/clydemcqueen/bluerov2_ignition/blob/4a107588bbb737dee66ed742ff2f76c0bb8e6c4b/models/bluerov2/model.sdf#L430)

Thruster 5 should be ccw and 6 should be cw. Note that use_angvel_cmd is false for both.

Sending these commands should result in constant downward thrust from thrusters 5 and 6. 5 should spin ccw and 6 should spin cw to balance out the torque:
```
ign topic -t /model/bluerov2/joint/thruster5_joint/cmd_thrust -m ignition.msgs.Double -p 'data: -9'
ign topic -t /model/bluerov2/joint/thruster6_joint/cmd_thrust -m ignition.msgs.Double -p 'data: -9'
```

The thrust values are correct, but both thrusters spin ccw. The torque does not balance, so the ROV frame rotates slightly.

I will send a 1-line PR.

clydemcqueen:
# 🦟 Bug fix

Fixes #1399 

## Summary
Preserve the signs of thrust and thrust_coefficient so that ThrustToAngularVec() is the inverse of AngularVelToThrust(). This provides support for clockwise thrusters.

I modified the `ThrusterTest` fixture to also preserve the signs of thrust and thrust_coefficient. I added one test. All tests pass.

`make codecheck` seems to generate a lot of unrelated cppcheck problems, perhaps I have something set up wrong?

`./tools/clang_tidy.sh` seemed to get stuck on `/usr/bin/../lib/gcc/x86_64-linux-gnu/9/../../../../include/c++/9/variant` -- I'm not sure what is going on there.

## Checklist
- [x] Added tests


clydemcqueen:
FYI @arjo129 @mabelzhang

clydemcqueen:
DCO check failed; I just uploaded my key, so it should succeed next time. Closing & reopening.

arjo129 (commented):
Thanks for the changes.

To fix the DCO check out the instructions here:
https://github.com/ignitionrobotics/ign-gazebo/pull/1402/checks?check_run_id=5612527595
You need to ` git rebase HEAD~2 --signoff` and the force push `git push --force-with-lease origin sfix_1399`.

Regarding codecheck as long as `make codecheck` returns a zero you can ignore the previous  warnings. 

A few other knits:
* We need to document the default behavior of the propeller being counterclockwise when positive and clockwise when negative in `src/systems/thruster/Thruster.hh`
* We need to verify if this behaviour is consistent when `<use_ang_vel>` is set.

clydemcqueen:
Thanks @arjo129. I will add the additional documentation and tests.

clydemcqueen:
I think this addresses all of the feedback:
* I added a test to cover the <use_angvel_cmd> case with a negative <thrust_coefficient>.
* I updated the documentation in Thruster.hh to note that a negative <thrust_coefficient> means clockwise rotation.

I also made a few other changes to the documentation in Thruster.hh:
* There were conflicting comments as to the default for <use_angvel_cmd>, I verified that it is false and updated the comment.
* I noted that the parameters <min_thrust_cmd> and <max_thrust_cmd> also apply to angular velocity in the <use_angvel_cmd> case. This is a bit awkward since the units are either N or rad/s. Also, the parameters should probably be named <min_cmd> and <max_cmd>, but deprecating the old parameters and adding new ones seemed like overkill.

I'm happy for other feedback.

arjo129 (approved):
Thanks for these changes. They look good to me.

src/systems/thruster/Thruster.hh
```diff
@@ -53,15 +53,18 @@ namespace systems
...
+  ///   terms of newtons. [Optional, Boolean, defaults to false]
   /// - <fluid_density> - The fluid density of the liquid in which the thruster
   ///   is operating in. [Optional, kg/m^3, defaults to 1000 kg/m^3]
   /// - <propeller_diameter> - The diameter of the propeller in meters.
   ///   [Optional, m, defaults to 0.02m]
   /// - <thrust_coefficient> - This is the coefficient which relates the angular
-  ///   velocity to actual thrust. [Optional, no units, defaults to 1.0]
+  ///   velocity to actual thrust. A positive coefficient spins the propeller
+  ///   counter-clockwise for a positive thrust force; a negative coefficient
+  ///   spins the propeller clockwise for positive thrust force. [Optional,
```
chapulina, line 66: Just a note that "clockwise" can be confusing unless the reference frame is clear. It's worth clarifying whether this is from the point of view from the vehicle, or looking at it from the outside.

chapulina (approved):
LGTM, I just have a nitpick about the documentation

src/systems/thruster/Thruster.hh
clydemcqueen: Geez, great catch @chapulina

I meant to apply the maritime nomenclature: viewed stern to bow, a clockwise (right-hand) propeller spins clockwise to provide forward (positive) thrust. But I got it flipped. I will fix that, and note the reference frame.

clydemcqueen:
@chapulina please take another look. ignition_gazebo-ci-pr_any-homebrew-amd64 failed 1 test but for unrelated reasons? Thanks

Rezenders:
Hello,

I am trying to build Ignition from source with this patch but I am having problems matching all the packages versions.
If I use `collection-garden.yaml` it gets everything from the main branches, it builds but then I have problems with packages integrating ros and ignition. I guess that is probably expected since the main branches receive updates daily.
So, what I have been trying to do is to figure out which specific commits I should checkout to build garden with this patch. Any suggestions?

clydemcqueen:
Hello, @Rezenders 

Please see PR#2 in bluerov2_ignition. It's not complete, but it does include the hashes you are looking for.

Rezenders:
Thanks for the reply. I will try it!

jeroenzwan:
Hi,
Missing_files_common

I am trying to also use this version with the hashes you referred to in a previous post. However, when I am compiling it seems to me that there are quite some files missing in the ign-common package. Above is a screenshot of the folders and the content of ImageHeightmap.hh. I am trying to compile using a system with Ubuntu 22.04 and ROS2 Humble.
Below you can see the error that happens when I try to build with `colcon build --merge-install --cmake-args ' -DBUILD_TESTING=OFF' ' -DCMAKE_BUILD_TYPE=Release'`
Screenshot from 2022-09-06 16-10-19
Do you have any suggestions?

mjcarroll:
You are missing the `common::graphics` component due to a lack of `freeimage`, `sudo apt install libfreeimage-dev` should fix it.

jeroenzwan:
Thanks @mjcarroll for your reply. However, I still have the same error.

jeroenzwan:
Update: I was indeed missing libfreeimage-dev, so I went through installing all the dependencies again and found I was missing some others as well. This resolved the issue, thanks for the help!

PR #1415, opened 2022-03-27
Set simulation time to Rendering

darksylinc:
See https://github.com/ignitionrobotics/ign-rendering/issues/556
See https://github.com/ignitionrobotics/ign-rendering/pull/584

# 🦟 Bug fix

There is no ticket filed for this issue, but there is one at ign-rendering#556

## Summary

Previously rendering was not being told of the current simulation time.

This is needed for proper particle FXs simulation and some postprocessing effects that rely on time.

This PR also accounts for a hack required only by ign-rendering6 so that SetTime() is honoured (otherwise legacy behavior is used, which uses real time instead of simulation time). Details of this legacy issue can be found in [my reply](https://github.com/ignitionrobotics/ign-rendering/pull/584#issuecomment-1079788789).

This PR should be merged AFTER ign-rendering6's ign-rendering#584

## Checklist
- [ ] Added tests


src/rendering/RenderUtil.cc
```diff
@@ -2499,6 +2501,12 @@ void RenderUtil::Init()
       this->dataPtr->scene->SetSkyEnabled(this->dataPtr->skyEnabled);
     }
   }
+
+#if IGNITION_RENDERING_MAJOR_VERSION <= 6
```
chapulina, line 2505: We only support one `ign-rendering` version with each branch, so this check shouldn't be necessary. `ign-gazebo6` uses `ign-rendering6`.

If this is something that needs to be updated when merging forward to `main` (future `ign-gazebo7`, which will use `ign-rendering7`), I think it's more visible to put a `TODO` comment

darksylinc: Yes it is necessary.

But I shall add a comment explaining why.

This is a hack. `ign-rendering6` was using real time to simulate particle FXs instead of simulation time.

However fixing this could cause unforeseen bugs to existing users; because they may not be calling SetTime at all (after all even Gazebo wasn't calling it). Fixing the bug could break stuff because existing apps never called SetTime.

Conversely, users already calling SetTime may be doing so incorrectly (after all SetTime was literally doing nothing, so users could never test if they were doing it correctly)

As a workaround, ign-rendering6 will default to the old behavior (use real time for particle and postprocessing FXs) but will switch to simulation time if SetTime( -1 ) is called (and ignore the -1).

We chose -1 because it is extremely unlikely an app in the wild is calling SetTime(-1) thus chances of breaking apps are almost 0.

But once we move to ign-rendering7, this workaround will be no more: Ignition will default to simulation time and SetTime( -1 ) may actually have consequences as it won't be ignored.

chapulina: So what I meant is that this check will always return true on this branch. Once this PR is ported to `main`, then it will always be false. So we could just remove the version check, and completely remove the `SetTime` call on `main`. 

Having this check here gives the impression that more than one `ign-rendering` version could be used with this branch, which is not true.

darksylinc: Oh, if you're confident the snippet won't sneak into main then sure.

darksylinc: OK I removed the `IGNITION_RENDERING_MAJOR_VERSION` ifdef.

I commented the code and added an `IGN_ASSERT` because I simply don't trust merges; so it doesn't accidentally sneak into newer branches.

iche033 (approved):

PR #1417, opened 2022-03-28
SceneBroadcaster: Use double for state publish frequency instead of int

### LINKED ISSUE #1414: SceneBroadcaster's _hertz options crash with doubles
## Environment
* OS Version:  Ubuntu 20.02
* Source or binary build? either
  
  

## Description
* Expected behavior:  I can pass double values to `<state_hertz>` and `<dynamic_state_hertz>`, like `0.01`
* Actual behavior:  A crash

## Steps to reproduce

Load a world that has a `< 0` double value for the frequency, for example

```
      <plugin
        <state_hertz>0.01</state_hertz>
      </plugin> 
```

## Output

Backtrace:

```
Thread 1 "INTEGRATION_sce" received signal SIGFPE, Arithmetic exception.                                                                 
0x00007fc1f83c4e04 in ignition::gazebo::v6::systems::SceneBroadcaster::Configure (this=0x563b60e176e0, _entity=<optimized out>, _sdf=..., 
    _ecm=...) at /usr/include/c++/9/chrono:331                                                                                              331               constexpr explicit duration(const _Rep2& __rep)                                                                           
(gdb) bt                                                                                                                                    
#0  0x00007fc1f83c4e04 in ignition::gazebo::v6::systems::SceneBroadcaster::Configure (this=0x563b60e176e0, _entity=<optimized out>, 
    _sdf=..., _ecm=...) at /usr/include/c++/9/chrono:331  
```

Thodoris1999:
…

# 🦟 Bug fix

Fixes #1414 

## Summary
Changed the `state_hertz` publishing frequency data type from `int` to `double`. Previously, worlds with those params set between 0 and 1 would crash due to being rounded to 0 and `state_hertz` being inverted. Note that dynamic_pose_hertz still gets rounded down to 0 as it is represented as `int` in `AdvertiseOptions`, but does not cause a crash.

## Checklist
- [ ] Added tests

### DIFF
--- a/src/systems/scene_broadcaster/SceneBroadcaster.cc
+++ b/src/systems/scene_broadcaster/SceneBroadcaster.cc
@@ -225,8 +225,8 @@ class ignition::gazebo::systems::SceneBroadcasterPrivate
   /// default update rate of 60Hz. The paused period has a key=true and a
   /// default update rate of 30Hz.
   public: std::map<bool, std::chrono::duration<double, std::ratio<1, 1000>>>
-      statePublishPeriod{{false, std::chrono::milliseconds(1000/60)},
-                         {true,  std::chrono::milliseconds(1000/30)}};
+      statePublishPeriod{{false, std::chrono::duration<double, std::ratio<1, 1000>>(1000/60.0)},
+                         {true,  std::chrono::duration<double, std::ratio<1, 1000>>(1000/30.0)}};
 
   /// \brief Flag used to indicate if the state service was called.
   public: bool stateServiceRequest{false};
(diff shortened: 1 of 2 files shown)


src/systems/scene_broadcaster/SceneBroadcaster.cc
```diff
@@ -261,16 +261,17 @@ void SceneBroadcaster::Configure(
   auto readHertz = _sdf->Get<int>("dynamic_pose_hertz", 60);
   this->dataPtr->dyPoseHertz = readHertz.first;
 
-  auto stateHerz = _sdf->Get<int>("state_hertz", 60);
+  auto stateHertz = _sdf->Get<double>("state_hertz", 60);
+  if (stateHertz.first < 0.0)
+  {
+    ignerr << "SceneBroadcaster state_hertz must be non-negative\n";
```
chapulina, line 267: How about also checking whether it's zero to avoid the division by zero?

And if this check fails, we shouldn't just print an error, but also make sure that the computations below don't end up in undefined behaviour. We could for example set the value to 1 and mention this in the error message.

chapulina (changes requested):
Thanks for the PR, the fix works for me!

Would you be able to add a test to `test/integration/scene_broadcaster_system.cc`? It will be tricky to check for precise publish times since the plugin uses the system clock, so I think the test could at least check that there's no crash.

src/systems/scene_broadcaster/SceneBroadcaster.cc
Thodoris1999: I see, I missed that

Thodoris1999:
I am a little new to this, is there an example for a test for crashing?

Thodoris1999:
I think I can make [death tests](https://google.github.io/googletest/advanced.html#death-tests) to check for crashes, is it ok? I am asking because I have not seen it in any other one of gazebo's test suites.

chapulina:
> I think I can make [death tests](https://google.github.io/googletest/advanced.html#death-tests) to check for crashes, is it ok? I am asking because I have not seen it in any other one of gazebo's test suites.

Oh I don't think we need to go that far. Just exercising that code path and having the test run successfully should be enough to make sure there's no crash. This way, if someone reverts the changes from this PR in the future, the test will crash and we'll notice the regression.

Thanks!

Thodoris1999:
I added a test with a simple SDF file. It passes with the current fix but if applied before the patch it crashes with `Floating point exception (core dumped)` as expected.

chapulina (approved):
Thanks for iterating!

PR #1452, opened 2022-04-20
Improve contact sensor / visualization performance

iche033:

# 🦟 Bug fix

## Summary

This PR improves contact sensor performance by using a new `GetByPhysicsId` function in `EntityFeatureMap` for faster retrieval of gazebo entities from physics ID.

The original call uses the `Get` function that requires casting to `ShapePtrType` which causes a big performance hit when done many times in a for loop. The cast is needed because the two types are actually different. 

With this change, I noticed that for our application with ~70 contact points, the computation time of the [for loop for building the contact map](https://github.com/ignitionrobotics/ign-gazebo/blob/ign-gazebo6/src/systems/physics/Physics.cc#L3231-L3245) drops from 2.x ms to 0.02x ms. This gives us ~10% RTF Improvement with a contact sensor in the world.

You can manually test this with shapes.world:

1. Launch shapes.sdf world
2. spawn 5 more **cylinders** into the world using the GUI tool: this gives ~60 contact points.
3. Go to top right 3 dots menu and open the `Visualize Contacts` GUI Plugin

Before this change, the RTF on my machine is roughly 3x%. After this change, it hovers around 8x%

## Checklist
- [x] Added tests

### DIFF
--- a/src/systems/physics/EntityFeatureMap.hh
+++ b/src/systems/physics/EntityFeatureMap.hh
@@ -203,6 +203,20 @@ namespace systems::physics_system
       return nullptr;
     }
 
+    /// \brief Get Gazebo entity that corresponds to the physics entity ID
+    /// \param[in] _id Physics entity ID
+    /// \return If found, returns the corresponding Gazebo entity. Otherwise,
+    /// kNullEntity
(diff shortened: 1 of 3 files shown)


ahcorde (approved):
Nice!

PR #1459, opened 2022-04-26
python: remove semicolons

dlech:
Python doesn't use semicolons (although it does ignore them).


arjo129 (approved):
Thanks. Looks good to me.

ahcorde (approved):

PR #1471, opened 2022-05-03
Refactor common sensor setup interactions

mjcarroll:
This addresses a bug uncovered when using the `Reset` implementation with the `ForceTorqueSensor`.

Not all sensor system implementations were updated as part of https://github.com/ignitionrobotics/ign-gazebo/pull/1281, one of which being the `ForceTorqueSensor`.  This makes the reset behavior incorrect and the sensor won't be respawned.

As part of this, I'm also experimenting with refactoring out the common functionality that is boilerplate in any GzSensors  system implementation.    To view the diff for a single system that was already updated, see https://github.com/ignitionrobotics/ign-gazebo/commit/20b15b619a3d67d607d8bbcb238b34689a8b9b9c

Once complete, this should close out: https://github.com/ignitionrobotics/ign-gazebo/issues/797 as all sensors will be updated.

### DIFF
--- a/src/systems/CMakeLists.txt
+++ b/src/systems/CMakeLists.txt
@@ -58,6 +58,12 @@ function(gz_add_system system_name)
       ignition-plugin${IGN_PLUGIN_VER}::register
   )
 
+  target_include_directories(${system_target}
+    PRIVATE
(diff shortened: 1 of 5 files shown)


mjcarroll:
-

PR #1475, opened 2022-05-04
Allow rendering to be forced externally

hidmic:
# 🎉 New feature

## Summary
`ignition::gazebo::systems::Sensors` will only update its rendering sensors when there's a need for it. While perfectly reasonable, this also means that any custom rendering sensor relying on pre/post render hooks to do its work must wait for a builtin sensor that is known to this system to require an update. 

This patch works around this by introducing a `ForceRender` event that other systems can emit to force a rendering pass.

## Test it
...

## Checklist
- [ ] Added tests

### DIFF
--- a/include/gz/sim/rendering/Events.hh
+++ b/include/gz/sim/rendering/Events.hh
@@ -44,9 +44,9 @@ namespace gz
       using SceneUpdate = gz::common::EventT<void(void),
           struct SceneUpdateTag>;
 
-      /// \brief The render event is emitted before rendering updates.
+      /// \brief The pre render event is emitted before rendering updates.
       /// The event is emitted in the rendering thread so rendering
-      /// calls can ben make in this event callback
+      /// calls can be made in this event callback
       ///
       /// For example:
       /// \code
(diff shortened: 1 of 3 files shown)


hidmic:
Early push, to get feedback. I'll go implement a custom rendering sensor on top of this now.

include/ignition/gazebo/rendering/Events.hh
```diff
@@ -65,6 +65,16 @@ namespace ignition
       /// \endcode
       using PostRender = ignition::common::EventT<void(void),
           struct PostRenderTag>;
+
+      /// \brief The force render event may be emitted outside the
+      /// rendering thread to force rendering calls.
```
chapulina, line 70: The approach looks reasonable to me.

I'd just improve the documentation a bit to explain that "forcing" rendering means that when the time comes, a plugin that performs rendering will do so even if it wasn't planning to.

hidmic:
A couple notes about tests and examples for this patch (which we're lacking). You can see in https://github.com/osrf/lrauv/pull/213 that putting together a plugin to make meaningful use of this new events is very tricky. So I think we have two choices: (a) we defer examples here and ticket a migration of https://github.com/osrf/lrauv/pull/213 into `gz-sensors` or (b) try to put together a dummy custom rendering sensor to motivate this changes. I'm biased towards (a), as I know (b) is going to be a rather long detour, full of booby traps.

iche033:
The changes look fine. One note about going with this approach is that there is no guarantee that the custom rendering sensor's updates are lockstepped with physics. Something to keep in mind in case you are seeing that your sensor is not hitting the target update rate.

caguero (approved):
It looks reasonable to me.

hidmic:
Alright, according to @caguero, failing checks are already happening on `master`. 

Moving forward!

PR #1480, opened 2022-05-11
Optimize sensor updates

iche033:
# 🦟 Bug fix

## Summary

This PR optimizes sensor updates by avoiding any unnecessary computation when:
* it's not time for the sensor to update
* there are not subscribers or event connections to the sensors

For non-rendering sensors, before these changes, the sensors themselves are throttled correctly but we are still doing updates (copying data from sim to sensor object) every iteration in `PostUpdate` function.

For rendering sensors, i.e. sensors system, I noticed that the `RGBDCameraSensor` and `GpuLidarSensor` are still doing updates when there are no subscribers so I added the check in the sensors system.  

After these changes, the `sensors_demo.sdf` world gives me consistent ~9x% RTF after closing all `ImageDisplay`s. Before the changes, the RTF still fluctuates between 3x% to 8x% after closing the `ImageDisplay`s (because `GpuLidarSensor` and `RGBDCameraSensor` are still updating)

I verified the behavior by enabling remotery in ign-sensors. This is the output from running `sensors_demo.sdf` world.

The output below shows the sensor updates after the `sensors_demo.sdf` world is launched and running. We see that `GpuLidarSensor` is no longer updating because there are no subscribers. On the other hand, all the other rendering sensors are updating because there are `ImageDisplay`s connecting to the sensors.
sensors_demo_profile_a

The output below shows the result after closing the RGBD cameras and thermal camera's `ImageDisplay`s. Now only the depth and RGB cameras are updating, which is the desired behavior.
sensors_demo_profile_b

## Alternatives considered

The alternative approach, which I implemented but then reverted, was to check for connection count inside the sensor classes in ign-sensors. The main reason I didn't go with this approach is because classes like `Lidar` (which `GpuLidarSensor` is derived from) also offers [Ranges](https://github.com/gazebosim/gz-sensors/blob/ign-sensors6/src/Lidar.cc#L409) APIs to get range data directly. If I skipped updates when there are no connections, `Ranges` returned invalid / empty data, and that broke integration tests.

## Checklist
- [ ] Added tests

### DIFF
--- a/src/systems/air_pressure/AirPressure.cc
+++ b/src/systems/air_pressure/AirPressure.cc
@@ -140,6 +140,24 @@ void AirPressure::PostUpdate(const UpdateInfo &_info,
 
   if (!_info.paused)
   {
+    // check to see if update is necessary
+    // we only update if there is at least one sensor that needs data
+    // and that sensor has subscribers.
+    // note: ign-sensors does its own throttling. Here the check is mainly
+    // to void doing work in the AirPressurePrivate::UpdatePressures function
+    bool needsUpdate = false;
+    for (auto &it : this->dataPtr->entitySensorMap)
+    {
+      if (it.second->NextDataUpdateTime() <= _info.simTime &&
+          it.second->HasConnections())
+      {
+        needsUpdate = true;
+        break;
+      }
+    }
+    if (!needsUpdate)
+      return;
+
     this->dataPtr->UpdateAirPressures(_ecm);
 
     for (auto &it : this->dataPtr->entitySensorMap)
--- a/src/systems/sensors/Sensors.cc
+++ b/src/systems/sensors/Sensors.cc
@@ -186,6 +186,11 @@ class ignition::gazebo::systems::SensorsPrivate
   /// \param[in] _ecm Entity component manager
   public: void UpdateBatteryState(const EntityComponentManager &_ecm);
 
+  /// \brief Check if sensor has subscribers
+  /// \param[in] _sensor Sensor to check
+  /// \return True if the sensor has subscribers, false otherwise
+  public: bool HasConnections(sensors::RenderingSensor *_sensor) const;
+
   /// \brief Use to optionally set the background color.
   public: std::optional<math::Color> backgroundColor;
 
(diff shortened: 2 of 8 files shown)


mjcarroll:
Looks like we are still getting some issues even with upstream release (6.4.0)

```
  [ 51%] Building CXX object src/systems/air_pressure/CMakeFiles/ignition-gazebo6-air-pressure-system.dir/AirPressure.cc.o
  /github/workspace/src/systems/air_pressure/AirPressure.cc: In member function 'virtual void ignition::gazebo::v6::systems::AirPressure::PostUpdate(const ignition::gazebo::v6::UpdateInfo&, const ignition::gazebo::v6::EntityComponentManager&)':
  /github/workspace/src/systems/air_pressure/AirPressure.cc:152:22: error: 'class ignition::sensors::v6::AirPressureSensor' has no member named 'HasConnections'
             it.second->HasConnections())
                        ^~~~~~~~~~~~~~
```

ahcorde:
@osrf-jenkins run tests please!

src/systems/air_pressure/AirPressure.cc
```diff
@@ -140,6 +140,24 @@ void AirPressure::PostUpdate(const UpdateInfo &_info,
 
   if (!_info.paused)
   {
+    // check to see if update is necessary
+    // we only update if there is at least one sensor that needs data
+    // and that sensor has subscribers.
+    // note: ign-sensors does its own throttling. Here the check is mainly
+    // to void doing work in the AirPressurePrivate::UpdatePressures function
```
chapulina, line 147:
```suggestion
    // to avoid doing work in the AirPressurePrivate::UpdatePressures function
```

src/systems/sensors/Sensors.cc
```diff
@@ -805,6 +833,39 @@ std::string Sensors::CreateSensor(const Entity &_entity,
...
+    auto s = dynamic_cast<sensors::ThermalCameraSensor *>(_sensor);
+    if (s)
+      return s->HasConnections();
+  }
+  {
+    auto s = dynamic_cast<sensors::CameraSensor *>(_sensor);
+    if (s)
+      return s->HasConnections();
+  }
+  return true;
```
chapulina, line 866: Consider printing a warning here, so we catch at runtime the situation when we forgot to add a sensor to this list.

chapulina (approved):
LGTM, just a couple of minor comments.

ahcorde (approved):

src/systems/air_pressure/AirPressure.cc
iche033: fixed. [8b611f8](https://github.com/gazebosim/gz-sim/pull/1480/commits/8b611f85a67d35b949a704e1f612736e4bf8285a)

azeey:
@iche033 `RgbdCameraTest.RgbdCameraBox` and `SensorsFixture.SensorsBatteryState` seem to be broken on ign-gazebo6 due to this PR. And `SensorsFixture.SensorsBatteryState` and `WideAngleCameraTest.WideAngleCameraBox` on `main` (reverting this PR on `main` fixes `SensorsFixture.SensorsBatteryState` and `WideAngleCameraTest.WideAngleCameraBox`). Do you mind taking a look?

iche033:
#1528 should fix `WideAngleCameraTest.WideAngleCameraBox`
#1529 should fix `SensorsFixture.SensorsBatteryState`

hmm I'm not able to reproduce `RgbdCameraTest.RgbdCameraBox`. It could be fixed by https://github.com/gazebosim/gz-sim/pull/1513?

azeey:
Thanks! Yes, I think `RgbdCameraTest.RgbdCameraBox` was fixed by #1513: https://build.osrfoundation.org/view/ign-fortress/job/ignition_gazebo-ci-ign-gazebo6-focal-amd64/63/#showFailuresLink

PR #1503, opened 2022-05-20
Add QML Debugging support

darksylinc:
The project must be rebuilt defining the QT_QML_DEBUG macro.

Once that's done, loading the GUI will wait until the debugger is
attached to port 40000.

QtCreator does this automatically when setting:
  Projects -> Build -> QML debugging and profiling -> Enable

Tested with QtCreator (Qt Quick plugin enabled) and working.

# 🦟 Bug fix

No ticket was created for this issue.

## Summary

Troubleshooting QML errors in Javascript code is a PITA without proper tools.
This PR addresses that.

Usually QML debugging is handled automatically by appending the right command line entry.

However IGN is special because it's a so loaded by a ruby script; hence all command line arguments added by QtCreator are lost (and may also result in failure to start).

A quick solution is to hardcode those commands when compiling with `QT_QML_DEBUG`.

## Checklist
- [ ] Added tests


mayman99:
Thanks for the PR!
I've a question, would it possible to expose the debug option thru the starting ruby script?

mayman99 (changes requested):
I think exposing this feature thru a tutorial or an example would be a better idea for usability, because if merged like this the only way users would know about the feature is by reading the source code.

darksylinc:
OK I added gazebosim/docs#251 as documentation that should accompany this new feature.

Is it satisfactory?

mayman99 (approved):
lgtm!
thanks for the PR!

PR #1520, opened 2022-06-06
ign -> gz Macro Migration : gz-sim

methylDragon:
See: https://github.com/gazebo-tooling/release-tools/issues/737

### DIFF
--- a/src/ign.hh
+++ b/src/ign.hh
@@ -25,7 +25,7 @@ extern "C" char *gzSimVersion();
 
 /// \brief Get the Gazebo version header.
 /// \return C-string containing the Gazebo version information.
-extern "C" char *gazeboVersionHeader();
+extern "C" char *simVersionHeader();
 
 /// \brief Set verbosity level
 /// \param[in] _verbosity 0 to 4
--- a/src/ign.hh
+++ b/src/ign.hh
@@ -21,7 +21,8 @@
 
 /// \brief External hook to read the library version.
 /// \return C-string representing the version. Ex.: 0.1.2
-extern "C" char *ignitionGazeboVersion();
+extern "C" char *gzSimVersion();
+extern "C" char GZ_DEPRECATED(7) *ignitionGazeboVersion();
 
 /// \brief Get the Gazebo version header.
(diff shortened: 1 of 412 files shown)


test/integration/log_system.cc
```diff
@@ -1523,7 +1523,7 @@ TEST_F(LogSystemTest, IGN_UTILS_TEST_DISABLED_ON_WIN32(LogResources))
   // Recorded models should exist
   EXPECT_GT(entryCount(recordPath), 2);
   EXPECT_TRUE(common::exists(common::joinPaths(recordPath, homeFake,
-      ".ignition", "fuel", "fuel.ignitionrobotics.org", "openrobotics",
+      ".gz", "fuel", "fuel.ignitionrobotics.org", "openrobotics",
```
chapulina, line 1526: We shouldn't change these yet

chapulina (changes requested):
LGTM besides the Fuel directories.

test/integration/log_system.cc
methylDragon: [`a3ecb2b` (#1520)](https://github.com/gazebosim/gz-sim/pull/1520/commits/a3ecb2b9b170b743a89c459faba90884b22eb70d)

src/ign.hh
```diff
@@ -21,7 +21,8 @@
 
 /// \brief External hook to read the library version.
 /// \return C-string representing the version. Ex.: 0.1.2
-extern "C" char *ignitionGazeboVersion();
+extern "C" char *gzSimVersion();
+extern "C" char GZ_DEPRECATED(7) *ignitionGazeboVersion();
```
chapulina, line 25: No need to tick-tock, it's not installed

chapulina (changes requested):

src/ign.hh
methylDragon: https://github.com/gazebosim/gz-sim/pull/1520/commits/b969c8bd754a44e9fe0aba50ae54a131ef015966

chapulina (approved):

PR #1523, opened 2022-06-07
Don't use 'EachNew' in ForceTorque PreUpdate function

mjcarroll:
# 🦟 Bug fix

Split from #1471 to address force-torque sensor incrementally.
Partially addresses #797

## Summary

Not all sensor system implementations were updated as part of https://github.com/gazebosim/gz-sim/pull/1281, one of which being the ForceTorqueSensor. This makes the reset behavior incorrect and the sensor won't be respawned.

## Checklist
- [ ] Added tests


src/systems/force_torque/ForceTorque.cc
```diff
@@ -112,7 +133,27 @@ void ForceTorque::PreUpdate(const UpdateInfo &/*_info*/,
...
+             << "] isn't in sensor map, this shouldn't happen." << std::endl;
+      continue;
+    }
+
+    // Set topic
+    _ecm.CreateComponent(entity, components::SensorTopic(it->second->Topic()));
+    // Enable JointTransmittedWrench to get force-torque measurements
+    auto jointEntity =
+        _ecm.Component<components::ParentEntity>(entity)->Data();
+    gzwarn << "Adding JointTransmittedWrench to: " << jointEntity << std::endl;
```
chapulina, line 153: A warning sounds like something that needs to be fixed. Maybe `gzdbg` would be better here?

src/systems/force_torque/ForceTorque.cc
```diff
@@ -84,13 +93,25 @@ class gz::sim::systems::ForceTorquePrivate
...
+    const components::ParentEntity *_parent);
+
+  /// \brief Create force-torque sensor
+  /// \param[in] _ecm Immutable reference to ECM.
+  public: void CreateSensors(const EntityComponentManager &_ecm);
 
   /// \brief Update FT sensor data based on physics data
   /// \param[in] _ecm Immutable reference to ECM.
-  public: void Update(const EntityComponentManager &_ecm);
+  public: void UpdateForceTorques(const EntityComponentManager &_ecm);
```
chapulina, line 114: On #1281 I tried to reduce the diff between the sensors by making these function names generic, see https://github.com/gazebosim/gz-sim/pull/1281#discussion_r785248236. But since you're already working on #1471 I think this won't stay here for long anyway.

chapulina (commented):
LGTM, I just have some minor comments.

It would be ideal to target this at Fortress and forward-port, to reduce the diff between the branches. If it's easier for you we can also backport this after merged.

src/systems/force_torque/ForceTorque.cc
mjcarroll: I will clean them all up in #1471.

mjcarroll: Actually, I will fix it here, so that we port back to fortress is cleaner.

src/systems/force_torque/ForceTorque.cc
```diff
@@ -178,99 +221,124 @@ Entity ForceTorquePrivate::GetLinkFromScopedName(
...
+    std::string topic = scopedName(_entity, _ecm) + "/forcetorque";
+    data.SetTopic(topic);
+  }
+  std::unique_ptr<sensors::ForceTorqueSensor> sensor =
+      this->sensorFactory.CreateSensor<
+      sensors::ForceTorqueSensor>(data);
+  if (nullptr == sensor)
+  {
+    gzerr << "Failed to create sensor [" << sensorScopedName << "]"
+           << std::endl;
```
chapulina, line 247: Should it return here to avoid a nullptr access?

src/systems/force_torque/ForceTorque.cc
```diff
@@ -178,99 +221,124 @@ Entity ForceTorquePrivate::GetLinkFromScopedName(
...
-            _ecm.Component<components::ChildLinkName>(jointEntity)->Data();
-        auto jointChildLinkEntity =
-            this->GetLinkFromScopedName(_ecm, jointChildName, modelEntity);
-        if (kNullEntity == jointChildLinkEntity)
+        });
+    this->initialized = true;
+  }
+  else
+  {
+    // Create air pressure sensors
```
chapulina, line 332:
```suggestion
    // Create force-torque sensors
```

src/systems/force_torque/ForceTorque.cc
```diff
@@ -84,9 +93,21 @@ class gz::sim::systems::ForceTorquePrivate
...
-  /// \brief Create FT sensor
-  /// \param[in] _ecm Mutable reference to ECM.
-  public: void CreateForceTorqueEntities(EntityComponentManager &_ecm);
+
+  /// \brief Create sensor
+  /// \param[in] _ecm Immutable reference to ECM.
+  /// \param[in] _entity Entity of the force-torque sensor
+  /// \param[in] _forceTorque ForceTorqueSensor component.
+  /// \param[in] _parent Parent entity component.
+  public: void AddForceTorque(
```
chapulina, line 102: You could call this `AddSensor` and rearrange the function order to reduce the diff between this sensor and others, like Imu. Not all sensors are following the same pattern, but for the ones that are, it's much easier to see their diff side-by-side.

chapulina (approved):
LGTM with the return comment addressed, and the other comment is up to you.

src/systems/force_torque/ForceTorque.cc
mjcarroll: Good catch, done.

src/systems/force_torque/ForceTorque.cc
mjcarroll: Good call, done. 6ef7ae9

mjcarroll:

PR #1527, opened 2022-06-08
Rename CMake project to gz

chapulina:
* See https://github.com/gazebo-tooling/release-tools/issues/718

## Shared libs

To test the warnings for `ignition-` prefixed physics engines, try:

```
$ ign gazebo rolling_shapes.sdf -v 2 --physics-engine ignition-physics-dartsim-plugin           
...                                              
[Wrn] [Physics.cc:760] Trying to load deprecated plugin [ignition-physics-dartsim-plugin]. Use [gz-physics-dartsim-plugin] instead.
...
```

And for the render engine:

```
$ ign gazebo shapes.sdf -v 2 --render-engine ignition-rendering-ogre
[Wrn] [ServerConfig.cc:543] Trying to load deprecated engine [ignition-rendering-ogre] for the server. Use [gz-rendering-ogre] instead.
[Wrn] [ServerConfig.cc:574] Trying to load deprecated engine [ignition-rendering-ogre] for the GUI. Use [gz-rendering-ogre] instead.
...
[GUI] [Wrn] [MainWindow.cc:955] Trying to load deprecated plugin [ignition-rendering-ogre]. Use [gz-rendering-ogre] instead.
...
```

I added a world that tries to load `ignition-gazebo` prefixed plugins to demonstrate that they still work:

```
$ ign gazebo deprecated_ignition.sdf  -v 2
...
[Wrn] [SystemLoader.cc:54] Trying to load deprecated plugin [ignition-gazebo-physics-system]. Using [gz-sim-physics-system] instead.
[Wrn] [SystemLoader.cc:54] Trying to load deprecated plugin [ignition-gazebo-user-commands-system]. Using [gz-sim-user-commands-system] instead.
[Wrn] [SystemLoader.cc:54] Trying to load deprecated plugin [ignition-gazebo-scene-broadcaster-system]. Using [gz-sim-scene-broadcaster-system] instead.
...
```

## CMake targets

Since the CMake targets are changing, I took the opportunity to update various parts of the code to use targets instead of other variables. This unveiled places where we were over-linking to things we didn't need, or relying on transient dependencies. That's why this PR has added arguments to `target_link_libraries` in various places.


src/ServerConfig.cc
```diff
@@ -558,9 +566,17 @@ const std::string &ServerConfig::RenderEngineGui() const
...
-  this->dataPtr->renderEngineGui = _renderEngineGui;
+  // Deprecated: accept ignition-prefixed engines
+  auto engine = _engine;
+  if (engine.find("ignition") != std::string::npos)
+  {
+    engine.replace(0, 8, "gz");
+    gzwarn << "Trying to load deprecated engine [" << _engine
+           << "] for the GUI. Use [" << engine << "] instead." << std::endl;
+  }
+  this->dataPtr->renderEngineGui = engine;
```
methylDragon, line 579: Is there a need to be accounting for `libignition-xxx` filenames as well?

src/SystemLoader.cc
```diff
@@ -46,6 +46,15 @@ class gz::sim::SystemLoaderPrivate
...
   {
+    // Deprecated: accept ignition-gazebo-prefixed systems
+    auto filename = _filename;
+    if (filename.find("ignition-gazebo") != std::string::npos)
+    {
+      filename.replace(0, 15, "gz-sim");
+      gzwarn << "Trying to load deprecated plugin [" << _filename
+             << "]. Using [" << filename << "] instead." << std::endl;
+    }
+
```
methylDragon, line 57: Is there a need to be accounting for `libignition-xxx` filenames as well?
It could break quite badly if `libignition-gazebo-xxx` -> `gz-simebo-xxx`

methylDragon: Perhaps?

```suggestion
    // Deprecated: accept ignition-gazebo-prefixed systems
    auto filename = _filename;
    auto pos = filename.find("ignition-gazebo");
    if (pos != std::string::npos)
    {
      filename.replace(pos, pos+15, "gz-sim");
      gzwarn << "Trying to load deprecated plugin [" << _filename
             << "]. Using [" << filename << "] instead." << std::endl;
    }

```

chapulina: Good call, will do 👍🏽

methylDragon (approved):

chapulina:
No new test failures! 🚀

PR #1549, opened 2022-06-22
Add system to an entity through Component Inspector

iche033:
# 🎉 New feature

depends on #1524

## Summary

Adds a tool button to the Component Inspector for adding Systems to a model. When the button is clicked, a dialog pops up and lets you fill the `name`, `filename`, and `innerxml` fields of the System / plugin. Once confirmed, it invokes a service call introduced in  #1524 that adds the System to the selected model entity.

## Test it

1. launch the `diff_drive_no_plugin` test world:

    ```
    ign gazebo -v 4 ./src/ign-gazebo/test/worlds/diff_drive_no_plugin.sdf
    ```

1. Click on the vehicle model -  The Component Inspector should now show properties of the vehicle model

1. Click the `+` button in the toolbar in the Component Inspector - A dialog should pop out

1. Fill the `name` field with:

    ```
    ignition::gazebo::systems::DiffDrive
    ```

1. Fill the `filename` field with:

    ```
    ignition-gazebo-diff-drive-system
    ```

1. Fill the `innerxml` field with:

    ```
        <left_joint>left_wheel_joint</left_joint>
        <right_joint>right_wheel_joint</right_joint>
        <wheel_separation>1.25</wheel_separation>
        <wheel_radius>0.3</wheel_radius>
        <max_linear_acceleration>1</max_linear_acceleration>
        <min_linear_acceleration>-1</min_linear_acceleration>
        <max_angular_acceleration>2</max_angular_acceleration>
        <min_angular_acceleration>-2</min_angular_acceleration>
        <max_linear_velocity>0.5</max_linear_velocity>
        <min_linear_velocity>-0.5</min_linear_velocity>
        <max_angular_velocity>1</max_angular_velocity>
        <min_angular_velocity>-1</min_angular_velocity>    
    ```

1.  Click Ok to add the System to the model - You should see the `+ System Plugin Info` appear in the Component Inspector to indicate that the model now has a System.

1.  Hit Play - The plugin should be running and read to accept vel commands

1. Publish a message to move the vehicle:

    ```
    ign topic -t "/model/vehicle/cmd_vel" -m ignition.msgs.Twist -p "linear: {x: 0.5}, angular: {z: 0.05}"
    ```

add_system_gui

## Checklist

- [ ] Added tests


src/gui/plugins/component_inspector/ComponentInspector.qml
```diff
@@ -227,6 +247,99 @@ Rectangle {
...
+      }
+    }
+
+    footer: DialogButtonBox {
+      id: buttons
+      standardButtons: Dialog.Ok | Dialog.Cancel
+    }
+
+    onOpened: {
+      buttons.standardButton(Dialog.Ok).enabled = false;
```
chapulina, line 326: After I inserted one plugin, I opened the dialog again to insert a plugin to a different model. It was nice that the dialog was already pre-populated, because I wanted to add the same plugin. But the `OK` button was disabled, so I couldn't add it.

src/gui/plugins/component_inspector/ComponentInspector.qml
```diff
@@ -216,6 +216,26 @@ Rectangle {
...
+        id: addSystemButton
+        checkable: false
+        text: "\u002B"
+        contentItem: Text {
+          text: addSystemButton.text
+          color: "#b5b5b5"
+          horizontalAlignment: Text.AlignHCenter
+          verticalAlignment: Text.AlignVCenter
+        }
+        visible: entityType == "model"
```
chapulina, line 229: It should be ok to add systems to models, the world, visuals and sensors.

src/gui/plugins/component_inspector/ComponentInspector.qml
```diff
@@ -216,6 +216,26 @@ Rectangle {
...
+        ToolTip.text: "Add a system to a model"
```
chapulina, line 230: To be more generic when we enable other entity types, maybe

```suggestion
        ToolTip.text: "Add a system to this entity"
```

src/gui/plugins/component_inspector/ComponentInspector.qml
```diff
@@ -227,6 +247,99 @@ Rectangle {
...
+          Layout.minimumWidth: 250
+          onTextEdited: {
+            addSystemDialog.updateButtonState();
+          }
+        }
+      }
+
+      Text {
+        id: innerxmlLabel
+        text: "Innerxml"
```
chapulina, line 298: Looks better on the UI

```suggestion
        text: "Inner XML"
```

src/gui/plugins/component_inspector/ComponentInspector.qml
```diff
@@ -227,6 +247,99 @@ Rectangle {
...
+        flickableDirection: Flickable.VerticalFlick
+        // contentY: 200
+        TextArea.flickable: TextArea {
+          id: innerxmlField
+          wrapMode: Text.WordWrap
+          selectByMouse: true
+          textFormat: TextEdit.PlainText
+          font.pointSize: 10
+        }
+        ScrollBar.vertical: ScrollBar {}
```
chapulina, line 316: How about setting the scrollbar's policy to always on? In certain situations I think it can be confusing for the user if the scrollbar isn't there


```
policy: ScrollBar.AlwaysOn
```

chapulina (commented):
Nice! It works for me! I just have some minor comments.

src/gui/plugins/component_inspector/ComponentInspector.qml
iche033: added. [da4f16a](https://github.com/gazebosim/gz-sim/pull/1549/commits/da4f16a0067dcce3bd0ae96d543786058646effc)

src/gui/plugins/component_inspector/ComponentInspector.qml
iche033: [da4f16a](https://github.com/gazebosim/gz-sim/pull/1549/commits/da4f16a0067dcce3bd0ae96d543786058646effc)

src/gui/plugins/component_inspector/ComponentInspector.qml
iche033: [da4f16a](https://github.com/gazebosim/gz-sim/pull/1549/commits/da4f16a0067dcce3bd0ae96d543786058646effc)

src/gui/plugins/component_inspector/ComponentInspector.qml
iche033: updated check. [da4f16a](https://github.com/gazebosim/gz-sim/pull/1549/commits/da4f16a0067dcce3bd0ae96d543786058646effc)

src/gui/plugins/component_inspector/ComponentInspector.qml
iche033: fixed. [da4f16a](https://github.com/gazebosim/gz-sim/pull/1549/commits/da4f16a0067dcce3bd0ae96d543786058646effc)

src/gui/plugins/component_inspector/ComponentInspector.qml
chapulina: Yeah that `material` type call looks mistaken.

How about `world`, does that work if we add it?

chapulina: Humm I tried adding `world`, but then when I add systems to the world they don't work properly. I tried adding `Buoyancy` to `shapes.sdf`, and also removed `Physics` from `rolling_shapes` and tried adding it from the GUI.

src/gui/plugins/component_inspector/ComponentInspector.qml
```diff
@@ -227,6 +247,100 @@ Rectangle {
...
+          Layout.row: 0
+          Layout.column: 1
+          Layout.minimumWidth: 250
+          onTextEdited: {
+            addSystemDialog.updateButtonState();
+          }
+        }
+
+        Text {
+          text: "Filename"
```
chapulina, line 279: So one usability issue with the current dialog is that the user needs to know the plugin filename and name to insert a plugin. Most beginners won't even know where to look for this.

We could try an approach similar to the GUI plugin menu, where we crawl all possible plugin sources (default directories, env vars...) and list all plugins to the user. We'll probably need to add a service from `SystemLoader` to provide this list to the GUI.

The other tricky part with the server plugins is that we can easily get the `Filename`, but the `Name` isn't so easy. I think one workaround would be to load the first plugin in the library when no name is provided here:

https://github.com/gazebosim/gz-sim/blob/6739800821af5e39767fe22e8da4b3e27fbaa284/src/SystemLoader.cc#L79-L88

I think this would increase the scope of this PR by a lot, so I'm ok leaving it for a follow-up PR. What do you think?

src/gui/plugins/component_inspector/ComponentInspector.qml
iche033: added support for world. [d7503c5](https://github.com/gazebosim/gz-sim/pull/1549/commits/d7503c5b099a57919774a329f1533423110fde45). I tested adding `Buoyancy` to `shapes.sdf` and I I think it's working for me. I see it appear in the `System Plugin Info` list in the Component Inspector, and when I add a new object into the world, it immediately shoots upwards.

iche033: This is what I did:
add_system_gui_shapes

src/gui/plugins/component_inspector/ComponentInspector.qml
iche033: yeah ok that sounds good. We can do this in a follow-up PR.

src/gui/plugins/component_inspector/ComponentInspector.qml
chapulina: >  when I add a new object into the world, it immediately shoots upwards.

How about the existing objects, they should also shoot upwards, right? It looks like the new world plugins aren't acting on pre-existing entities. It's the same issue as adding physics to `rolling_shapes`.

iche033: > How about the existing objects, they should also shoot upwards, right?

I think that depends on how the system is written? I see that in the Buoyancy plugin for example, it uses `EachNew` to do some preprocessing. ?However, when a system is inserted at runtime, that logic will not be executed for existing objects.

chapulina: Ahh makes sense

chapulina (commented):
LGTM so far, but let's not merge it until we have a dropdown to choose the plugin.

chapulina (approved):
Just pending #1581

src/gui/plugins/component_inspector/ComponentInspector.qml
chapulina: Done in #1581

chapulina (approved):

PR #1560, opened 2022-06-28
Allow SDF model to be constructed in a single shot

mxgrey:
# 🎉 New feature

## Summary

This PR tweaks the behavior of the physics system so that when constructing a model it will first check if a `ModelSdf` component is available for the entity. If so, it uses that component to fill in all the model information, overriding the fields that are given by the other entity components (e.g. name and pose).

When creating a link, the physics engine will first check if a link of the same name for that model was already constructed. If so, the construction step is skipped, the link is queried from the model, and it gets added to the entity map. Same for joints and collision objects.

The net result of these behavioral changes is that physics engines which can only construct models all at once are able to be used for the typical use cases, while physics engines that can construct models incrementally can still do so when the user wants them to.

## Test it
Try running with the dartsim plugin. Even though the behavior changed, there is no noticeable difference in the outcome.

The ability to test these changes with a `bullet-featherstone` plugin is forthcoming. The implementation is finished, but I am debugging some segfaults from inside of bullet at the moment.

## Checklist
- [ ] Added tests

### DIFF
--- a/src/systems/physics/Physics.cc
+++ b/src/systems/physics/Physics.cc
@@ -59,6 +59,7 @@
 #include <gz/physics/FixedJoint.hh>
 #include <gz/physics/GetContacts.hh>
 #include <gz/physics/GetBoundingBox.hh>
+#include <gz/physics/GetEntities.hh>
 #include <gz/physics/Joint.hh>
 #include <gz/physics/Link.hh>
 #include <gz/physics/RemoveEntities.hh>
@@ -166,9 +167,11 @@ class gz::sim::systems::PhysicsPrivate
           gz::physics::LinkFrameSemantics,
           gz::physics::ForwardStep,
           gz::physics::RemoveModelFromWorld,
-          gz::physics::sdf::ConstructSdfLink,
           gz::physics::sdf::ConstructSdfModel,
-          gz::physics::sdf::ConstructSdfWorld
+          gz::physics::sdf::ConstructSdfWorld,
+          gz::physics::GetLinkFromModel,
+          gz::physics::GetJointFromModel,
+          gz::physics::GetShapeFromLink
           >{};
 
   /// \brief Engine type with just the minimum features.
@@ -572,7 +575,6 @@ class gz::sim::systems::PhysicsPrivate
 
   //////////////////////////////////////////////////
   // Meshes
-
   /// \brief Feature list for meshes.
   /// Include MinimumFeatureList so created collision can be automatically
   /// up-cast.
@@ -581,8 +583,14 @@ class gz::sim::systems::PhysicsPrivate
             physics::mesh::AttachMeshShapeFeature>{};
 
   //////////////////////////////////////////////////
-  // Heightmap
+  // Construct Links
+  /// \brief Feature list for constructing links
+  public: struct ConstructSdfLinkFeatureList : gz::physics::FeatureList<
+            MinimumFeatureList,
+            gz::physics::sdf::ConstructSdfLink>{};
 
+  //////////////////////////////////////////////////
+  // Heightmap
   /// \brief Feature list for heightmaps.
   /// Include MinimumFeatureList so created collision can be automatically
   /// up-cast.
@@ -632,7 +640,8 @@ class gz::sim::systems::PhysicsPrivate
             MinimumFeatureList,
             JointFeatureList,
             BoundingBoxFeatureList,
-            NestedModelFeatureList>;
+            NestedModelFeatureList,
+            ConstructSdfLinkFeatureList>;
 
   /// \brief A map between model entity ids in the ECM to Model Entities in
   /// ign-physics.
@@ -1052,8 +1061,17 @@ void PhysicsPrivate::CreateModelEntities(const EntityComponentManager &_ecm,
         }
         // TODO(anyone) Don't load models unless they have collisions
 
+
         // Check if parent world / model exists
         sdf::Model model;
+        if (const auto *modelSdfComp =
+            _ecm.Component<components::ModelSdf>(_entity))
+        {
+          model = modelSdfComp->Data();
+        }
+
+        // Component values should override whatever values were put into the
+        // ModelSdf component.
         model.SetName(_name->Data());
         model.SetRawPose(_pose->Data());
         auto staticComp = _ecm.Component<components::Static>(_entity);
@@ -1208,8 +1226,31 @@ void PhysicsPrivate::CreateLinkEntities(const EntityComponentManager &_ecm,
                   << "] not found on model map." << std::endl;
           return true;
         }
+
+        auto basicModelPtrPhys = this->entityModelMap.Get(_parent->Data());
+
+        if (const auto existingLink = basicModelPtrPhys->GetLink(_name->Data()))
+        {
+          // No need to create this link because it was already created when
+          // parsing the model (links in models are required to have unique
+          // names). Instead we will register its existence and move along.
+          this->entityLinkMap.AddEntity(_entity, existingLink);
+          this->topLevelModelMap.insert(
+            std::make_pair(_entity, topLevelModel(_entity, _ecm)));
+          return true;
+        }
+
         auto modelPtrPhys =
-            this->entityModelMap.Get(_parent->Data());
+            this->entityModelMap
+              .EntityCast<ConstructSdfLinkFeatureList>(_parent->Data());
+
+        if (!modelPtrPhys)
+        {
+          gzwarn << "Cannot create a new link [" << _name->Data() << "] "
+                 << "because the physics engine plugin does not support "
+                 << "link construction during runtime" << std::endl;
+          return true;
+        }
 
         sdf::Link link;
         link.SetName(_name->Data());
@@ -1279,6 +1320,29 @@ void PhysicsPrivate::CreateCollisionEntities(const EntityComponentManager &_ecm,
           return true;
         }
         auto linkPtrPhys = this->entityLinkMap.Get(_parent->Data());
+
+        if (const auto existingShape = linkPtrPhys->GetShape(_name->Data()))
+        {
+          // No need to create this collision shape because it was already
+          // created when parsing the model.
+
+          // TODO(MXG): There's some significant design inconsistency in this
+          // CreateCollisionEntities function. There's a mix of using the
+          // ConstructSdfCollision feature and other collision attachment
+          // features like AttachHeightMapFeature and AttachMeshShape. This
+          // makes it confusing for physics plugin implementers to know where
+          // and how to implement the construction of collision shapes.
+          //
+          // In the current implementation, if a physics engine implements
+          // constructing a mesh or heightmap in ConstructSdfCollision then that
+          // ability won't actually get leveraged by gz-sim because it
+          // specifically uses AttachMeshShapeFeature and AttachHeightMapFeature
+          // for those shapes. The physics plugin implementer needs to know to
+          // implement those specific features instead.
+          this->topLevelModelMap.insert(
+            std::make_pair(_entity, topLevelModel(_entity, _ecm)));
+          return true;
+        }
 
         // Make a copy of the collision DOM so we can set its pose which has
         // been resolved and is now expressed w.r.t the parent link of the
@@ -1559,12 +1623,11 @@ void PhysicsPrivate::CreateJointEntities(const EntityComponentManager &_ecm,
                   << "] not found on model map." << std::endl;
           return true;
         }
-        auto modelPtrPhys = this->entityModelMap.Get(_parentModel->Data());
 
-        auto modelJointFeature =
+        auto modelPtrPhys =
             this->entityModelMap.EntityCast<JointFeatureList>(
                 _parentModel->Data());
-        if (!modelJointFeature)
+        if (!modelPtrPhys)
         {
           static bool informed{false};
           if (!informed)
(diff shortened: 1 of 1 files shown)


src/systems/physics/Physics.cc
```diff
@@ -1280,6 +1321,29 @@ void PhysicsPrivate::CreateCollisionEntities(const EntityComponentManager &_ecm,
...
+          // features like AttachHeightMapFeature and AttachMeshShape. This
+          // makes it confusing for physics plugin implementers to know where
+          // and how to implement the construction of collision shapes.
+          //
+          // In the current implementation, if a physics engine implements
+          // constructing a mesh or heightmap in ConstructSdfCollision then that
+          // ability won't actually get leveraged by gz-sim because it
+          // specifically uses AttachMeshShapeFeature and AttachHeightMapFeature
+          // for those shapes. The physics plugin implementer needs to know to
+          // implement those specific features instead.
```
scpeters, line 1341: the reason for this is that the mesh and heightmap SDFormat elements contain URI's that in many cases point to fuel resources, but libsdformat doesn't depend on fuel. The `Attach*ShapeFeature`s accept `gz::common` types that contain the mesh and heightmap info, and `gz-sim` is responsible for fetching the data and populating those types.

Properly implementing the `ConstructSdfMesh` and `ConstructSdfHeightmap` would involve resolving those URIs, which seems beyond the responsibility of a physics plugin, so perhaps we should just deprecate those features from gz-physics

mxgrey: Yeah, I think the motivation totally makes sense.

I think what I'd propose is to move the implementations of the uri resolving functionality into a reusable utility function in the `SDFormat` library. Maybe even have them as methods in `MeshShape` and `HeightmapShape` classes for maximum visibility. The utility functions could just output the same data structures that are being passed into `AttachMeshShape` and `AttachHeightmapShape` so the physics plugin developer can easily recycle the implementations of those functions for `ConstructSdfCollision`.

ahcorde:
Joint is not placed in the right pose (`bullet-featherstone`), tested with `log_record_dbl_pendulum.sdf`

bullet-featherstone-bad-pendulim

mxgrey:
The seg fault issue with the classic bullet plugin is fixed by gazebosim/gz-physics#393 and the TPE failing to load is fixed by [4b01010](https://github.com/gazebosim/gz-sim/pull/1560/commits/4b010109285a9bb8582f66cd7b8e4de2e98e625f).

I think the joint offset issue for `bullet-featherstone` shouldn't be a blocker for this PR since that will need to be fixed in gazebosim/gz-physics#373 and wouldn't involve any changes to this PR.

mxgrey:

src/systems/physics/Physics.cc
```diff
@@ -166,9 +167,10 @@ class gz::sim::systems::PhysicsPrivate
           gz::physics::LinkFrameSemantics,
           gz::physics::ForwardStep,
           gz::physics::RemoveModelFromWorld,
-          gz::physics::sdf::ConstructSdfLink,
           gz::physics::sdf::ConstructSdfModel,
-          gz::physics::sdf::ConstructSdfWorld
+          gz::physics::sdf::ConstructSdfWorld,
+          gz::physics::GetLinkFromModel,
+          gz::physics::GetShapeFromLink
```
chapulina, line 173: Note that modifying `MinimumFeatureList` isn't backwards compatible, as documented above. So I'm adding the `Breaking change` label.

src/systems/physics/Physics.cc
chapulina: I've been confused by this in the past (https://github.com/gazebosim/gz-physics/issues/9):

"_For example, what does [ConstructCollision](https://github.com/ignitionrobotics/ign-physics/blob/a99c99d267972afee4501c061e31dd9821fb131d/sdf/include/ignition/physics/sdf/ConstructCollision.hh#L36) do? Does it create and attach a collision to a link? How is that different from [AttachMeshShape](https://github.com/ignitionrobotics/ign-physics/blob/a99c99d267972afee4501c061e31dd9821fb131d/mesh/include/ignition/physics/mesh/MeshShape.hh#L106)? If they do similar things, why are their names so different?_"

> mesh and heightmap SDFormat elements contain URI's that in many cases point to fuel resources, but libsdformat doesn't depend on fuel

SDF has the ability of resolving Fuel URIs for `<include>`s, we pass it a Fuel-Tools callback so it knows how to resolve those:

https://github.com/gazebosim/gz-sim/blob/417609c88d7110864a7a10ae7ea9f6d7ebf924ab/src/Server.cc#L66-L68

So it could also resolve the URIs for other assets.

I think the main problem is that SDFormat doesn't depend on Gazebo Common, which is a huge dependency and brings external graphics libraries, so it can't return  `gz::common` types.

I guess one option is passing those resolved URIs to `gz-rendering` and `gz-physics`, and each implementation decides how to load the assets. They can use `gz::common` if they want. It may cause a lot of duplication across various plugins though, or inconsistencies on how the meshes are loaded if they use different loaders.

mjcarroll:
@mxgrey we have a whole slew of test failures here that aren't in `main`.  I spent a little bit of time, but didn't make much headway.

chapulina:

This PR can either be merged into `main` and backported to `gz-physics6` after the stable release, or retargeted to `gz-physics6` and merged after the stable release (i.e. October).

chapulina:

mjcarroll:

mjcarroll:

src/systems/physics/Physics.cc
scpeters: > I think the main problem is that SDFormat doesn't depend on Gazebo Common, which is a huge dependency and brings external graphics libraries, so it can't return `gz::common` types.

yeah, I think this is the real crux of the problem. Perhaps we could add basic Mesh and HeightmapData classes to gz-math, while keeping the parsers in gz-common? Then libsdformat could return the gz-math types, and the find-file callback function could use fuel and gz-common's parsers to resolve the URI into the math object

azeey: I think @mjcarroll has a solution with https://github.com/gazebosim/sdformat/pull/1147 so that the SDF DOM object passed into gz-physics has all the resource paths resolved.

scpeters: > I think @mjcarroll has a solution with [gazebosim/sdformat#1147](https://github.com/gazebosim/sdformat/pull/1147) so that the SDF DOM object passed into gz-physics has all the resource paths resolved.

that means we don't need to use fuel, but it's still just a file path provided by `libsdformat`, but something will still need to parse that file, and in general we can't assume that the physics implementation will be able to parse it, so I think that wouldn't resolve the matter of properly implementing `ConstructSdfMesh` or `ConstructSdfHeightmap` in the general case so that `AttachMeshShape` and `AttacheHeightmapShape` aren't needed

src/systems/physics/Physics.cc
```diff
@@ -1312,6 +1344,34 @@ void PhysicsPrivate::CreateCollisionEntities(const EntityComponentManager &_ecm,
...
+          // constructing a mesh or heightmap in ConstructSdfCollision then that
+          // ability won't actually get leveraged by gz-sim because it
+          // specifically uses AttachMeshShapeFeature and AttachHeightMapFeature
+          // for those shapes. The physics plugin implementer needs to know to
+          // implement those specific features instead.
+          auto linkCollisionFeature =
+              this->entityLinkMap.EntityCast<CollisionFeatureList>(
+                  _parent->Data());
+          this->entityCollisionMap.AddEntity(
+            _entity, linkCollisionFeature->GetCollision(_name->Data()));
```
scpeters, line 1369: I changed `GetCollision` to `GetShape` on this line, and it works fine for me. I think we can revert the `GetCollision` API from `gz/physics/sdf/ConstructCollision.hh`

ahcorde:
| | |
|------|-----|
| ackermann_steering | ok |
| acoustic_comms_demo | vehicles are not moving properly |
| acoustic_comms_moving_targets.sdf  | wierd |
| acoustic_comms.sdf   | ok |
| actor_crowd | ok |
| linear_battery_demo.sdf | ok |
| logical_audio_sensor_plugin.sdf | ok |
| logical_camera_sensor.sdf | ok |
| log_record_dbl_pendulum.sdf | ok |
| log_record_keyboard.sdf | ok |
| actor_crowd.sdf  | ok |
| log_record_resources.sdf | ok |
| actor.sdf  | ok |
| log_record_shapes.sdf | ok |
| actors_population.sdf    | ok |
| mecanum_drive.sdf | not working |
| minimal_scene.sdf | ok |
| apply_joint_force.sdf    | vel can be setted but when zero is setted the joint will not stop (damping is not applied) |
| model_photo_shoot.sdf | segfault |
| apply_link_wrench.sdf    | ok |
| multicopter_velocity_control.sdf | small quad can takeoff but the hexacopter cannot |
| auv_controls.sdf | doing the wrong thing |
| multi_lrauv_race.sdf | the vehicle in the middle is not working properly the other two sometimes are doing wierd stuff too |
| boundingbox_camera.sdf   | ok |
| nested_model.sdf | ok |
| breadcrumbs.sdf  | box is getting some initial speed and the vehicle is not include in the physics engine |
| optical_tactile_sensor_plugin.sdf | ok |
| buoyancy_engine.sdf  | boxes are moving up and down, not sure if this is ok |
| particle_emitter.sdf | ok |
| buoyancy.sdf | motor is rotating but the model is stuck in the same position |
| pendulum_links.sdf | removing the velocity plugin works well, commands are doing wierd things |
| camera_sensor.sdf    | ok |
| perfect_comms.sdf | ok |
| camera_video_record_dbl_pendulum.sdf   | ok |
| performer_detector.sdf | ok |
| physics_options.sdf | ok |
| collada_world_exporter.sdf | ok |
| plane_propeller_demo.sdf | plane doing wierd things, crazy since the beginning of the simulation |
| contact_sensor.sdf   | ok |
| plot_3d.sdf | ok |
| conveyor.sdf | feature not implemented |
| polylines.sdf | polyline not supported as a collision shape |
| debug_shapes.sdf | ok |
| pose_publisher.sdf | ok |
| default.sdf  | ok |
| quadcopter.sdf | ok |
| dem_monterey_bay.sdf | not supported |
| rf_comms.sdf | ok |
| dem_moon.sdf | not supported |
| rolling_shapes.sdf | ok |
| dem_volcano.sdf  | not supported |
| segmentation_camera.sdf | ok |
| deprecated_ignition.sdf  | ok |
| sensors_demo.sdf | ok |
| depth_camera_sensor.sdf  | ok |
| sensors.sdf | ok |
| detachable_joint.sdf | ok |
| shader_param.sdf | segfautl |
| diff_drive.sdf   | ok |
| shapes_bitmask.sdf | not implemented |
| diff_drive_skid.sdf | ok |
| elevator.sdf | ok (need to adjust PID) |
| shapes.sdf | ok |
| empty_gui.sdf    | ok |
| skid_steer_mecanum.sdf | not turning properly |
| empty.sdf  | ok |
| sky.sdf | ok |
| follow_actor.sdf | ok |
| spaces.sdf | ok |
| fuel.sdf   | ok (heightmap not supported) |
| spherical_coordinates.sdf | when spawn a model, the model is crazy |
| fuel_textured_mesh.sdf   | ok |
| thermal_camera.sdf | error reading world, might be related with the migration, this world should work |
| gpu_lidar_retro_values_sensor.sdf | ok |
| gpu_lidar_sensor.sdf | ok |
| touch_plugin.sdf | ok |
| graded_buoyancy.sdf  | some physics errors |
| track_drive.sdf | models are crazy |
| grid.sdf   | ok |
| tracked_vehicle_simple.sdf | segfault |
| heightmap.sdf    | not supported |
| trajectory_follower.sdf | working but not doing the same as dartsim |
| import_mesh.sdf  | ok |
| triggered_camera_sensor.sdf | ok |
| joint_controller.sdf | ok |
| triggered_publisher.sdf | ok |
| joint_position_controller.sdf  | ok |
| trisphere_cycle_wheel_slip.sdf | ok |
| joint_trajectory_controller.sdf    | ok |
| tunnel.sdf | segfault |
| kinetic_energy_monitor.sdf | not working (missing features?) |
| velocity_control.sdf | ok |
| levels_no_performers.sdf |
| video_record_dbl_pendulum.sdf | ok |
| levels.sdf   | ok |
| visibility.sdf | ok |
| lift_drag_battery.sdf    | joint rotating but not taking off |
| visualize_contacts.sdf | ok |
| lift_drag_nested.sdf | issues building model |
| visualize_lidar.sdf | ok (car jumps at the beginning) |
| lift_drag.sdf    | joint rotates but the model is not taking off |
| wide_angle_camera.sdf | ok |
| lightmap.sdf | not able to see anything in the minimal scene (rendering issue?) |
| wind.sdf | ok |
| lights.sdf | ok |
| world_joint.sdf | segfault |


NOTES:
 - There are some tutorials to record logs but it will be great to add some instructions to the sdf files

src/systems/physics/Physics.cc
ahcorde: https://github.com/gazebosim/gz-physics/pull/444

mjcarroll:

livanov93:
@mxgrey @ahcorde What needs to be done to merge this one?

mjcarroll:

mjcarroll:
> @mxgrey @ahcorde What needs to be done to merge this one?

I _believe_ that I got all the tests passing, but CI isn't reflecting that yet, giving it another go.

mjcarroll:

ahcorde (approved):

PR #1563, opened 2022-06-30
Remove deprecated image data api

luca-della-vedova:
# 🦟 Bug fix

## Summary
Same as the [gz-gui PR](https://github.com/gazebosim/gz-gui/pull/430). Migrate to new and safer Image API to avoid deprecation warnings.
Depends on [gz-common PR](https://github.com/gazebosim/gz-common/pull/372).

In the process I actually found out the test that was using the deprecated API was actually doing nothing since it was comparing an image to itself. I fixed that so the test should now do what it's supposed to, with the consequence that it actually fails since I guess bytewise equality in a whole image is quite a strict constraint.
I'm not too familiar with that system so I'm not sure if the test can be fixed or should be disabled.

## Checklist
- [ ] Added tests


luca-della-vedova:
Replaced by https://github.com/gazebosim/gz-sim/pull/1570.
Will be effective as of next gazebo6->main forward port

PR #1580, opened 2022-07-07
Fix component updates

### LINKED ISSUE #1022: Modifying lights in component inspector not working
## Environment
* OS Version: Ubuntu 20.04
* Source or binary build?
  
Source, `main` branch, commit 1fb320715525a86361f2710ffa6aadb4c56a58c0

## Description
* Expected behavior: Lights to change color when using the component inspector
* Actual behavior: Most of the time nothing happens (sometimes will change color but only seems to be once). When changing one of the RGBA sliders then changing another, the first slider will reset back. For example, if I change R then B, R seems to reset back.

Peek 2021-09-10 16-08

## Steps to reproduce

1. Apply fix from https://github.com/ignitionrobotics/ign-gazebo/pull/1018
2. `ign gazebo -v 4 lights.sdf`
3. Attempt to change the color of any of the lights

jennuine:
# 🦟 Bug fix

Fixes #1022

## Summary

To speed up performance, removed components were cached/stored here: 
https://github.com/gazebosim/gz-sim/blob/1f6c1263fea66cd1736f420741acbe030a3f3089/include/ignition/gazebo/detail/View.hh#L140-L149

Then when the same component was re-added: https://github.com/gazebosim/gz-sim/blob/1f6c1263fea66cd1736f420741acbe030a3f3089/src/EntityComponentManager.cc#L1088-L1089

The component data was moved from `invalidData` to `validData` but the data was never updated.

The fix is to check that component needs updated in the `EntityComponentManager::SetState` functions then retrieve the component and deserialize the data again (to update the re-added component).

---

To run unit tests:
```bash
./path/to/build/UNIT_EntityComponentManager_TEST --gtest_filter="*AddRemoveAddComponents*"
```
---

To test through the GUI: 
```bash
ign gazebo -v 4 lights.sdf
```

**Before PR**
before_pr

**After PR**
after_pr

Thanks @chapulina for the help!

## Checklist
- [x] Added tests

### DIFF
--- a/src/EntityComponentManager.cc
+++ b/src/EntityComponentManager.cc
@@ -1085,6 +1085,7 @@ bool EntityComponentManager::CreateComponentImplementation(
 
       for (auto &viewPair : this->dataPtr->views)
       {
+        viewPair.second.first->UpdateComponent(_entity, _data);
         viewPair.second.first->NotifyComponentAddition(_entity,
             this->IsNewEntity(_entity), _componentTypeId);
       }
--- a/src/EntityComponentManager.cc
+++ b/src/EntityComponentManager.cc
@@ -1735,7 +1735,6 @@ void EntityComponentManager::SetState(
       // Get Component
       auto comp = this->ComponentImplementation(entity, type);
 
-
       // Create if new
       if (nullptr == comp)
       {
--- a/src/EntityComponentManager.cc
+++ b/src/EntityComponentManager.cc
@@ -1838,17 +1838,19 @@ void EntityComponentManager::SetState(
 
         auto updateData = this->CreateComponentImplementation(entity,
             newComp->TypeId(), newComp.get());
-        if (updateData)
-        {
-          // Set comp so we deserialize the data below again
-          comp = this->ComponentImplementation(entity, compIter.first);
-        }
+        // if (updateData)
+        // {
+        //   // Set comp so we deserialize the data below again
+        //   comp = this->ComponentImplementation(entity, compIter.first);
+        // }
       }
       // Update component value
-      if (comp)
+      // if (comp)
+      else
       {
-        std::istringstream istr2(compMsg.component());
-        comp->Deserialize(istr2);
+        // std::istringstream istr2(compMsg.component());
+        // comp->Deserialize(istr2);
+        comp->Deserialize(istr);
         this->SetChanged(entity, compIter.first,
             _stateMsg.has_one_time_component_changes() ?
             ComponentState::OneTimeChange :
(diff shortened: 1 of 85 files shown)


src/EntityComponentManager.cc
```diff
@@ -493,7 +493,16 @@ Entity EntityComponentManager::CloneImpl(Entity _entity, Entity _parent,
...
 
-    this->CreateComponentImplementation(clonedEntity, type, clonedComp.get());
+    auto updateData =
+      this->CreateComponentImplementation(clonedEntity, type, clonedComp.get());
+    if (updateData)
+    {
+      // When a cloned entity is removed, it erases all components/data so a new
+      // cloned entity should not have components to be updated
+      ignerr << "The component's data needs to be updated but this is not "
+             << "expected. Please submit an issue including your use case "
```
ahcorde, line 503: What do you think about adding here the url the github repo ?

ahcorde (changes requested):
I can confirm that this is working like a charm! Thank you!

I had a minor comment about the error message

src/EntityComponentManager.cc
chapulina: Or alternatively, this could say something like "... not expected. This is an internal error."

I don't think we usually ask that users submit users, it's kind of implied. What do you think?

src/systems/user_commands/UserCommands.cc
```diff
@@ -1823,13 +1823,23 @@ bool VisualCommand::Execute()
...
 
-  if (visualMsg->id() == kNullEntity)
+  Entity visualEntity = kNullEntity;
+  std::string entityName = visualMsg->name();
+  if (!entityName.empty())
+  {
+    visualEntity =
+        this->iface->ecm->EntityByComponents(components::Name(entityName));
+  }
+  else if (visualMsg->id() != kNullEntity)
```
chapulina, line 1833: It's better to check the ID before the name. IDs are unique, while names aren't. The name usually needs to be used in conjunction with the `parent_name` or `parent_id` to narrow down what the exact entity.

src/EntityComponentManager.cc
jennuine: bb7f37e0289cdde223571b2444a7be71594a55f0 I modified the error message to follow this format: https://github.com/gazebosim/gz-sim/blob/afd087055f87d3e4c199d8e786fbccd68d3c80c1/src/EntityComponentManager.cc#L1165-L1167

src/systems/user_commands/UserCommands.cc
jennuine: I used `parent_name`  bb7f37e0289cdde223571b2444a7be71594a55f0

PR #1595, opened 2022-07-15
Helper function to get an entity from an entity message

chapulina:
# 🎉 New feature

* Broken off https://github.com/gazebosim/gz-sim/pull/1593
* Requires https://github.com/gazebosim/gz-msgs/pull/275

## Summary

The `Entity` message has some [documentation](https://github.com/gazebosim/gz-msgs/blob/e8f5791a5d721b8a241e7934f233da00b4a6e4e8/proto/ignition/msgs/entity.proto#L25-L36) explaining how the message is meant to be used, but we don't have a helper function implementing that, leaving it up to plugin developers to parse the message as they see fit.

Over time, we've accumulated a couple of  different implementations for the message parsing, and it would be good to standardize it in a single place. On `ign-gazebo6`, there are a couple of implementations in the `UserCommands` system:

* [topLevelEntityFromMessage](https://github.com/gazebosim/gz-sim/blob/f39bc25da4a93a0b2ccf5fca8c3e1b6863185a9d/src/systems/user_commands/UserCommands.cc#L96-L149)
* [scopedEntityFromMsg](https://github.com/gazebosim/gz-sim/blob/ign-gazebo6/src/systems/user_commands/UserCommands.cc#L1857-L1918)

The implementation here is a combination of the 2 above, which matches the description of the message.

## Test it

See the added tests.

## Checklist
- [x] Added tests

### DIFF
--- a/src/Util.cc
+++ b/src/Util.cc
@@ -527,6 +527,86 @@ std::string validTopic(const std::vector<std::string> &_topics)
   }
   return std::string();
 }
+
+//////////////////////////////////////////////////
+Entity entityFromMsg(const EntityComponentManager &_ecm,
+    const msgs::Entity &_msg)
+{
+  if (_msg.id() != kNullEntity)
+  {
+    return _msg.id();
+  }
+
+  // If there's no ID, check name + type
+  if (_msg.type() == msgs::Entity::NONE)
+  {
+    return kNullEntity;
+  }
+
+  auto entities = entitiesFromScopedName(_msg.name(), _ecm);
+  if (entities.empty())
+  {
+    return kNullEntity;
+  }
+
+  for (const auto entity : entities)
+  {
+    if (_msg.type() == msgs::Entity::LIGHT &&
+        _ecm.Component<components::Light>(entity))
+    {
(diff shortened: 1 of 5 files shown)


src/Util.cc
```diff
@@ -527,6 +527,86 @@ std::string validTopic(const std::vector<std::string> &_topics)
...
+    return kNullEntity;
+  }
+
+  auto entities = entitiesFromScopedName(_msg.name(), _ecm);
+  if (entities.empty())
+  {
+    return kNullEntity;
+  }
+
+  for (const auto entity : entities)
```
ahcorde, line 552:
```suggestion
  for (const auto &entity : entities)
```
```suggestion
  for (const auto entity : entities)
```

ahcorde (changes requested):

src/Util.cc
chapulina: I think this isn't very important because `Entity` is just a POD. But I guess this could change in the future, so it doesn't hurt

ahcorde (approved):

PR #1608, opened 2022-07-22
Acoustic comms plugin

adityapande-1995:

# 🎉 New feature

## Summary
This plugin adds a simple acoustic comms plugin to gz-sim.

## Test it
Added a small test case which contains 2 pairs of boxes which communicate with each other. One set in inside the max range of comms, and the other is outside it. The one outside the range should not receive any messages.

## TODO
* The current calculation for ``distanceToTransmitter`` considers the current distance between the 2 bodies, when actually it should be the distance between current receiver position and position of the sender at the time at which the packet was sent, as the sender might be moving. This is not a big problem if the 2 bodies are at rest or moving slowly compared to speed of sound.

Since the packet does not contain a field related to position of sender at the time the packet was sent, I;m not sure how to implement this. 

## Checklist
- [x] Added tests

### DIFF
--- a/src/systems/acoustic_comms/AcousticComms.cc
+++ b/src/systems/acoustic_comms/AcousticComms.cc
@@ -37,7 +37,7 @@ class AcousticComms::Implementation
   // Default max range for acoustic comms in metres.
   public: double maxRange = 1000.0;
 
-  // Default speed of sound in air (m/s).
+  // Default speed of sound in air in metres/sec.
   public: double speedOfSound = 343.0;
 };
 
--- a/src/systems/acoustic_comms/AcousticComms.cc
+++ b/src/systems/acoustic_comms/AcousticComms.cc
@@ -20,8 +20,6 @@
  * Research Institute (MBARI) and the David and Lucile Packard Foundation
  */
 
-#include <cmath>
-
 #include <gz/common/Profiler.hh>
 #include <gz/plugin/Register.hh>
 #include <gz/sim/Entity.hh>
@@ -41,20 +39,7 @@ class AcousticComms::Implementation
 
   // Default speed of sound in air (m/s).
   public: double speedOfSound = 343.0;
-
-  public: double DistanceBetweenBodies(
-              math::Vector3<double> _src,
-              math::Vector3<double> _dst);
 };
-
-double AcousticComms::Implementation::DistanceBetweenBodies(
-    math::Vector3<double> _src, math::Vector3<double> _dst)
-{
-  double x = _src.X() - _dst.X();
-  double y = _src.Y() - _dst.Y();
-  double z = _src.Z() - _dst.Z();
-  return sqrt(pow(x, 2) + pow(y, 2) + pow(z, 2));
-}
 
 AcousticComms::AcousticComms()
   : dataPtr(gz::utils::MakeUniqueImpl<Implementation>())
@@ -136,21 +121,21 @@ void AcousticComms::Step(
         auto poseDst = worldPose(itDst->second.entity, _ecm).Pos();
 
         auto distanceToTransmitter =
-          this->dataPtr->DistanceBetweenBodies(poseSrc,
-              poseDst);
+          (poseSrc - poseDst).Length();
 
         // Calculate distance covered by the message.
         std::chrono::steady_clock::time_point currTime(_info.simTime);
         auto timeOfTransmission = msg->mutable_header()->stamp();
 
-        double currTimestamp =
-          currTime.time_since_epoch().count() / 1000000000.0;
-        double packetTimestamp =
-          static_cast<double>(timeOfTransmission.sec()) +
-          static_cast<double>(timeOfTransmission.nsec()) / 1000000000.0;
+        auto currTimestamp =
+          std::chrono::nanoseconds(currTime.time_since_epoch().count());
+        auto packetTimestamp =
+          std::chrono::seconds(timeOfTransmission.sec()) +
+          std::chrono::nanoseconds(timeOfTransmission.nsec());
 
-        double deltaT = currTimestamp - packetTimestamp;
-        double distanceCoveredByMessage = deltaT *
+        std::chrono::duration<double> deltaT =
+          currTimestamp - packetTimestamp;
+        double distanceCoveredByMessage = deltaT.count() *
           this->dataPtr->speedOfSound;
 
         // Only check msgs that haven't exceeded the maxRange.
--- a/src/systems/acoustic_comms/AcousticComms.cc
+++ b/src/systems/acoustic_comms/AcousticComms.cc
@@ -33,26 +33,29 @@ using namespace gz;
 using namespace sim;
 using namespace systems;
 
+/// \brief Private Acoustic comms data class.
 class AcousticComms::Implementation
 {
-  // Default max range for acoustic comms in metres.
+  /// \brief Default max range for acoustic comms in metres.
   public: double maxRange = 1000.0;
 
-  // Default speed of sound in air in metres/sec.
+  /// \brief Default speed of sound in air in metres/sec.
   public: double speedOfSound = 343.0;
 
-  // Position of the transmitter at the time the message was
-  // sent, or first processed.
+  /// \brief Position of the transmitter at the time the message was
+  /// sent, or first processed.
   public: std::unordered_map
           <std::shared_ptr<msgs::Dataframe>, math::Vector3d>
           poseSrcAtMsgTimestamp;
 };
 
+//////////////////////////////////////////////////
 AcousticComms::AcousticComms()
   : dataPtr(gz::utils::MakeUniqueImpl<Implementation>())
 {
 }
 
+//////////////////////////////////////////////////
 void AcousticComms::Load(
     const Entity &_entity,
     std::shared_ptr<const sdf::Element> _sdf,
@@ -69,6 +72,7 @@ void AcousticComms::Load(
   }
 }
 
+//////////////////////////////////////////////////
 void AcousticComms::Step(
     const UpdateInfo &_info,
     const comms::Registry &_currentRegistry,
--- a/src/systems/acoustic_comms/AcousticComms.cc
+++ b/src/systems/acoustic_comms/AcousticComms.cc
@@ -19,6 +19,7 @@
  * Development of this module has been funded by the Monterey Bay Aquarium
  * Research Institute (MBARI) and the David and Lucile Packard Foundation
  */
+#include <unordered_map>
 
 #include <gz/common/Profiler.hh>
 #include <gz/plugin/Register.hh>
@@ -39,6 +40,12 @@ class AcousticComms::Implementation
 
   // Default speed of sound in air in metres/sec.
   public: double speedOfSound = 343.0;
+
+  // Position of the transmitter at the time the message was
+  // sent, or first processed.
+  public: std::unordered_map
+          <std::shared_ptr<msgs::Dataframe>, math::Vector3d>
+          poseSrcAtMsgTimestamp;
 };
 
 AcousticComms::AcousticComms()
@@ -124,24 +131,17 @@ void AcousticComms::Step(
         // If it has reached neither the destination nor the maxRange,
         // it is considered in transit.
 
-        // Check if the message header contains the position of sender
-        // at the time the message was sent. If yes, use that as poseSrc,
-        // else use current position of the sender.
-        math::Vector3d poseSrc =
-          worldPose(itSrc->second.entity, _ecm).Pos();
-
-        for (int i = 0; i < msg->header().data_size(); i++)
+        if (this->dataPtr->poseSrcAtMsgTimestamp.count(msg) == 0)
         {
-          if (msg->header().data(i).key() == "transmitter_position"
-              && msg->header().data(i).value_size() == 3)
-          {
-            auto poseSrcX = std::stod(msg->header().data(i).value().Get(0));
-            auto poseSrcY = std::stod(msg->header().data(i).value().Get(1));
-            auto poseSrcZ = std::stod(msg->header().data(i).value().Get(2));
-            poseSrc = math::Vector3d(poseSrcX, poseSrcY, poseSrcZ);
-            break;
-          }
+          // This message is being processed for the first time.
+          // Record the current position of the sender and use it
+          // for distance calculations.
+          this->dataPtr->poseSrcAtMsgTimestamp[msg] =
+            worldPose(itSrc->second.entity, _ecm).Pos();
         }
+
+        math::Vector3d poseSrc =
+          this->dataPtr->poseSrcAtMsgTimestamp[msg];
 
         // Calculate distance between the bodies.
         auto poseDst = worldPose(itDst->second.entity, _ecm).Pos();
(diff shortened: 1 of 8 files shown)


src/systems/acoustic_comms/AcousticComms.cc
```diff
@@ -0,0 +1,186 @@
...
+};
+
+double AcousticComms::Implementation::DistanceBetweenBodies(
+    math::Vector3<double> _src, math::Vector3<double> _dst)
+{
+  double x = _src.X() - _dst.X();
+  double y = _src.Y() - _dst.Y();
+  double z = _src.Z() - _dst.Z();
+  return sqrt(pow(x, 2) + pow(y, 2) + pow(z, 2));
+}
```
hidmic, line 57: @adityapande-1995 hmm, why not `(_src - _dst).Length()`?

src/systems/acoustic_comms/AcousticComms.cc
```diff
@@ -0,0 +1,186 @@
...
+
+        // Calculate distance covered by the message.
+        std::chrono::steady_clock::time_point currTime(_info.simTime);
+        auto timeOfTransmission = msg->mutable_header()->stamp();
+
+        double currTimestamp =
+          currTime.time_since_epoch().count() / 1000000000.0;
+        double packetTimestamp =
+          static_cast<double>(timeOfTransmission.sec()) +
+          static_cast<double>(timeOfTransmission.nsec()) / 1000000000.0;
```
hidmic, line 150: @adityapande-1995 hmm, consider using `std::chrono` types wherever possible. Floating point epsilons can be problematic. 

```suggestion
        auto packetTimestamp =
          std::chrono::seconds(timeOfTransmission.sec()) +
          std::chrono::nanoseconds(timeOfTransmission.nsec());
```

test/integration/acoustic_comms.cc
```diff
@@ -0,0 +1,188 @@
...
+
+  // Verify subscriber received no msgs.
+  int sleep = 0;
+  while (sleep++ < 5)
+  {
+    std::lock_guard<std::mutex> lock(mutex);
+    std::this_thread::sleep_for(std::chrono::milliseconds(100));
+  }
+  EXPECT_EQ(0, msgCounter);
+}
```
hidmic, line 188: @adityapande-1995 meta: it may be worth adding a test with a moving sender and another test with a moving receiver.

src/systems/acoustic_comms/AcousticComms.cc
```diff
@@ -0,0 +1,186 @@
...
+
+        // Is the destination address attached to a model?
+        bool dstAddressAttachedToModel =
+          dstAddressBound && itDst->second.entity != kNullEntity;
+
+        if (!dstAddressAttachedToModel)
+          continue;
+
+        // Calculate distance between the bodies.
+        auto poseSrc = worldPose(itSrc->second.entity, _ecm).Pos();
```
hidmic, line 135: @adityapande-1995 like you've pointed out elsewhere, we need the position of the sender at the time the message was emitted to model attenuation. Generally speaking, different communication models may need different bits of information  to be sent with the messages. I'm not sure what @caguero had in mind, but I'd think we need to compose messages. Type erasure would be handy, but templating the comms infrastructure would do too.

src/systems/acoustic_comms/AcousticComms.cc
adityapande-1995: Done https://github.com/gazebosim/gz-sim/pull/1608/commits/8cb0a46f91b638ed3387697c2864afa499428156

test/integration/acoustic_comms.cc
adityapande-1995: Done : https://github.com/gazebosim/gz-sim/pull/1608/commits/920506df253f0ea251ae5b9a6e7d5bccc0522594

src/systems/acoustic_comms/AcousticComms.cc
adityapande-1995: Checks for ``transmitter_position`` in header of the dataframe msg : https://github.com/gazebosim/gz-sim/pull/1608/commits/aa779fa203d3613e72ef305848ba157c1aa7e53f

caguero (commented):
At a high level, the code looks good to me and I like the strategy for deciding when to keep the message, drop it or deliver it.

My only suggestion is to try to remove the need for the transmitter to populate the message with its own pose. I think it would be way more elegant if that's something that the plugin can do automatically for us. We have access to the ECM and the entity ID of the sender, so it should be possible.

adityapande-1995:
As per offline discussion with @caguero , we decided that making the users specify the position of the sender in the header file is cumbersome. Instead, what I have now is : The plugin will keep track the position of sender when a msg is first processed, and use that for distance calculations throughout the life of the message. 

This means the difference between actual position of the sender (when the msg was sent) and the one we use will only differ by one spin cycle of the plugin, which is still good.

src/systems/acoustic_comms/AcousticComms.cc
```diff
@@ -0,0 +1,204 @@
...
+  public: double speedOfSound = 343.0;
+
+  // Position of the transmitter at the time the message was
+  // sent, or first processed.
+  public: std::unordered_map
+          <std::shared_ptr<msgs::Dataframe>, math::Vector3d>
+          poseSrcAtMsgTimestamp;
+};
+
+AcousticComms::AcousticComms()
```
caguero, line 51: Nit: Use the delimiters on top of each function.
```
```

src/systems/acoustic_comms/AcousticComms.cc
```diff
@@ -0,0 +1,204 @@
...
+          else
+          {
+            // This message is still in transit, should be kept in the
+            // outbound buffer of source and not moved to inbound of
+            // the destination.
+            newOutbound.push_back(msg);
+          }
+        }
+        else
+        {
```
caguero, line 184: I think there's a situation that isn't covered here: If source and destination are at a distance > `maxRange` but the destination is moving closer to the sender (in the direction of the message). If I'm not wrong, you'll discard the message instantly but you're not accounting that potentially the destination might be within range while the message is traveling.

adityapande-1995: I am actually, ``distanceToTransmitter`` is the distance between :
* Position of sender when the message was sent, (this is constant throughout the lifetime of the message)
* Current position of the destination. This is when the plugin is spinning, so always contains the current position of destination. So even if the destination has just moved towards the source, that distance should be reflected accurately.

caguero: You're right, the key is to use `distanceCoveredByMessage` to decide when to drop the message and not `distanceToTransmitter` as I originally thought you were using.

src/systems/acoustic_comms/AcousticComms.cc
adityapande-1995: Done : https://github.com/gazebosim/gz-sim/pull/1608/commits/3db98f2e47a03311b01cdcac79f9be6e8b14d8e3

chapulina:

adityapande-1995:

adityapande-1995:
I don't think the failures in CI are related to the changes made here.

test/integration/acoustic_comms.cc
```diff
@@ -0,0 +1,346 @@
...
+  }
+
+  // Verify subscriber received all msgs.
+  int sleep = 0;
+  bool done = false;
+  while (!done && sleep++ < 3)
+  {
+    std::lock_guard<std::mutex> lock(mutex);
+    done = msgCounter == pubCount;
+    std::this_thread::sleep_for(std::chrono::milliseconds(100));
```
hidmic, line 113: @adityapande-1995 hmm, don't we have to unlock the mutex before sleeping? Same elsewhere.

test/integration/acoustic_comms.cc
```diff
@@ -0,0 +1,346 @@
...
+    server.Run(true, 100, false);
+  }
+
+  // Verify subscriber received no msgs.
+  int sleep = 0;
+  while (sleep++ < 3)
+  {
+    std::lock_guard<std::mutex> lock(mutex);
+    std::this_thread::sleep_for(std::chrono::milliseconds(100));
+  }
```
hidmic, line 186: @adityapande-1995 hmm, do we even have to lock and loop?

test/integration/acoustic_comms.cc
```diff
@@ -0,0 +1,346 @@
...
+  {
+    // Verify msg content
+    std::lock_guard<std::mutex> lock(mutex);
+    std::string expected = "hello world " + std::to_string(msgCounter);
+
+    gz::msgs::StringMsg receivedMsg;
+    receivedMsg.ParseFromString(_msg.data());
+    EXPECT_EQ(expected, receivedMsg.data());
+    msgCounter++;
+  };
```
hidmic, line 300: @adityapande-1995 meta: I see lots of duplication, can we use a fixture?

src/systems/acoustic_comms/AcousticComms.cc
```diff
@@ -0,0 +1,208 @@
...
+    EventManager &_eventMgr)
+{
+  if (_sdf->HasElement("max_range"))
+  {
+    this->dataPtr->maxRange = _sdf->Get<double>("max_range");
+  }
+  if (_sdf->HasElement("speed_of_sound"))
+  {
+    this->dataPtr->speedOfSound = _sdf->Get<double>("speed_of_sound");
+  }
```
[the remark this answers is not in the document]
adityapande-1995: Done https://github.com/gazebosim/gz-sim/pull/1608/commits/1546da990d97519c34043a91d8c7307065109d68

src/systems/acoustic_comms/AcousticComms.cc
```diff
@@ -0,0 +1,208 @@
...
+        if (this->dataPtr->poseSrcAtMsgTimestamp.count(msg) == 0)
+        {
+          // This message is being processed for the first time.
+          // Record the current position of the sender and use it
+          // for distance calculations.
+          this->dataPtr->poseSrcAtMsgTimestamp[msg] =
+            worldPose(itSrc->second.entity, _ecm).Pos();
+        }
+
+        math::Vector3d poseSrc =
```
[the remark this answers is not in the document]
adityapande-1995: Done https://github.com/gazebosim/gz-sim/pull/1608/commits/1546da990d97519c34043a91d8c7307065109d68

src/systems/acoustic_comms/AcousticComms.cc
```diff
@@ -0,0 +1,208 @@
...
+          this->dataPtr->poseSrcAtMsgTimestamp[msg];
+
+        // Calculate distance between the bodies.
+        auto poseDst = worldPose(itDst->second.entity, _ecm).Pos();
+        auto distanceToTransmitter =
+          (poseSrc - poseDst).Length();
```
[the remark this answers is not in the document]
adityapande-1995: Done https://github.com/gazebosim/gz-sim/pull/1608/commits/1546da990d97519c34043a91d8c7307065109d68

src/systems/acoustic_comms/AcousticComms.cc
```diff
@@ -0,0 +1,208 @@
...
+
+        // Calculate distance covered by the message.
+        std::chrono::steady_clock::time_point currTime(_info.simTime);
```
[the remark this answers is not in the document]
adityapande-1995: Done https://github.com/gazebosim/gz-sim/pull/1608/commits/1546da990d97519c34043a91d8c7307065109d68

src/systems/acoustic_comms/AcousticComms.cc
```diff
@@ -0,0 +1,208 @@
...
+        auto timeOfTransmission = msg->mutable_header()->stamp();
+
+        auto currTimestamp =
+          std::chrono::nanoseconds(currTime.time_since_epoch().count());
```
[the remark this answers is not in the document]
adityapande-1995: Done https://github.com/gazebosim/gz-sim/pull/1608/commits/1546da990d97519c34043a91d8c7307065109d68

test/integration/acoustic_comms.cc
adityapande-1995: Done https://github.com/gazebosim/gz-sim/pull/1608/commits/929dad6a4871be21eef1df68f1d1385b64eec0f7

adityapande-1995: Done https://github.com/gazebosim/gz-sim/pull/1608/commits/929dad6a4871be21eef1df68f1d1385b64eec0f7

test/integration/acoustic_comms.cc
adityapande-1995: Done https://github.com/gazebosim/gz-sim/pull/1608/commits/929dad6a4871be21eef1df68f1d1385b64eec0f7

test/integration/acoustic_comms.cc
adityapande-1995: Done https://github.com/gazebosim/gz-sim/pull/1608/commits/8f643c5cff0393961e2c7e68d5f4b1792a193efb

test/integration/acoustic_comms.cc
```diff
@@ -0,0 +1,155 @@
...
+using namespace sim;
+
+/////////////////////////////////////////////////
+class AcousticCommsTest : public InternalFixture<::testing::Test>
+{
+  public:
+    void CommsTestFixture(
+        std::string worldFile,
+        std::string srcAddr, std::string dstAddr,
+        int numMsgs = 3)
```
hidmic, line 41: @adityapande-1995 FYI: since you've fully parameterized the test, you could also inherit from `::testing::TestWithParam<AcousticCommsTestDefinition>` where:

```c++
struct AcousticCommsTestDefinition
{
    std::string worldFile;
    std::string srcAddr;
    std::string dstAddr;
    int numMsgs;
};
```

FTR here are the docs for [value-parameterized tests in `gtest`](https://google.github.io/googletest/advanced.html#value-parameterized-tests).

test/integration/acoustic_comms.cc
```diff
@@ -0,0 +1,155 @@
...
+      while (!done && sleep++ < 3)
+      {
+        {
+          std::lock_guard<std::mutex> lock(mutex);
+          done = (msgCounter == pubCount) && (pubCount != 0);
+        }
+        std::this_thread::sleep_for(std::chrono::milliseconds(100));
+      }
+      EXPECT_EQ(pubCount, msgCounter);
+    }
```
hidmic, line 118: @adityapande-1995 nit: consider
```suggestion
      bool done = false;
      for (int sleep = 0; !done && sleep < 3; ++sleep)
      {
        std::this_thread::sleep_for(std::chrono::milliseconds(100));
        {
          std::lock_guard<std::mutex> lock(mutex);
          done = (msgCounter == pubCount) && (pubCount != 0);
        }
      }
      EXPECT_TRUE(done);
      EXPECT_EQ(pubCount, msgCounter);
    }
```

hidmic (approved):
LGTM barring minor comments

test/integration/acoustic_comms.cc
adityapande-1995: Done https://github.com/gazebosim/gz-sim/pull/1608/commits/d99fc281e15767b744f58f01847e757076a6d042

test/integration/acoustic_comms.cc
adityapande-1995: Done https://github.com/gazebosim/gz-sim/pull/1608/commits/6aceed49c6a5ef97c9e4484b3eec276b1e2d1a05, looks much better now.

hidmic (approved):

PR #1616, opened 2022-07-27
Add EnvironmentalData component

hidmic:
# 🎉 New feature

## Summary

This patch adds an `EnvironmentalData` component that can hold a `DataFrame` worth of user-defined data for plugins to leverage. Additionally, preload and (interactive) load plugins are included.

Needs https://github.com/gazebosim/gz-common/pull/402.

## Test it

Reviewers may try the loader GUI plugin manually. 

## Checklist
- [x] Added tests


src/gui/plugins/environmental_data_loader/EnvironmentalDataLoader.cc
```diff
@@ -0,0 +1,250 @@
...
+
+  gui::App()->findChild<gui::MainWindow *>()->installEventFilter(this);
+}
+
+void EnvironmentalDataLoader::Update(const UpdateInfo &,
+                                     EntityComponentManager &_ecm)
+{
+  if (this->dataPtr->needsUpdate)
+  {
+    std::lock_guard<std::mutex> lock(this->dataPtr->mutex);
```
caguero, line 91: Should we move this lock before the `if` above?

src/gui/plugins/environmental_data_loader/EnvironmentalDataLoader.cc
```diff
@@ -0,0 +1,250 @@
...
+namespace gz
+{
+namespace sim
+{
+inline namespace GZ_SIM_VERSION_NAMESPACE
+{
+/// \brief Private data class for EnvironmentalDataLoader
+class EnvironmentalDataLoaderPrivate
+{
+  public: QString dataPath;
```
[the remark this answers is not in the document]
hidmic: Done in https://github.com/gazebosim/gz-sim/pull/1616/commits/267150c869fd5e866cfe0331357bd3852379f466.

src/gui/plugins/environmental_data_loader/EnvironmentalDataLoader.cc
```diff
@@ -0,0 +1,250 @@
...
+
+  public: std::mutex mutex;
+
+  public: std::atomic<bool> needsUpdate{false};
+};
+}
+}
+}
+
+EnvironmentalDataLoader::EnvironmentalDataLoader()
```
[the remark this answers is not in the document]
hidmic: Done in https://github.com/gazebosim/gz-sim/pull/1616/commits/267150c869fd5e866cfe0331357bd3852379f466.

src/gui/plugins/environmental_data_loader/EnvironmentalDataLoader.cc
hidmic: @caguero that's the point of using an atomic boolean: avoid unnecessary mutex locking on every simulation step.

chapulina:

src/gui/plugins/environmental_data_loader/EnvironmentalDataLoader.cc
caguero: Ah, I didn't realize `needsLoad` was atomic. Then you can move the `lock` after setting `needsLoad` to false, right?

caguero (commented):
Looking good, just waiting for CI to be happy after releasing `common`.

src/gui/plugins/environmental_data_loader/EnvironmentalDataLoader.cc
hidmic: @caguero Yes and no. No in the general case, as you could potentially race with another thread to reach the lock. Yes in this particular case as we know there won't more than one thread exercising this execution path (IIRC there's just one thread stepping the simulation). 

I went for the generally correct case out of habit. We can move it out of the lock if you want.

hidmic:
It doesn't look like CI will pass. I'll go ahead and cherry-pick this to our custom branch for the time being. We can come back and merge this once CI stabilizes again.

src/gui/plugins/environmental_data_loader/EnvironmentalDataLoader.cc
```diff
@@ -0,0 +1,277 @@
...
+    this->dataPtr->needsLoad = false;
+
+    std::ifstream dataFile(this->dataPtr->dataPath.toStdString());
+    gzmsg << "Loading environmental data from "
+          << this->dataPtr->dataPath.toStdString()
+          << std::endl;
+    try
+    {
+      using ComponentT = components::EnvironmentalData;
+      auto component = ComponentT{common::IO<ComponentT::Type>::ReadFrom(
```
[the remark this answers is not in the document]
hidmic: > What if two instances are loaded?

@arjo129 last component wins, there can only be one component of a given type per entity (and we always target the world entity). See [here](https://github.com/gazebosim/gz-sim/blob/main/include/gz/sim/detail/EntityComponentManager.hh#L87) then [here](https://github.com/gazebosim/gz-sim/blob/main/src/EntityComponentManager.cc#L1048).

include/gz/sim/components/Environment.hh
```diff
@@ -0,0 +1,76 @@
...
+    /// instantiation that is guaranteed to outlive
+    /// them.
+    static std::shared_ptr<EnvironmentalData>
+    MakeShared(FrameT _frame, ReferenceT _reference);
+
+    /// \brief Environmental data frame.
+    FrameT frame;
+
+    /// \brief Spatial reference for data coordinates. 
+    ReferenceT reference;
```
arjo129, line 61: I assume `CoordinateType::SPHERICAL` means we use lat-long. On the other hand `LOCAL` frame makes no sense here. `GLOBAL` and `ECEF` also makes sense, however I don't know if we want to support `ECEF` right away?

include/gz/sim/components/Environment.hh
```diff
@@ -0,0 +1,76 @@
...
+  /// \brief Environment data across time and space. This is useful to
+  /// introduce physical quantities that may be of interest even if not
+  /// modelled in simulation. 
+  struct EnvironmentalData
+  {
+    using T = math::InMemoryTimeVaryingVolumetricGrid<double>;
+    using FrameT = common::DataFrame<std::string, T>;
+    using ReferenceT = math::SphericalCoordinates::CoordinateType;
+
+    /// \brief Instantiate environmental data.
```
arjo129, line 48: Knit: There's a bit of whitespace at EOL issue here. Should be caught when running `make codecheck`

arjo129 (commented):
Just a few clarifications I wanted to ask about.

src/systems/environment_preload/EnvironmentPreload.cc
```diff
@@ -0,0 +1,172 @@
...
+{
+  this->dataPtr->sdf = _sdf;
+}
+
+//////////////////////////////////////////////////
+void EnvironmentPreload::PreUpdate(
+  const gz::sim::UpdateInfo &,
+  gz::sim::EntityComponentManager &_ecm)
+{
+  if (!std::exchange(this->dataPtr->loaded, true))
```
arjo129, line 70: Is there a reason we implement `PreUpdate` when we probably could get away with doing this in `Configure`?

arjo129 (commented):
Sorry as I work through my sensor implmentation I spotted something else.

src/systems/environment_preload/EnvironmentPreload.cc
hidmic: @arjo129 it used to be that way. Thing is, we can't query the world SDF (root) path until it's fully loaded, and that has to wait for all world plugins (including this one) to load and configure.

include/gz/sim/components/Environment.hh
hidmic: @arjo129 I'll start by saying that I chose to use `gz::math::SphericalCoordinates` as a reference only because it was convenient. Regarding coordinate system types, what I gather from https://github.com/gazebosim/gz-math/blob/6de6120e4d610607b71cb03242862aa07d2582bb/src/SphericalCoordinates.cc#L464-L473 is that `LOCAL` is a rotation away (that specified by heading offsets in spherical coordinates) from `GLOBAL`. In any case, no code should have to deal with these coordinate systems explicitly. It's just how the data coordinates are expressed. 

> I assume `CoordinateType::SPHERICAL` means we use lat-long.

It means the data uses lat-lon coordinates. Code may as well be using global cartesian coordinates. As long as the conversion gets carried out before looking up the frame (see https://github.com/osrf/lrauv/pull/241/files#diff-294ae3632db44c65a140d81aa17a033ddf905909f1093087116e231742ba4c06R1690-R1695 for an example), it'll be fine. Same for `ECEF`.

arjo129:
Oh well then it's fine.

include/gz/sim/components/Environment.hh
hidmic: Done in https://github.com/gazebosim/gz-sim/pull/1616/commits/36c444dd354fa540784d1dbdd4e7fbda19ced8eb.

test/integration/environment_preload_system.cc
```diff
@@ -0,0 +1,89 @@
+/*
+ * Copyright (C) 2019 Open Source Robotics Foundation
```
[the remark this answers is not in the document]
hidmic: Done in https://github.com/gazebosim/gz-sim/pull/1616/commits/ebe9d63a7dde5487a56bbc250ac9da60e5ae69f4.

hidmic:
Pending on https://github.com/gazebosim/gz-common/pull/460.

hidmic:
Re-targeting to Garden.

nkoenig:
@osrf-jenkins run tests

hidmic:

nkoenig:
aha. Looks like we timed it perfectly.

nkoenig:
@hidmic , now the problem is that the `io` component is not being built in the release process.

hidmic:
@osrf-jenkins run tests

hidmic:
We have some failures on Jenkins:

- https://build.osrfoundation.org/job/ignition_gazebo-ci-pr_any-ubuntu_auto-amd64/9791 fails a bunch of tests, seemingly unrelated to this patch.
- https://build.osrfoundation.org/job/ignition_gazebo-ci-pr_any-homebrew-amd64/9405 fails the build with `#include nested too deeply`. I suspect a case sensitivity issue: it's mixing up `io.hh` (generated) with `Io.hh` (checked out).
- https://build.osrfoundation.org/job/ign_gazebo-pr-win/5169/ same as above. 
- https://build.osrfoundation.org/job/ignition_gazebo-abichecker-any_to_any-ubuntu_auto-amd64/5792/ fails to unknown reasons. I suspect the same as above.

@nkoenig @caguero I can change the header name over at `gz-common`, but I'm not sure what's going with those failing tests.

hidmic:
See https://github.com/gazebosim/gz-common/pull/471.

hidmic:
Let's try again:

@osrf-jenkins run tests

mjcarroll:
> I managed to reproduce in a VM. DLL names start with gz-sim7 instead of gz-sim. This is breaking a lot of tests, not just the one this patch adds. Was this naming change intentional?

I think there are large swath of Windows tests that are disabled.  This is likely the root cause.

This is out of the scope of what you are trying to do here.  Can you open an issue noting this is happening and I think you can disable your test here.

hidmic:
Alright, CI is finally mostly green. Test failures on Windows are unrelated. @arjo129 @caguero may I ask you to do a final review?

caguero (approved):
Looks good to me!

hidmic:
@Crola1702 looks like the `io` component in `gz-common5` was not built/released.

nuclearsandwich:
> @Crola1702 looks like the `io` component in `gz-common5` was not built/released.

@Crola1702 and I looked at this together, the dependency on libgz-common5-io-dev wasn't added to the debian build dependency data for this package which caused the regression. https://github.com/gazebo-release/gz-sim7-release/pull/18 should resolve that.

PR #1617, opened 2022-07-27
Adding tests for hydrodynamics

### LINKED ISSUE #1273: Add tests for Hydrodynamics plugin
The plugin was added in #749 without tests. Since then, we've forked the plugin upstream in https://github.com/osrf/lrauv/blob/main/lrauv_ignition_plugins/src/HydrodynamicsPlugin.hh and made tweaks to fix bugs in the code. We should eventually port those fixes here and add tests verifying that the plugin behaves as expected.

## Desired behavior

A new integration test that tests various aspects of the hydrodynamics plugin.

## Implementation suggestion

The test should verify the hydrodynamics in isolation, so we're sure we're testing the effects of that plugin. Some ideas on how to do that:

* Set gravity to zero so we don't need buoyancy or a ground plane (that's similar to the Thruster tests added in #1190 )
* Apply forces / velocities directly to the vehicle without using extra plugins
* Measure the theoretical resulting velocities and accelerations based on the provided parameters

## Additional context

* This may be useful for #1247

quarkytale:
## Summary
Updated the plugin with downstream changes from [LRAUV Hydrodynamics](https://github.com/osrf/lrauv/blob/main/lrauv_ignition_plugins/src/HydrodynamicsPlugin.cc) plugin with the added mass contribution sign and smoothening `stateDot` calculation.
Added tests for verifying:
1. The plugin affects the motion of a body when a force is applied. Here comparing two models, first having hydrodynamics applied and second without it.
2. Making sure that the transforms of the hydrodynamics plugin are correct by comparing 3 models in different positions and orientations having the same final velocities.

### Setup
Using basic shapes: sphere and cylinders in the test world.
To get hydrodynamic plugin parameters for the models,
- Choose any rigid body mass and dimensions
- Inertia: sphere = `2/5 MR^2`, cylinder = `1/2 MR^2`
- Added mass in z-direction: sphere = `2/3 ρπR^3`, cylinder = `ρπR^2 L` -> zDotW
- Linear drag coefficient using Stoke's law, `C=6πηR` -> zW
- Quadratic drag coefficient,`C=1/4 ρA` ->zWW

### Resources
1. [Added_Mass_Derivation](http://web.mit.edu/2.016/www/handouts/Added_Mass_Derivation_050916.pdf)
3. [Viscosity](https://en.wikipedia.org/wiki/Viscosity)
4. [Linear Drag](https://homework.study.com/explanation/an-object-moving-in-a-liquid-experiences-a-linear-drag-force-d-bv-direction-opposite-the-motion-where-b-is-a-constant-called-the-drag-coefficient-for-a-sphere-of-radius-r-the-drag-constant-can-be-computed-as-b-6pnr-where-n-is-the-viscosity-of-the.html#:~:text=Question%3A-,An%20object%20moving%20in%20a%20liquid%20experiences%20a%20linear%20drag,the%20viscosity%20of%20the%20liquid.)
5. [Quadratic Drag Model](https://dynref.engr.illinois.edu/afp.html#:~:text=Quadratic%20drag%20model%23afp%E2%80%91sq&text=This%20is%20referred%20to%20as,%E2%88%92%20c%20v%202%20v%20%5E%20.)

## Test it
Run the test `./build/ignition-gazebo6/bin/INTEGRATION_hydrodynamics`
or check the world directly `ign gazebo src/ign-gazebo/test/worlds/hydrodynamics.sdf
` 
## Checklist
- [x] Added tests


test/integration/hydrodynamics.cc
```diff
@@ -0,0 +1,143 @@
...
+#include "gz/sim/Server.hh"
+#include "gz/sim/SystemLoader.hh"
+#include "gz/sim/TestFixture.hh"
+#include "gz/sim/Util.hh"
+#include "gz/sim/World.hh"
+
+#include "test_config.hh"
+#include "../helpers/EnvTestFixture.hh"
+
+#define PI 3.141592653
```
chapulina, line 38: We can use `IGN_PI` / `GZ_PI`

test/integration/hydrodynamics.cc
```diff
@@ -0,0 +1,143 @@
...
+  Model model;
+  Link body;
+  std::vector<math::Vector3d> bodyVels;
+  double dt{0.0};
+  fixture.
+  OnConfigure(
+    [&](const gz::sim::Entity &_worldEntity,
+      const std::shared_ptr<const sdf::Element> &/*_sdf*/,
+      gz::sim::EntityComponentManager &_ecm,
+      gz::sim::EventManager &/*eventMgr*/)
```
chapulina, line 78: Avoid using the `gz` and `sim` namespaces while already `using` them (lines 40/41). This makes it easier to port the code across branches, because older ones are still in `ignition::gazebo`

test/integration/hydrodynamics.cc
```diff
@@ -0,0 +1,143 @@
...
+  
+  // drag force, F = 6 * pi * a * nu * v
+  // drag force, F = 0.5 * rho * u^2 * Cd * A
+  // terminal velocity, u = 2 * sqrt((3 * pi * a * nu * v)/(rho * Cd * A))
+  
+  for (unsigned int i = 0; i < bodyVels.size(); ++i)
+  {
+    auto bodyVel = bodyVels[i];
+    // It takes a few iterations to reach the terminal velocity
+    if (i > 25)
```
chapulina, line 127: Can we just start from `i = 25` above?

chapulina (commented):
Could you target this PR at `ign-gazebo6` (Fortress)? That's where the plugin was first added. Then after merged we forward-port.

src/systems/hydrodynamics/Hydrodynamics.cc
```diff
@@ -340,13 +344,16 @@ void Hydrodynamics::PreUpdate(
   state(4) = localRotationalVelocity.Y();
   state(5) = localRotationalVelocity.Z();
 
-  auto dt = static_cast<double>(_info.dt.count())/1e9;
-  stateDot = (state - this->dataPtr->prevState)/dt;
+  auto dt = (double)_info.dt.count()/1e9;
```
arjo129, line 347: Don't use a c-style cast here. Also it might be worth using an incantation like: `auto dt = std::chrono::duration<double>(info.dt).count();` (I'm not sure if its correct but `std::chrono` has magical functions) insteat of hard coding` 1e9`.

src/systems/hydrodynamics/Hydrodynamics.cc
```diff
@@ -340,13 +344,16 @@ void Hydrodynamics::PreUpdate(
...
+  auto alpha = 0.9;
```
arjo129, line 348: Put a todo to make this configureable.

arjo129 (approved):
The test expectations seem reasonable. You might want to put in a disclaimer that added mass was flipped as thats a behaviour change. In any case we should look at deprecating added mass her in Garden and onwards.

src/systems/hydrodynamics/Hydrodynamics.cc
```diff
@@ -340,13 +344,16 @@ void Hydrodynamics::PreUpdate(
...
+  stateDot = alpha * (state - this->dataPtr->prevState)/dt
+    + (1-alpha) * this->dataPtr->prevStateDot;
```
JoanAguilar, line 350: I'm curious, what's the reason to filter `stateDot` here?

JoanAguilar (approved):
The tests look to me like good high-level tests. I'll defer any language-specific comments to more experienced reviewers.

src/systems/hydrodynamics/Hydrodynamics.cc
quarkytale: This is to smoothen the rate of change of velocity, we can directly capture `WorldLinearAccelerationt` too.

JoanAguilar: I'm saying because having the rolling average filter here means that `stateDot` depends on all previously visited states (rather than just the last one). I'm not sure if or how much we want to worry about this, though.

quarkytale: That's a valid point. It was discussed in the Buoy Sim project that this might introduce lag and instability.
@arjo129 How should I proceed here? I'm thinking to get this in to be at par with the current changes (though I want to make sure this is generally sound and not specific for LRAUV). And have a separate effort to make this suitable for buoy as well.

arjo129: We can go ahead and remove this. It will not effect LRAUV much. Alternatively we can add an alpha. Finally I thought we were deprecating the added mass within this plugin and switching to the new added mass formulation.

quarkytale: So I compared the previous implementation, the filter and `WorldLinearAcceleration` and the previous implementation is definitely the most stable. Going with that and we can iterate on it in future, thanks!

PR #1624, opened 2022-07-30
Just the ABI breaking parts of #1560

mjcarroll:
It's feature freeze, so getting the important stuff in.

#1560 is introducing may test failures, so this just lifts the minimal API breaking parts out.


chapulina (approved):
Let's hope CI is happy. We should document the change in `MinimumFeatureList` on the migration guide.

It doesn't make a great migration experience for downstream developers to be changing these lists, but it looks to me that the new required features should be quick to implement.

PR #1635, opened 2022-08-09
ign -> gz : Remove redundant namespace references

methylDragon:
Part of: https://github.com/gazebo-tooling/release-tools/issues/784

### DIFF
--- a/src/ServerConfig.cc
+++ b/src/ServerConfig.cc
@@ -499,7 +499,7 @@ unsigned int ServerConfig::Seed() const
 void ServerConfig::SetSeed(unsigned int _seed)
 {
   this->dataPtr->seed = _seed;
-  ignition::math::Rand::Seed(_seed);
+  math::Rand::Seed(_seed);
 }
 
 /////////////////////////////////////////////////
@@ -800,7 +800,7 @@ parsePluginsFromDoc(const tinyxml2::XMLDocument &_doc)
 
 /////////////////////////////////////////////////
 std::list<ServerConfig::PluginInfo>
-ignition::gazebo::parsePluginsFromFile(const std::string &_fname)
+gazebo::parsePluginsFromFile(const std::string &_fname)
 {
   tinyxml2::XMLDocument doc;
   doc.LoadFile(_fname.c_str());
@@ -809,7 +809,7 @@ ignition::gazebo::parsePluginsFromFile(const std::string &_fname)
 
 /////////////////////////////////////////////////
 std::list<ServerConfig::PluginInfo>
-ignition::gazebo::parsePluginsFromString(const std::string &_str)
+gazebo::parsePluginsFromString(const std::string &_str)
 {
   tinyxml2::XMLDocument doc;
   doc.Parse(_str.c_str());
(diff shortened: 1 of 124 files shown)


src/SystemLoader_TEST.cc
```diff
@@ -43,7 +43,7 @@ TEST(SystemLoader, Constructor)
       "<world name='default'>"
       "<plugin filename='libignition-gazebo") +
       IGNITION_GAZEBO_MAJOR_VERSION_STR + "-physics-system.so' "
-      "name='ignition::gazebo::systems::Physics'></plugin>"
+      "name='gazebo::systems::Physics'></plugin>"
```
chapulina, line 46: Within string

src/gui/plugins/component_inspector/ComponentInspector.cc
```diff
@@ -110,7 +110,7 @@ using namespace gazebo;
 
 //////////////////////////////////////////////////
 template<>
-void ignition::gazebo::setData(QStandardItem *_item,
+void gazebo::setData(QStandardItem *_item,
```
chapulina, line 113: We're `using namespace gazebo` here, but this wasn't updated. It would be nice to get rid of those which will need to be changed to `sim` on version 7

src/gui/plugins/entity_tree/EntityTree.cc
```diff
@@ -434,5 +434,5 @@ bool EntityTree::eventFilter(QObject *_obj, QEvent *_event)
 }
 
 // Register this plugin
-IGNITION_ADD_PLUGIN(ignition::gazebo::EntityTree,
+IGNITION_ADD_PLUGIN(EntityTree,
                     ignition::gui::Plugin)
```
chapulina, line 438: Interesting that it didn't catch some of these `ignition::`. Or are we keeping it because the compiler can't tell `ignition::gui` from `ignition::gazebo::gui`?

chapulina (changes requested):

src/gui/plugins/entity_tree/EntityTree.cc
methylDragon: Yep! It 💥 otherwise.

Most cases where it doesn't catch is from me putting in an explicit escape hatch (:

src/SystemLoader_TEST.cc
methylDragon: Oh.. I gotta catch `'` rather than just `"` too, oops.
I'll revert it tomorrow

src/gui/plugins/component_inspector/ComponentInspector.cc
methylDragon: It broke on just `setData` unfortunately.. I forgot why, but the compiler throws a fit

src/SystemLoader_TEST.cc
methylDragon: https://github.com/gazebosim/gz-sim/compare/355b08bcddceeabb3ca91481e04feef2e7b98f69..5529d93c595010177c80f1cde37bab6bdbdc5337

The components::Name is from a different escape hatch targeting another part of `ignition-gazebo`... I'm surprised it compiled properly the first time

src/ServerConfig.cc
```diff
@@ -819,28 +819,28 @@ ignition::gazebo::parsePluginsFromString(const std::string &_str)
...
 /////////////////////////////////////////////////
 std::list<ServerConfig::PluginInfo>
-ignition::gazebo::loadPluginInfo(bool _isPlayback)
+gazebo::loadPluginInfo(bool _isPlayback)
 {
   std::list<ServerConfig::PluginInfo> ret;
 
   // 1. Check contents of environment variable
   std::string envConfig;
-  bool configSet = ignition::common::env(gazebo::kServerConfigPathEnv,
```
azeey, line 828: If the `gazebo` namespace is still valid, wouldn't it be better to leave `gazebo::kServerConfigPathEnv` intact?

chapulina: We're trying to future-proof to reduce the diff for when we change to `sim`

src/LevelManager.cc
```diff
@@ -154,7 +154,7 @@ void LevelManager::ReadLevelPerformerInfo()
   const std::string kPluginName{"ignition::gazebo"};
 
   sdf::ElementPtr pluginElem;
-  // Get the ignition::gazebo plugin element
```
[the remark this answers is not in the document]
methylDragon: Does this mean I should revert the change? :O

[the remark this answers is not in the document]
chapulina: Yeah this line is referring to `kPluginName`  3 lines above, it doesn't make much sense without the `ignition::`.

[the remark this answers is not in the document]
methylDragon: https://github.com/gazebosim/gz-sim/compare/5529d93c595010177c80f1cde37bab6bdbdc5337..2f3afcbcaf597bd07f415f96688214ce938fd43e fixed!

methylDragon:
Seems this is good to go too!

chapulina (approved):

PR #1636, opened 2022-08-09
Add support for specifying log record period

iche033:

# 🎉 New feature

## Summary

Current log recorder records states on every update. By specifying the log record period, the user is able to control how often the simulator records states.  A new command line arg `--record-period` is added (gazebo-classic also has the same arg name)

## Test it
Run the `INTEGRATION_log_system` test.

You can also try enabling log recording through the command line, e.g.

```sh
# record at 10ms instead of the default 1ms and save the log file in a `log_test` dir
ign gazebo -v 4 -r --record --record-path ./log_test --record-period 0.01  pendulum_links.sdf
```

then playback the log file:

```sh
ign gazebo -v 4 --playback  ./log_test
```

You should see that the playback is now choppy due to reduced recording rate.

## Checklist
- [x] Added tests

### DIFF
--- a/include/ignition/gazebo/EntityComponentManager.hh
+++ b/include/ignition/gazebo/EntityComponentManager.hh
@@ -590,6 +590,11 @@ namespace ignition
       /// \return True if there are any components with one-time changes.
       public: bool HasOneTimeComponentChanges() const;
 
+      /// \brief Get whether there are periodic component changes. These changes
+      /// may happen frequently and are processed periodically.
+      /// \return True if there are any components with periodic changes.
+      public: bool HasPeriodicComponentChanges() const;
+
       /// \brief Get the components types that are marked as periodic changes.
       /// \return All the components that at least one entity marked as
       /// periodic changes.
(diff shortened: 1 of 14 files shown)


caguero (commented):
Just for testing I tried `0.1` and `1` record-periods. I was expecting very choppy results but I think the playback stops for a lot of time. Not sure if there's something going on.

iche033:
> Just for testing I tried `0.1` and `1` record-periods. I was expecting very choppy results but I think they playback stops for a lot of time. Not sure if there's something going on.

I can reproduce the issue. Just dumping my findings here: Found that it's to do with the `scene_broadcaster` in playback mode. It's not able to pick up [periodic changed states](https://github.com/gazebosim/gz-sim/blob/ign-gazebo6/src/systems/scene_broadcaster/SceneBroadcaster.cc#L377-L380) and send them to the gui. The logic has implicit assumption that states are updated often enough so that every time it's about to send the states to the gui, there are state changes available. However, log playback now updates at lower rate. You will only see the pendulum move if by some timing coincidence that log playback updates at the same iteration as when scenebroadcaster is about to publish its states.

I haven't found a good fix yet

iche033:
playback scene broadcaster issue should be fixed in [f4728cc](https://github.com/gazebosim/gz-sim/pull/1636/commits/f4728ccccaeba0c39b9ee6fec56dbadaff99df9e). Now when no periodic changes are available when scene broadcaster is publishing,  I force it to do an offcycle state update the next time it sees periodic changes. This should ensure all changed states from log playback are published to the GUI.

caguero (approved):
Works for me now with `0.1` and `1`.

PR #1642, opened 2022-08-10
Replace pose for TransformControl with GzPose

AzulRadio:
Signed-off-by: youhy <haoyuan2019@outlook.com>

# New feature

## Summary

As title. The behavior should be exactly the same as before.

transform_control_replace_demo

## Checklist
- [ ] Added tests


src/gui/plugins/transform_control/TransformControl.qml
```diff
@@ -339,153 +339,40 @@ ToolBar {
...
-            TransformControl.OnSnapUpdate(
-              xEntry.value, yEntry.value, zEntry.value,
-              rollEntry.value, pitchEntry.value, yawEntry.value,
-              0, 0, 0
-            )
-          }
-        }
-        Text {
-          text: "Rotation (deg)"
+          text: "Rotation (rad)"
```
jennuine, line 344: This is not working correctly since `TransformControl.OnSnapUpdate` takes rpy in as degrees: https://github.com/gazebosim/gz-sim/blob/6d1582f8371f064e1aa0c461fc9031ae6c046363/src/gui/plugins/transform_control/TransformControl.hh#L64-L69

Also, we should update it so that it shows as degrees since its easier to read/write.

jennuine (commented):
The sizing needs to be fixed, the values showing in the spin boxes are cut off
image

src/gui/plugins/transform_control/TransformControl.qml
```diff
@@ -339,153 +339,41 @@ ToolBar {
...
+          readOnly: false
+          useRadian: false
+          spinMax: 180
+
+          xValue: 1.0
+          yValue: 1.0
+          zValue: 1.0
+          rollValue: 45
+          pitchValue: 45
+          yawValue: 45
```
jennuine, line 368: nit: let's move `useRadian` from above to here with the  `rpy` elements for better clarity
```suggestion
          useRadian: false
          rollValue: 45
          pitchValue: 45
          yawValue: 45
```

src/gui/plugins/transform_control/TransformControl.qml
```diff
@@ -339,153 +339,41 @@ ToolBar {
...
+
+          onGzPoseSet: {
+            // _x, _y, _z, _roll, _pitch, _yaw are parameters of signal gzPoseSet
+            // from gz-gui GzPose.qml
+            TransformControl.OnSnapUpdate(_x, _y, _z,_roll, _pitch, _yaw, 0, 0, 0)
           }
-        }
+          expand: true
```
jennuine, line 375: nit: can you move this up with the other variables? (e.g., above `readOnly`)

jennuine (commented):
Nice work with allowing degrees in `GzPose`! 

The top of each spin box is still cut off. I saw in 5fd186ec1bd775ca2c71831853fca2a5caa1eba4 the width was adjusted but it's most likely the height that needs to be updated. 
image

Also, I think we should revert the width so that only 2 decimal places are showing (like how it was before) but this is only my preference/opinion. I'll leave it up to you to decide.

AzulRadio:
@jennuine The width and height both worked for me but didn't work for you. So there is no way I know what size would work for you and the only thing I can do is to increase the height and width a bit. 

About the precision, I doubt if we can have 2 decimals without editing GzPose again. So let's just keep it 4 decimals for now.

jennuine:
This is looking good @AzulRadio but requires a `gz-gui3` release before it can be merged. It would be nice to get https://github.com/gazebosim/gz-gui/pull/466 in before then so I'll take over this PR next week. Thanks for all your hard work! :rocket:

jennuine:
Moving this to TODO to focus on the Garden release. 

`GzPose` will need some additional updates so we don't have recession of functionality (eg., spinbox min/max values for each element)

azeey:
@jennuine / @AzulRadio this has been stale for a long time. Do you think there's a path forward or should we close the PR?

jennuine:
Since this has been stale for so long, going to close it. Feel free to re-open if anyone wants to pick it back up.

PR #1652, opened 2022-08-17
Adding thrust coefficient calculation

marcoag:
Signed-off-by: Marco A. Gutierrez <marco@openrobotics.org>

# 🎉 New feature

Work towards https://github.com/osrf/lrauv/issues/225

## Summary

Thrust coefficient in Thruster plugin can only be set to a fixed number, by default 1. This PR adds a calculation that makes it depend on the velocity of the vehicle. The calculation implemented here is based Fossen's equations described in "Guidance and Control of Ocean Vehicles".

## Checklist
- [ ] Added tests


src/systems/thruster/Thruster.cc
```diff
@@ -98,15 +98,36 @@ class ignition::gazebo::systems::ThrusterPrivateData
   /// thrust
   public: double thrustCoefficient = 1;
 
+  /// \brief True if the thrust coefficient was set by conifguration.
+  public: bool thrustCoefficientSet = false;
+
+  /// \brief  Relative speed reduction between the water at the propeller vs
```
ahcorde, line 104:
```suggestion
  /// \brief Relative speed reduction between the water at the propeller vs
```

src/systems/thruster/Thruster.cc
```diff
@@ -98,15 +98,36 @@ class ignition::gazebo::systems::ThrusterPrivateData
...
+  /// behind the vessel.
+  public: double wakeFraction = 0.2;
+
+  /// Constant given by the open water propeller diagram. Used in the
```
ahcorde, line 108:
```suggestion
  /// \brief Constant given by the open water propeller diagram. Used in the
```

src/systems/thruster/Thruster.cc
```diff
@@ -98,15 +98,36 @@ class ignition::gazebo::systems::ThrusterPrivateData
...
+  /// calculation of the thrust coeficiennt.
+  public: double alpha_1 = 1;
+
+  /// Constant given by the open water propeller diagram. Used in the
+  /// calculation of the thrust coeficiennt.
```
ahcorde, line 113:
```suggestion
  /// \brief Constant given by the open water propeller diagram. Used in the
  /// calculation of the thrust coefficient.
```

src/systems/thruster/Thruster.cc
```diff
@@ -98,15 +98,36 @@ class ignition::gazebo::systems::ThrusterPrivateData
...
+  public: double alpha_2 = 0;
+
   /// \brief Density of fluid in kgm^-3, default: 1000kgm^-3
   public: double fluidDensity = 1000;
 
   /// \brief Diameter of propeller in m, default: 0.02
   public: double propellerDiameter = 0.02;
 
+  /// \brief Linear velocity of the vehicle.
+  public: double linearVelocity = 0;
```
ahcorde, line 123:
```suggestion
  public: double linearVelocity = 0.0;
```

src/systems/thruster/Thruster.cc
```diff
@@ -171,6 +193,40 @@ void Thruster::Configure(
...
+  if (_sdf->HasElement("alpha_1"))
+  {
+    this->dataPtr->alpha_1 = _sdf->Get<double>("alpha_1");
+    if (this->dataPtr->thrustCoefficientSet)
+    {
+      ignwarn << " The [alpha_2] value will be ignored as a "
+              << "[thrust_coefficient] was also defined through the SDF file."
+              << " If you want the system to use the alpha values to calculate"
+              << " and update the thrust coefficient please remove the "
+              << "[thrust_coefficient] value  from the SDF file." << std::endl;
```
ahcorde, line 212:
```suggestion
              << "[thrust_coefficient] value from the SDF file." << std::endl;
```

src/systems/thruster/Thruster.cc
```diff
@@ -171,6 +193,40 @@ void Thruster::Configure(
...
+  if (_sdf->HasElement("alpha_2"))
+  {
+    this->dataPtr->alpha_2 = _sdf->Get<double>("alpha_2");
+    if (this->dataPtr->thrustCoefficientSet)
+    {
+      ignwarn << " The [alpha_2] value will be ignored as a "
+              << "[thrust_coefficient] was also defined through the SDF file."
+              << " If you want the system to use the alpha values to calculate"
+              << " and update the thrust coefficient please remove the "
+              << "[thrust_coefficient] value  from the SDF file." << std::endl;
```
ahcorde, line 226:
```suggestion
              << "[thrust_coefficient] value from the SDF file." << std::endl;
```

src/systems/thruster/Thruster.cc
```diff
@@ -98,15 +98,36 @@ class ignition::gazebo::systems::ThrusterPrivateData
...
 
+  /// \brief True if the thrust coefficient was set by conifguration.
+  public: bool thrustCoefficientSet = false;
+
+  /// \brief  Relative speed reduction between the water at the propeller vs
+  /// behind the vessel.
+  public: double wakeFraction = 0.2;
+
+  /// Constant given by the open water propeller diagram. Used in the
+  /// calculation of the thrust coeficiennt.
```
ahcorde, line 109:
```suggestion
  /// calculation of the thrust coefficient.
```

src/systems/thruster/Thruster.hh
```diff
@@ -68,6 +68,15 @@ namespace systems
...
   ///                      defaults to -1000N]
+  /// - <wake_fraction>  - Relative speed reduction between the water
+  ///                      at the propeller vs behind the vessel.
+  ///                      [Optional, defults to 0.2]
+  /// - <alpha_1> - Constant given by the open water propeller diagram. Used
+  ///               in the calculation of the thrust coeficiennt.
+  ///               [Optional, defults to 1]
+  /// - <alpha_2> - Constant given by the open water propeller diagram. Used
+  ///               in the calculation of the thrust coeficiennt.
+  ///               [Optional, defults to 0]
```
ahcorde, line 79:
```suggestion
  /// - <alpha_1> - Constant given by the open water propeller diagram. Used
  ///               in the calculation of the thrust coefficient.
  ///               [Optional, defults to 1]
  /// - <alpha_2> - Constant given by the open water propeller diagram. Used
  ///               in the calculation of the thrust coefficient.
  ///               [Optional, defults to 0]
```

ahcorde (changes requested):

src/systems/thruster/Thruster.cc
```diff
@@ -333,6 +396,17 @@ double ThrusterPrivateData::ThrustToAngularVec(double _thrust)
...
 }
 
+/////////////////////////////////////////////////
+void ThrusterPrivateData::UpdateThrustCoefficient()
+{
+  double calculatedThrustCoefficient = this->alpha_1 + this->alpha_2 *
+      (((1 - this->wakeFraction) * this->linearVelocity)
+      / (this->propellerAngVel * this->propellerDiameter));
+  if (!std::isnan(calculatedThrustCoefficient)
+      && calculatedThrustCoefficient!=0)
```
hidmic, line 406: @marcoag meta: hmm, something's off here. If the thrust coefficient is zero, then no thrust should apply. If the propeller angular velocity is zero, then no thrust should apply and the thrust coefficient is not defined (and it doesn't matter anyways). I see this cause-effect relationship is reversed in this plugin: we assume thrust is always applicable and compute propeller angular velocity. I think we should set that straight. Then you would not need this conditional.

src/systems/thruster/Thruster.hh
```diff
@@ -70,6 +70,15 @@ namespace systems
...
   ///                      defaults to -1000N]
+  /// - <wake_fraction>  - Relative speed reduction between the water
+  ///                      at the propeller vs behind the vessel.
+  ///                      [Optional, defults to 0.2]
+  /// - <alpha_1> - Constant given by the open water propeller diagram. Used
+  ///               in the calculation of the thrust coefficient.
+  ///               [Optional, defults to 1]
+  /// - <alpha_2> - Constant given by the open water propeller diagram. Used
+  ///               in the calculation of the thrust coefficient.
+  ///               [Optional, defults to 0]
```
hidmic, line 81: @marcoag bibliographic references here would be great.

src/systems/thruster/Thruster.cc
marcoag: When the plugin starts the angular velocity is zero, making the thrust coefficient nan. Since it's used later on to calculate the propeller angular velocity then the model would never start moving. The intent of this check was to break this loop and use the default thrust coefficient which is 1 and won't affect the final result. I think it might suffice with the `isnan` check for this or another option is to only update the thrust coefficient once the propeller angular velocity is not 0, maybe the later makes more sense?

hidmic: Ah, almost forgot to reply. 

I guess my point is that the current implementation is not a good approximation. It takes thrust commands instead of propeller angular velocities, and applies forces instantaneously. That's why we are running into this odd corner cases. It makes me wonder, why use this improved thrust coefficient if the dynamics are not there.

Anyways, I think that if we want to keep everything as-is, we could (a) set propeller angular velocity to zero if thrust is zero, bypassing all calculations, and (b) solve the quadratic-like equation to get propeller angular velocity in all other cases, as if it had reached steady state instantaneously (like with forces).

marcoag: I don't know the reason to take thrust commands instead of propeller angular velocities here, the later would be even easier since it seems the thrust gets converter into an angular velocity anyway.

a) The propeller angular velocity would be zero if thrust is zero using calculations, so there doesn't seem a to be a need to bypass them. 

Still the problem is a chicken egg dependency between velocity and thrust coefficient, due to a mutual dependency, if any of them is zero it would make the other one zero, preventing the vehicle to move when starting from stationary. With the conditional above I was trying to avoid this problem. I guess it makes more sense to only calculate when `angular_velocity != 0`  and use the default thrust coefficient (or latest) when that happens. Will add this change for now.

src/systems/thruster/Thruster.hh
marcoag: Added in https://github.com/gazebosim/gz-sim/pull/1652/commits/7351aab3928bb9c294b9c3df115b023f2c989098.

src/systems/thruster/Thruster.cc
hidmic: As per offline discussin, let's go with this as-is. I ran a couple, very much simplified numerical experiments and the coefficient converges even if we start from a guess. I can't swear on convergence, but if the sim looks OK we should be fine.

> solve the quadratic-like equation to get propeller angular velocity in all other cases, as if it had reached steady state instantaneously (like with forces).

Re: to myself. This is not super easy, I missed the linear velocity term in the thrust coefficient model.

src/systems/thruster/Thruster.cc
```diff
@@ -321,6 +379,13 @@ void ThrusterPrivateData::OnCmdThrust(const msgs::Double &_msg)
 /////////////////////////////////////////////////
 double ThrusterPrivateData::ThrustToAngularVec(double _thrust)
 {
+  // Only update if the thrust coefficient was not set by configuration
+  // and angular velocity is not zero. Some velocity is needed to calculate
+  // the thrust coefficient otherwise it will never start moving.
+  if (!this->thrustCoefficientSet && this->propellerAngVel !=0)
```
hidmic, line 385: @marcoag could the thrust coefficient ever become zero while moving (in other words, could either $\alpha_1$ or $\alpha_2$ be negative)? If so, maybe we should reset thrust coefficient to one if propeller angular velocity ever becomes zero. WDYT?

hidmic: Also, I'm surprised the compiler didn't complain about a floating point comparison to zero. AFAIK `std::abs(this->propellerAngVel) < std::numeric_limits<double>::epsilon()` would be a slightly more robust test.

hidmic (commented):
LGTM but for two relatively minor comments

src/systems/thruster/Thruster.cc
marcoag: It did threw a warning, I assumed anything that is not a strict zero would work. 

Changed to your suggestion: [51337cd](https://github.com/gazebosim/gz-sim/pull/1652/commits/51337cd036d004d73e14e96ecb362978e7089074).

hidmic (approved):
LGTM pending green CI

Crola1702:
:man_farmer:  Hi there!

I was taking a look at [this issue](https://build.osrfoundation.org/job/ignition_gazebo-ci-ign-gazebo6-focal-amd64/103/) on the buildfarm and I noticed we have a new gcc warning.

* Gcc warning: 

```
comparing floating point with == or != is unsafe [-Wfloat-equal] 247 | if (_calculateCoefficient && angularVelocity != 0) | ^
```

On [thruster.cc:247](https://github.com/gazebosim/gz-sim/commit/4dd1858d3a7f35c06f51535b4539690356a9da5f#diff-e2f32b0dccaccdba6cba24acfc16cc2b3e52b1f989ceb4e99ffee0fccc67c687R247).

Do you mind taking a look at how to solve/supress it? Thanks

j-rivero:
> Do you mind taking a look at how to solve/supress it? Thanks

I think that @ahcorde arrived to this https://github.com/gazebosim/gz-sim/pull/1754

PR #1655, opened 2022-08-17
Uses new XDisplacement func to check X displacement

onurtore:
# 🦟 Bug fix

Fixes  https://github.com/gazebosim/gz-common/pull/393
## Summary

Uses new XDisplacement func introduced in [here](https://github.com/gazebosim/gz-common/pull/411) to check XDisplacement. 

## Checklist
- [ ] Added tests

### DIFF
--- a/src/rendering/SceneManager.cc
+++ b/src/rendering/SceneManager.cc
@@ -1078,13 +1078,7 @@ rendering::VisualPtr SceneManager::CreateActor(Entity _id,
             std::string rootNodeName = meshSkel->RootNode()->Name();
             common::SkeletonAnimation *skelAnim =
                 meshSkel->Animation(trajInfo.AnimIndex());
-            common::NodeAnimation *rootNode = skelAnim->NodeAnimationByName(
-                rootNodeName);
-            math::Matrix4d lastPos = rootNode->KeyFrame(
-                rootNode->FrameCount() - 1).second;
-            math::Matrix4d firstPos = rootNode->KeyFrame(0).second;
-            if (!math::equal(firstPos.Translation().X(),
-                lastPos.Translation().X()))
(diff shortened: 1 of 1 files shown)


onurtore:
@Ahcorde, reminder for the tag.

ahcorde:
It requires this PR https://github.com/gazebosim/gz-common/pull/411

azeey:
@luca-della-vedova Is this something we need for Harmonic?

luca-della-vedova:
> @luca-della-vedova Is this something we need for Harmonic?

This is also a controversial change.

Without this change, we check only the root node for a "sufficient" displacement in X before we set the `interpolate_x` flag in the actor animation.
The purpose is to make sure that we don't end up in cases where we enable X interpolation yet there is a very small (or zero) range of x movement in the root node.

This function instead would return true if _any_ of the nodes in the actor moves in X by a non negligible amount, which is probably always going to happen, for example if a standing person was to move its hand a bit that would already be above threshold and this would enable X interpolation.
However, when doing the actual interpolation we will still need a single value to divide by, and that value will be the X displacement of the root node ([here](https://github.com/gazebosim/gz-sim/blob/588155992f46081ada9157b9fb68665ba8591e13/src/rendering/SceneManager.cc#L1801)).

For this reason I'm afraid this PR would end up actually creating issues since we will end up in that code path even when the root node has 0 (or very small) displacement, causing issues that could go from numerical instability (i.e. when dividing by very small values) all the way to division by zero, NaN propagation and exceptions being thrown down the line.

For this reason my recommendation would be not to merge this PR.

azeey:
Thanks for the explanation @luca-della-vedova, even though I don't have enough context to fully understand it. I'll go ahead and close the PR.

PR #1717, opened 2022-09-16
Update tutorial for `model_and_optimize_meshes`

AndrejOrsula:
Small updates to the tutorial for "Model and optimize meshes in Blender for simulations".

Mostly phrasing, but I also added some missing shortcuts for the utilised operations. The COLLADA (`.dae`) exporter of Blender sometimes has issues with exporting materials that Gazebo can use, so I suggested the reader to use Wavefront (`.obj`) if that becomes an issue.


mjcarroll (approved):
Thanks for the improvements!

PR #1718, opened 2022-09-16
Fix tutorial for `blender_sdf_exporter`

AndrejOrsula:
# 🦟 Bug fix

## Summary
I encountered 2 issues when trying the tutorial.

1. The COLLADA (`.dae`) would get exported correctly under `./meshes/model.dae`. However, SDF file refer to `./model.dae` as a relative URI for the mesh, which prevented Gazebo from finding it (assuming that only the parent directory of the model is part of ` GZ_SIM_RESOURCE_PATH`). Corrected the path.
2. The script always expected a model to have an assigned material, which would make it fail for objects without any material. Added a simple check to prevent throwing an exception when exporting models without a material.

Tested both with Blender 2.82 and 3.2

## Checklist
- [ ] Added tests

### DIFF
--- a/examples/scripts/blender/sdf_exporter.py
+++ b/examples/scripts/blender/sdf_exporter.py
@@ -1,11 +1,11 @@
-import bpy
+import xml.etree.ElementTree as ET
 from os import path
-from bpy_extras.io_utils import ImportHelper
+from xml.dom import minidom
+
+import bpy
 from bpy.props import StringProperty
 from bpy.types import Operator
-
-import xml.etree.ElementTree as ET
-from xml.dom import minidom
+from bpy_extras.io_utils import ImportHelper
 
 # Tested Blender version: 2.82/3.2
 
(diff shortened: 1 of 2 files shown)


examples/scripts/blender/sdf_exporter.py
```diff
@@ -22,7 +22,7 @@ def export_sdf(prefix_path):
     meshes_folder_prefix = 'meshes/'
     
     # Exports the dae file and its associated textures
-    bpy.ops.wm.collada_export(filepath=prefix_path+meshes_folder_prefix+dae_filename, check_existing=False, filter_blender=False, filter_image=False, filter_movie=False, filter_python=False, filter_font=False, filter_sound=False, filter_text=False, filter_btx=False, filter_collada=True, filter_folder=True, filemode=8)
+    bpy.ops.wm.collada_export(filepath=os.path.join(prefix_path, meshes_folder_prefix, dae_filename), check_existing=False, filter_blender=False, filter_image=False, filter_movie=False, filter_python=False, filter_font=False, filter_sound=False, filter_text=False, filter_btx=False, filter_collada=True, filter_folder=True, filemode=8)
```
mabelzhang, line 25: Would be good to break this long line into 80-character lines.

examples/scripts/blender/sdf_exporter.py
```diff
@@ -46,21 +46,22 @@ def export_sdf(prefix_path):
         geometry = ET.SubElement(visual, "geometry")
         mesh = ET.SubElement(geometry, "mesh")
         uri = ET.SubElement(mesh, "uri")
-        uri.text = dae_filename
+        uri.text = os.path.join(meshes_folder_prefix, dae_filename)
```
mabelzhang, line 49: While we're here, could you apply the same fix for line 76 `albedo_map.text`, line 81 `light_map.text`, and line 143 `uri.text` for the collision? Right now, the visual tag exports `meshes/model.dae`, but the collision tag still just says `model.dae`.

mabelzhang (commented):
Thanks for the fixes! For consistency, we should make the same fixes in a few other places in the script. I found a few, but that might not be all of them.

examples/scripts/blender/sdf_exporter.py
AndrejOrsula: Thank you! I missed that one. Fixed in https://github.com/gazebosim/gz-sim/pull/1718/commits/f8dc7631639b392fe043450dec7caa4fd4c74209

This is the SDF that gets generated now:
```xml
<?xml version="1.0" ?>
<sdf version="1.8">
  <model name="test">
    <link name="testlink">
      <visual name="LPCoke_Cube.000">
        <geometry>
          <mesh>
            <uri>meshes/model.dae</uri>
            <submesh>
              <name>LPCoke_Cube.000</name>
            </submesh>
          </mesh>
        </geometry>
        <material>
          <diffuse>1.0 1.0 1.0 1.0</diffuse>
          <specular>0.0 0.0 0.0 1.0</specular>
          <pbr>
            <metal>
              <albedo_map>meshes/coke.png</albedo_map>
            </metal>
          </pbr>
        </material>
      </visual>
      <collision name="collision">
        <geometry>
          <mesh>
            <uri>meshes/model.dae</uri>
          </mesh>
        </geometry>
        <surface/>
        <contact/>
        <collide_bitmask>0x01</collide_bitmask>
      </collision>
    </link>
  </model>
  <static>true</static>
</sdf>
```

examples/scripts/blender/sdf_exporter.py
AndrejOrsula: I formatted everything via the last commit https://github.com/gazebosim/gz-sim/pull/1718/commits/e17968bfe942ac466954d51c54d8bd9bda448cae using `isort` and `black` with default settings.

Originally, I did not want to do this because of the number of changes this generates (although this commit fixes a lot of things like quote and indentation inconsistency). Please, let me know if I should revert it and fix only the line lengths.

nkoenig (approved):

PR #1719, opened 2022-09-16
Update tutorial for `blender_distort_meshes`

AndrejOrsula:
Tiny edits to the tutorial for "Blender Python script for programmatic mesh distortion".

- `ign` -> `gz`
- Added a new version of Blender that this script was tested with (3.2)
- Reduced the `distort_extent` in the CLI example from the `0.2` to modest `0.005`. I think `0.2` might be a tiny bit too excessive for the `coke.obj` model used in this example.

image


mabelzhang (approved):
Thanks for the fix!

PR #807, opened 2021-05-05
👨‍🌾 Add a potential temporary fix for flaky tests on macOS

azeey:
# 🦟 Bug fix

Relates to #806 

## Summary
A potential solution to the problem of flaky tests on macOS is to include all known components in libignition-gazebo.so. I have tested this locally with Address Sanitizer and it seems to work.

## Checklist
- [ ] Signed all commits for DCO
- [ ] Added tests
- [ ] Updated documentation (as needed)
- [ ] Updated migration guide (as needed)
- [ ] `codecheck` passed (See [contributing](https://ignitionrobotics.org/docs/all/contributing#contributing-code))
- [ ] All tests passed (See [test coverage](https://ignitionrobotics.org/docs/all/contributing#test-coverage))
- [ ] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Aignitionrobotics+repo%3Aosrf%2Fsdformat+archived%3Afalse+) to support the maintainers

**Note to maintainers**: Remember to use **Squash-Merge**

PR #1728, opened 2022-09-22
Fix macOS test segfaults

azeey:
# 🦟 Bug fix

## Summary
The segfaults happen when components are registered by plugins and component creation is attempted after the plugin that registered the component has been unloaded and it's component storage object has become invalid. Including this header insures that all components are registered by the core library ahead of any plugin. 

This is a temporary solution. A better solution would be to keep track of all the plugins that have registered a component. When a plugin gets unloaded, we would unregister that component in the destructor (right before the plugin is unloaded), and check our list if there are other plugins that have registered the component. If so, we would register the component again, this time using the component storage object from a currently loaded plugin.

This is similar to https://github.com/gazebosim/gz-sim/pull/807 but has code to try to mitigate other issues. This PR targets `gz-sim7` because the segfaults appear to be not flaky anymore in this branch.

## Checklist
- [ ] Added tests

### DIFF
--- a/src/SdfEntityCreator.cc
+++ b/src/SdfEntityCreator.cc
@@ -22,6 +22,16 @@
 #include "gz/sim/Events.hh"
 #include "gz/sim/SdfEntityCreator.hh"
 
+#if __APPLE__
+// This is here to avoid segfaults on macOS tests. The segfaults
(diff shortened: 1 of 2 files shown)


ahcorde (changes requested):
linters

PR #1747, opened 2022-10-04
modifying examples using  new detachable joint plugin

liamhan0905:
Related [PR](https://github.com/gazebosim/gz-sim/pull/1687). This PR should be merged first

Modifies some examples and tutorials that uses the `topic` tag.


ahcorde (changes requested):
Why you don't include these changes in the original PR ?

liamhan0905:
good point. I thought this would make the other PR more decoupled. But since this is directly correlated, I'll close this PR and add the changes to the original PR

PR #1755, opened 2022-10-10
Acoustic comms : Packet collision timeout

adityapande-1995:

# 🎉 New feature

## Summary
This PR adds a new tag ``<collision_time_interval>`` to the acoustic comms plugin. If an address endpoint receives two data packets within that time interval, the second one will be dropped. The timeout value defaults to zero, so no packets are ever dropped if the tag is unspecified.

The time must be in seconds.

## Test it
Added an integration test using ``acoustic_comms_packet_collision.sdf``. If multiple packets are sent during a 1 second time interval, they will be dropped. In the test, I send 3 packets, and only the first one of them is received.

## Checklist
- [x] Added tests


hidmic (commented):
LGTM, but I don't follow what phenomena we are trying to model here.

src/systems/acoustic_comms/AcousticComms.cc
```diff
@@ -175,8 +190,41 @@ void AcousticComms::Step(
...
+              // This is the first message received by this address.
+              receivedSuccessfully = true;
+            }
+            else
+            {
+              // A previous msg was already received at this address.
+              std::chrono::duration<double> timeGap = currTimestamp -
+                this->dataPtr->lastMsgReceivedTime[msg->dst_address()];
+
+              auto dropInterval = std::chrono::duration<double>(
```
arjo129, line 209: Knit: From a logical stand point wouldn't the dropped packet also clog the reciever for a fixed amount of time.

arjo129 (commented):
Seems reasonable to me. However, I would encourage us to actually allow users to set a "bit-rate" so that the packet collision also depends on the length of the packet. I don't believe this is too much extra work but would lead to a much more realistic 
and usable model.

src/systems/acoustic_comms/AcousticComms.cc
adityapande-1995: Done https://github.com/gazebosim/gz-sim/pull/1755/commits/25b975ae3b6d7c0f48ad3a428bf1a951179b01dc

hidmic (approved):

adityapande-1995:
I'm pretty sure this is not breaking ABI, but the ABI checker is failing weirdly.

mjcarroll:
> I'm pretty sure this is not breaking ABI, but the ABI checker is failing weirdly.

ABI checker is having some issues at the moment.

PR #1784, opened 2022-11-07
Adds support for hydrodynamic cross terms

arjo129:
# 🎉 New feature

Closes #<NUMBER>

## Summary
The current plugin implementation only supports diagonal parameters. This PR adds support for cross terms in the added mass and the damping matrix.

## What this means for existing users of this plugin

Nothing. The plugin should continue to work as expected. Hence the test suite is not changed.

## Cross terms

For reference read: Fossen T. "Guidance and Control of Ocean Vehicles".

Fossens equations rely on so called "stability derivatives" in the damping and the added mass terms. These are given by parameters such as `<xUU> `. What this means is `xUU` is the Quadratic Damping term (hence the repeated Us along the X axis for a velocity vector along the X axis (U, V, W are used by SNAME (Society of Naval Architects and Marine Engineers) to represent linear velocity along X, Y and Z axis repspectively). Now, in the case of many ocean going vehicles a push along the U (X-direction velocity) may result in translations and rotations in other axis (for instance my boat might yaw to the right). This may seem counter intuitive but fluid-solid interaction is often complex an unintuitive. In previous versions of the plugin, we only supported the parameters along the diagonal of the damping and added mass matrix. While this is often more than sufficient for simple simulations, there are cases where we may like to simulate cross terms. The adding of cross terms now opens up a lot more parameters. For instance, now we will have terms like `xQR` (quadratic damping in x axis with respect to angular velocity in roll and yaw axis).

## Checklist
- [ ] Added tests

### DIFF
--- a/src/systems/hydrodynamics/Hydrodynamics.cc
+++ b/src/systems/hydrodynamics/Hydrodynamics.cc
@@ -276,7 +276,7 @@ void Hydrodynamics::PreUpdate(
 
   // These variables follow Fossen's scheme in "Guidance and Control
   // of Ocean Vehicles." The `state` vector contains the ship's current velocity
-  // in the formate [x_vel, y_vel, z_vel, roll_vel, pitch_vel, yaw_vel].
+  // in the format [x_vel, y_vel, z_vel, roll_vel, pitch_vel, yaw_vel].
   // `stateDot` consists of the first derivative in time of the state vector.
   // `Cmat` corresponds to the Centripetal matrix
   // `Dmat` is the drag matrix
--- a/src/systems/hydrodynamics/Hydrodynamics.cc
+++ b/src/systems/hydrodynamics/Hydrodynamics.cc
@@ -42,60 +42,16 @@ using namespace systems;
 /// \brief Private Hydrodynamics data class.
 class ignition::gazebo::systems::HydrodynamicsPrivateData
 {
-  /// \brief Values to set via Plugin Parameters.
-  /// Plugin Parameter: Added mass in surge, X_\dot{u}.
-  public: double paramXdotU;
+ /// \brief Contains the quadratic stability derivatives like $X_{uv}$,
+  /// Y_{vv}, Y_{wv} etc.
+  public: double stabilityQuadraticDerivative[216];
 
-  /// \brief Plugin Parameter: Added mass in sway, Y_\dot{v}.
-  public: double paramYdotV;
+  /// \brief Contains the quadratic stability derivatives like $X_{u|u|},
+  /// Y_{v|v|}, Y_{w|v|}$ etc.
+  public: double stabilityQuadraticAbsDerivative[216];
 
-  /// \brief Plugin Parameter: Added mass in heave, Z_\dot{w}.
-  public: double paramZdotW;
-
-  /// \brief Plugin Parameter: Added mass in roll, K_\dot{p}.
-  public: double paramKdotP;
-
-  /// \brief Plugin Parameter: Added mass in pitch, M_\dot{q}.
-  public: double paramMdotQ;
-
-  /// \brief Plugin Parameter: Added mass in yaw, N_\dot{r}.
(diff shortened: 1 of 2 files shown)


src/systems/hydrodynamics/Hydrodynamics.cc
```diff
@@ -363,40 +331,50 @@ void Hydrodynamics::PreUpdate(
...
+          this->dataPtr->stabilityQuadraticAbsDerivative[index] * abs(state(k));
+        coeff -= absCoeff;
+        auto velCoeff =
+          this->dataPtr->stabilityQuadraticDerivative[index] * state(k);
+        coeff -= velCoeff;
+      }
+
+      Dmat(i, j) = coeff;
+    }
+  }
```
hidmic, line 377: @arjo129 meta: I'll start by saying I know next to nothing about underwater vessel modelling. That said, I can't relate this equations to those in Fossen's books (ie. Guidance of Ocean Vehicles and Marine Control Systems). I see you are including all cross terms, but none of the models proposed includes a pure quadratic term nor the means to estimate most of the rest (which begs the question, why bother?). 

Mind to explain?

arjo129: [So the LRAUV's original model does](https://github.com/osrf/lrauv/wiki/Vehicle-specifications-and-hydrodynamic-parameters), we had not implemented it as we were interested in getting the hydrodynamics plugin up and running early on. It is often hard to estimate and verify the cross terms, however there are definitely techniques that can be used.

src/systems/hydrodynamics/Hydrodynamics.hh
```diff
@@ -62,6 +63,20 @@ namespace systems
...
   ///   * <nRR>   - Quadratic damping, 2nd order, yaw component [kg/m^2]
   ///   * <nR>    - Linear damping, 1st order, yaw component [kg/m]
+  /// ### Cross terms
+  /// In general we support cross terms as well. These are terms which act on
+  /// non-diagonal sides. We use the SNAMe convention of naming search terms.
+  /// (x, y, z) correspond to the respective axis. (p, q, r) correspond to
+  /// roll, pitch and yaw. Similarly U, V, W represent velocity vectors in
+  /// X, Y and Z axis while P, Q, R representangular velocity in roll, pitch
+  /// and yaw axis respectively.
+  ///   * Added Mass: <{x|y|z|k|m|n}Dot{U|V|W|P|Q|R}> e.g. <xDotR>
```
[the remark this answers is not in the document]
arjo129: Thanks for spotting this. k,m,n represent roll pitch and yaw. I made the correction above instead.

arjo129:
Test failures seem unrelated.

hidmic (approved):
LGTM, though CI is a mess right now.

azeey:
@arjo129 looks like this PR has changed the behavior of the Hydrodynamics plugin as mentioned in https://github.com/gazebosim/gz-sim/issues/1871#issuecomment-1419984589. Is that a regression or the intended outcome?

srmainwaring:
> looks like this PR has changed the behavior of the Hydrodynamics plugin

Although it's a change to previous behaviour I think the new parameter labels more accurately describe the implementation. For the case in #1871 we've updated the parameters labels in the ROV to use the new definitions and everything is working fine.

It may be a case of adding a note to emphasise that it's a breaking change and provide a pointer to the type of error you might see if you don't update the params (the rapid increase in force causing the collision box to overflow tends to lead you on a wild goose chase initially!).

azeey:
FYI, another affected user https://answers.gazebosim.org/question/28783/help-with-hydrodynamics-plugin-in-gazebosim-garden/. I'll refer them to this PR.

arjo129:
Thanks for Pointing this out @azeey. I dont think behavior change was intended, especially since we didnt update the example. As @srmainwaring pointed out these new names are definitely more sensible. I can add support back for the old names with a warning.

PR #1431, opened 2022-04-08
Add interface to allow systems to declare parameters

ivanpauno:
# 🎉 New feature

Based on https://github.com/ignitionrobotics/ign-transport/pull/305.
Replacement of https://github.com/ignitionrobotics/ign-gazebo/pull/1280.

## Summary

This PR adds:

- An interface that can be implemented by systems that allows them to declare parameters (`ISystemConfigureParameters`):
  - The gazebo simulation runner creates an instance of `ignition::transport::parameters::ParametersRegistry` (see https://github.com/ignitionrobotics/ign-transport/pull/305) with a namespace `/world/<world_name>` that provides the following services:
  - `/world/<world_name>/list_parameters` service: List available parameters names and types.
  - `/world/<world_name>/get_parameter` service: Get the type and value of a parameter.
  - `/world/<world_name>/set_parameter` service: Set a parameter: parameter name, value and type need to be provided.

Summary of what's needed to declare and use a parameter can be seen in the wheel slip demo [branch](https://github.com/ignitionrobotics/ign-gazebo/tree/ivanpauno/parameters-component-wheel-slip-demo6): [diff](https://github.com/ignitionrobotics/ign-gazebo/compare/b7b290cfd1e6984f66216bc1434a4df697844fd7..317f44880c837c7e887fcc35858a1b165e2bfb77).

## Test it

I haven't written automated tests yet, but there's a demo using parameters in the wheel slip plugin:

- https://github.com/ignitionrobotics/ign-gazebo/tree/ivanpauno/parameters-component-wheel-slip-demo6
- https://github.com/ignitionrobotics/ign-msgs/tree/ivanpauno/parameters-component-wheel-slip-demo8

Steps to run the demo:

- Build from source ignition `fortress` checking out the two branches specified above (I can provide PRs for another version if needed).
- Launch the trisphere cycle world:
  ```
  ign gazebo trisphere_cycle_wheel_slip.sdf
  ```
- In another terminal with the workspace sourced:
  - List available parameters:
    ```
    ign param -r /world/wheel_slip -l
    ```
    <details>
      <summary>Expected output</summary>

      > systems.wheel_slip.trisphere_cycle1.wheel_front            [ignition.msgs.WheelSlipParameters]
      > systems.wheel_slip.trisphere_cycle1.wheel_rear_right            [ignition.msgs.WheelSlipParameters]
      > systems.wheel_slip.trisphere_cycle0.wheel_rear_right            [ignition.msgs.WheelSlipParameters]
[...]

PR #1789, opened 2022-11-10
Ci matching branch/parameters component6

scpeters:
testing for #1431


scpeters:
this is no longer needed sing ign-transport 11.3.1 was released

PR #1793, opened 2022-11-11
Acoustic comms : Propagation model

adityapande-1995:
# 🎉 New feature

## Summary
This PR adds a propagation model for acoustic comms, using RFComms implementation as a guide. The difference is in the way signal loss is calculated and modulation is implemented.

Here, we use the sonar equation to calculate signal to noise ratio (SNR). SNR is used to calculate ``Eb/N0``, which gives us the bit error rate using BPSK modulation. FInally we calculate the packet drop probability.

This feaure is turned off by default and oly works if the SDF tags are used when declaring the plugin.

A short comparison of propagation model approaches :
* Acoustic comms propagation model (This PR) : 
  ``SNR using sonar equation`` -> ``Eb/N0`` -> ``bit error rate (BPSK)`` -> ``packetDropProbability``
* RFComms propagation model (Taken as a reference) : 
  ``Log path loss equation`` -> ``bit error rate (QPSK)`` -> ``packetDropProbability``

## Test it
Added a test with a world file ``acoustic_comms_propagation.sdf``. It is impossible to get the packets to drop with the default values and the LRAUvs being that close to each other (< 5 m). Moving them apart would make the tests run a lot slower.

I've exaggerated the noise and source power to make the packets drop.

## Note
This is still an experimental feature, and I'm no expert in communications. If there is a better model for propagation, I'd be happy to improve this.

## Checklist
- [x] Added tests

### DIFF
--- a/src/systems/acoustic_comms/AcousticComms.cc
+++ b/src/systems/acoustic_comms/AcousticComms.cc
@@ -83,6 +83,9 @@ class AcousticComms::Implementation
 
   /// \brief Flag to store if the propagation model should be used.
   public: bool usePropagationModel = false;
+
+  /// \brief Seed value for random sampling.
+  public: unsigned int seed = 0;
 };
 
 //////////////////////////////////////////////////
@@ -125,6 +128,7 @@ bool AcousticComms::Implementation::propagationModel(
     1.0 - std::exp(static_cast<double>(_numBytes) *
                    std::log(1 - ber));
 
+  gz::math::Rand::Seed(this->seed);
   double randDraw = gz::math::Rand::DblUniform();
   return randDraw > packetDropProb;
 }
@@ -165,6 +169,7 @@ void AcousticComms::Load(
     this->dataPtr->noiseLevel = propElement->Get<double>("noise_level");
     this->dataPtr->spectralEfficiency =
       propElement->Get<double>("spectral_efficiency");
+    this->dataPtr->seed = propElement->Get<int>("seed");
   }
 
   gzmsg << "AcousticComms configured with max range : " <<
--- a/src/systems/acoustic_comms/AcousticComms.cc
+++ b/src/systems/acoustic_comms/AcousticComms.cc
@@ -67,7 +67,7 @@ class AcousticComms::Implementation
 
   /// \brief Source power in Watts.
   /// \ref https://www.sciencedirect.com/topics/
-  /// computer-science/underwater-acoustic-signal 
+  /// computer-science/underwater-acoustic-signal
   public: double sourcePower = 2000;
 
   /// \brief Ratio of the noise intensity at the
@@ -76,14 +76,14 @@ class AcousticComms::Implementation
 
   /// \brief Ratio of the total noise power at the array to the
   /// noise received by the array along its main response axis.
-  /// \ref https://ieeexplore.ieee.org/document/5664178 
+  /// \ref https://ieeexplore.ieee.org/document/5664178
   public: double directivityIndex = 4;
 
   /// \brief Information rate that can be transmitted over a given
-  /// bandwidth in a specific communication system, in (bits/sec)/Hz. 
+  /// bandwidth in a specific communication system, in (bits/sec)/Hz.
   /// \ref: https://www.ncbi.nlm.nih.gov/pmc/articles/PMC5514747/
   public: double spectralEfficiency = 7.0;
-  
+
   /// \brief Flag to store if the propagation model should be used.
   public: bool usePropagationModel = false;
 };
--- a/src/systems/acoustic_comms/AcousticComms.cc
+++ b/src/systems/acoustic_comms/AcousticComms.cc
@@ -61,7 +61,61 @@ class AcousticComms::Implementation
            std::tuple<std::chrono::duration<double>,
             std::chrono::duration<double>>>
           lastMsgReceivedInfo;
+
+  public: bool propagationModel(double distToSource);
+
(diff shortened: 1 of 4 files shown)


src/systems/acoustic_comms/AcousticComms.cc
```diff
@@ -90,6 +158,19 @@ void AcousticComms::Load(
       _sdf->Get<double>("collision_time_per_byte");
   }
 
+  if (_sdf->HasElement("propagation_model"))
+  {
+    this->dataPtr->usePropagationModel = true;
+    sdf::ElementPtr propElement = _sdf->Clone()->
+                                   GetElement("propagation_model");
```
adityapande-1995, line 165: Open to suggestions on how cloning can be avoided here. ``GetElement`` cannot be applied to const ``Element`` members it seems, and ``_sdf->GetElement("foo")`` throws an error that " `this` argument discards qualifiers."

src/systems/acoustic_comms/AcousticComms.cc
```diff
@@ -61,8 +61,76 @@ class AcousticComms::Implementation
...
+  // SNR = SL - TL - (NL - DI)
+  // SNR : Signal to noise ratio.
+  // SL : Source level. Ratio of the transmitted intensity from
+  //      the source to a reference intensity (1 m from source),
+  //      converted to dB.
+  // TL : Transmission loss (dB)
+  // NL : Noise level.
+  // DI : Receiver directivity index.
+
+  double sl = 171.5 + 10 * std::log10(this->sourcePower);
```
arjo129, line 107: Might be worth mentioning where 171.5 comes from. IIUC, this is the intensity measured at 1 yard (why use yards? :skull_and_crossbones:) from the source of the sound.

src/systems/acoustic_comms/AcousticComms.cc
```diff
@@ -61,8 +61,76 @@ class AcousticComms::Implementation
...
+  double tl = 20 * std::log10(_distToSource);
+
+  // Calculate SNR.
+  auto snr = sl - tl - (this->noiseLevel - this->directivityIndex);
```
arjo129, line 111: Doesn't the directivity index take into account the _direction_ of a sonar array? If we aren't modelling the directional response then we might as well account for this in our noise term.

arjo129 (commented):
Just did a quick pass. Couple of small knits.

src/systems/acoustic_comms/AcousticComms.cc
```diff
@@ -61,8 +61,76 @@ class AcousticComms::Implementation
            std::tuple<std::chrono::duration<double>,
             std::chrono::duration<double>>>
           lastMsgReceivedInfo;
+
+  public: bool propagationModel(double _distToSource,
```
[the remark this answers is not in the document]
adityapande-1995: Done https://github.com/gazebosim/gz-sim/pull/1793/commits/924e18e878bba1c14c1854b1d5612489951a2791

src/systems/acoustic_comms/AcousticComms.cc
adityapande-1995: Oh right just saw that. Changed it to 1m, updated docstring : https://github.com/gazebosim/gz-sim/pull/1793/commits/924e18e878bba1c14c1854b1d5612489951a2791

src/systems/acoustic_comms/AcousticComms.cc
adityapande-1995: So do you mean the directivity index should default to zero (i.e. not modelling directional nature), unless specified ?

arjo129: I mean that unless we are factoring in direction to calculate the directivity index, it makes little sense to add this term.

adityapande-1995: I see. Yeah I'll get rid of this term entirely.

adityapande-1995: Removed : https://github.com/gazebosim/gz-sim/pull/1793/commits/1f39deff7c5084ab996d253bdca954bad6b40a44

test/integration/acoustic_comms.cc
```diff
@@ -161,6 +161,10 @@ INSTANTIATE_TEST_SUITE_P(
     // Message packets will be dropped as they are sent too fast
     // compared to "collision_time_interval".
     AcousticCommsTestDefinition(
-      "acoustic_comms_packet_collision.sdf", "addr2", "addr1", 3, 1)
+      "acoustic_comms_packet_collision.sdf", "addr2", "addr1", 3, 1),
+    // Source power is decreased and noise bumped up to make the packets
+    // drop.
+    AcousticCommsTestDefinition(
+      "acoustic_comms_propagation.sdf", "addr2", "addr1", 3, 0)
```
arjo129, line 168: Given the non-determinism in the system there is a teeny-tiny chance this _could_ become flaky. One possibility is to set  the [RNG to a fixed seed](https://github.com/gazebosim/gz-math/blob/a2f014e292a3b68e61384eb4bbea28ff4ad11553/include/gz/math/Rand.hh#L53), alternatively you could just check for a less than equal.

For the seed option exposing the seed to the end user may also help downstream users test individual cases better.

arjo129 (changes requested):
One more tiny issue. After that it looks like its good to go.

arjo129 (approved):
Lgtm. Test failures unrelated.

PR #1814, opened 2022-11-25
Add multichannel lookup for environment sensors.

arjo129:

# 🎉 New feature

Closes #<NUMBER>

## Summary
Often when sensing we may be interested in vector quantities. This PR extends the environmental_sensor to support vector quantities. Some additional work was required handle transformations. In particular if the data is something like Ocean Currents or Wind then the data has to be transformed to the robot's local coordinate frame and the robot's velocity also has to be accounted for.

## Checklist
- [x] Added tests


src/systems/environmental_sensor_system/EnvironmentalSensorSystem.cc
```diff
@@ -76,19 +78,115 @@ class EnvironmentalSensor : public gz::sensors::Sensor
...
     {
-      this->fieldName =
+      this->fieldName[0] =
         _sdf.Element()->Get<std::string>("environment_variable");
     }
+    else if (_sdf.Element() != nullptr &&
+      this->numberOfFields == 2)
+    {
+      if (
+        _sdf.Element()->HasElement("x_variable") &&
```
hidmic, line 153: @arjo129 meta: I find tag naming a bit inconsistent (`environment_variable` vs. `x_variable`, `y_variable`). Considering that variable names are likely to follow a pattern (e.g. `<name>_<component>`), why not enforce a convention? Then we wouldn't need to specify each component, only the `environment_variable` e.g. `wind_speed`.

This will also keep the interface future-proof for when file formats that natively support vectorial data are added (like netCDF).

src/systems/environmental_sensor_system/EnvironmentalSensorSystem.hh
```diff
@@ -34,6 +34,34 @@ class EnvironmentalSensorSystemPrivate;
...
+///    frame vector3 fields.
+///  <transform_type> - When handling vector2 or vector3 types it may be in our
+///    interest to transform them to local coordinate frames. For instance
+///    measurements of ocean currents or wind depend on the orientation and
+///    velocity of the robot. This field can have 4 values:
+///      * GLOBAL - Don't transform the vectors to a local frame.
+///      * LOCAL - Transform the vector to a local frame.
+///      * ADD_VELOCITY_GLOBAL - Don't transform to local frame but account for
+///          velocity change.
+///      * ADD_VELOCITY_LOCAL - Transform to local frame and account for robot's
```
hidmic, line 62: @arjo129 nit: why assume it's a robot?

src/systems/environmental_sensor_system/EnvironmentalSensorSystem.cc
```diff
@@ -113,30 +211,87 @@ class EnvironmentalSensor : public gz::sensors::Sensor
...
+      frame->add_value((this->frameId == "") ? this->Name() : this->frameId);
+
+      if (!dataPoints[0].has_value() || !dataPoints[1].has_value())
+      {
+        gzwarn << "Failed to acquire value perhaps out of field?\n";
+        return false;
+      }
+
+      math::Vector3d vector(dataPoints[0].value(), dataPoints[1].value(), 0);
+      auto transformed = transformFrame(this->transformType, this->objectPose,
```
hidmic, line 263: @arjo129 meta: hmm, a 3D transform for 2D data? On z = 0? Why z = 0?

hidmic: Thinking out loud, Gazebo performs rigid body simulation in 3D. Does it make sense to have 2D data? Or is 2D data actually 3D data in some plane (or all planes)?

hidmic (commented):
First pass. I don't yet buy the overall design.

src/systems/environmental_sensor_system/EnvironmentalSensorSystem.cc
arjo129: Hmm how would that work with our current data? The CSV column headers would be fixed?

src/systems/environmental_sensor_system/EnvironmentalSensorSystem.cc
arjo129: I'm not too sure how we should handle a 2D case. Our current example format for MBARI uses such a 2D case which is why I default to zero.

src/systems/environmental_sensor_system/EnvironmentalSensorSystem.cc
arjo129: Oh ok I get it you would define the `wind_speed_x` and the `wind_speed_y` in the CSV. We would need to change the header of our CSV files to match and then just specify `<environment_variable>wind_speed</environment_variable>` in the SDFormat file.

arjo129: Hmmm... I just checked the DVL it seems we have the x_column, y_column and z_column. I'd like this to be consistent with it.

hidmic: IMO it's not an exact analogue because the DVL will always work with 3D data whereas this has to deal with various dimensions (and even then perhaps the DVL should change to use multi-dimensional variables if supported).

That said, in the interest of time, I'm OK with not switching to a convention yet, but at least tag naming should be consistent across dimensions (i.e. it should follow some pattern and not be completely different if it's 1D or 2D).

src/systems/environmental_sensor_system/EnvironmentalSensorSystem.cc
hidmic: > Perhaps if we see a z component is missing we assume that the thing defaults to zero.

That sounds pretty reasonable.

arjo129: 4b2ea71 does this.

src/systems/environmental_sensor_system/EnvironmentalSensorSystem.cc
```diff
@@ -76,19 +78,115 @@ class EnvironmentalSensor : public gz::sensors::Sensor
     // Load common sensor params
     gz::sensors::Sensor::Load(_sdf);
 
-    this->pub = this->node.Advertise<gz::msgs::Double>(this->Topic());
+    // Handle the field type
+    if (_sdf.Element() != nullptr &&
+      _sdf.Element()->HasElement("field_type"))
+    {
+      auto dataType = _sdf.Element()->Get<std::string>("field_type");
```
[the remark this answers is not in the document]
arjo129: Renaming to "output_format"

src/systems/environmental_sensor_system/EnvironmentalSensorSystem.cc
arjo129: In the interest of time I'm keeping `x_variable`, `y_variable` and `z_variable`. When NetCDF is supported we could just use `environment_variable`

src/systems/environmental_sensor_system/EnvironmentalSensorSystem.cc
```diff
@@ -76,19 +78,89 @@ class EnvironmentalSensor : public gz::sensors::Sensor
...
+      }
+      else
+      {
+        gzerr << "Invalid output format " << format << "given."
+          << "valid options are scalar or vector3. "
+          << "Defaulting to 1." << std::endl;
+      }
+    }
+
+    // Hansle the transform type
```
hidmic, line 102: @arjo129 nit: typo: 
```suggestion
    // Handle the transform type
```

src/systems/environmental_sensor_system/EnvironmentalSensorSystem.cc
```diff
@@ -76,19 +78,89 @@ class EnvironmentalSensor : public gz::sensors::Sensor
     // Load common sensor params
     gz::sensors::Sensor::Load(_sdf);
 
-    this->pub = this->node.Advertise<gz::msgs::Double>(this->Topic());
+    // Handle the field type
+    if (_sdf.Element() != nullptr &&
+      _sdf.Element()->HasElement("output_format"))
+    {
+      std::string format = _sdf.Element()->Get<std::string>("output_format");
```
hidmic, line 85: @arjo129 meta: output or input?

src/systems/environmental_sensor_system/EnvironmentalSensorSystem.cc
```diff
@@ -76,19 +78,89 @@ class EnvironmentalSensor : public gz::sensors::Sensor
...
+        break;
+      default:
+        gzerr << "Unable to create publisher too many fields" << "\n";
+    }
 
     // If "environment_variable" is defined then remap
     // sensor from existing data.
     if (_sdf.Element() != nullptr &&
+      this->numberOfFields == 1 &&
       _sdf.Element()->HasElement("environment_variable"))
```
hidmic, line 137: @arjo129 hmm, should this be `x_variable` instead? If we want to keep `environment_variable` for scalar data, I'd suggest `environment_variable_x` and the like for vector data.

src/systems/environmental_sensor_system/EnvironmentalSensorSystem.cc
```diff
@@ -98,7 +170,7 @@ class EnvironmentalSensor : public gz::sensors::Sensor
       this->frameId = _sdf.Element()->Get<std::string>("frame_id");
     }
 
-    gzdbg << "Loaded environmental sensor for " << this->fieldName
+    gzdbg << "Loaded environmental sensor for " << this->fieldName[0]
```
hidmic, line 173: @arjo129 hmm, the first field may not necessarily be descriptive for vector data.

src/systems/environmental_sensor_system/EnvironmentalSensorSystem.cc
```diff
@@ -113,30 +185,72 @@ class EnvironmentalSensor : public gz::sensors::Sensor
...
-    frame->set_key("frame_id");
-    frame->add_value((this->frameId == "") ? this->Name() : this->frameId);
-    auto data = this->gridField->frame[this->fieldName].LookUp(
-      this->session.value(), this->position);
-    if (!data.has_value())
+      dataPoints[i] = this->gridField->frame[this->fieldName[i]].LookUp(
+        this->session[i].value(), this->position);
+    }
+
+    if(this->numberOfFields == 1) {
```
hidmic, line 213: @arjo129 nit: 
```suggestion
    if (this->numberOfFields == 1) {
```
Have you ran `codecheck`?

src/systems/environmental_sensor_system/TransformTypes.hh
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  switch (_type) {
+      case ADD_VELOCITY_LOCAL:
+      case LOCAL:
+          result = _pose.Rot().Inverse() * (_reading + offset);
+          break;
+      default:
+          result = (_reading + offset);
+  }
```
hidmic, line 86: @arjo129 meta: it would almost seem like local/global is orthogonal to whether velocity should be computed relative some observer or the world? Would it make sense to split them?

src/systems/environmental_sensor_system/TransformTypes_TEST.cc
```diff
@@ -0,0 +1,73 @@
...
+
+using namespace gz;
+using namespace sim;
+
+TEST(TransformationTypes, transformFrame)
+{
+  // Create a vehicle rotated 90 degrees in the yaw axis
+  math::Pose3d pose(math::Vector3d(0, 0, 0), math::Quaterniond(0, 0, GZ_PI_2));
+
+  // Vehicle should moving at 1m/s along x-axis (In global frame)
```
hidmic, line 28: @arjo129 nit: 
```suggestion
  // Vehicle should be moving at 1m/s along x-axis (In global frame)
```

src/systems/environmental_sensor_system/TransformTypes_TEST.cc
```diff
@@ -0,0 +1,73 @@
...
+  math::Vector3d velocity(1, 0, 0);
+
+  // Imagine a current moving against the vehicle
```
hidmic, line 31: @arjo129 `imagine` is an interesting choice of words.

hidmic (commented):
Overall LGTM. CI failures look unrelated (would a rebase or merging back fix them?).

src/systems/environmental_sensor_system/EnvironmentalSensorSystem.cc
hidmic: Ok, fair enough.

src/systems/environmental_sensor_system/TransformTypes.hh
arjo129: Thats true, but in the interest of time we can check this in.

hidmic (approved):
LGTM with green CI

PR #1826, opened 2022-12-01
Re-add namespace for GUI render event

jennuine:
# 🦟 Bug fix

## Summary
A required namespace was accidentally removed in https://github.com/gazebosim/gz-sim/pull/1635 , this caused the assumption that the render event was a `ignition::gazebo::gui` event when it needed to be an `ignition::gui` event.

To test: 
```bash
ign gazebo shapes.sdf
```
then use the `Screenshot` plugin.

Before PR, nothing happened when clicking the screenshot icon (no popup stating where image was saved and no image saved to `~/.ignition/gui/pictures`).

## Checklist
- [ ] Added tests

### DIFF
--- a/src/gui/plugins/scene3d/Scene3D.cc
+++ b/src/gui/plugins/scene3d/Scene3D.cc
@@ -778,7 +778,7 @@ void IgnRenderer::Render()
 
   if (ignition::gui::App())
   {
-    gui::events::Render event;
+    ignition::gui::events::Render event;
     ignition::gui::App()->sendEvent(
         ignition::gui::App()->findChild<ignition::gui::MainWindow *>(),
         &event);


src/gui/plugins/scene3d/Scene3D.cc
```diff
@@ -778,7 +778,7 @@ void IgnRenderer::Render()
 
   if (ignition::gui::App())
   {
-    gui::events::Render event;
+    ignition::gui::events::Render event;
```
iche033, line 781: there are other places in ign-gazebo3 that uses gui::events::Render. Do they need to be updated as well? e.g. https://github.com/gazebosim/gz-sim/blob/ign-gazebo3/src/gui/plugins/grid_config/GridConfig.cc#L90

jennuine: Nope, the bug lived in the sender. Anything listening to the render event doesn't need updated since both a `ignition::gui::events::Render` and `ignition::gazebo::gui::events::Render` are being sent: https://github.com/gazebosim/gz-sim/blob/ign-gazebo3/include/ignition/gazebo/gui/GuiEvents.hh

Since both events are being broadcasted, plugins listening to either render event will be caught now (in fortress, the `ignition::gazebo::gui::events::Render` is removed)

PR #1840, opened 2022-12-16
Disable tests that require dartsim on windows

azeey:
The dartsim component of gz-physics is currently not built on CI, so these tests should be disabled.


iche033 (approved):
looks good to me. 

The `TransformsTestinWater` was already passing on windows. Looking at the checks in the test, I think it makes sense that it may just pass without physics running.

azeey:
> looks good to me.
> 
> The `TransformsTestinWater` was already passing on windows. Looking at the checks in the test, I think it makes sense that it may just pass without physics running.

Yeah, the comparisons are against each other, not against an expected velocity. I still think we should disable it though since it would be giving a false impression if it passes without physics.

PR #1888, opened 2023-02-09
Document behaviour changes introduced #1784

arjo129:
# 🦟 Bug fix

## Summary
The PR in #1784 claimed not to introduce behavior changes, but actually ended up introducing behavior changes. Specifically, it would break a lot of marine simulation code that runs on our tools. The issue in question is a matter of naming parameters. In the past `<xUU>` used to refer to the quadratic drag term in the diagonal axis. This term would be multiplied by the absolute velocity. After #1784 this term was multiplied by the velocity (no absolute). An equivalent term of `<xUabsU>` This behavior change likely breaks all previous maritime simulations that rely on our hydrodynamics plugin.

There are several options:
1. Revert #1784
2. Make `<xUU>` mirror `<xUabsU>`
3. Document the change and mark it as breaking.

This PR goes with option 3 as there are already code bases using `<xUabsU>` nota tion instead of `<xUU>`. We also add a warning if someone does use the `<xUU>` term that points to this PR for details.

## Checklist
- [ ] Added tests

### DIFF
--- a/src/systems/hydrodynamics/Hydrodynamics.cc
+++ b/src/systems/hydrodynamics/Hydrodynamics.cc
@@ -195,16 +195,16 @@ void Hydrodynamics::Configure(
         SdfParamDouble(_sdf, prefix, 0);
       for(auto k = 0; k < 6; k++)
       {
-        // TODO(arjo): 
         auto fieldName = prefix + snameConventionVel[k];
         this->dataPtr->stabilityQuadraticDerivative[i*36 + j*6 + k] =
           SdfParamDouble(
             _sdf,
             fieldName,
             0);
+
         if (_sdf->HasElement(fieldName)) {
           warnBehaviourChange = true;
-        } 
+        }
 
         this->dataPtr->stabilityQuadraticAbsDerivative[i*36 + j*6 + k] =
           SdfParamDouble(
@@ -220,9 +220,9 @@ void Hydrodynamics::Configure(
   {
     ignwarn << "You are using parameters that may cause instabilities"
       << "in your simulation. If your simulation crashes we recommend"
-      << "renaming <xUU> -> <xUabsU> and likewise for other information"
+      << "renaming <xUU> -> <xUabsU> and likewise for other axis"
       << "for more information see:"
-      << "\t<INSERT LINK HERE>\n";
+      << "\thttps://github.com/gazebosim/gz-sim/pull/1888\n";
   }
 
   // Added mass according to Fossen's equations (p 37)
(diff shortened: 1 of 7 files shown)


ejalaa12:
This is an important and good thing to document quickly.
I had to find (through the source code) that the new parameter to use are <xUabsU> instead of <xUU>.

This also begs the question: in which scenario would you need <xUU> ? Basically, it would make the drag dependent on the direction of the movement. For example, in a forward movement, the drag slows down the link, but in a backward movement it would accelerate it. Which is weird, no?

Shouldn't the quadratic term always be multiplied the absolute velocity?

arjo129:
@quarkytale @mabelzhang pinging again for :heavy_check_mark: .

quarkytale:
I think tests aren't accounting for the range of weight typically the use-cases have. Lower weights might be preventing large translations.

arjo129:
> Lower weights might be preventing large translations

Doesn't quite make sense. Since `F = ma` won't `m` and `a`

> This also begs the question: in which scenario would you need ? Basically, it would make the drag dependent on the direction of the movement. For example, in a forward movement, the drag slows down the link, but in a backward movement it would accelerate it. Which is weird, no?

For diagonal terms this is true. Cross terms add forces due to motion in one axis along another axis.

quarkytale:
Well you are right. Refining that, the range of added mass and drag terms. The sim crashed for me when I had those values in range of 10^3 and more, less than that kept the sim running (but incorrect behavior of accelerating in backward motion). In the test's world we should probably have another model with higher values.

src/systems/hydrodynamics/Hydrodynamics.cc
```diff
@@ -207,6 +215,16 @@ void Hydrodynamics::Configure(
     }
   }
 
+
+  if (warnBehaviourChange)
+  {
+    ignwarn << "You are using parameters that may cause instabilities"
+      << "in your simulation. If your simulation crashes we recommend"
+      << "renaming <xUU> -> <xUabsU> and likewise for other axis"
```
quarkytale, line 223:
```suggestion
    ignwarn << "You are using parameters that may cause instabilities "
      << "in your simulation. If your simulation crashes we recommend "
      << "renaming <xUU> -> <xUabsU> and likewise for other axis "
```

quarkytale (approved):
LGTM after minor formatting

arjo129:
BTW Might be worth adding an assertion on the following line to check direction of motion. 
https://github.com/gazebosim/gz-sim/blob/c03c43261094d3d78a3706df46d10dbede80a918/test/integration/hydrodynamics.cc#L136

PR #1907, opened 2023-02-24
Added magnetometer value based on location

ahcorde:
# 🦟 Bug fix

## Summary

The magnetometer sensor is publishing a value but it's not based on the location. This PR calculate the value based on the location.

## Checklist
- [ ] Added tests


src/systems/magnetometer/Magnetometer.cc
```diff
@@ -286,6 +417,34 @@ void MagnetometerPrivate::Update(
...
+            gzwarn << "Failed to update NavSat sensor enity [" << _entity
+                    << "]. Spherical coordinates not set." << std::endl;
+            return true;
+          }
+
+          auto lat_rad = GZ_DTOR(latLonEle.value().X());
+          auto lon_rad = GZ_DTOR(latLonEle.value().Y());
+
+          // Magnetic declination and inclination (radians)
+          float declination_rad = get_mag_declination(lat_rad * 180 / M_PI, lon_rad * 180 / M_PI) * M_PI / 180;
```
mjcarroll, line 434: `GZ_RTOD`

src/systems/magnetometer/Magnetometer.cc
```diff
@@ -47,6 +47,73 @@ using namespace gz;
 using namespace sim;
 using namespace systems;
 
+/** set this always to the sampling in degrees for the table below */
+#define SAMPLING_RES		10.0f
```
mjcarroll, line 51: Prefer `static constexpr` I think for constants.

mjcarroll:
I think that we need some documentation on the derivation of these tables.  It's worth-while to note that these are for a) earth only and b) only valid in a certain window (I think they are updated every 5 years).

A more generic solution would allow for the consumption of the world magnetic model: https://www.ncei.noaa.gov/products/world-magnetic-model, but I understand if that's out of the scope of the current PR.

ahcorde:
I used this code from this other Gazebo classic plugin https://github.com/PX4/PX4-SITL_gazebo-classic/blob/5610c3fb441a2f3babc8ad7a63c8c4ce3e40abfa/src/geo_mag_declination.cpp

mjcarroll:
> I used this code from this other Gazebo classic plugin

Ah, makes sense, all good then.  I think the only other thing I would request is a switch to be able to switch between a constant magnetic declination/field and a varying one.  That way people can opt into the new behavior (in the unlikely case someone was depending on a constant field?)

ahcorde:
> > I used this code from this other Gazebo classic plugin
> 
> Ah, makes sense, all good then. I think the only other thing I would request is a switch to be able to switch between a constant magnetic declination/field and a varying one. That way people can opt into the new behavior (in the unlikely case someone was depending on a constant field?)

the spherical coordinates should turn on and off this feature.

mjcarroll:
> the spherical coordinates should turn on and off this feature.

Perfect, I think that is reasonable.

mjcarroll (approved):
LGTM with green ci

mjcarroll:
CI Failures here aren't from this PR.

peci1:
> the spherical coordinates should turn on and off this feature.

https://github.com/gazebosim/gz-sim/blob/8515827c5730501da03372d8e0a15615ef932cf2/src/systems/magnetometer/Magnetometer.cc#L459-L465

So it is true that missing spherical coords do disable this feature, but they would probably flood the console with these warnings... Or am I missing something?

PR #1913, opened 2023-03-03
Add Actor class

iche033:
# 🎉 New feature

Part of https://github.com/gazebosim/gz-sim/issues/325

## Summary

Add Actor class. Currently there are only a small set of API functions due to limited number of actor components available.

## Test it

Run the `UNIT_Actor_TEST` and `INTEGRATION_actor` tests.

## Checklist
- [x] Added tests

### DIFF
--- a/src/Actor.cc
+++ b/src/Actor.cc
@@ -1,0 +1,132 @@
+/*
+ * Copyright (C) 2023 Open Source Robotics Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+#include "ignition/gazebo/components/Actor.hh"
+#include "ignition/gazebo/components/Name.hh"
+#include "ignition/gazebo/components/ParentEntity.hh"
+#include "ignition/gazebo/components/Pose.hh"
+
+#include "ignition/gazebo/Actor.hh"
+#include "ignition/gazebo/Util.hh"
+
+using namespace ignition;
+using namespace gazebo;
+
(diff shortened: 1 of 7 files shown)


ahcorde (approved):

src/Actor.cc
```diff
@@ -0,0 +1,132 @@
...
+std::optional<math::Pose3d> Actor::Pose(
+    const EntityComponentManager &_ecm) const
+{
+  auto pose = _ecm.Component<components::Pose>(this->dataPtr->id);
+
+  if (!pose)
+    return std::nullopt;
+
+  return std::optional<math::Pose3d>(pose->Data());
+}
```
luca-della-vedova, line 76: I've personally been bitten a few times in the past by the fact that, if `TrajectoryPose` components are populated, actor's pose is actually not the `Pose` component but a combination of `TrajectoryPose` and `Pose`, as done [here](https://github.com/gazebosim/gz-sim/blob/13c582a2b428fbf1beb0028cd295eb346e568869/src/rendering/RenderUtil.cc#L1341).

For example if an actor has a pose in the origin but a `TrajectoryPose` component that moves it around, this function would return a constant, which is not what a user might expect.

iche033: good call, also reminds me about trajectory pose component! I updated the `Pose` function documentation to explain how to retrieve the actual world pose and also added `SetTrajectoryPose` / `TrajectoryPose` APIs.

luca-della-vedova: Awesome thanks a lot! I was going to followup asking if it would make sense to have an API to get the world pose but looking at the documentation of the new `TrajectoryPose` function it seems it still wouldn't behave as expected if the trajectory was set through an SDF script, so it seems there is no way to get an Actor's world pose that works in all cases?

iche033: I think the easiest way is to let RenderUtil compute the actor's world pose, create and populate a `WorldPose` component for the actor, and update the ECM. In the Actor class, we just retrieve the world pose from that component. I'll go ahead and give this a try

iche033: added `WorldPose` API in [2849026](https://github.com/gazebosim/gz-sim/pull/1913/commits/2849026f08ca136a146e64bfc9d21594dfccca47)

Here's the WorldPose component being updated in `actor.sdf` world:
actor_world_pose

I also tested with `follow_actor.sdf` which manually updates the trajectory pose using the `TrajectoryPose` component, and the `WorldPose` components correctly reflects the actual pose.

luca-della-vedova (approved):
Awesome thanks for the API updates!

PR #1911, opened 2023-03-02
Address a few Windows CI Issues

mjcarroll:
* Disabled the failing Hydrodynamics test (everything else in the file is disabled)
 * Suppress some MSVC-specific warnings.
 * Fixes using incorrect export header

PR #1917, opened 2023-03-06
Get Windows to green on gz-sim7

mjcarroll:
Follow up to #1911

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -68,7 +68,6 @@ set (sources
   Util.cc
   View.cc
   World.cc
-  cmd/ModelCommandAPI.cc
   ${PROTO_PRIVATE_SRC}
   ${network_sources}
   ${comms_sources}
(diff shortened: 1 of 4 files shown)


src/CMakeLists.txt
```diff
@@ -68,7 +68,6 @@ set (sources
   Util.cc
   View.cc
   World.cc
-  cmd/ModelCommandAPI.cc
   ${PROTO_PRIVATE_SRC}
```
azeey, line 71: Ah, I see why `ModelCommandAPI` was behaving differently with msvc warnings than `gz.cc`

src/SdfGenerator_TEST.cc
```diff
@@ -514,7 +514,7 @@ TEST_F(ElementUpdateFixture, ConfigOverride)
     auto uri = inclElem->Get<std::string>("uri");
     EXPECT_FALSE(uri.empty());
     const std::string version = common::split(uri, "/").back();
-    EXPECT_NO_THROW(std::stol(version));
+    EXPECT_NO_THROW(auto version_parsed = std::stol(version));
```
azeey, line 517: nit: 
```suggestion
    EXPECT_NO_THROW(auto versionParsed = std::stol(version));
```

azeey (commented):
I've restarted the windows build.

azeey (approved):
I believe most, if not all, of the CI failures are due to a Fuel server issue.

PR #1927, opened 2023-03-13
Proposal to add deadband to thruster

ejalaa12:
# 🎉 New feature

This PR is a proposal to add a deadband to the thruster system.
Real thrusters have a physical deadband that are not currently simulated.

## Summary

This implements a limit below which the thrusters will not spin nor generate thrust. 

It adds an additional parameter in the sdf called `<deadband>`.
 - If this tag is not present, the deadband will not be applied.
 - otherwise the value in the tag is the absolute value of thrust below which the output will be set to 0.

An additional topic called `/enable_deadband` is also created to enable or disable the deadband effect during runtime.

## Test it

Unit test are coming...

## Checklist
- [ ] Added tests

### DIFF
--- a/src/systems/thruster/Thruster.cc
+++ b/src/systems/thruster/Thruster.cc
@@ -14,6 +14,8 @@
  * limitations under the License.
  *
  */
+#include <cstdlib>
+#include <gz/transport/TopicUtils.hh>
 #include <memory>
 #include <mutex>
 #include <limits>
@@ -144,6 +146,15 @@ class gz::sim::systems::ThrusterPrivateData
   /// \brief Linear velocity of the vehicle.
   public: double linearVelocity = 0.0;
 
+  /// \brief deadband
+  public: double deadband = 0.0;
+
+  /// \brief Flag to enable/disable deadband
+  public: bool enableDeadband = false;
+
+  /// \brief Topic name used to enable/disable the deadband
+  public: std::string deadbandTopic = "";
+
   /// \brief Topic name used to control thrust. Optional
   public: std::string topic = "";
 
@@ -158,6 +169,9 @@ class gz::sim::systems::ThrusterPrivateData
 
   /// \brief Callback for handling thrust update
   public: void OnCmdThrust(const msgs::Double &_msg);
+
+  /// \brief Callback for handling deadband enable/disable update
+  public: void OnDeabandEnable(const msgs::Boolean &_msg);
 
   /// \brief Recalculates and updates the thrust coefficient.
   public: void UpdateThrustCoefficient();
@@ -179,6 +193,10 @@ class gz::sim::systems::ThrusterPrivateData
   /// \return True if battery is charged, false otherwise. If no battery found,
   /// returns true.
   public: bool HasSufficientBattery(const EntityComponentManager &_ecm) const;
+          
+  /// \brief Applies the deadband to the thrust and angular velocity by setting
+  /// those values to zero if their absolute value is below the deadband
+  public: void ApplyDeadband(double &_thrust, double &_angVel);
 };
 
 /////////////////////////////////////////////////
@@ -276,6 +294,13 @@ void Thruster::Configure(
               << "[thrust_coefficient] value from the SDF file." << std::endl;
     }
   }
+  
+  // Get deadband, default to 0
+  if (_sdf->HasElement("deadband"))
+  {
+    this->dataPtr->deadband = _sdf->Get<double>("deadband");
+    this->dataPtr->enableDeadband = true;
+  }
 
   // Get a custom topic.
   if (_sdf->HasElement("topic"))
@@ -312,6 +337,8 @@ void Thruster::Configure(
     // Subscribe to specified topic for force commands
     thrusterTopic = gz::transport::TopicUtils::AsValidTopic(
       ns + "/" + this->dataPtr->topic);
+    this->dataPtr->deadbandTopic = gz::transport::TopicUtils::AsValidTopic(
+      ns + "/" + this->dataPtr->topic + "/enable_deadband");
     if (this->dataPtr->opmode == ThrusterPrivateData::OperationMode::ForceCmd)
     {
       this->dataPtr->node.Subscribe(
@@ -347,6 +374,9 @@ void Thruster::Configure(
 
     feedbackTopic = gz::transport::TopicUtils::AsValidTopic(
       "/model/" + ns + "/joint/" + jointName + "/ang_vel");
+    
+    this->dataPtr->deadbandTopic = gz::transport::TopicUtils::AsValidTopic(
+      "/model/" + ns + "/joint/" + jointName + "/enable_deadband");
   }
   else
   {
@@ -362,10 +392,19 @@ void Thruster::Configure(
 
     feedbackTopic = gz::transport::TopicUtils::AsValidTopic(
         "/model/" + ns + "/joint/" + jointName + "/force");
+
+    this->dataPtr->deadbandTopic = gz::transport::TopicUtils::AsValidTopic(
+      "/model/" + ns + "/joint/" + jointName + "/enable_deadband");
   }
 
   gzmsg << "Thruster listening to commands on [" << thrusterTopic << "]"
         << std::endl;
+  this->dataPtr->node.Subscribe(
+      this->dataPtr->deadbandTopic,
+      &ThrusterPrivateData::OnDeabandEnable,
+      this->dataPtr.get());
+  gzmsg << "Thruster listening to enable_deadband on [" 
+        << this->dataPtr->deadbandTopic << "]" << std::endl;
 
   this->dataPtr->pub = this->dataPtr->node.Advertise<msgs::Double>(
       feedbackTopic);
@@ -476,6 +515,19 @@ void ThrusterPrivateData::OnCmdThrust(const gz::msgs::Double &_msg)
 }
 
 /////////////////////////////////////////////////
+void ThrusterPrivateData::OnDeabandEnable(const gz::msgs::Boolean &_msg)
+{
+  if (_msg.data() != this->enableDeadband)
+  {
+    if (_msg.data()) gzmsg << "Enabling deadband." << std::endl;
+    else gzmsg << "Disabling deadband." << std::endl;
+    
+    this->enableDeadband = _msg.data();
+  }
+
+}
+
+/////////////////////////////////////////////////
 void ThrusterPrivateData::OnCmdAngVel(const gz::msgs::Double &_msg)
 {
   std::lock_guard<std::mutex> lock(mtx);
@@ -551,6 +603,16 @@ bool ThrusterPrivateData::HasSufficientBattery(
 }
 
 /////////////////////////////////////////////////
+void ThrusterPrivateData::ApplyDeadband(double &_thrust, double &_angVel)
+{
+    if (abs(_thrust) < this->deadband)
+    {
+        _thrust = 0.0;
+        _angVel = 0.0;
+    }
+}
+
+/////////////////////////////////////////////////
 void Thruster::PreUpdate(
   const gz::sim::UpdateInfo &_info,
   gz::sim::EntityComponentManager &_ecm)
@@ -560,6 +622,9 @@ void Thruster::PreUpdate(
 
   if (!this->dataPtr->enabled)
   {
+    return;
+  }
+  if (!_ecm.HasEntity(this->dataPtr->linkEntity)){
     return;
   }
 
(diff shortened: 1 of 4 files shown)


src/systems/thruster/Thruster.cc
```diff
@@ -475,6 +514,19 @@ void ThrusterPrivateData::OnCmdThrust(const gz::msgs::Double &_msg)
   this->propellerAngVel = this->ThrustToAngularVec(this->thrust);
 }
 
+/////////////////////////////////////////////////
+void ThrusterPrivateData::OnDeabandEnable(const gz::msgs::Boolean &_msg)
+{
+  if (_msg.data() != this->enableDeadband)
```
caguero, line 520: I recommend to use a mutex to protect `this->enableDeadband`.

src/systems/thruster/Thruster.hh
```diff
@@ -85,6 +85,8 @@ namespace systems
   ///                      [Optional, defaults to 1000N or 1000rad/s]
   /// - <min_thrust_cmd> - Minimum input thrust or angular velocity command.
   ///                      [Optional, defaults to -1000N or -1000rad/s]
+  /// - <deadband> - Deadband of the thruster. Absolute value below which the 
+  ///                thruster won't spin nor generate thrust
```
caguero, line 89: Could you also mention that can be enabled/disabled via messages?

src/systems/thruster/Thruster.cc
```diff
@@ -475,6 +514,19 @@ void ThrusterPrivateData::OnCmdThrust(const gz::msgs::Double &_msg)
...
+  {
+    if (_msg.data()) gzmsg << "Enabling deadband." << std::endl;
```
caguero, line 522: Nit: For consistency, could you move to the next line the `gzmsg` instructions?

caguero (commented):
Thanks for the contribution! I did a first pass and left a few minor comments.

src/systems/thruster/Thruster.cc
```diff
@@ -159,6 +170,9 @@ class gz::sim::systems::ThrusterPrivateData
   /// \brief Callback for handling thrust update
   public: void OnCmdThrust(const msgs::Double &_msg);
 
+  /// \brief Callback for handling deadband enable/disable update
```
ahcorde, line 173: doc argument

src/systems/thruster/Thruster.cc
```diff
@@ -179,6 +193,10 @@ class gz::sim::systems::ThrusterPrivateData
   /// \return True if battery is charged, false otherwise. If no battery found,
   /// returns true.
   public: bool HasSufficientBattery(const EntityComponentManager &_ecm) const;
+          
+  /// \brief Applies the deadband to the thrust and angular velocity by setting
+  /// those values to zero if their absolute value is below the deadband
```
ahcorde, line 198: doc arguments

src/systems/thruster/Thruster.cc
```diff
@@ -347,6 +374,9 @@ void Thruster::Configure(
 
     feedbackTopic = gz::transport::TopicUtils::AsValidTopic(
       "/model/" + ns + "/joint/" + jointName + "/ang_vel");
+    
+    this->dataPtr->deadbandTopic = gz::transport::TopicUtils::AsValidTopic(
```
ahcorde, line 378: check if the output is `empty()` and handle the error

src/systems/thruster/Thruster.cc
```diff
@@ -362,10 +392,19 @@ void Thruster::Configure(
 
     feedbackTopic = gz::transport::TopicUtils::AsValidTopic(
         "/model/" + ns + "/joint/" + jointName + "/force");
+
+    this->dataPtr->deadbandTopic = gz::transport::TopicUtils::AsValidTopic(
```
ahcorde, line 396: check if the output is `empty()` and handle the error

ahcorde (changes requested):

src/systems/thruster/Thruster.cc
ejalaa12: is this really necessary? as the only places where there is write-access to `this->enableDeadband` is in this method and when loading the sdf (which is necessary before the callback exist). So I don't see any conflict. Or did I miss something?

src/systems/thruster/Thruster.hh
ejalaa12: sure !

src/systems/thruster/Thruster.cc
ejalaa12: I went ahead and ran `make codecheck` and fixed a few other code style issues.

src/systems/thruster/Thruster.cc
ejalaa12: same remark as above

ejalaa12:
I just pushed the changes.

ejalaa12:
Sure I'm working on it and I'll update the PR by next week

ejalaa12:
@arjo129 just added the unit test for that feature, thanks for reviewing

test/integration/thruster.cc
```diff
@@ -285,14 +314,50 @@ void ThrusterTest::TestWorld(const std::string &_world,
...
+  {
+    // disable the deadband
+    db_msg.set_data(false);
+    db_pub.Publish(db_msg);
+    // And we send a command that is still below the deadband threshold
+    msg.set_data(_deadband / 2.0);
+    pub.Publish(msg);
+    // Check movements
+    fixture.Server()->Run(true, 1000, false);
+    std::this_thread::sleep_for(std::chrono::milliseconds(100));
```
arjo129, line 347: To the best of my knowledge, there is no need for this sleep. You already run the server in blocking mode.

test/integration/thruster.cc
```diff
@@ -285,14 +314,50 @@ void ThrusterTest::TestWorld(const std::string &_world,
...
+    EXPECT_EQ(1000u, propellerAngVels.size());
+    EXPECT_EQ(1000u, propellerLinVels.size());
+    modelPoses.clear();
+    propellerAngVels.clear();
+    propellerLinVels.clear();
+  }
+  // re-enable the deadband
+  db_msg.set_data(true);
+  db_pub.Publish(db_msg);
+
```
arjo129, line 360: And then what does this achieve? We are not checking any assertions here.

test/integration/thruster.cc
```diff
@@ -285,14 +314,50 @@ void ThrusterTest::TestWorld(const std::string &_world,
...
+    // disable the deadband
+    db_msg.set_data(false);
+    db_pub.Publish(db_msg);
+    // And we send a command that is still below the deadband threshold
+    msg.set_data(_deadband / 2.0);
+    pub.Publish(msg);
+    // Check movements
+    fixture.Server()->Run(true, 1000, false);
+    std::this_thread::sleep_for(std::chrono::milliseconds(100));
+    EXPECT_LT(0.1, modelPoses.back().Pos().X());
```
arjo129, line 348: A few comments of what these expectations mean would be helpful. We usually check velocity as opposed to position as that results in less flaky tests.

src/systems/thruster/Thruster.cc
arjo129: Yes. That'd be a good idea.

src/systems/thruster/Thruster.cc
```diff
@@ -362,10 +396,19 @@ void Thruster::Configure(
...
+      "/model/" + ns + "/joint/" + jointName + "/enable_deadband");
   }
 
   gzmsg << "Thruster listening to commands on [" << thrusterTopic << "]"
         << std::endl;
+  this->dataPtr->node.Subscribe(
+      this->dataPtr->deadbandTopic,
```
arjo129, line 407: If the `deadbandTopic` is empty, we should not subscribe. This does happen during the tests and we get a complaint from `gz-transport`

arjo129 (changes requested):
I like the idea for the feature. However, this does not compile :( (see my comments and CI). After fixing the syntax errors, I am able to run the tests. There are still some clean ups needed. A bit of documentation on the expectations would help me understand the test better.

test/integration/thruster.cc
```diff
@@ -372,7 +437,7 @@ TEST_F(ThrusterTest, GZ_UTILS_TEST_DISABLED_ON_WIN32(VelocityControl))
 /////////////////////////////////////////////////
 TEST_F(ThrusterTest, GZ_UTILS_TEST_DISABLED_ON_WIN32(BatteryIntegration))
 {
-  const std::string ns = "lowbattery";
+:string ns = "lowbattery";
```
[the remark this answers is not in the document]
ejalaa12: This was indeed a typo. I don't know how I let it pass. Sorry i'll fix this in the next commit

test/integration/thruster.cc
ejalaa12: I simply copied this block from a few lines above: https://github.com/gazebosim/gz-sim/blob/4eeed347a12e3e63f61b7f0148ce41b947d8ec98/test/integration/thruster.cc#L207-L208

I also don't know why it was necessary there. I'll try re-running the test without the sleep and tell you if it works.

test/integration/thruster.cc
ejalaa12: that is indeed not necessary. I'll remove those lines in the next commit

test/integration/thruster.cc
ejalaa12: Note that the sleep is also present here:
https://github.com/gazebosim/gz-sim/blob/42c285ef1a7b35d7e64c8fcdc4f270fb2b42b9e3/test/integration/thruster.cc#L216-L217

arjo129: Thanks for pointing this out.

test/integration/thruster.cc
ejalaa12: After giving it some thoughts, one reason I had this at the end of the test, was in case someone wants to extend the test and forgot to re-enable the deadband. But I guess, its the tester responsibility to make sure they are in the conditions they want. So I'll remove it anyway.

test/integration/thruster.cc
ejalaa12: That is a good point, I'll use the same equations for checking the propeller speed used earlier in the test.

src/systems/thruster/Thruster.cc
ejalaa12: The complaints do not come from the plugin but from the tests them-selves.
In the test I create a publisher to `deadbandTopic` which is empty in most unit test except the one I added. So I added a check for emptiness there.
I also did not subscribe to the topic in the plugin, if the topic is empty, which is also a good idea.

ejalaa12:
@arjo129 I addressed your remarks in the new commit. Btw, should I rebase this PR on gz-sim7 ?

ejalaa12:
I had to do a force-push to reword my previous commits in order to sign them off

src/systems/thruster/Thruster.cc
arjo129: Thanks! I didn't notice that.

ejalaa12:
@arjo129 did you have time to look at my latest commits ?

ejalaa12:
@arjo129 friendly ping :)

src/systems/thruster/Thruster.cc
arjo129: In practice, this is unlikely to be a problem.

However there are several issues at play here. First, callbacks from `gz-transport` are run on a separate thread. This means that when a message arrives and is processed there is no guarantee that it will run on the same CPU core. Thus if you don't have any synchronization primitive there is no guarantee that `this->enableDeadband` will be read correctly on the next round due to cache coherency issues (see: https://blog.the-pans.com/std-atomic-from-bottom-up/).

test/integration/thruster.cc
arjo129: To clarify, the sleep is also probably not needed in the above test cases you've pointed out. I believe its best new code does not use the sleep.

arjo129:
Sorry for the (really looong) turn around time. There are still two unresolved threads. Please address those.

ejalaa12:
Hey @arjo129, NP, thanks for your last comment. 
I addressed the unresolved threads.

arjo129 (changes requested):
Apologies for being obsessive about this, but spotted a few more minor issues. If you're observing flakiness with the test you need only check the last item in the buffer.

ejalaa12:
@arjo129 no worries, thanks for sticking up with this PR.
I protected the mentioned block with a mutex, and adjusted the pid settings in the world file to make the commented section work.

However I had to rollback the removal of the std::sleep_for, otherwise it breaks the other tests, i'm not entirely sure why. We can probably open a separate issue for this.

arjo129 (approved):
Thanks for iterating on this @ejalaa12. LGTM. We can look into the sleep related issue elsewhere.

mjcarroll (approved):

ejalaa12:
Thanks for the merge ! :)
Happy to contribute to this project 👍🏼

PR #1938, opened 2023-03-21
Optimize render updates and use of thread mutexes in Sensors system

iche033:
~**Marked as draft to debug test failures**~ Fixed

# 🦟 Bug fix

Related to: https://github.com/gazebosim/gz-sensors/issues/332

## Summary

While profiling the sensors system, I noticed that it spends a quite bit of time
* waiting in the [Sensors::Update](https://github.com/gazebosim/gz-sim/blob/gz-sim7/src/systems/sensors/Sensors.cc#L602-L606) function for rendering operation to finish and release the [renderMutex](https://github.com/gazebosim/gz-sim/blob/gz-sim7/src/systems/sensors/Sensors.cc#L606). This is not necessary since the `RenderUtil` calls inside the `Update` function should be thread-safe (protected by mutex inside RenderUtil).
* doing scene tree updates (`scene->PreRender()`, `scene->PostRender`, `renderUtil.Update()`) when there are no sensor or render event subscribers

This PR optimizes the threading and render updates in the sensors system by:
* ~removing~ reducing the scope of the `renderMutex` lock in `Sensors::Update` function
~* use a separate mutex for the render condition variable to reduce places that need to lock the `renderMutex`~
* changing a few bool variables to atomic, also to reduce the need to lock mutexes
* performing scene tree updates only when necessary

## Test results

Here are the RTF changes that I got from running different worlds (i7 3.4GHz 32GB RAM, NVIDIA GeForce GTX 1070 8GB VRAM). I added a best fit line to the RTF values. There is an improvement in RTF and noticeably less jitter.

*  `sensors_demo.sdf` world (with lidar visualization on)

    * **Before**
sensors_demo_before

    * **After**
sensors_demo_after

*  [`profile_lidar.sdf`](https://gist.github.com/iche033/2b3aa693687ed8f1742720bfd1fbb6a0) world (5 lidars inside [Depot](https://app.gazebosim.org/OpenRobotics/fuel/models/Depot) model. All lidars are subscribed)

    * **Before**
profile_lidar_before

    * **After**
profile_lidar_after

## Checklist
- [x] Added tests

### DIFF
--- a/src/systems/sensors/Sensors.cc
+++ b/src/systems/sensors/Sensors.cc
@@ -17,8 +17,6 @@
 
 #include "Sensors.hh"
 
-#include <chrono>
-
 #include <map>
 #include <set>
 #include <unordered_map>
@@ -32,8 +30,6 @@
 #include <sdf/Sensor.hh>
 
 #include <gz/math/Helpers.hh>
-
-#include <gz/common/Timer.hh>
 
 #include <gz/rendering/Scene.hh>
 #include <gz/sensors/BoundingBoxCameraSensor.hh>
@@ -127,6 +123,8 @@ class gz::sim::systems::SensorsPrivate
 
   /// \brief Mutex to protect rendering data
   public: std::mutex renderMutex;
+
+  /// \brief Mutex to protect renderCv
   public: std::mutex cvMutex;
 
   /// \brief Condition variable to signal rendering thread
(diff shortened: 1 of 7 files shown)


src/systems/sensors/Sensors.cc
```diff
@@ -140,15 +139,17 @@ class gz::sim::systems::SensorsPrivate
...
+  public: std::chrono::steady_clock::duration nextUpdateTime;
+
   /// \brief Sensors to include in the next rendering iteration
-  public: std::vector<sensors::RenderingSensor *> activeSensors;
+  public: std::set<sensors::SensorId> activeSensors;
 
-  /// \brief Mutex to protect sensorMask
-  public: std::mutex sensorMaskMutex;
+  /// \brief Sensors to be updated next
+  public: std::set<sensors::SensorId> sensorsToUpdate;
```
mjcarroll, line 149: Can probably use `unordred_set` here

mjcarroll (commented):
Overall approach looks good, I'm going to allocate a little more time to review this afternoon as this is pretty tricky logic.

mjcarroll (approved):
Finally got to think about this one enough, and I'm happy with the approach. Thanks for tackling this tricky bit of logic!

ahcorde:
@iche033 do you have any plans to backport this to fortress?

iche033:
> @iche033 do you have any plans to backport this to fortress?

yes I'll backport this to fortress

Issue #1410, opened 2022-03-24
Rendering engines are not unloaded or cleaned up when `Server` is destroyed


azeey:
## Environment
* OS Version: Ubuntu 20.04 
* Source or binary build? ign-gazebo5
    

## Description
* Expected behavior: The rendering engine to be unloaded when `Server` (or more specifically, `RenderUtil`) is destroyed. 
* Actual behavior: The rendering engine and all the scenes within it remain intact after `Server` is destroyed. Loading the same world in two tests in the same file causes the test to segfault. @marcoag ran into this while working on #1331.

## Steps to reproduce
Add these two tests to `test/integration/depth_camera.cc` (or any integration test file) and run with `--gtest_filter="*Cleanup*"`
```cpp

TEST(RenderingCleanup, Cleanup1)
{
  // Start server
  ServerConfig serverConfig;
  const auto sdfFile = std::string(PROJECT_SOURCE_PATH) +
    "/test/worlds/depth_camera_sensor.sdf";
  serverConfig.SetSdfFile(sdfFile);

  Server server(serverConfig);

  server.Run(true, 10, false);
}

TEST(RenderingCleanup, Cleanup2)
{
  // Start server
  ServerConfig serverConfig;
  const auto sdfFile = std::string(PROJECT_SOURCE_PATH) +
    "/test/worlds/depth_camera_sensor.sdf";
  serverConfig.SetSdfFile(sdfFile);

  Server server(serverConfig);

  server.Run(true, 10, false);
}

```

## Output
```
❯ bin/INTEGRATION_depth_camera --gtest_filter="*Cleanup*"
Running main() from /home/addisu/ws/edifice/src/ign-gazebo/test/gtest/src/gtest_main.cc
Note: Google Test filter = *Cleanup*
[==========] Running 2 tests from 1 test suite.
[----------] Global test environment set-up.
[----------] 2 tests from RenderingCleanup
[ RUN      ] RenderingCleanup.Cleanup1
[       OK ] RenderingCleanup.Cleanup1 (1280 ms)
[ RUN      ] RenderingCleanup.Cleanup2
[Err] [SceneManager.cc:179] Visual: [ground_plane] already exists
[Err] [SceneManager.cc:179] Visual: [box] already exists
[Err] [SceneManager.cc:179] Visual: [depth_camera] already exists
[Err] [BaseStorage.hh:927] Another item already exists with name: sun
[...]

PR #1960, opened 2023-04-07
Unload render engine when the sensors system exits

iche033:
# 🦟 Bug fix

Fixes https://github.com/gazebosim/gz-sim/issues/1410

Depends on: 
* https://github.com/gazebosim/gz-sensors/pull/340
* https://github.com/gazebosim/gz-rendering/pull/840

## Summary

The render engine is never destroyed / unloaded when the server shuts down. This leads to issues such as memory leak and not being able to relaunch the server with the sensors system, e.g. as seen in #1410, rendering tests n gz-sim can not be run one after another in the same process.

This PR adds a function in RenderUtil to destroy the scene and unload the engine.

To test:

Run the `PERFORMANCE_sensors_system` test that repeatedly launches the server with the sensors system.

Note that to run the tests, you'll need the above PRs in gz-rendering and gz-sensors otherwise gz will crash. 

## Checklist
- [x] Added tests

### DIFF
--- a/test/performance/sensors_system.cc
+++ b/test/performance/sensors_system.cc
@@ -80,6 +80,8 @@ class SensorsSystemFixture : public InternalFixture<::testing::Test>
 // memory leaks
 TEST_F(SensorsSystemFixture, GZ_UTILS_TEST_DISABLED_ON_MAC(MemLeak))
 {
+  gz::common::Console::SetVerbosity(4);
+
   // max memory change allowed
   const double resMaxPercentChange = 1.0;
   const double shareMaxPercentChange = 1.0;
--- a/test/performance/sensors_system.cc
+++ b/test/performance/sensors_system.cc
@@ -116,7 +116,8 @@ TEST_F(SensorsSystemFixture, GZ_UTILS_TEST_DISABLED_ON_MAC(MemLeak))
   // to increase memory usage
   for (unsigned int i = 1; i < residentMemV.size() - 1; ++i)
   {
-    double resPercentChange = (residentMemV[i+1] - residentMemV[i]) / residentMemV[i];
+    double resPercentChange =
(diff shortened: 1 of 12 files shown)


test/performance/sensors_system.cc
```diff
@@ -0,0 +1,135 @@
...
+  // Start from index 1 (instead of 0) since there is a big jump in memory usage
+  // when the system is first loaded. Make sure subsequent runs do not continue
+  // to increase memory usage
+  for (unsigned int i = 1; i < residentMemV.size() - 1; ++i)
+  {
+    double resPercentChange =
+        (residentMemV[i+1] - residentMemV[i]) / residentMemV[i];
+    double sharePercentChange = (shareMemV[i+1] - shareMemV[i]) / shareMemV[i];
+
+    gzdbg << "ResPercentChange[" << resPercentChange << "] "
```
mjcarroll, line 123: Since you aren't explicitly setting the console log level, I don't know that these are guaranteed to come out as part of the test results?

iche033: added a call to set verbosity level to make sure we see these console messages [bd3ff2e](https://github.com/gazebosim/gz-sim/pull/1960/commits/bd3ff2eb5491808362623851ffcde138b07229e7)

mjcarroll:
I think this is still failing `PERFORMANCE_sensors_system` on the buildfarm, so it may need another look.

iche033:
> I think this is still failing `PERFORMANCE_sensors_system` on the buildfarm, so it may need another look.

removed bad `ifdef` logic in [e26e577](https://github.com/gazebosim/gz-sim/pull/1960/commits/e26e577c115a56e69d75a4c281413494907bb98b), which should hopefully fix it.

iche033:
ubuntu build fixed

mjcarroll (approved):

PR #1962, opened 2023-04-11
Speed up Resource Spawner load time by fetching model list asynchronously

### LINKED ISSUE #1253: Speed up resource spawner load time
## Environment
* OS Version:  all
* Source or binary build? all

## Description
* Expected behavior:  Resource spawner GUI plugin shows models soon after being loaded
* Actual behavior:  It shows "Please wait... Loading models from Fuel" for minutes sometimes.

One idea for a fix would be to load the Open Robotics models first, and then all models from other users subsequently. We should populate the menu while more models are loaded on the background, instead of waiting until all models from all owners are loaded. We may need to add APIs to `ign-fuel-tools` for that.

## Steps to reproduce

1. `ign gazebo`
2. Insert `Resource Spawner` from top-right corner
3. Wait

azeey:
# 🦟 Bug fix

Fixes #1253 

## Summary
The main reason that the Resource Spawner took a long time to load is that it tried to fetch the list of all available models on Fuel instead of just the selected owner. And it did that while blocking the Qt thread, so the user was unable to interact with the GUI while the model list was being fetched. The approach I've taken is to only fetch the list of models for the default owner ("openrobotics") and allow users to add/remove any owner they want. The fetching is also done in a separate PR so as to not block the GUI.

This currently does not preserve the list of owners added by the user.  I thought it would be better to add that in a follow up PR at a future time.

Needs https://github.com/gazebosim/gz-fuel-tools/pull/350

## Checklist
- [ ] Added tests


mjcarroll:
This one has conflicts now.

mjcarroll (approved):
Great improvement. LGTM with green CI!

PR #1967, opened 2023-04-21
Use ignition::gazebo:: in class instantiation

scpeters:
# 🦟 Bug fix

Fixes macOS build.

## Summary

The macOS build has been broken since https://github.com/gazebosim/gz-sim/pull/1646:

* [Build Status](https://build.osrfoundation.org/view/ign-citadel/job/ignition_gazebo-ci-ign-gazebo3-homebrew-amd64/282/) https://build.osrfoundation.org/view/ign-citadel/job/ignition_gazebo-ci-ign-gazebo3-homebrew-amd64/282/

In other packages, I have avoided using the `gz::sim` namespace when implementing classes, so doing it here seems to fix the macOS build. The ABI build is broken too, but that needs a different fix.

## Checklist
- [X] Signed all commits for DCO
- [ ] Added tests
- [ ] Updated documentation (as needed)
- [ ] Updated migration guide (as needed)
- [ ] Consider updating Python bindings (if the library has them)
- [X] `codecheck` passed (See [contributing](https://gazebosim.org/docs/all/contributing#contributing-code))
- [X] All tests passed (See [test coverage](https://gazebosim.org/docs/all/contributing#test-coverage))
- [ ] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Agazebosim+archived%3Afalse+) to support the maintainers

**Note to maintainers**: Remember to use **Squash-Merge** and edit the commit message to match the pull request summary while retaining `Signed-off-by` messages.

PR #1983, opened 2023-05-02
Backport gz/sim.hh fix, add redirection header

scpeters:
# 🦟 Bug fix

Backports #1978 and adds a missing header file

## Summary

This backports a fix for using the `gz/sim.hh` header file, but this header file does not yet exist on the `ign-gazebo3` branch as a consequence of how gz-cmake derives the auto-generated header name from the cmake project name. For ignition-gazebo*, the auto-generated header name is `gz/gazebo.hh`. Rather than try to modify gz-cmake, I've added a redirection header file from `gz/sim.hh` to `gz/gazebo.hh` for this branch. This should not be merged forward to `gz-sim*`. This is similar to the [sdf/sdf.hh](https://github.com/gazebosim/sdformat/blob/sdf13/include/sdf/sdf.hh) redirection header file used in SDFormat.

## Checklist
- [X] Added tests

**Note to maintainers**: Remember to use **Rebase-and-merge**.

### DIFF
--- a/include/gz/sim.hh
+++ b/include/gz/sim.hh
@@ -1,0 +1,29 @@
+/*
+ * Copyright (C) 2023 Open Source Robotics Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+#ifndef GZ_SIM_HH_
+#define GZ_SIM_HH_
+
+// Since gz-cmake derives the auto-generated header name from the cmake
+// project name, for ignition-gazebo*, the auto-generated header name
+// is <gz/gazebo.hh>.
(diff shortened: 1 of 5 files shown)


scpeters:
there's a problem that the `gz/gazebo.hh` header isn't installed either

scpeters:
> there's a problem that the `gz/gazebo.hh` header isn't installed either

actually I think it's fine

mjcarroll (approved):

scpeters:
I think this will fix the abichecker build after it is merged, but the job currently fails in this PR when trying to compile the `ign-gazebo3` branch

scpeters:
homebrew is also failing, but that should be fixed by #1967. I will merge this now, and we should see clean CI in #1967

include/gz/sim.hh
```diff
@@ -0,0 +1,29 @@
...
+ *
+ */
+
+#ifndef GZ_SIM_HH_
+#define GZ_SIM_HH_
+
+// Since gz-cmake derives the auto-generated header name from the cmake
+// project name, for ignition-gazebo*, the auto-generated header name
+// is <gz/gazebo.hh>.
+// This header file provides a redirection from <gz/sim.hh> to <gz/gazebo.hh>
```
azeey, line 24: Do we plan to have `gz/gazebo.hh` in gz-sim that forwards to `gz/sim.hh`? Otherwise, if users start using the header with `ign-gazebo3` or `ign-gazebo6` and then want to migrate to gz-sim, they will encounter build errors.

scpeters: yes, they would encounter build errors. I suppose we could add a `gz/gazebo.hh` header that redirects to `gz/sim.hh` and generates a deprecation warning when used

scpeters: it doesn't seem likely to me but it is possible

PR #2000, opened 2023-05-30
Update gz-sim to use new message generation pipeline

mjcarroll:
Allows gz-sim to take advantage of the new message generation pipeline in https://github.com/gazebosim/gz-msgs/pull/339


test/integration/optical_tactile_plugin.cc
```diff
@@ -18,11 +18,13 @@
 #include <gtest/gtest.h>
 
 #include <gz/msgs/contacts.pb.h>
+#include <gz/msgs/point_cloud_packed.pb.h>
```
iche033, line 21:
```suggestion
#include <gz/msgs/pointcloud_packed.pb.h>
```

test/integration/reset.cc
```diff
@@ -17,6 +17,9 @@
 
 #include <gtest/gtest.h>
 
+#include <gz/msgs/boolean.pb.h>
+#include <gz/msgs/world_control.pb.h>
```
iche033, line 21: this comment is actually for `test/integration/reset_sensors.cc`:

We'll also need to add `#include <gz/msgs/world_control.pb.h>` to that file to fix a compile error

PR #2010, opened 2023-06-09
Tracks periodic changes in scene broadcaster

arjo129:
This commit proposes a change to the scene broadcaster which enables tracking of all components with periodic changes. This way if a component has a periodic change the scene broadcaster does not miss it.

For more info see this discussion https://github.com/gazebosim/gz-sim/pull/2001#issuecomment-1581573728 where @azeey proposes this solution.

This commit is WIP (Work In Progress) and I still need to handle entity/component removal and add tests.

### DIFF
--- a/include/ignition/gazebo/EntityComponentManager.hh
+++ b/include/ignition/gazebo/EntityComponentManager.hh
@@ -602,8 +602,12 @@ namespace ignition
           ComponentTypesWithPeriodicChanges() const;
 
       /// \brief Get all components with periodic changes.
-      /// \param[in] _f Callback function to be called for each matching entity
-      /// and component that has experienced a change.
+      /// \param[in] _changes A list of components with the latest periodic
+      /// changes. If a component has a periodic change, it is added to the
+      /// hash map. It the component or entity was removed, it is removed from
+      /// the hashmap. This way the hashmap stores a list of components and
+      /// entities which have had periodic changes in the past and still
+      /// exit within the ECM.
       public: void PeriodicChangeEntityComponentMap(std::unordered_map<Entity,
                         std::unordered_set<ComponentTypeId>> &_changes) const;
 
@@ -637,10 +641,10 @@ namespace ignition
       /// \brief Get a message with the serialized state of the given entities
       /// and components.
       /// \details The header of the message will not be populated, it is the
-      /// responsibility of the caller to timestamp it before use.
-      /// \param[out] _state The serialized state message to populate.
+      /// responsibility of the caller to timestamp it before use. Additionally,
+      /// changes such as addition or removal will not be populated.
+      /// \param[inout] _state The serialized state message to populate.
       /// \param[in] _entityMap A map of entities and components to serialize.
-      /// False will get only components and entities that have changed.
       public: void State(
                   msgs::SerializedStateMap &_state,
                   const std::unordered_map<Entity,
--- a/include/ignition/gazebo/EntityComponentManager.hh
+++ b/include/ignition/gazebo/EntityComponentManager.hh
@@ -606,7 +606,7 @@ namespace ignition
       /// and component that has experienced a change.
       public: void EachPeriodicChange(const std::function<
                   void(const Entity &_entity,
-                       components::BaseComponent *_component)> _f) const; 
+                       const ComponentTypeId &_type)> _f) const; 
 
       /// \brief Set the absolute state of the ECM from a serialized message.
       /// Entities / components that are in the new state but not in the old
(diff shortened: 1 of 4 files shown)


src/EntityComponentManager.cc
```diff
@@ -963,6 +963,27 @@ std::unordered_set<ComponentTypeId>
...
+    
+    for (const auto entity: entities) {
+
+      const components::BaseComponent* implementation =  
+        this->ComponentImplementation(entity, componentType);
+      
+      if(!implementation)
+        continue;
+      
+      _f(entity, const_cast<components::BaseComponent*>(implementation));
```
azeey, line 982: Do we need the `const_cast`?  Why not change callback the signature to `const components::BaseComponent*`?

src/systems/scene_broadcaster/SceneBroadcaster.cc
```diff
@@ -257,8 +257,54 @@ class ignition::gazebo::systems::SceneBroadcasterPrivate
   /// \brief Flag used to indicate if periodic changes need to be published
   /// This is currently only used in playback mode.
   public: bool pubPeriodicChanges{false};
+
+  /// \brief Stores a cache of components that are changed. (This prevents dropping of
+  /// periodic change components which may not be frequent enough)
+  public: std::unordered_map<Entity,
+    std::unordered_map<ComponentTypeId,
+    std::unique_ptr<components::BaseComponent>>> changedComponents;
```
azeey, line 265: I see we're storing clones of the component. But this will incur unnecessary clones since if there are multiple changes to the component between SceneBroadcaster publications, we'd be replacing the cloned component with new clones multiple times. I think it would be more efficient if we just  stored the `ComponentTypeId` and get the actual component based on the type id when we're ready to publish the state message.

arjo129: Thanks for pointing this out I've found a cleaner way to do this in [467048c](https://github.com/gazebosim/gz-sim/pull/2010/commits/467048c5a309ae6e1fbf3e3ab27e51d225ca8145).

src/EntityComponentManager.cc
arjo129: Thanks for pointing this out I've found a cleaner way to do this in [467048c](https://github.com/gazebosim/gz-sim/pull/2010/commits/467048c5a309ae6e1fbf3e3ab27e51d225ca8145)

arjo129:
:warning: Do not merge yet. Scene broadcaster test is failing. Will investigate why.

jrutgeer:
@arjo129 

> Do not merge yet. Scene broadcaster test is failing. Will investigate why.

I did not check the test, but other than that it seems to work: I can publish position command messages and the pose is updated accordingly.

arjo129:
Test failures are unrelated. This looks like its good for a round of review.

include/ignition/gazebo/EntityComponentManager.hh
```diff
@@ -601,6 +601,16 @@ namespace ignition
       public: std::unordered_set<ComponentTypeId>
           ComponentTypesWithPeriodicChanges() const;
 
+      /// \brief Get all components with periodic changes.
+      /// \param[in] _changes A list of components with the latest periodic
+      /// changes. If a component has a periodic change, it is added to the
+      /// hash map. It the component or entity was removed, it is removed from
+      /// the hashmap. This way the hashmap stores a list of components and
+      /// entities which have had periodic changes in the past and still
+      /// exit within the ECM.
```
mjcarroll, line 610:
```suggestion
      /// exist within the ECM.
```

src/systems/scene_broadcaster/SceneBroadcaster.cc
```diff
@@ -384,14 +395,8 @@ void SceneBroadcaster::PostUpdate(const UpdateInfo &_info,
...
 
-      if (_manager.HasPeriodicComponentChanges())
-      {
-        auto periodicComponents = _manager.ComponentTypesWithPeriodicChanges();
-        _manager.State(*this->dataPtr->stepMsg.mutable_state(),
-             {}, periodicComponents);
-        this->dataPtr->pubPeriodicChanges = false;
-      }
-      else
+      // TODO(arjo): We may not need this.
```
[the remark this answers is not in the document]
arjo129: I believe this was a work around for when `PeriodicChange` is not published with every frame in the case of log files. With the new change tracker, I don't see why we would need to do this. I am however not sure and haven;t actually tested with a recorded playback. Perhaps I will if I get time.

Issue #306, opened 2020-08-21
Mouse interaction with simulated models


diegoferigo:
Other widely used simulators like mujoco and pybullet allow to interact in a dynamic sense (as opposed as pure kinematics) with models part of the simulation. This is something that also Gazebo Classic never had. The most that could be done was selecting links and applying wrenches (I though there was a `closethegap` ticket for this but maybe I was wrong).

You can see a nice mujoco demo here: https://www.youtube.com/watch?v=0ORsj_E17B0

chapulina:
> The most that could be done was selecting links and applying wrenches (I though there was a closethegap ticket for this but maybe I was wrong).

That's Gazebo classic's Apply Force and Torque tool, it's true that we're missing a close-the-gap issue for it

https://classic.gazebosim.org/tutorials?tut=apply_force_torque&cat=tools_utilities

The dialog approach allows more precision, while the mouse dragging interaction is more intuitive. I think both are great tools to add.

scpeters:
for reference, bullet's example use a ball joint (aka point-to-point constraint) between the mouse cursor and a point on the body where the user clicked:

* https://github.com/bulletphysics/bullet3/blob/39b8de74df93721add193e5b3d9ebee579faebf8/examples/CommonInterfaces/CommonRigidBodyBase.h#L231-L247
* https://github.com/bulletphysics/bullet3/blob/39b8de74df93721add193e5b3d9ebee579faebf8/examples/CommonInterfaces/CommonRigidBodyBase.h#L313-L376

PR #2014, opened 2023-06-14
Apply Force and Torque GUI plugin

Henrique-BO:
# 🎉 New feature

Partially addresses #306 

## Summary

Adds a new ApplyForceTorque GUI plugin, essentially implementing the [Apply Force and Torque](https://classic.gazebosim.org/tutorials?tut=apply_force_torque&cat=tools_utilities) interface that we have in Gazebo Classic.

This allows users to apply a specified force and torque to a chosen link in a model upon pressing a button. The implementation takes advantage of the existing ApplyLinkWrench system, sending the wrenches to the `/world/<world_name>/wrench` topic. The ApplyLinkWrench system is automatically loaded once this plugin is loaded.

However, a current limitation is that applying the force with an offset from the link origin is not supported, since it is not implemented in ApplyLinkWrench.

## Test it

Open any test world.

```
gz sim src/gz-sim/examples/worlds/apply_link_wrench.sdf
```

Load the Apply Force Torque plugin from the plugin dropdown. 

On the interface, input the desired force and torque. The model is selected from the scene, and the specific link is specified through a dropdown on the interface. Buttons are available to apply either just a force, just a torque or both at the same time.

ezgif com-video-to-gif

## Checklist
- [ ] Added tests


src/gui/plugins/apply_force_torque/ApplyForceTorque.qml
```diff
@@ -0,0 +1,300 @@
...
+    maximumValue: 1000000
+    minimumValue: -1000000
+    value: 0
+    decimals: 2
+    stepSize: 1.0
+    onValueChanged: ApplyForceTorque.UpdateForce(
+      forceX.value, forceY.value, forceZ.value)
+  }
+
+  Label {
```
azeey, line 105: I think it's best to leave out the offset portion until we actually support it.

src/gui/plugins/apply_force_torque/ApplyForceTorque.qml
```diff
@@ -0,0 +1,300 @@
...
+    id: forceXText
+    color: "dimgrey"
+    text: qsTr("X (N)")
+  }
+
+  GzSpinBox {
+    Layout.columnSpan: 2
+    Layout.fillWidth: true
+    id: forceX
+    maximumValue: 1000000
```
azeey, line 96: Can we make these constants, so we can change them easily if needed?

src/gui/plugins/apply_force_torque/ApplyForceTorque.qml
```diff
@@ -0,0 +1,300 @@
...
+    id: linkCombo
+    Layout.fillWidth: true
+    // enabled: ApplyForceTorque.configured
+    model: ApplyForceTorque.linkNameList
+    currentIndex: ApplyForceTorque.linkIndex
+    onCurrentIndexChanged: {
+      ApplyForceTorque.linkIndex = currentIndex
+    }
+    // ToolTip.visible: hovered
+    // ToolTip.text: qsTr("Link to which the wrench should be applied")
```
azeey, line 65: Remove?

src/gui/plugins/apply_force_torque/ApplyForceTorque.qml
```diff
@@ -0,0 +1,300 @@
...
+    id: linkText
+    color: "dimgrey"
+    text: qsTr("Link:")
+  }
+
+  ComboBox {
+    Layout.columnSpan: 6
+    id: linkCombo
+    Layout.fillWidth: true
+    // enabled: ApplyForceTorque.configured
```
azeey, line 58: Remove?

src/gui/plugins/apply_force_torque/ApplyForceTorque.qml
```diff
@@ -0,0 +1,300 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+*/
+import QtQuick 2.9
+import QtQuick.Controls 2.1
+import QtQuick.Dialogs 1.0
+import QtQuick.Controls.Material 2.1
+import QtQuick.Layouts 1.3
+import "qrc:/qml"
```
azeey, line 22: Is this needed?

src/gui/plugins/apply_force_torque/ApplyForceTorque.cc
```diff
@@ -0,0 +1,336 @@
...
+        return;
+      }
+
+      this->dataPtr->modelName = QString::fromStdString(
+        selectedModel.Name(_ecm));
+
+      // Put all of the model's links into the list
+      auto links = selectedModel.Links(_ecm);
+      unsigned int i{0};
+      while (i < selectedModel.LinkCount(_ecm))
```
azeey, line 173: nitpick:
```suggestion
      while (i < links.size())
```
since we're actually indexing into `links`.

src/gui/plugins/apply_force_torque/ApplyForceTorque.cc
```diff
@@ -0,0 +1,336 @@
...
+                }
+              }
+              gzerr << "World not found" << std::endl;
+              return false;
+            });
+        auto topic = transport::TopicUtils::AsValidTopic(
+          "/world/" + worldName + "/wrench");
+        this->dataPtr->pub =
+            this->dataPtr->node.Advertise<msgs::EntityWrench>(topic);
+        gzdbg << "Created publisher to " << topic << std::endl;
```
azeey, line 211: Creation of the publisher should be done in `LoadConfig` and you can get the worldName using `gui::worldNames` similar to https://github.com/gazebosim/gz-sim/blob/064f69ed7c441b90abbf0f84eeb0bfaeec628da9/src/gui/plugins/spawn/Spawn.cc#L186-L188

src/gui/plugins/apply_force_torque/ApplyForceTorque.cc
```diff
@@ -0,0 +1,336 @@
...
+
+/////////////////////////////////////////////////
+bool ApplyForceTorque::eventFilter(QObject *_obj, QEvent *_event)
+{
+  if (_event->type() ==
+    gz::sim::gui::events::EntitiesSelected::kType)
+  {
+    gz::sim::gui::events::EntitiesSelected *_e =
+        static_cast<gz::sim::gui::events::EntitiesSelected*>(_event);
+    this->dataPtr->selectedEntities = _e->Data();
```
azeey, line 118: If we only need one selected entity, can we just store that instead of the vector? If you need to know whether the selected entity has been set or not, you can use `std::optional`.

src/gui/plugins/apply_force_torque/ApplyForceTorque.cc
```diff
@@ -0,0 +1,336 @@
...
+        auto topic = transport::TopicUtils::AsValidTopic(
+          "/world/" + worldName + "/wrench");
+        this->dataPtr->pub =
+            this->dataPtr->node.Advertise<msgs::EntityWrench>(topic);
+        gzdbg << "Created publisher to " << topic << std::endl;
+      }
+    }
+  }
+
+  this->ModelNameChanged();
```
azeey, line 216: Do you need `emit` for these?

src/gui/plugins/apply_force_torque/ApplyForceTorque.qml
```diff
@@ -0,0 +1,300 @@
...
+
+  Label {
+    Layout.columnSpan: 2
+    horizontalAlignment: Text.AlignRight
+    id: forceXText
+    color: "dimgrey"
+    text: qsTr("X (N)")
+  }
+
+  GzSpinBox {
```
azeey, line 92: Can we have a message here that the force would be applied at the link origin, and not the COM?

azeey (commented):
Looks great!! I haven't tested it yet, but wanted to give it a first pass.

src/gui/plugins/apply_force_torque/ApplyForceTorque.cc
```diff
@@ -0,0 +1,336 @@
...
+    public: void PublishWrench(bool _applyForce, bool _applyTorque);
+
+    /// \brief Transport node
+    public: transport::Node node;
+
+    /// \brief Publisher for EntityWrench messages
+    public: transport::Node::Publisher pub;
+
+    /// \brief To synchronize member access
+    public: std::mutex mutex;
```
azeey, line 54: Since everything here is happening from a single Qt thread, I don't think you need this mutex.

src/gui/plugins/apply_force_torque/ApplyForceTorque.qml
```diff
@@ -0,0 +1,300 @@
...
+
+  Button {
+    text: qsTr("Apply All")
+    Layout.columnSpan: 8
+    Layout.fillWidth: true
+    onClicked: function() {
+      ApplyForceTorque.ApplyAll()
+    }
+  }
+}
```
arjo129, line 300: Knit: New line at end of file.

src/gui/plugins/apply_force_torque/ApplyForceTorque.hh
```diff
@@ -0,0 +1,132 @@
...
+namespace gz
+{
+namespace sim
+{
+  class ApplyForceTorquePrivate;
+
+  /// \brief Publish wrench to "/world/apply_link_wrench/wrench" topic.
+  ///
+  /// ## Configuration
+  /// This plugin doesn't accept any custom configuration.
```
arjo129, line 34: Knit: Might be worth mentioning which plugin provides the `/world/apply_link_wrench/wrench` topic.

arjo129 (commented):
Minor documentation related feedback. Which plugin is needed in order to apply the force?

src/gui/plugins/apply_force_torque/ApplyForceTorque.qml
Henrique-BO: Should we remove it entirely from the interface for now or just block the user from inputing values (like with a `enabled: false`)?

src/gui/plugins/apply_force_torque/ApplyForceTorque.cc
```diff
@@ -0,0 +1,336 @@
...
+                if (model == entity)
+                {
+                  worldName = _name->Data();
+                  return true;
+                }
+              }
+              gzerr << "World not found" << std::endl;
+              return false;
+            });
+        auto topic = transport::TopicUtils::AsValidTopic(
```
ahcorde, line 207: check if topic is valid. it should return an empty string if the function fails

src/gui/plugins/apply_force_torque/ApplyForceTorque.cc
```diff
@@ -0,0 +1,336 @@
...
+#include <gz/sim/Model.hh>
+#include <gz/sim/World.hh>
+#include <gz/sim/EntityComponentManager.hh>
+#include <gz/sim/components/World.hh>
+#include <gz/sim/components/Name.hh>
+#include <gz/msgs/Utility.hh>
+#include <gz/msgs/entity_wrench.pb.h>
+#include <gz/plugin/Register.hh>
+#include <gz/transport/Node.hh>
+
```
ahcorde, line 35: include `gz/mah/Vector3.hh`,

ahcorde (changes requested):

quarkytale (commented):
Tried it out, works as expected

Henrique-BO:
Thanks for the feedback. I addressed the requested changes in 6fbbfc0.

quarkytale:

src/gui/plugins/apply_force_torque/ApplyForceTorque.cc
```diff
@@ -0,0 +1,297 @@
...
+    }
+    this->dataPtr->pub =
+      this->dataPtr->node.Advertise<msgs::EntityWrench>(topic);
+    gzdbg << "Created publisher to " << topic << std::endl;
+  }
+
+  gz::gui::App()->findChild<gz::gui::MainWindow *>
+    ()->installEventFilter(this);
+  gz::gui::App()->findChild<gz::gui::MainWindow *>
+    ()->QuickWindow()->installEventFilter(this);
```
azeey, line 117: I'm not sure if we need to install the event filter on QuickWindow.

src/gui/plugins/apply_force_torque/ApplyForceTorque.cc
azeey: alas, I was wrong about this. The `EntitiesSelected` and `DeselectAllEntities` can come from the main Qt thread (via the `EntityTree` plugin) as well as the rendering thread (via the `SelectEntities` plugin), so a mutex is necessary. Sorry about that :(

src/gui/plugins/apply_force_torque/ApplyForceTorque.qml
azeey: Can you address this?

src/gui/plugins/apply_force_torque/ApplyForceTorque.qml
```diff
@@ -0,0 +1,243 @@
...
+      ApplyForceTorque.linkIndex = currentIndex
+    }
+  }
+
+  // Force
+  Text {
+    Layout.columnSpan: 8
+    id: forceText
+    color: "dimgrey"
+    text: qsTr("Force (applied to link origin):")
```
azeey, line 80: I checked how Gazebo-classic works and there, the application point is the center of mass. Would it be possible to do the same here?

src/gui/plugins/apply_force_torque/ApplyForceTorque.cc
```diff
@@ -0,0 +1,297 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+*/
+
+#include <string>
+
+#include <gz/gui/Application.hh>
```
ahcorde, line 20: alphabetize headers

src/gui/plugins/apply_force_torque/ApplyForceTorque.cc
```diff
@@ -0,0 +1,297 @@
...
+
+/////////////////////////////////////////////////
+void ApplyForceTorque::Update(const UpdateInfo &/*_info*/,
+  EntityComponentManager &_ecm)
+{
+  if (this->dataPtr->changedEntity)
+  {
+    this->dataPtr->changedEntity = false;
+
+    this->dataPtr->modelName = "";
```
ahcorde, line 149:
```suggestion
    this->dataPtr->modelName{""};
```

src/gui/plugins/apply_force_torque/ApplyForceTorque.hh
```diff
@@ -0,0 +1,127 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+#ifndef GZ_GUI_APPLYFORCETORQUE_HH_
+#define GZ_GUI_APPLYFORCETORQUE_HH_
+
+#include <memory>
+
```
ahcorde, line 22: include this headers `QString`, `QStringList`. `Qobject`, `QEvent`, 

`#include <ignition/gazebo/EntityComponentManager.hh>`

ahcorde (changes requested):

src/gui/plugins/apply_force_torque/ApplyForceTorque.cc
Henrique-BO: I don't think this works, since this is an assignment, not an initialization. I get a compilation error if I try this.

src/gui/plugins/apply_force_torque/ApplyForceTorque.qml
Henrique-BO: The current version has only the necessary imports

Henrique-BO:
Changes addressed in ca1c904.

src/gui/plugins/apply_force_torque/ApplyForceTorque.cc
```diff
@@ -0,0 +1,321 @@
...
+  if (!this->selectedEntity.has_value())
+  {
+    gzdbg << "No link selected" << std::endl;
+    return;
+  }
+
+  math::Vector3d forceToApply =
+    _applyForce ? this->force : math::Vector3d::Zero;
+  math::Vector3d torqueToApply =
+    _applyTorque ? this->torque : math::Vector3d::Zero +
```
azeey, line 300: Parenthesize the ternary expression. Otherwise, I'm not sure if the cross product term is being applied in both conditions.

azeey (commented):
I checked gazebo-classic once more and it looks like the force/torque is applied in the body fixed frame. I don't want to block this PR on that. Maybe we can add that later, but I think we should indicate that the force/torque is applied in the world frame in the GUI.

Henrique-BO:
You're right, I hadn't noticed that. I changed the application to the body fixed frame and added an indication to the GUI, so it's more consistent with Gazebo-Classic.

src/gui/plugins/apply_force_torque/ApplyForceTorque.cc
```diff
@@ -0,0 +1,318 @@
...
+  msgs::EntityWrench msg;
+  msg.mutable_entity()->set_id(*this->selectedEntity);
+  msgs::Set(msg.mutable_wrench()->mutable_force(), forceToApply);
+  msgs::Set(msg.mutable_wrench()->mutable_torque(), torqueToApply);
+
+  this->pub.Publish(msg);
+}
+
+// Register this plugin
+GZ_ADD_PLUGIN(ApplyForceTorque, gz::gui::Plugin);
```
azeey, line 318: The `;` is causing a warning.
```suggestion
GZ_ADD_PLUGIN(ApplyForceTorque, gz::gui::Plugin)
```

Henrique-BO: Done in 3911d73

azeey (approved):

ahcorde (approved):

PR #2026, opened 2023-06-30
Add force offset support to ApplyLinkWrench system and to Link API

Henrique-BO:
# 🎉 New feature

- Builds on top of #1593

## Summary
Adds support for specifying the application point of a force in the ApplyLinkWrench system. This is done through an offset expressed in the link frame (thus, relative to the link origin).

This uses the `force_offset` field that was already present in the Wrench message, but wasn't being used in the system. Also, an overloaded method `AddWorldWrench` was added to the Link interface. 

## Test it

Open a test world that has the ApplyLinkWrench system:

```
gz sim src/gz-sim/examples/worlds/apply_link_wrench.sdf
```

Send a message to the `/world/<world_name>/wrench` topic:

```
gz topic -t "/world/apply_link_wrench/wrench" -m gz.msgs.EntityWrench  -p "entity: {name: 'cylinder::link', type: LINK}, wrench: {force: {z: 100}, force_offset: {x: 0, y: 100, z: 0}}"
```

Screencast from 30-06-2023 18_08_34

## Checklist
- [x] Added tests


ahcorde (changes requested):
can you add a test to this new method?

Henrique-BO:
Added a test in 34101f7. Since there already was a `LinkAddWorldForce` test doing very similar checks and setup, I just added a few lines to the existing function. Should we keep it like this (and so maybe rename the test) or is it better to have a separate test?

azeey:

quarkytale (approved):
Works for me

include/gz/sim/Link.hh
```diff
@@ -306,6 +306,19 @@ namespace gz
...
 
+      /// \brief Add a wrench expressed in world coordinates and applied to
+      /// the link at the given offset from its center of mass. This wrench
+      /// is applied for one simulation step.
+      /// \param[in] _ecm Mutable Entity-component manager.
+      /// \param[in] _force Force to be applied expressed in world coordinates
+      /// \param[in] _offset The point of application of the force expressed
+      /// in the link-fixed frame and relative to the center of mass.
+      /// \param[in] _torque Torque to be applied expressed in world coordinates
+      public: void AddWorldWrench(EntityComponentManager &_ecm,
```
azeey, line 317: The `AddWorldWrench` above applies the force at the link origin. In this overload, the offset is relative to the COM.  It might be a good idea to change the name to reflect that. Maybe `AddWorldWrenchRelativeToCOM`? /cc @scpeters.

Henrique-BO: Renamed the function in 361cd4f

include/gz/sim/Link.hh
```diff
@@ -306,6 +306,20 @@ namespace gz
                                  const math::Vector3d &_force,
                                  const math::Vector3d &_torque) const;
 
+      /// \brief Add a wrench expressed in world coordinates and applied to
+      /// the link at the given offset from its center of mass. This wrench
+      /// is applied for one simulation step.
+      /// \param[in] _ecm Mutable Entity-component manager.
+      /// \param[in] _force Force to be applied expressed in world coordinates
+      /// \param[in] _offset The point of application of the force expressed
```
azeey, line 314: I'm not trying to be nitpicky, but the frame of expression for these quantities has always been a point of confusion, so I want to be a little more careful. I see that the offset is relative to the COM, but it's expressed in the link frame. Wouldn't it be more consistent if it was expressed in the COM frame? i.e, if you have the following SDF
```xml
<model name="test">
  <link name="base">
    <pose>1 0 0   1.57 0 0</pose>
    <inertial>
      <pose>0 1 0    0 1.57 0</pose>
      ...
      </inertial>
   </link>
</model>

With the current function, we'd be ignoring the pose of the `<inertial>` tag.

Comment (Henrique-BO) [reply include/gz/sim/Link.hh:314]: I agree, expressing the offset in the COM frame is more intuitive and consistent. Changed it in 336b119.

File: src/systems/apply_link_wrench/ApplyLinkWrench.cc [hunk 3]
```diff
@@ -80,16 +80,17 @@ class gz::sim::systems::ApplyLinkWrenchPrivate
 /// \param[in] _msg Entity message. If it's a link, that link is returned. If
 /// it's a model, its canonical link is returned.
 /// \param[out] Force to apply.
+/// \param[out] Offset of the force application point relative to the
```
azeey, line 83: This implies that the offset in `msgs::EntityWrench` and by association, in the `component::ExternalWorldWrench`. is relative to the COM, but the point of application is the link origin. Am I right? If so, I think we should avoid having two different reference points. If the point of application is the link origin, the offset should also be from the link origin.

Henrique-BO: Changed in ee693bc so that the Link API and the ApplyLinkWrench system interpret the offset in the link frame. This way, the semantics of the offset are consistent between them, the `msgs::EntityWrench` message and the `component::ExternalWorldWrench` component.

include/gz/sim/Link.hh
```diff
@@ -306,6 +306,19 @@ namespace gz
...
+      /// the link at an offset from the link's origin. This wrench
+      /// is applied for one simulation step.
+      /// \param[in] _ecm Mutable Entity-component manager.
+      /// \param[in] _force Force to be applied expressed in world coordinates
+      /// \param[in] _offset The point of application of the force expressed
+      /// in the link frame
+      /// \param[in] _torque Torque to be applied expressed in world coordinates
+      public: void AddWorldWrench(EntityComponentManager &_ecm,
+                                const math::Vector3d &_force,
+                                const math::Vector3d &_offset,
```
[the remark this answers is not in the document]
Henrique-BO: Done in f36ac69

azeey (approved):

azeey:
The windows node was killed after building for 3 hours (https://build.osrfoundation.org/job/ign_gazebo-pr-win/5927/). Going to merge without it. /cc @j-rivero.

PR #2030, opened 2023-07-11
Fix TransformControl's BlockOrbit behavior

Henrique-BO:
# 🦟 Bug fix

## Summary

Currently, the TransformControl plugin sends a `BlockOrbit` event on every rendering call, even if it does not wish to block the camera orbit. Because of this, if another plugin sends a `BlockOrbit` event, it might get overwritten by TransformControl's and not have the desired effect.

This PR fixes this by making TransformControl send `BlockOrbit` `false` events only once (to unblock after a transform), allowing other plugins to block the camera while TransformControl is idle. `true` events are unchanged, and are still sent on every rendering call.

## Checklist
- [ ] Added tests

### DIFF
--- a/src/gui/plugins/transform_control/TransformControl.cc
+++ b/src/gui/plugins/transform_control/TransformControl.cc
@@ -173,6 +173,9 @@ namespace gz::sim
 
     /// \brief Block orbit
     public: bool blockOrbit = false;
+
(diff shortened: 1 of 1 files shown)


azeey (approved):
Great solution!

PR #2042, opened 2023-07-19
Adds bindings for the Sensor convenience class

Voldivh:
# 🎉 New feature

## Summary
This PR adds bindings for the `Sensor` convenience class.

## Checklist
- [ ] Added tests


python/src/gz/sim/Sensor.cc
```diff
@@ -0,0 +1,62 @@
...
+  py::class_<gz::sim::Sensor>(module, "Sensor")
+  .def(py::init<gz::sim::Entity>())
+  .def("entity", &gz::sim::Sensor::Entity,
+      "Get the entity which this sensor is related to.")
+  .def("reset_entity", &gz::sim::Sensor::ResetEntity,
+      "Reset Entity to a new one.")
+  .def("valid", &gz::sim::Sensor::Valid,
+      py::arg("ecm"),
+      "Check whether this sensor correctly refers to an entity that"
+      "has a components::Sensor.")
```
ahcorde, line 44:
```suggestion
      "Check whether this sensor correctly refers to an entity that "
      "has a components::Sensor.")
```

python/src/gz/sim/Sensor.cc
```diff
@@ -0,0 +1,62 @@
...
+  .def("name", &gz::sim::Sensor::Name,
+      py::arg("ecm"),
+      "Get the sensor's unscoped name.")
+  .def("pose", &gz::sim::Sensor::Pose,
+      py::arg("ecm"),
+      "Get the pose of the sensor."
+      "If the sensor has a trajectory, this will only return the origin"
+      "pose of the trajectory and not the actual world pose of the sensor.")
```
ahcorde, line 52:
```suggestion
      "Get the pose of the sensor. "
      "If the sensor has a trajectory, this will only return the origin "
      "pose of the trajectory and not the actual world pose of the sensor.")
```

python/src/gz/sim/Sensor.hh
```diff
@@ -0,0 +1,40 @@
+/*
+ * Copyright (C) 2021 Open Source Robotics Foundation
```
ahcorde, line 2:
```suggestion
 * Copyright (C) 2023 Open Source Robotics Foundation
```

ahcorde (changes requested):
tests ?

python/src/gz/sim/Sensor.cc
Voldivh: Done.

python/src/gz/sim/Sensor.hh
Voldivh: Done.

python/src/gz/sim/Sensor.cc
```diff
@@ -0,0 +1,73 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+
+#include <pybind11/pybind11.h>
+#include <pybind11/stl.h>
+
+#include <iostream>
```
ahcorde, line 21: where is this used? remove ?

python/src/gz/sim/Sensor.hh
ahcorde: I still see 2021

ahcorde (changes requested):

python/src/gz/sim/Sensor.hh
Voldivh: Weird I might have changed in other place and forgot about this one. It should be done now.

python/src/gz/sim/Sensor.cc
Voldivh: Thanks and done.

python/test/sensor_TEST.py
```diff
@@ -0,0 +1,79 @@
...
+            self.assertNotEqual(K_NULL_ENTITY, sensor.entity())
+            # Valid Test
+            self.assertTrue(sensor.valid(_ecm))
+            # Name Test
+            self.assertEqual('sensor_test', sensor.name(_ecm))
+            # Pose Test
+            self.assertEqual(Pose3d(0, 1, 0, 0, 0, 0), sensor.pose(_ecm))
+            # Topic Test
+            # This is None because the entity does not have a
+            # components::SensorTopic component.
```
azeey, line 57: Do we know why this is? If it's because this is a Joint sensor, can you create an issue so we can fix it later?

azeey (approved):
LGTM! Just one comment about `SensorTopic`.

python/test/sensor_TEST.py
Voldivh: After a little lookup, I'm not completely sure why this is happening, it could be due to the reason you just commented but I'm not 100% sure of it. How should we proceed on this case?

azeey: Looks like they're created by sensor systems (https://github.com/gazebosim/gz-sim/blob/87e87bd6676538bb963ccd793c89a1b25cffa270/src/systems/force_torque/ForceTorque.cc#L149), so we'd need the `ForceTorque` sensor in this case. 

Putting
```
    <plugin filename="gz-sim-physics-system" name="gz::sim::systems::Physics" />
    <plugin filename="gz-sim-forcetorque-system" name="gz::sim::systems::ForceTorque" />
```
to the .sdf file should fix this.

Voldivh: Ok, I'll try it out

PR #2047, opened 2023-07-25
Adds a warning if the `Server` method of a `TestFixture` is called before `Finalize`

arjo129:
## Summary
This PR adds a warning if someone were to call `Server()` before `Finalize`. The rationale for this is because I made the mistake of not calling `Finalize` ina program I was writing. It took me some time to track down the root cause of the problem. I figure if we warn users it should be immediately obvious.

## Checklist
- [ ] Added tests

### DIFF
--- a/src/TestFixture.cc
+++ b/src/TestFixture.cc
@@ -210,6 +210,13 @@ TestFixture &TestFixture::OnPostUpdate(std::function<void(
 //////////////////////////////////////////////////
 std::shared_ptr<Server> TestFixture::Server() const
 {
+  if (!this->dataPtr->finalized)
+  {
+    ignwarn << "Fixture has not been finalized, any functions you attempted"
+      << "to hook into will not be run. It is recommended to call Finalize()"
+      << "before accessing the server."
+      << std::endl;
+  }
   return this->dataPtr->server;
 }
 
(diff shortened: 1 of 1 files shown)


src/TestFixture.cc
```diff
@@ -210,6 +210,13 @@ TestFixture &TestFixture::OnPostUpdate(std::function<void(
 //////////////////////////////////////////////////
 std::shared_ptr<Server> TestFixture::Server() const
 {
+  if (!this->dataPtr->finalized)
+  {
+    ignwarn << "Fixture has not been finalized, any functions you attempted"
```
azeey, line 215: Can we use `gzwarn` instead to make forward porting easier?

arjo129: I don't think `gzwarn` is available in citadel. It might be worth considering back porting the functionality to reduce the maintenance overhead.

mjcarroll: It's not.  If we go ahead and do the forward port while it's "fresh" we can minimize some pain

azeey: Ah, sorry. For some reason I thought they were. @mjcarroll would there be an issue if we added those to gz-common3 at some point?

PR #2052, opened 2023-07-26
Use sdf::Element::FindElement instead of GetElement in ApplyLinkWrench

azeey:
# 🦟 Bug fix

Fixes #<NUMBER>

## Summary
I noticed the error
```
Error [Utils.cc:174] Missing element description for [persistent]
```
when testing #2014.  /cc @Henrique-BO 

## Checklist
- [ ] Added tests

### DIFF
--- a/src/systems/apply_link_wrench/ApplyLinkWrench.cc
+++ b/src/systems/apply_link_wrench/ApplyLinkWrench.cc
@@ -148,8 +148,7 @@ void ApplyLinkWrench::Configure(const Entity &_entity,
   this->dataPtr->verbose = _sdf->Get<bool>("verbose", true).first;
 
   // Initial wrenches
-  auto ptr = const_cast<sdf::Element *>(_sdf.get());
-  for (auto elem = ptr->GetElement("persistent");
+  for (auto elem = _sdf->FindElement("persistent");
        elem != nullptr;
        elem = elem->GetNextElement("persistent"))
   {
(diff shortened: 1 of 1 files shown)


src/systems/apply_link_wrench/ApplyLinkWrench.cc
```diff
@@ -163,7 +162,7 @@ void ApplyLinkWrench::Configure(const Entity &_entity,
 
     msg.mutable_entity()->set_name(elem->Get<std::string>("entity_name"));
 
-    auto typeStr = elem->GetElement("entity_type")->Get<std::string>();
+    auto typeStr = elem->FindElement("entity_type")->Get<std::string>();
```
iche033, line 165: Check the element returned since `FindElement` may return `nullptr`?

src/systems/apply_link_wrench/ApplyLinkWrench.cc
```diff
@@ -182,12 +181,12 @@ void ApplyLinkWrench::Configure(const Entity &_entity,
     if (elem->HasElement("force"))
     {
       msgs::Set(msg.mutable_wrench()->mutable_force(),
-          elem->GetElement("force")->Get<math::Vector3d>());
+          elem->FindElement("force")->Get<math::Vector3d>());
```
iche033, line 184: should be fine to use `GetElement` here since there is a `HasElement` check above to ensure that the element exists?

src/systems/apply_link_wrench/ApplyLinkWrench.cc
azeey: It's checked with `HasElement` above so it should be fine.

src/systems/apply_link_wrench/ApplyLinkWrench.cc
azeey: We should avoid using `GetElement` since it potentially creates the element. `FindElement` should be safe to use, especially after checking with `HasElement`. `GetElement` also requires `elem` be non-const, which is not the case here.

iche033: ah just realized that `FindElement` essentially just calls the same underlying `GetElementImp` function as `GetElement`

iche033 (approved):

PR #2069, opened 2023-08-03
Avoid nullptr dereference if TouchPlugin is not attached to a model entity.

jrutgeer:
# 🦟 Bug fix

## Summary

Two small fixes:

- Avoid nullptr dereference in case the plugin is not attached to a model (as `sdfConfig` is initialized only if `TouchPlugin::Configure()` is successful, see code snippet below),
- Avoid execution of `preUpdate()` code if the system was not initialized correctly.

I also changed the documentation somewhat to better explain what the plugin actually does.

</br>

https://github.com/gazebosim/gz-sim/blob/f2ff574664574100989e3633e4debe16431fe00f/src/systems/touch_plugin/TouchPlugin.cc#L358-L370

### DIFF
--- a/src/systems/touch_plugin/TouchPlugin.cc
+++ b/src/systems/touch_plugin/TouchPlugin.cc
@@ -103,7 +103,7 @@ class gz::sim::systems::TouchPluginPrivate
   public: bool initialized{false};
 
   /// \brief Set during Load to true if the configuration for the plugin is
-  /// valid and the post update can run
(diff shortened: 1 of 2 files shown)


azeey (approved):
Good catch. Thanks!

mjcarroll (approved):

azeey:
#2070 seems to have fixed the ABI failure.

PR #2081, opened 2023-08-17
Add support for writing systems in Python

### LINKED ISSUE #790: Scripting: plugins that aren't written in C++
## Desired behavior

In the interest of making Gazebo more approachable and quicker to use, we could offer a scripting interface that lets users write plugins in Python (and potentially other languages too, like Ruby and Javascript).

The user could load their Python script through the SDF's `<plugin>` tag, just like they do with C++ libraries right now.

## Alternatives considered

Users could use ROS to interact with Gazebo using various scripting languages. We could also just offer a scripting API to Ignition Transport, so users can interact with simulation asynchronously.

These 2 approaches limit what the user can do though, because they're constrained to the transport API and must run async.

## Implementation suggestion

First we'll need to expose C++ APIs that are commonly used by plugins to scripts, such as:

* The [EntityComponentManager](https://github.com/ignitionrobotics/ign-gazebo/blob/ign-gazebo5/include/ignition/gazebo/EntityComponentManager.hh)'s API so users can access and modify entities and components
* Component data types so that they can be read and manipulated. Components commonly include Math, SDFormat and Msgs types. 
* All the [System](https://github.com/ignitionrobotics/ign-gazebo/blob/ign-gazebo5/include/ignition/gazebo/System.hh) callbacks need to be inherited by the Python plugin so that they're called at the correct time during simulation.

We can either add C APIs to all of these and then call those C functions from scripts, or use SWIG to generate APIs in other languages from C++. 

Finally, we'll need to support loading those scripts from the `<plugin>` tag.

## Additional context

Related to https://github.com/ignitionrobotics/ign-gazebo/issues/789

azeey:
# 🎉 New feature

Closes #790
Replaces #2045, which was automatically closed by Github when the branch it was targeting  was deleted.

## Summary

Creates a C++ system that is able to load a python module and pass along the System interface calls, such as `Configure`, `PreUpdate`, etc. The convention is pretty simple. The python module exposes a `get_system` function which returns an instance of a class that implements the desired System interface functions. 

I'd appreciate feedback on the convention/general interface between `PythonSystemLoader` and python systems.

TODO:
- [x] Add tests

## Test it

Run `examples/worlds/python_system_loader.sdf`. See the file for instructions on environmental variables that need to be set.

## Checklist
- [ ] Added tests


examples/scripts/python_api/systems/test_system.py
```diff
@@ -0,0 +1,31 @@
+from gz.math7 import Vector3d
+from gz.sim8 import Model, Link
+import random
+
+
+class TestSystem(object):
+    def __init__(self):
+        self.id = random.randint(1, 100)
+
+    def configure(self, entity, sdf, ecm, event_mgr):
```
mjcarroll, line 10: I think we could probably leave out the event manager for now as it is generally not used much and used async when it is.  The `async` parts concern me a bit with the interaction with python.

src/systems/python_system_loader/PythonSystemLoader.cc
```diff
@@ -0,0 +1,198 @@
...
+
+  if (_method)
+  {
+    try
+    {
+      _method(std::forward<Args>(_args)...);
+    }
+    catch (const pybind11::error_already_set &_err)
+    {
+      gzerr << _err.what() << std::endl;
```
mjcarroll, line 162: Will the error message contain the name of the method and python module here?  It would certainly help with tracking down errors, otherwise they will all have the same name, correct?

src/systems/python_system_loader/PythonSystemLoader.cc
```diff
@@ -0,0 +1,198 @@
...
+
+//////////////////////////////////////////////////
+void PythonSystemLoader::Update(const UpdateInfo &_info,
+                                EntityComponentManager &_ecm)
+{
+  CallPythonMethod(this->updateMethod, _info, &_ecm);
+}
+
+//////////////////////////////////////////////////
+void PythonSystemLoader::PostUpdate(const UpdateInfo &_info,
```
mjcarroll, line 183: Did we determine what happens when there are multiple python modules?  Is that supported/unsupported?

Postupdate is the most interesting case because of parallel execution

examples/scripts/python_api/systems/test_system.py
azeey: My thinking is that if we ever create bindings for `SdfEntityCreater`, the `event_mgr` needs to be passed along. Currently, it's just an empty binding, so there's not much you can do with it in python other than passing it to some other C++ function. This is also the case for the ECM. 

If we don't include it now, adding it later will mean all existing scripts have to be updated.

src/systems/python_system_loader/PythonSystemLoader.cc
azeey: It should, but I'll double check.

src/systems/python_system_loader/PythonSystemLoader.cc
azeey: So, we're currently creating an instance of ` pybind11::scoped_interpreter` in `ServerPrivate` and it seems to work fine with multiple python modules, since there's only one `Server` instance. However, if we want to have multiple `Server`s in the same process, this will not work. So maybe we need a Singleton for this.

examples/scripts/python_api/systems/test_system.py
mjcarroll: Ah, okay, in that case lets keep it.

mjcarroll (approved):

src/systems/python_system_loader/PythonSystemLoader.cc
mjcarroll: I think we can generally assume that multiple Servers per process isn't likely at the moment. We may want to add a note somewhere in the event that someone attempts it, though.

mjcarroll:
Hmm, approval is more of approval-of-approach.  I think we still need a few tests here (which you have already noted)

caguero (approved):
It works for me, this is very cool. I just left very minor comments.

I think the `PythonSystemLoader` is a good option. I was thinking if it would make sense to promote this upstream somehow but I didn't find a better approach than this one.

src/systems/python_system_loader/PythonSystemLoader.cc
azeey: I threw an exception in the python file and this is what I got
```
[Err] [PythonSystemLoader.cc:174] NameError: name 'entity' is not defined

At:
  ~/ws/harmonic/src/gz-sim/python/test/plugins/test_model_system.py(51): pre_update

```
I think that's sufficient.

src/systems/python_system_loader/PythonSystemLoader.cc
```diff
@@ -0,0 +1,198 @@
...
+  if (_method)
+  {
+    try
+    {
+      _method(std::forward<Args>(_args)...);
+    }
+    catch (const pybind11::error_already_set &_err)
+    {
+      gzerr << _err.what() << std::endl;
+      return;
```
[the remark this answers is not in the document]
azeey: Done in 2b7469e67487b3b05e7559455a17f4886f41fabe

src/systems/python_system_loader/PythonSystemLoader.cc
azeey: Some of the tests were failing because they initialized two `Server` instances. I fixed it by using `pybind11::initialize_interpreter()` instead of `pybind11::scoped_interpreter` so we can skip initialization of the interpreter is already initialized.

src/systems/python_system_loader/PythonSystemLoader.hh
```diff
@@ -0,0 +1,96 @@
...
+  // Documentation inherited
+  public: void Reset(const UpdateInfo &_info,
+                     EntityComponentManager &_ecm) final;
+
+  /// \brief Function that calls each of the python equivalents of Configure,
+  /// PreUpdate, etc.
+  private: template <typename ...Args>
+  void CallPythonMethod(pybind11::object _method, Args&&...);
+
+  private: bool validConfig{false};
```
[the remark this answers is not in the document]
azeey: Done in 2b7469e67487b3b05e7559455a17f4886f41fabe

azeey (commented):
There were a couple of GIL issues that showed up when I started using gz-transport inside the python system. The fix requires releasing the GIL when we run `PostUpdate` threads and acquire the GIL from the threads each `PostUpdate` runs in. This also requires an update to gz-transport so that the GIL is acquired before calling subscription callbacks since that also happens from a different thread than the main thread that contains the Python interpreter.

test/integration/python_system_loader.cc
```diff
@@ -0,0 +1,109 @@
...
+namespace components = gz::sim::components;
+
+class PythonSystemLoaderTest : public InternalFixture<::testing::Test>
+{
+};
+
+/////////////////////////////////////////////////
+void worldReset()
+{
+  gz::msgs::WorldControl req;
```
ahcorde, line 39: include message: gz::msgs::WorldControl and gz::msgs::Boolean

test/integration/python_system_loader.cc
```diff
@@ -0,0 +1,109 @@
...
+}
+
+/////////////////////////////////////////////////
+TEST_F(PythonSystemLoaderTest, LoadMultipleSystems)
+{
+  common::setenv("GZ_SIM_SYSTEM_PLUGIN_PATH",
+                 common::joinPaths(std::string(PROJECT_SOURCE_PATH), "python",
+                                   "test", "plugins"));
+
+  sim::ServerConfig serverConfig;
```
ahcorde, line 61: include: ServerConfig, gz/common/Filesystem.hh, components::Name, components::Model, sim::worldPose

test/integration/python_system_loader.cc
```diff
@@ -0,0 +1,109 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+#include <gtest/gtest.h>
+
```
ahcorde, line 19: include optional, string, chrono

test/integration/python_system_loader.cc
azeey: Done in https://github.com/gazebosim/gz-sim/pull/2081/commits/f30128ff50d9453168ec88e99e3d1b41126a7bc4

test/integration/python_system_loader.cc
azeey: Done in https://github.com/gazebosim/gz-sim/pull/2081/commits/f30128ff50d9453168ec88e99e3d1b41126a7bc4

test/integration/python_system_loader.cc
azeey: Some of the headers were already there, added the rest in https://github.com/gazebosim/gz-sim/pull/2081/commits/f30128ff50d9453168ec88e99e3d1b41126a7bc4

src/systems/python_system_loader/PythonSystemLoader.cc
```diff
@@ -0,0 +1,217 @@
...
+  CallPythonMethod(this->postUpdateMethod, _info, &_ecm);
+}
+//////////////////////////////////////////////////
+void PythonSystemLoader::Reset(const UpdateInfo &_info,
+                               EntityComponentManager &_ecm)
+{
+  CallPythonMethod(this->resetMethod, _info, &_ecm);
+}
+
+GZ_ADD_PLUGIN(PythonSystemLoader, System, ISystemConfigure, ISystemPreUpdate,
```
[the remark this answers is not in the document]
azeey: Do we need to if it's already in "PythonSystemLoader.hh"? Added anyway https://github.com/gazebosim/gz-sim/pull/2081/commits/f30128ff50d9453168ec88e99e3d1b41126a7bc4

src/systems/python_system_loader/PythonSystemLoader.cc
```diff
@@ -0,0 +1,217 @@
...
+  {
+    if (py::hasattr(this->pythonSystem, "shutdown"))
+    {
+      this->pythonSystem.attr("shutdown")();
+    }
+  }
+}
+
+void PythonSystemLoader::Configure(
+    const Entity &_entity, const std::shared_ptr<const sdf::Element> &_sdf,
```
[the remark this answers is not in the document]
azeey: Some were already there, added the rest in https://github.com/gazebosim/gz-sim/pull/2081/commits/f30128ff50d9453168ec88e99e3d1b41126a7bc4

src/ServerPrivate.hh
```diff
@@ -22,6 +22,7 @@
 #include <gz/msgs/server_control.pb.h>
 #include <gz/msgs/stringmsg_v.pb.h>
 
+
```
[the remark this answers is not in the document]
azeey: Done in https://github.com/gazebosim/gz-sim/pull/2081/commits/f30128ff50d9453168ec88e99e3d1b41126a7bc4

ahcorde (approved):

tutorials/python_interfaces.md
```diff
@@ -81,3 +81,66 @@ iterations  1000
...
+
+
+def get_system():
+    return TestSystem()
+```
+
+The system can be added to SDFormat model or world with:
+
+```xml
+
```
ahcorde, line 136:
```suggestion
```

tutorials/python_interfaces.md
```diff
@@ -81,3 +81,66 @@ iterations  1000
...
+includes `GZ_SIM_SYSTEM_PLUGIN_PATH` as well as `PYTHONPATH`. The module is
+expected to provide a function called `get_system` that returns an instance of
+a class that implements the various interfaces in `gz::sim::System`.
+
+Example python system:
+
+<!-- TODO(azeey) Allow including python files in doxygen -->
+<!-- \include examples/scripts/python_api/systems/test_system.py -->
+```python
+
```
ahcorde, line 99:
```suggestion
```python
```

azeey: Done in https://github.com/gazebosim/gz-sim/pull/2081/commits/24c3d52966e4a1514fcc6d0c30e79daf1050e754

scpeters:
I opened #2899 since the behavior of some Link APIs was changed in https://github.com/gazebosim/gz-sim/pull/2081/commits/9bf542eb39cb931e1f74c9bf0963216a14123c14 without updating the API documentation

PR #2103, opened 2023-08-31
Port: 7 to main

mjcarroll:
# ➡️ Forward port

Port `gz-sim7` to `main`

Branch comparison: https://github.com/gazebosim/gz-sim/compare/main...gz-sim7


src/Server.cc
```diff
@@ -160,6 +160,14 @@ Server::Server(const ServerConfig &_config)
     return;
   }
 
+  if (this->dataPtr->sdfRoot.WorldCount() == 0)
```
azeey, line 163: Ah, this shouldn't have made it in gz-sim7 since it's now possible to load a model file.

mjcarroll: I will go ahead and add a PR to fix that so we don't lose it

mjcarroll: I cherry-picked it in here and squashed it down into the merge commit.

azeey (approved):

iche033 (approved):

PR #2104, opened 2023-08-31
Remove forward-ported restriction on model loading

mjcarroll:
This shouldn't have been ported to gz-sim7: https://github.com/gazebosim/gz-sim/pull/2103#discussion_r1311043266

### DIFF
--- a/src/Server.cc
+++ b/src/Server.cc
@@ -160,14 +160,6 @@ Server::Server(const ServerConfig &_config)
     return;
   }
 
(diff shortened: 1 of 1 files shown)


azeey (approved):

iche033 (approved):

mjcarroll:
Some new warnings about not using `gz::physics::getEngineInstallDir()`, but I think that is out of scope for here.

PR #2141, opened 2023-09-13
Fix another deadlock in sensors system

iche033:
# 🦟 Bug fix

## Summary

This PR:

* Fixed a situation deadlock that occurs when waiting on the `updateAvailable` variable with a condition variable while that variable is also being written.  
* ~~Reduced scope of a lock (`timeLock`). Not needed for fixing deadlock - can be done in a separate PR if needed.~~ 
    * Update: Reverted these changes.

More info on the deadlock:

These two operations can happen concurrently (from 2 threads):

```cpp
        std::unique_lock<std::mutex> cvLock(this->dataPtr->renderMutex);
        // updateAvailable is true here
        this->dataPtr->renderCv.wait(cvLock, [this] {
          return !this->dataPtr->running || !this->dataPtr->updateAvailable; });
```

```cpp
        // updateAvailable is set to false
        this->dataPtr->updateAvailable = true;
        this->dataPtr->renderCv.notify_one();
```

The `wait` condition may miss the `notify_one()` call causing it to sleep forever. To solve this, we need to make sure these two groups of operations are performed one after another. This fix is to group and lock the `updateAvailable = [true|false]` and `notify_one` calls using the same lock that's used by the `wait` condition variable.

## Test it

Ran into this deadlock when adding the DvL sensor system to the Harmonic demo world. The DVL sensor is implemented as a custom rendering sensor. It changes the timing of the sensors system which revealed this deadlock issue. 

Test with this pull request that adds a DVL sensor to the harmonic demo world:
https://github.com/gazebosim/harmonic_demo/pull/9

Without the changes in this PR, the deadlock happens within 1 minute of running the simulation for me.

## Checklist
- [ ] Added tests

### DIFF
--- a/src/systems/sensors/Sensors.cc
+++ b/src/systems/sensors/Sensors.cc
@@ -361,13 +361,16 @@ void SensorsPrivate::RunOnce()
 
     // safety check to see if reset occurred while we're rendering
     // avoid publishing outdated data if reset occurred
-    std::unique_lock<std::mutex> timeLock(this->renderMutex);
-    if (updateTimeApplied <= this->updateTime)
     {
-      // publish data
-      GZ_PROFILE("RunOnce");
-      this->sensorManager.RunOnce(updateTimeApplied);
-      this->eventManager->Emit<events::Render>();
+      std::unique_lock<std::mutex> timeLock(this->renderMutex);
+      if (updateTimeApplied <= this->updateTime)
+      {
+        timeLock.unlock();
+        // publish data
+        GZ_PROFILE("RunOnce");
+        this->sensorManager.RunOnce(updateTimeApplied);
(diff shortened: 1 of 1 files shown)


mjcarroll:
This causes `SensorsFixture/UpdateRate` to fail, which is a bit concerning.

iche033:
> This causes `SensorsFixture/UpdateRate` to fail, which is a bit concerning.

I ran CI again and confirmed that it failed again. I reverted the lock scope changes in [b8ba42f](https://github.com/gazebosim/gz-sim/pull/2141/commits/b8ba42f1126e3e55887562c641a8c454048c4e04) and that seems to fix it. I ran CI two more times and test passes.

mjcarroll (approved):

PR #2166, opened 2023-09-23
Server: explicitly skip auto-inertial calculation

scpeters:
# 🦟 Bug fix

Needed when https://github.com/gazebosim/sdformat/pull/1325 is merged

## Summary

Automatic inertial calculation may not work in `Root::Load`, so explicitly configure the loader to skip auto-inertial calculation. Computation of auto-inertial values are currently skipped by default, but the default behavior is proposed to be changed in https://github.com/gazebosim/sdformat/pull/1325, after which, this pull request would be required.

## Checklist
- [X] Signed all commits for DCO
- [ ] Added tests
- [ ] Updated documentation (as needed)
- [ ] Updated migration guide (as needed)
- [ ] Consider updating Python bindings (if the library has them)
- [ ] `codecheck` passed (See [contributing](https://gazebosim.org/docs/all/contributing#contributing-code))
- [ ] All tests passed (See [test coverage](https://gazebosim.org/docs/all/contributing#test-coverage))
- [ ] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Agazebosim+archived%3Afalse+) to support the maintainers

**Note to maintainers**: Remember to use **Squash-Merge** and edit the commit message to match the pull request summary while retaining `Signed-off-by` messages.

PR #2170, opened 2023-09-25
Don't merge: testing #2166 with ci_matching_branch

scpeters:
Testing #2166 against https://github.com/gazebosim/sdformat/pull/1325 using `ci_matching_branch/*` names and https://github.com/osrf/homebrew-simulation/commit/4d2d61f8ed6ab09f314f3a39e62adc956c834bd2

Server: explicitly skip auto-inertial calculation

Automatic inertial calculation may not work in
Root::Load, so explicitly configure the loader
to skip auto-inertial calculation.

**Do not merge**

### DIFF
--- a/src/Server.cc
+++ b/src/Server.cc
@@ -121,6 +121,8 @@ Server::Server(const ServerConfig &_config)
       }
       gzmsg <<  msg;
       sdf::ParserConfig sdfParserConfig = sdf::ParserConfig::GlobalConfig();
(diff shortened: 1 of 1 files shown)


scpeters:
this didn't actually use https://github.com/gazebosim/sdformat/pull/1325 because it used bottles instead of building from source; let me try again

scpeters:

PR #2182, opened 2023-09-29
Bugfix: Fixed Centre of Mass and Inertia Matrix Calculation Bug `MeshInertiaCalculator::CalculateMassProperties()` function

jasmeet0915:
# 🦟 Bug fix

Previous Understanding:
The `MeshInertiaCalculator.cc` computes the inertia w.r.t the mesh origin. This can lead to incorrect inertia values in case the Mesh Origin was not at the Center of Mass as mentioned [here](https://github.com/gazebosim/gz-sim/blob/gz-sim8/tutorials/auto_inertia_calculation.md#key-points-on-mesh-inertia-calculator).
Since many times users may tend to use meshes with origins, not at the center, this PR adds and uses a function to transform the Inertia Matrix to the COM in such cases.

Actual Scenario:
There was a bug in the center of mass and inertia matrix calculation. This PR fixes bugs in a moment of inertia matrix calculation and adds a test for the inertia calculation of a cylinder model with the origin set not at the center of mass.

Some extra functions were also added according to the previous understanding but are not needed after the bugfix:

- `CalculateMeshCentroid()`
- `TransformInertiaMatrixToCOM()`

> **Note:** I have still kept the functions as a part of this PR as they might come in handy later but they can be removed if required.

Below you can see multiple cylinder meshes with origins set at different places inside or outside the mesh getting similar inertia values:

inertia_different_origins

## Checklist
- [x] Added tests


azeey:
@jasmeet0915 would it be possible to add a test for this?

jasmeet0915:
@azeey I have added a test in [a161374](https://github.com/gazebosim/gz-sim/pull/2182/commits/a1613745c33b586e2cadfdc9faa9137cdd3475c9)

src/MeshInertiaCalculator.hh
```diff
@@ -103,6 +103,22 @@ namespace gz
...
+        /// \param[out] _massMatrix MassMatrix object to hold mass &
+        /// moment of inertia of the mesh
+        /// \param[in] _centerOfMass The centroid (center of mass) of the mesh
+        /// to which the inertia matrix has to be transformed.
+        /// \param[in] _inertiaOrigin The point about which the inertia matrix
+        /// was calculated. This would be the origin of the mesh.
+        public: void TransformInertiaMatrixToCOM(
+          gz::math::MassMatrix3d &_massMatrix,
+          gz::math::Pose3d &_centreOfMass,
+          gz::math::Pose3d &_inertiaOrigin);
```
iche033, line 120: can these be const ref?

jasmeet0915: Done in [01aa539](https://github.com/gazebosim/gz-sim/pull/2182/commits/01aa5395895d1188ea1b1b4269c17661d699cd2c)

iche033:
code looks good to me. Maybe @azeey can take a one final look before merging.

iche033 (approved):

scpeters:
is there documentation of the algorithm / integrals used in `MeshInertiaCalculator`? In particular I'm not sure where the terms in CalculateMassProperties are coming from

jasmeet0915:
> is there documentation of the algorithm / integrals used in `MeshInertiaCalculator`? In particular I'm not sure where the terms in CalculateMassProperties are coming from

@scpeters The method used was referred from this doc: https://www.geometrictools.com/Documentation/PolyhedralMassProperties.pdf
You can find the documentation of all the integrals in the same.

src/MeshInertiaCalculator.cc
```diff
@@ -101,12 +107,61 @@ void MeshInertiaCalculator::CalculateMeshCentroid(
   _centreOfMass.SetZ(centroid.Z());
 }
 
+//////////////////////////////////////////////////
+void MeshInertiaCalculator::TransformInertiaMatrixToCOM(
```
azeey, line 111: Since `CalculateMeshCentroid` and `TransformInertiaMatrixToCOM` are unused and untested, I think we should remove them.

azeey (commented):
Looks good! Thanks for the fix. My only suggestion is we remove unused code.

jasmeet0915:
> Looks good! Thanks for the fix. My only suggestion is we remove unused code.

@azeey I have removed the unused code in this commit: 7424bab

src/MeshInertiaCalculator.cc
jasmeet0915: Removed the unused functions in 7424bab. Would create a PR over at gz-math as soon as possible for the `TransformInertiaMatrixToCOM()` function.

azeey (approved):
Thanks @jasmeet0915!!

PR #2183, opened 2023-09-30
Standardize Doxygen parameter formatting for systems A-N

mabelzhang:
# 🦟 Bug fix

We were looking to improve the Doxygen pages for system plugins and noticed some huge paragraphs that meant to be bullet lists, and inconsistencies in headings.

This PR standardizes the formatting for SDF parameters for systems starting with letter A to N (which is not to say I will do the same for O to Z, as this isn't really the task I'm tackling... just fixing on the way as I familiarize myself with how systems are documented).
A-N is 46 systems. There are only 24 remaining that others can fight over.

## Summary

This is the standard being adopted (comes from `AckermannSteering`, not because it's the first in the alphabetical list, but because its rendered page looks the best):
- Heading size should be `## System Parameters`, to appear the same size as the preset "Detailed Description" heading
- Parameters should be in bullets, otherwise there's a variety of rendered results, especially when you have sub-bullets too, or if you do a `\sa`, or they are all in one huge paragraph.
- Parameter font should be in code face, ideally surrounded by angle brackets, e.g. `- <param>`
- Most parameters are followed by a colon, as in `- <param>:`, but I let some pass with hyphen, because it wasn't worth the effort.
- I wasn't too picky about indentation after the first line, because it wasn't worth the effort

## To test
```
colcon build --packages-select gz-sim8
```
In a browser, navigate to `build/gz-sim8/doxygen/html/index.html`
Click on the “gz::sim::systems” link.
Look at the corresponding system pages.

Compare to the live render here before the fix: https://gazebosim.org/api/sim/8/

## Checklist
- [ ] Added tests

### DIFF
--- a/src/systems/acoustic_comms/AcousticComms.hh
+++ b/src/systems/acoustic_comms/AcousticComms.hh
@@ -94,6 +94,7 @@ namespace systems
   ///    </propagation_model>
   ///
   ///  </plugin>
+  /// ```
 
   class AcousticComms:
     public gz::sim::comms::ICommsModel
--- a/src/systems/acoustic_comms/AcousticComms.hh
+++ b/src/systems/acoustic_comms/AcousticComms.hh
@@ -76,6 +76,7 @@ namespace systems
   ///       * <seed> : Seed value to be used for random sampling.
   ///
   /// Here's an example:
+  /// ```
   ///  <plugin
   ///    filename="gz-sim-acoustic-comms-system"
   ///    name="gz::sim::systems::AcousticComms">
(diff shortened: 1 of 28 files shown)


src/systems/acoustic_comms/AcousticComms.hh
```diff
@@ -76,6 +76,7 @@ namespace systems
   ///       * <seed> : Seed value to be used for random sampling.
   ///
   /// Here's an example:
+  /// ```
```
ahcorde, line 79: where are the closing quotes?

ahcorde (changes requested):

azeey (approved):
Other than the @ahcorde's unresolved issue, LGTM!

src/systems/acoustic_comms/AcousticComms.hh
mabelzhang: Good catch! That might have broken that page. Oops!! 0f65637

mabelzhang:
Addressed

mabelzhang:
Merging is blocked until change requested changes to green check :)

mjcarroll (approved):

PR #2189, opened 2023-10-03
Lift Drag Bug Fix

### LINKED ISSUE #2188: Missing coeffficients in LiftDrag Plugin
I was going through the code for the liftdrag plugin and noticed that the cm coefficient was set to 0. Above it is a comment marking it as todo, saying "it needs testing". I was wondering what was not working well that it had to be set to 0? The issue is that the effect of the coefficient is neglected for the remainder of the function, meaning the plugin does not work as accurate as it should.

The lines in question:
https://github.com/gazebosim/gz-sim/blob/f6efeefb4ab99284d1a053061235f8c2fa02a5da/src/systems/lift_drag/LiftDrag.cc#L438C3-L440C12

frede791:
# 🦟 Bug fix

Fixes #2188 

## Summary
This PR fixes a bug in the liftdrag plugin that prevented the moment coefficient (cm) from being correctly calculated. The reason for this bug is an mathematical error that calculated cos^2 rather than cos for the cosSweepAngle. Using this updated value, I added in missing calculations from the liftdrag plugin in gazebo-classic, so that the two are now identical in functionality. I also took the opportunity to update the header file to the current formatting and fixed a spelling mistake I found. I also added in the cm_delta parameter that was not present in the current version of the liftdrag plugin but is used by some drones (such as the tailsitter) and may thus find use in the future. 

In addition, wind has also been added as a parameter to be considered in calculations, bringing in line with what is present in the gazebo-classic lift drag plugin.

## Checklist

### DIFF
--- a/src/systems/lift_drag/LiftDrag.cc
+++ b/src/systems/lift_drag/LiftDrag.cc
@@ -494,6 +494,30 @@ void LiftDragPrivate::Update(EntityComponentManager &_ecm)
   const auto totalTorque = torque + cpWorld.Cross(force);
   Link link(this->linkEntity);
   link.AddWorldWrench(_ecm, force, totalTorque);
+
+  // Debug
+  // auto linkName = _ecm.Component<components::Name>(this->linkEntity)->Data();
+  // gzdbg << "=============================\n";
+  // gzdbg << "Link: [" << linkName << "] pose: [" << pose
+  //        << "] dynamic pressure: [" << q << "]\n";
+  // gzdbg << "spd: [" << vel.Length() << "] vel: [" << vel << "]\n";
+  // gzdbg << "LD plane spd: [" << velInLDPlane.Length() << "] vel : ["
+  //        << velInLDPlane << "]\n";
+  // gzdbg << "forward (inertial): " << forwardI << "\n";
+  // gzdbg << "upward (inertial): " << upwardI << "\n";
+  // gzdbg << "q: " << q << "\n";
+  // gzdbg << "cl: " << cl << "\n";
+  // gzdbg << "lift dir (inertial): " << liftI << "\n";
+  // gzdbg << "Span direction (normal to LD plane): " << spanwiseI << "\n";
+  // gzdbg << "sweep: " << sweep << "\n";
+  // gzdbg << "alpha: " << alpha << "\n";
+  // gzdbg << "lift: " << lift << "\n";
+  // gzdbg << "drag: " << drag << " cd: " << cd << " cda: "
+  //        << this->cda << "\n";
+  // gzdbg << "moment: " << moment << "\n";
+  // gzdbg << "force: " << force << "\n";
+  // gzdbg << "torque: " << torque << "\n";
+  // gzdbg << "totalTorque: " << totalTorque << "\n";
 }
 
 //////////////////////////////////////////////////
--- a/src/systems/lift_drag/LiftDrag.cc
+++ b/src/systems/lift_drag/LiftDrag.cc
@@ -282,7 +282,8 @@ void LiftDragPrivate::Update(EntityComponentManager &_ecm)
 
   const auto &pose = worldPose->Data();
   const auto cpWorld = pose.Rot().RotateVector(this->cp);
-  const auto vel = worldLinVel->Data() + worldAngVel->Data().Cross(cpWorld) - windLinearVel->Data();
+  const auto vel = worldLinVel->Data() + worldAngVel->Data().Cross(
+    cpWorld) - windLinearVel->Data();
 
   if (vel.Length() <= 0.01)
     return;
@@ -293,7 +294,8 @@ void LiftDragPrivate::Update(EntityComponentManager &_ecm)
   const auto forwardI = pose.Rot().RotateVector(this->forward);
 
   if (forwardI.Dot(vel) <= 0.0){
-    // Only calculate lift or drag if the wind relative velocity is in the same direction
+    // Only calculate lift or drag if the wind relative velocity
+    // is in the same direction
     return;
   }
 
--- a/src/systems/lift_drag/LiftDrag.cc
+++ b/src/systems/lift_drag/LiftDrag.cc
@@ -41,6 +41,7 @@
 #include "gz/sim/components/Name.hh"
 #include "gz/sim/components/ExternalWorldWrenchCmd.hh"
 #include "gz/sim/components/Pose.hh"
+#include "gz/sim/components/Wind.hh"
 
 using namespace gz;
 using namespace sim;
@@ -264,6 +265,11 @@ void LiftDragPrivate::Update(EntityComponentManager &_ecm)
   const auto worldPose =
       _ecm.Component<components::WorldPose>(this->linkEntity);
 
+  // get wind as a component from the _ecm
+  Entity windEntity = _ecm.EntityByComponents(components::Wind());
+  auto windLinearVel =
+          _ecm.Component<components::WorldLinearVelocity>(windEntity);
+
   components::JointPosition *controlJointPosition = nullptr;
   if (this->controlJointEntity != kNullEntity)
   {
@@ -276,7 +282,7 @@ void LiftDragPrivate::Update(EntityComponentManager &_ecm)
 
   const auto &pose = worldPose->Data();
   const auto cpWorld = pose.Rot().RotateVector(this->cp);
-  const auto vel = worldLinVel->Data() + worldAngVel->Data().Cross(cpWorld);
+  const auto vel = worldLinVel->Data() + worldAngVel->Data().Cross(cpWorld) - windLinearVel->Data();
 
   if (vel.Length() <= 0.01)
     return;
@@ -285,6 +291,11 @@ void LiftDragPrivate::Update(EntityComponentManager &_ecm)
 
   // rotate forward and upward vectors into world frame
   const auto forwardI = pose.Rot().RotateVector(this->forward);
+
+  if (forwardI.Dot(vel) <= 0.0){
+    // Only calculate lift or drag if the wind relative velocity is in the same direction
+    return;
+  }
 
   math::Vector3d upwardI;
   if (this->radialSymmetry)
--- a/src/systems/lift_drag/LiftDrag.cc
+++ b/src/systems/lift_drag/LiftDrag.cc
@@ -88,7 +88,7 @@ class gz::sim::systems::LiftDragPrivate
   /// \brief Cm-alpha rate after stall
   public: double cmaStall = 0.0;
 
-  /// \brief How much Cm changes with a change in control 
+  /// \brief How much Cm changes with a change in control
   /// surface deflection angle
   public: double cm_delta = 0.0;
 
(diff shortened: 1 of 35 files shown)


src/systems/lift_drag/LiftDrag.cc
```diff
@@ -155,6 +160,7 @@ void LiftDragPrivate::Load(const EntityComponentManager &_ecm,
   this->area = _sdf->Get<double>("area", this->area).first;
   this->alpha0 = _sdf->Get<double>("a0", this->alpha0).first;
   this->cp = _sdf->Get<math::Vector3d>("cp", this->cp).first;
+  this->cm_delta = _sdf->Get<double>("cm_delta",this->cm_delta).first;
```
ahcorde, line 163:
```suggestion
  this->cm_delta = _sdf->Get<double>("cm_delta", this->cm_delta).first;
```

src/systems/lift_drag/LiftDrag.hh
```diff
@@ -35,34 +35,36 @@ namespace systems
   /// \brief The LiftDrag system computes lift and drag forces enabling
   /// simulation of aerodynamic robots.
   ///
-  /// The following parameters are used by the system:
+  /// ##System Parameters
```
ahcorde, line 38:
```suggestion
  /// ## System Parameters
```

ahcorde (changes requested):

frede791:
Added in the wind velocity as a factor from the original liftdrag plugin.

arjo129 (changes requested):
Run `make codecheck` .

See here: https://gazebosim.org/docs/harmonic/contributing

https://github.com/gazebosim/gz-sim/actions/runs/6396048528/job/17419401043?pr=2189

frede791:
Removed long lines and fixed trailing white spaces.

src/systems/lift_drag/LiftDrag.cc
```diff
@@ -259,6 +265,11 @@ void LiftDragPrivate::Update(EntityComponentManager &_ecm)
   const auto worldPose =
       _ecm.Component<components::WorldPose>(this->linkEntity);
 
+  // get wind as a component from the _ecm
+  Entity windEntity = _ecm.EntityByComponents(components::Wind());
+  auto windLinearVel =
```
arjo129, line 270: What if there is no wind component? This is very common as many of the worlds will not define a wind component. Check if `windLinearVel` ia a nullptr before using it.

src/systems/lift_drag/LiftDrag.cc
```diff
@@ -281,6 +293,12 @@ void LiftDragPrivate::Update(EntityComponentManager &_ecm)
   // rotate forward and upward vectors into world frame
   const auto forwardI = pose.Rot().RotateVector(this->forward);
 
+  if (forwardI.Dot(vel) <= 0.0){
```
arjo129, line 296: I don't follow this. Why? Is returning the right behavior here?

src/systems/lift_drag/LiftDrag.cc
```diff
@@ -435,15 +453,13 @@ void LiftDragPrivate::Update(EntityComponentManager &_ecm)
   else
     cm = this->cma * alpha * cosSweepAngle;
 
-  /// \todo(anyone): implement cm
-  /// for now, reset cm to zero, as cm needs testing
-  cm = 0.0;
+  // Take into account the effect of control surface deflection angle to cm
+  cm += this->cm_delta * controlJointPosition->Data()[0];
```
arjo129, line 457: Check that controlJointPosition actually has data before accessing it.

src/systems/lift_drag/LiftDrag.cc
```diff
@@ -259,6 +265,11 @@ void LiftDragPrivate::Update(EntityComponentManager &_ecm)
   const auto worldPose =
       _ecm.Component<components::WorldPose>(this->linkEntity);
 
+  // get wind as a component from the _ecm
+  Entity windEntity = _ecm.EntityByComponents(components::Wind());
```
arjo129, line 269: What if no intity has wind. Check this is not equal to `kNullEntity`.

arjo129 (changes requested):
Currently this plugin causes a segfault in the tests - I think its because you access components without checking if they exist. 
To test your plugin doesnt break current integration tests run the following steps:
* `colcon build`
* `. install/setup.bash`
* run `build/gz-sim7/bin/INTEGRATION_lift_drag_system`

src/systems/lift_drag/LiftDrag.cc
frede791: If relative velocity of the vehicle forms a negative dot product with the forward vector then no lift can be generated and it does not need to be calculated.

src/systems/lift_drag/LiftDrag.cc
```diff
@@ -469,30 +485,6 @@ void LiftDragPrivate::Update(EntityComponentManager &_ecm)
   const auto totalTorque = torque + cpWorld.Cross(force);
   Link link(this->linkEntity);
   link.AddWorldWrench(_ecm, force, totalTorque);
-
```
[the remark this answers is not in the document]
frede791: In my other PR, I was asked to remove unused gzdbg commands by @ahcorde.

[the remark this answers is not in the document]
arjo129: This was not introduced by you. Whoever wrote it may have had a reason. Only touch stuff that effects your change.

src/systems/lift_drag/LiftDrag.cc
arjo129: Noted! Thanks. Makes sense.

frede791:
Improved based on comments. Integration tests and codecheck both pass. Commented feedback needs a reply if possible.

src/systems/lift_drag/LiftDrag.cc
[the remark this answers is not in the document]
frede791: So I should leave the gzdbg commands in?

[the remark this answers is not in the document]
frede791: gzdbg commands are back in the file.

arjo129 (approved):
Thanks for iterating on this so quickly. LGTM!

ahcorde (approved):

srmainwaring:
@ahcorde, I did not spot this change before it was merged into Harmonic, but have now tested with some of the models set up for ArduPilot and have some questions:

### Sweep angle calculation

I think the calculation for the sweep angle adjustment should not include the square root. In the original Gazebo classic version of the plugin the variable was named `cosSweepAngle2` (i.e. squared). 

- https://github.com/gazebosim/gazebo-classic/blob/7af0e34ff724b3447f85eba039f596cf6895aa8e/plugins/LiftDragPlugin.cc#L150C10-L150C24

It was change to `cosSweepAngle` in:

-  https://github.com/gazebosim/gazebo-classic/commit/9b98cb34ea74235eaa0b3a71c6b3751ffc1dd48b#diff-66d0f4a4af1ce5c7c5c7cffb63821f5a1aa944650b15be8e0ddcd44ee33e6603L167

I cannot find the PR corresponding to the change, so it's not clear why it was renamed.

The argument for using the square of the angle is that the sweep angle correction arises from using the velocity normal to the leading edge of the wing in the lift calculation, rather than the free stream velocity. As the velocity appears quadratically in the dynamic pressure the correction requires `cos(sweep_angle)^2` not `cos(sweep_angle)`. The adjustment is then applied as a correction to the lift coefficient rather than the velocity used in the dynamic pressure calculation which is possibly where the confusion arises, as it appears that it's taking a component of the resultant force, not the input velocity.

Unfortunately there is no reference provided to the source of the calculations used in the original derivation of the plugin, so it's hard to confirm the formula forming the basis for the calculations. 

### Moment calculation

This calculation is not enabled in the Gazebo classic version:

- https://github.com/gazebosim/gazebo-classic/blob/7ccef40e24831eb5cd974b489ee279fc064eacc4/plugins/LiftDragPlugin.cc#L346

There is not much explanation aside from it not being checked. I think this term is supposed to account for the movement of the centre of pressure with velocity, and allows the lift and drag coefficients to be calculated and applied at the quarter chord point rather than the unknown centre of pressure (so the cp parameter might be better named quarter chord?).

In any case there are a few models that will not work with this calculation enabled. The Zephyr is one of them: https://app.gazebosim.org/OpenRobotics/fuel/models/Zephyr%20Delta%20Wing. The solution to get this model working again is to zero `<cma>0.0</cma>` so the adjustment is zero. It should not affect any PX4 mdoels as AFAICT they all set the parameter to zero. The models used with ArduPilot are being updated to also zero this adjustment (which is backwards compatible with Garden).

An announcement that this change may break some models would be helpful.

arjo129:
Hi Rhys,
Thanks for pointing that out. It seems my review fell short of certain
checks. I think it seems like this PR has introduced a regression. For
moment calculations, the easiest thing to solve the incompatibilities is to
default cma to zero. We can revert the change that added a regression.

srmainwaring:
> On another note perhaps we should not allow any changes into existing plugins on the release branches. i.e. A change like this should target `main` so that errors get caught during the tutorial party

Maybe, although in this case the changes looked reasonable because of the variable naming. Btw I'm still not 100% certain of the intended sweep angle calculation. If someone has a good reference that would settle it that would be great. From what I understand the original motivation for swept wings arose to reduce the effective stream velocity over the wings in transonic / supersonic flight (so argument for taking components of the velocity). Either way a comment above the calculation about the intention would make things clearer, and we could revert the variable name back to it's original `cosSweepAngle2` if we decide that's the correct value. It does affect the lift and drag forces as well as the moment, so there should be a test to catch that.

`<cma>` does default to zero, the regression occurs for models where it is set to a non-zero value (overriding the default). In Garden the parameter was ignored, in Harmonic it isn't. I don't think it affects many models, as it appears in most cases to be set to zero anyway.

arjo129:
No `<cma>` currently [does not default to zero](https://github.com/gazebosim/gz-sim/blob/8f84b05435a93763dd1eb527a5e0b1865e7d1311/src/systems/lift_drag/LiftDrag.cc#L77). And our documentation does not seem to have been updated for it.

I dont know about cos vs cos2 @frede791 is there a reason why you'd use cos over cos2 otherwise I'll revert that change.

srmainwaring:
> No `<cma>` currently [does not default to zero](https://github.com/gazebosim/gz-sim/blob/8f84b05435a93763dd1eb527a5e0b1865e7d1311/src/systems/lift_drag/LiftDrag.cc#L77).

You're right - I misread that. Thanks for the the PR defaulting `<cma>` to zero, that'll provide consistency between major versions.

srmainwaring:
> I dont know about cos vs cos2 @frede791 is there a reason why you'd use cos over cos2 otherwise I'll revert that change.

Reference for simple sweep theory: https://web.archive.org/web/20060727083150/http://www.desktopaero.com/appliedaero/potential3d/sweeptheory.html

Should relabel variable to `cosSweepAngle2` and revert change.

frede791:
@arjo129 Thanks for catching this.

PR #2201, opened 2023-10-11
Fix sensors system parallel updates 

iche033:
# 🦟 Bug fix

depends on #2200 

## Summary

While profiling the Sensors system, I noticed that the render updates are no longer happening in parallel but instead it's blocking the main thread. 

The remotery graph below illustrate the issue. The `SimulationRunner` is being blocked for the entire duration of the sensors system's update

sensors_system_remotery_before

With changes in this PR, the  `SimulationRunner` will now continue to step while sensor updates take place:

sensors_system_remotery_after

More info:

It came down to this [lock](https://github.com/gazebosim/gz-sim/blob/8a808009f6672db964acd27e62b710f13535b063/src/systems/sensors/Sensors.cc#L668) that blocks the `PostUpdate` thread longer than it should. The fix is to use a new mutex (`renderUtilMutex`) that's specific `RenderUtil` updates. 

After reducing the scope of the lock, I noticed a race condition that sometimes the RenderUtil's states are [overriden](https://github.com/gazebosim/gz-sim/blob/8a808009f6672db964acd27e62b710f13535b063/src/systems/sensors/Sensors.cc#L669) (in the `PostUpdate` thread) before the changes are [applied](https://github.com/gazebosim/gz-sim/blob/8a808009f6672db964acd27e62b710f13535b063/src/systems/sensors/Sensors.cc#L305) to the scene (in the rendering thread). An `updateTimeCv` condition variable is added to make sure that does not happen. 

To test:

Run the `INTEGRATION_sensors_system_update_rate` test multiple times to verify the sensors' timestamps are correct.

On my desktop machine, the RTF values (displayed in the GUI)  for the `sensors_demo.sdf`  world are:

* Before: 77-78%
* After: 95-98%

Here is another [test world](https://gist.github.com/iche033/3af9b79fc45a28217df62cd93ab79fab) consisting of 2 1980x1280 cameras @ 10Hz.

The raw RTF values over 2000 iterations are plotted in the graphs below. 

Before:

RTF dips on every camera update

sensors_demo_parallel_before

After:

There is still jitter but less than before

sensors_demo_parallel_after

## Checklist
- [ ] Added tests

### DIFF
--- a/src/systems/sensors/Sensors.cc
+++ b/src/systems/sensors/Sensors.cc
@@ -389,9 +389,12 @@ void SensorsPrivate::RunOnce()
     this->activeSensors.clear();
   }
 
-  this->updateAvailable = false;
   this->forceUpdate = false;
-  this->renderCv.notify_one();
+  {
+    std::unique_lock<std::mutex> cvLock(this->renderMutex);
+    this->updateAvailable = false;
+    this->renderCv.notify_one();
+  }
 }
 
 //////////////////////////////////////////////////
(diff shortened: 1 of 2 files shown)


ahcorde (approved):

iche033:
For some reason, github is stuck processing updates from my last push. Going to close and reopen the PR.
github_processing_updates

PR #2207, opened 2023-10-14
Add tutorial for using components in systems

mabelzhang:
# 🎉 New feature

Practices:
- I tried to keep it as straight-forward as possible, only providing brief explanation when needed. This is so that tutorials are not explanations, following the [Divio documentation](https://documentation.divio.com/) principles.
- Links to other pages are provided when context and examples are needed, so that a user who isn’t a dev team developer can understand what we’re talking about. But, IANAD, IANAL, IANAU (I am not a user).
- For the sake of maintenance, all links are `\ref` links or relative links, except for links going to GitHub source code.

## Summary

This PR adds two pages.

Parent page is an overview of using components in systems. It has a high-level overview of what people need to know, linking to prerequisites and resources. It then talks about the ECM object in practical terms as the gateway to access entities and components.

It then links to a list of case studies, of which there is just 1 item right now, for `JointForceCmd` component (child page). The assumption is that we will be adding more case studies to this list for other components.

The child page uses the simplest built-in system plugin `ApplyJointForce` to explain how to use the `JointForceCmd` component through the ECM, by finding the Joint and then modifying the component.

## Test it

```
colcon build --merge-install --packages-select gz-sim8 --cmake-target doc
```

Navigate browser to `build/gz-sim8/doxygen/html/usingcomponents.html` (parent page) and `build/gz-sim8/doxygen/html/jointforcecmdcomponent.html` (child page).

## Checklist
- [ ] Added tests


tutorials/component_jointforcecmd.md
```diff
@@ -0,0 +1,79 @@
...
+
+In this case, we use the joint entity found above to look for and modify its
+`JointForceCmd` component.
+This will apply a force command to the joint.
+
+In `PreUpdate()`, look for the component:
+\snippet src/systems/apply_joint_force/ApplyJointForce.cc jointForceComponent
+
+Create it if it does not exist yet, and modify it:
+\snippet src/systems/apply_joint_force/ApplyJointForce.cc modifyComponent
```
ahcorde, line 62:
```suggestion
Create it if it does not exist yet, and modify it:

\snippet src/systems/apply_joint_force/ApplyJointForce.cc modifyComponent
```

ahcorde (changes requested):

mabelzhang:
DCO was complaining because the 4 automatic signoffs used gmails. I just set it to pass.

Ready for another look.

mabelzhang:
DCO complained again, so I fixed it for real.

tutorials/using_components.md
```diff
@@ -0,0 +1,69 @@
...
+
+An ECM object is passed into all of the interfaces in a system, including
+`ISystemConfigure()` and `ISystem*Update()`.
+The signatures of those interfaces are specified in
+\ref gz/sim/System.hh and explained in the
+[tutorial on creating system plugins](./createsystemplugins.html).
+We will assume that these interfaces are implemented in functions called
+`Configure()` and `*Update()` in a system.
+
+Note that because entities are created when the world is loaded, at the time
```
azeey, line 59: I think we can be more precise about this.  `Configure` is called after all the elements in the parent element of the plugin are loaded, i.e, If the plugin is attached to a `<model>`, all the elements in the `<model>` would have been loaded before `Configure` is called. The same applies if the plugin is attached to `<world>`. I believe the only time you have to use `*Update` for configuration is when the plugin depends on entities outside the parent element of the plugin. For example, the `Touch` plugin does this because it works by checking if the model containing the plugin comes in contact with another model.

mabelzhang: Thanks! I did not know about the details. Updated in 1c1e023ba

azeey (approved):

PR #2212, opened 2023-10-17
 Standardize Doxygen parameter formatting for systems O-Z

mabelzhang:
# 🦟 Bug fix

Follow up to https://github.com/gazebosim/gz-sim/pull/2183
Fixes the rest of the alphabet.

## Summary
See #2183

## Checklist
- [ ] Added tests

### DIFF
--- a/src/systems/thruster/Thruster.hh
+++ b/src/systems/thruster/Thruster.hh
@@ -123,15 +123,17 @@ namespace systems
   /// gz sim auv_controls.sdf
   /// ```
   /// To control the rudder of the craft run the following:
-  /// ```
-  /// gz topic -t /model/tethys/joint/vertical_fins_joint/0/cmd_pos \
-  ///    -m gz.msgs.Double -p 'data: -0.17'
-  /// ```
+  /** ```
+      gz topic -t /model/tethys/joint/vertical_fins_joint/0/cmd_pos \
+         -m gz.msgs.Double -p 'data: -0.17'
+      ```
+  **/
   /// To apply a thrust you may run the following command:
-  /// ```
-  /// gz topic -t /model/tethys/joint/propeller_joint/cmd_thrust \
-  ///    -m gz.msgs.Double -p 'data: -31'
-  /// ```
+  /** ```
+      gz topic -t /model/tethys/joint/propeller_joint/cmd_thrust \
+         -m gz.msgs.Double -p 'data: -31'
+      ```
+  **/
   /// The vehicle should move in a circle.
   class Thruster:
     public gz::sim::System,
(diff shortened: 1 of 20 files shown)


ahcorde (changes requested):
linters are failing

ahcorde (approved):

src/systems/thruster/Thruster.hh
```diff
@@ -112,15 +123,17 @@ namespace systems
   /// gz sim auv_controls.sdf
   /// ```
   /// To control the rudder of the craft run the following:
-  /// ```
-  /// gz topic -t /model/tethys/joint/vertical_fins_joint/0/cmd_pos
-  ///    -m gz.msgs.Double -p 'data: -0.17'
-  /// ```
+  /** ```
```
azeey, line 126: Is it necessary to mix comment styles?

mabelzhang: There were 2 gcc warnings about multi-line comment, when in fact it's not a multi-line comment, it's just a multi-line code block. gcc doesn't like the `\` character at the end of the line in the code block. It's two lines of code, and if we don't add the `\`, when people copy-paste into the terminal, it probably will give an error. If we put them into one line, then it'll be over 80 characters.

So I tried to make it a multi-line comment to see if those warnings go away. But looks like the build didn't produce outputs... maybe I'll rerun it.

mabelzhang: These were the warnings https://build.osrfoundation.org/job/gz_sim-ci-pr_any-jammy-amd64/309/gcc/

mabelzhang:

src/systems/thruster/Thruster.hh
mabelzhang: This comment style has gotten rid of the warnings https://build.osrfoundation.org/job/gz_sim-ci-pr_any-jammy-amd64/315/

There are old warnings about buoyancy_engine from #2183. I'll get rid of those warnings the same way too. I guess that's why they didn't add the `\` before.
Sigh, it's a tradeoff between the code lines being copy-pastable in documentation, versus the code looking clean. I think correctness in documentation should take precedence because probably more people look at that.

azeey: Okay, I agree that correctness in documentation is more important, so let's keep the comment style that makes it work. Maybe someday we can convince cppcheck that long comments are okay, and put the command on one line.

azeey:
@mabelzhang could you fix the compiler warnings on BuoyancyEngine.hh?

src/systems/thruster/Thruster.hh
mabelzhang: 67c84499c

mabelzhang:
Got distracted. Let's see the new CI...

azeey (approved):
LGTM with green CI

mabelzhang:
CI is good. (The warnings are the same as those in other PRs.)

azeey:
It would be great if we can backport this to gz-sim7

PR #2219, opened 2023-11-01
Add tutorial for using the Pose component

mabelzhang:
# 🎉 New feature

Depends on #2207, which adds the main component tutorial page. This PR is a sub-page.

## Summary

Adds a tutorial on how to use the Pose component, using OdometryPublisher system as an example, because that system depends on only 1 component, good as a standalone example.

## Test it

```
colcon build --merge-install --packages-select gz-sim8 --cmake-target doc
```

Navigate browser to `build/gz-sim8/doxygen/html/usingcomponents.html` (parent page) and `build/gz-sim8/doxygen/html/posecomponent.html` (child page).

## Checklist
- [ ] Added tests


ahcorde (approved):

src/systems/odometry_publisher/OdometryPublisher.cc
```diff
@@ -287,13 +292,15 @@ void OdometryPublisher::PreUpdate(const gz::sim::UpdateInfo &_info,
   }
 
   // Create the pose component if it does not exist.
+  //! [createComponent]
```
azeey, line 295: We actually don't need to create the `Pose` component since it's created by default on every `Model` entity, so this code can be removed. We don't have documentation on the default list of components (https://github.com/gazebosim/gz-sim/issues/2306), but we should at least remove counter examples.

mabelzhang: [9c0b2c4](https://github.com/gazebosim/gz-sim/pull/2219/commits/9c0b2c46ff46e5fd6924bcb7fcf819ba1a2ca0f5)

azeey:
@azeey to take another look.

Issue #2165, opened 2023-09-22
Removing model with `JointPositionController` plugin leads to 'entity does not exist' error messages.


jrutgeer:
When a model that uses the `JointPositionController` plugin is removed from the simulation through the UserCommands service, repeating error messages appear:

```
[Err] [EntityComponentManager.cc:1083] Trying to create a component of type [8319580315957903596] attached to entity [12], but this entity does not exist. This create component request will be ignored.
```

See the attached example files:

Run `gz sim -r world.sdf`
Spawn a model with the spawn script: `./spawn_boxes`
Remove the boxes with the remove script: `./remove_boxes`

--> The error messages start.

Edit the `boxes.sdf` file and remove the `JointPositionController` plugin.
Repeat the test.

--> No error messages after removal.

[example.tar.gz](https://github.com/gazebosim/gz-sim/files/12704007/example.tar.gz)

arjo129:
I think if we solve #2217 this will no longer be an issue.

jrutgeer:
Closing as duplicate of https://github.com/gazebosim/gz-sim/issues/2217.

PR #2232, opened 2023-11-07
Remove systems if their parent entity is removed

### LINKED ISSUE #2217: System plugins keep executing after their model is removed.
## Environment
* OS Version: Ubuntu 22.04
* Harmonic, source built

## Description
Many system plugins need to be attached to a model (e.g. to access that model's links, collisions, etc).
However, when that model is removed, the plugin keeps running.
Given typically a model plugin needs to access entities of the model (that no longer exist), this leads to crashes.

## Expected behavior
When a model is removed, all of its plugins are removed as well, or at the least their pre/update/post methods are no longer called.

## Steps to reproduce
- Create a trivial system plugin, e.g. 

```c++
void TrivialPlugin::PreUpdate(const gz::sim::UpdateInfo &_info,
                         gz::sim::EntityComponentManager &_ecm)
{
  if (_info.paused) return;

  static int counter = 0;
  counter++;
  if (1000 == counter)
  {
    counter = 0;
    std::cout << "TrivialPlugin::PreUpdate" << std::endl;
  }
}
```

- Add this plugin to a model.

- Start the simulation

- Select the model in the Entity Tree, right click + 'remove'

- The model is removed, but the `cout` message is still being output to the terminal.

arjo129:
# 🦟 Bug fix

Fixes #2217

## Summary
In particular if a user despawns an entity, the associated plugin gets removed. This should prevent issues like #2165. TBH I'm not sure if this is the right way forward as a system should technically be able to access any entity in a traditional ECS.

no_crash

The PR has now been reworked and greatly simplified. All we do is stop all worker threads if an entity is removed and then recreaye remaining threads.

## Checklist
- [x] Added tests


jrutgeer:
@arjo129 

I thought of this possible alternative implementation:

- Only the system plugin itself knows which entities are necessary for its functioning,
   - This could e.g. be the model  entity it is attached to, but it could as well be any other entity,
   - Hence it's the system plugin's responsibility to identify those entities,
- To identify these entities, the system plugin adds a component `used_by_plugin` to each entity it acts upon,
   -  A `used_by_plugin` component identifies a plugin which acts on a component,
- Upon removal of entities, the ECM checks whether they have `used_by_plugin` components,
  - If so, the SystemManager is triggered to remove the corresponding plugin.

A possible extension could be to also have a kind of `severity` flag in the component, to allow for different reactions from the SystemManager depending on which entities are removed.

E.g.: consider a system attached to a model and which acts on some subentities of that model:
- If the model entity is removed, the system should also be removed,
- But if a subentity is removed, maybe the system should not be removed yet, but rather perform some 'reconfigure' action instead.

arjo129:
I like the thought process behind your suggested API addition. However, in my mind that would be an added feature. We need to (1) be able to remove systems safely. That is something that will take a fair deal of work right now and what this PR hopes to (eventually achieve). Your proposed API could be bolted on top in Ionic once we are able to perform safe removal. The current state of this PR is we can safely remove some plugins, but it causes weird breakages in the `PosePublisher` plugin which I need to investigate prior to making any changes. It is also entirely possible that I'm doing something wrong with the barriers. Either ways I need to revisit this another day when my head is clearer and I have more time.

image

arjo129:
I think this works now... Will have to write some unit tests to verify. But it does not go :boom: on the warehouse world.

azeey:
I haven't had a chance to look into this deeply, but one thing to keep is how this interacts with the levels feature. If a model that contains a system is unloaded by the levels feature and gets loaded back again, what would be the behavior? Would the system get configured and start anew? If that works properly, that would cover most cases, but we might also want the ability for a system to stay loaded even when its parent entity is removed so that if the parent entity is loaded back in, the system might continue from where it left off.

Bogdanov-am:
Is any work ongoing in this PR? I'm currently working on a simulation that requires frequent loading and unloading of models. The problem of unloaded plugins really worries me.

arjo129:
I just need to write some tests for this.

arjo129:
Tests have been written. I think this is ready for feedback.

landersson:
Also hoping to see this merged soon.

arjo129:
If you'd like to get this checked in faster you could help with reviewing it. One thing to do is to checkout the relevant branch and run it on example worlds to see that things work as expected.

landersson:
> If you'd like to get this checked in faster you could help with reviewing it. One thing to do is to checkout the relevant branch and run it on example worlds to see that things work as expected.

Unfortunately, I don't have any experience with the gz-sim codebase and can't really comment on whether this is a suitable solution or not. 

Just need a way to remove models without plugins like JointStatePublisher crashing.

arjo129:
> Unfortunately, I don't have any experience with the gz-sim codebase and can't really comment on whether this is a suitable solution or not.

You can try it out and tell me if it crashes! There may be plugins I've not tested on!

src/SystemManager.hh
```diff
@@ -41,6 +43,21 @@ namespace gz
     // Inline bracket to help doxygen filtering.
     inline namespace GZ_SIM_VERSION_NAMESPACE {
 
+    /// \brief Helper container to keep track of
+    /// system interfaces and their parents
+    template<typename IFace>
+    struct SystemHolder {
```
scpeters, line 49: `SystemHolder` isn't the clearest name; it sounds very much like `SystemContainer`, but it doesn't hold systems. I think `SystemIfaceWithParent` is more accurate, though not super concise

src/systems/pose_publisher/PosePublisher.cc
```diff
@@ -155,16 +155,6 @@ class gz::sim::systems::PosePublisherPrivate
   /// improves performance by avoiding memory allocation
   public: std::vector<std::pair<Entity, math::Pose3d>> staticPoses;
 
-  /// \brief A variable that gets populated with poses. This also here as a
```
landersson, line 158: What is the reason for turning these member variables into local variables in the PosePublisherPrivate::PublishPoses() method?

arjo129: Spooky action at a distance. Described in commit message for: https://github.com/gazebosim/gz-sim/pull/2232/commits/a4c89e3ce4b6cf230ed6dd84649e0192e5d93e33

> This commit removes an "optimization" in the PosePublisher that was
causing the system to segfault when an entity was removed. In reality
this commit is unrelated to the previous few commits. However, without this change
the previous change would cause removal of an entity to crash demos that
used pose publishers. I think the crux of the issue is that protobufs takeover
memory management of objects leading to a scenario where if a reference
was freed elsewhere, the protobuf would continue to hold said reference.
When we try to do operations on the protobuf we end up double-freeing
the memory and hence there is a 💥.

src/SystemManager.hh
landersson: Not sure that adding SystemHolder is neccessary. All the info about system->parentId mapping is already stored in the SystemManager::systems container. 

So in SystemManager::ProcessRemovedEntities(), what about just filtering SystemManager::systems and then clearing+rebuilding the interface containers (systemsConfigure, systemsReset) etc from scratch, like in SystemManager::ActivatePendingSystems() ?

src/systems/pose_publisher/PosePublisher.cc
landersson: Ok, my concern is that this change didn't actually fix the problem, but just made it more unlikely to happen.

arjo129: I came to this conclusion by using the debugger. I'm pretty sure it 100% solves the problem of the crash within this plugin.

src/SystemManager.hh
scpeters: you are correct that the `SystemInternal` data structure held by the `systems` member variable does include the parent entity, but this `SystemHolder` data structure doesn't store `SystemInternal` data. My first comment suggests renaming this data structure, since it doesn't hold a `gz::sim::System` or `SystemInternal`; rather it stores an interface pointer and `Entity` Id

scpeters: and to add why this is needed; it's currently augmenting the data structures that hold a vector of interface pointers

arjo129: I think @landersson suggestion may have some merit. Unfortunately I discovered a bug in my `SystemContainer` implementation which is taking me a bit of time to debug  while implementing it.

src/SystemCointainer.hpp
scpeters, line 1: rename to `SystemContainer.hpp`

src/SystemContainer_TEST.cc
```diff
@@ -0,0 +1,138 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+*/
+#include <gtest/gtest.h>
+#include <unordered_set>
+#include "SystemCointainer.hpp"
```
scpeters, line 19:
```suggestion
#include "SystemContainer.hpp"
```
after renaming the header file

src/SystemManager.cc
```diff
@@ -16,10 +16,13 @@
 */
 
 #include <list>
+#include <mutex>
 #include <set>
 
 #include <gz/common/StringUtils.hh>
 
+#include "SystemCointainer.hpp"
```
scpeters, line 24:
```suggestion
#include "SystemContainer.hpp"
```

src/SystemManager.hh
```diff
@@ -32,6 +33,7 @@
 #include "gz/sim/SystemLoader.hh"
 #include "gz/sim/Types.hh"
 
+#include "SystemCointainer.hpp"
```
scpeters, line 36:
```suggestion
#include "SystemContainer.hpp"
```

scpeters (commented):
nit: fix spelling in header file name

src/SystemManager.hh
scpeters: > I think @landersson 's suggestion may have some merit (thanks @landersson).

👍 

> Unfortunately I discovered a bug in my `SystemContainer` implementation which is taking me a bit of time to debug while implementing it.
> 
> Essentially since we dont run remove entities very often it should be pretty ok to flush and rebuild the `SystemIface` list. Either ways, I will debug `SystemContainer` first (Fixed in [4b4332d](https://github.com/gazebosim/gz-sim/pull/2232/commits/4b4332da7785985283a26ca79614c9fc370bbf7d)).

ok sounds good

arjo129: Unfortunately I think its been too long since I touched this PR. The runner does need access to the parent information hence we need the `SystemIfaceWithParent` class. Although it'd have been nice not to have it.

src/SystemContainer_TEST.cc
arjo129: 2416a19d571504576bbe8c7a5dfd40c63eaca13e

src/SystemManager.cc
arjo129: [079d252](https://github.com/gazebosim/gz-sim/pull/2232/commits/079d2521544a0c3303fabcd4cccee3f7dd06ad7f)

src/SystemCointainer.hpp
arjo129: [079d252](https://github.com/gazebosim/gz-sim/pull/2232/commits/079d2521544a0c3303fabcd4cccee3f7dd06ad7f)

src/SystemContainer.hpp
```diff
@@ -0,0 +1,162 @@
...
+      }
+
+      private: std::vector<T> systems;
+
+      private: std::vector<bool> occupied;
+
+      private: std::vector<std::size_t> freeSpots;
+
+      //////////////////////////////////////////
+      class iterator : public std::iterator<
```
scpeters, line 105: I just noticed clang is giving deprecation warnings about using `std::iterator`, which is [deprecated as of c++17](https://en.cppreference.com/w/cpp/iterator/iterator)

* [Build Status](https://build.osrfoundation.org/job/gz_sim-ci-pr_any-homebrew-amd64/429/) https://build.osrfoundation.org/job/gz_sim-ci-pr_any-homebrew-amd64/429/clang/

arjo129: [5001971](https://github.com/gazebosim/gz-sim/pull/2232/commits/500197164f398279d65028af9a9549bfda48c6a8)

src/SimulationRunner.cc
```diff
@@ -565,10 +567,22 @@ void SimulationRunner::ProcessSystemQueue()
         this->postUpdateStartBarrier->Wait();
         if (this->postUpdateThreadsRunning)
         {
-          system->PostUpdate(this->currentInfo, this->entityCompMgr);
+          auto terminate = this->threadsToTerminate.find(parentEntity);
+          if (terminate != this->threadsToTerminate.end()) {
+            terminate->second--;
+            if (terminate->second == 0) {
+              this->threadsToTerminate.erase(terminate);
```
arjo129, line 574: :warning: Threads to terminate is not thread safe. Rework this bit.

arjo129: [0089730](https://github.com/gazebosim/gz-sim/pull/2232/commits/008973093d0e87d592d1876f3fcf5d6b363d11d1)

src/SystemContainer.hpp
arjo129: [5001971](https://github.com/gazebosim/gz-sim/pull/2232/commits/500197164f398279d65028af9a9549bfda48c6a8)

src/SystemManager.cc
```diff
@@ -409,3 +426,93 @@ void SystemManager::ProcessPendingEntitySystems()
...
+    [&](const SystemIfaceWithParent<ISystemPreUpdate>& system) {
+      return _ecm.IsMarkedForRemoval(system.parent);
+    });
+  RemoveFromVectorIf(this->systemsUpdate,
+    [&](const SystemIfaceWithParent<ISystemUpdate>& system) {
+      return _ecm.IsMarkedForRemoval(system.parent);
+    });
+  RemoveFromVectorIf(this->systemsPostupdate,
+    [&](const SystemIfaceWithParent<ISystemPostUpdate>& system) {
+      // Post update system. Make sure that the threadsToTerminate
```
scpeters, line 493: this comment seems incomplete

scpeters (commented):
this looks good to me! I just noted two places where comments could be improved, and then I will be ready to approve

src/SystemManager.cc
scpeters:
```suggestion
      // If system with a PostUpdate is marked for removal, mark its thread for
      // termination.
```

scpeters (approved):

src/SystemContainer.hpp
```diff
@@ -0,0 +1,162 @@
...
+    ///  -> TODO(arjoc): It should be possible to use a cache
+    ///  to make each iterator increment have O(1) time.
+    ///  For now worst case iterator increment is O(n).
+    ///  That being said for most cases the increment
+    ///  should be O(1). Eitherways, run does not iterate
+    ///  through systems but rather ISystem* interfaces.
+    ///  The only two places we iterate over this is during
+    ///  reset and TotalByEntity in the tests.
+    template<typename T>
+    class SystemContainer
```
azeey, line 50: Can you a put this class inside a `inline namespace GZ_SIM_VERSION_NAMESPACE`. You'll have to include `config.hh` as well.

src/SystemContainer.hpp
azeey, line 1: Headers have a `.hh` extension in this codebase.

src/SystemContainer.hpp
```diff
@@ -0,0 +1,162 @@
...
+      }
+
+      private: std::vector<T> systems;
+
+      private: std::vector<bool> occupied;
+
+      private: std::vector<std::size_t> freeSpots;
+
+      //////////////////////////////////////////
+      class iterator {
```
azeey, line 106: nit: capitalize?

src/SystemManager.cc
```diff
@@ -409,3 +426,94 @@ void SystemManager::ProcessPendingEntitySystems()
...
+    else
+    {
+      vec[i-j] = vec[i];
+    }
+  }
+
+  while (vec.size() > originalSize - j)
+  {
+    vec.pop_back();
+  }
```
azeey, line 462: Can you use `std::vector::erase` and `std::remove_if`? https://stackoverflow.com/a/39019851/283225

src/systems/pose_publisher/PosePublisher.cc
```diff
@@ -155,16 +155,6 @@ class gz::sim::systems::PosePublisherPrivate
...
-  /// member variable to avoid repeated memory allocations and improve
-  /// performance.
-  public: msgs::Pose poseMsg;
-
-  /// \brief A variable that gets populated with poses. This also here as a
-  /// member variable to avoid repeated memory allocations and improve
-  /// performance.
-  public: msgs::Pose_V poseVMsg;
```
azeey, line 166: This is definitely concerning. Do we know the root cause? I would be hesitant to merge this if we have to put this kind of condition on system authors.

src/SystemContainer.hpp
scpeters: I'm guessing it doesn't want to be capitalized in order to use [std::iterator_traits](https://en.cppreference.com/w/cpp/iterator/iterator_traits)?

azeey: That shouldn't be necessary. [Here's](https://stackoverflow.com/a/67650380/283225) a counter example. AFAIK, there are no requirements on the name of the class, but the `using` directives below need to be as they are.

scpeters: ok thanks, I wasn't exactly sure how it worked. I guess it gets the iterators from `begin()` and `end()` so it would infer the types from them

src/systems/pose_publisher/PosePublisher.cc
scpeters: I've tried to reproduce the failure by reverting the changes to `PosePublisher` on this branch, running the [pose_publisher.sdf](https://github.com/gazebosim/gz-sim/blob/gz-sim8/examples/worlds/pose_publisher.sdf) example world, echoing the double pendulum pose topic, and then removing the pendulum, and I haven't seen any problems in testing on macOS. I suppose I could try to reproduce on Ubuntu

arjo129: From [a4c89e3](https://github.com/gazebosim/gz-sim/pull/2232/commits/a4c89e3ce4b6cf230ed6dd84649e0192e5d93e33):

> This commit removes an "optimization" in the PosePublisher that was
causing the system to segfault when an entity was removed. In reality
this commit is unrelated to the previous few commits. However, without this change
the previous change would cause removal of an entity to crash demos that
used pose publishers. I think the crux of the issue is that protobufs takeover
memory management of objects leading to a scenario where if a reference
was freed elsewhere, the protobuf would continue to hold said reference.
When we try to do operations on the protobuf we end up double-freeing
the memory and hence there is a 💥.

Any form of removal will require that system authors think about lifetimes. I could get the warehouse world to segfault on ubuntu when I made this change. Its been several months now so I havent tried recently.

src/SystemContainer.hpp
arjo129: [d6462a2](https://github.com/gazebosim/gz-sim/pull/2232/commits/d6462a2a953f62dafc0936838276f13b6a324841)

src/SystemContainer.hpp
arjo129: [d6462a2](https://github.com/gazebosim/gz-sim/pull/2232/commits/d6462a2a953f62dafc0936838276f13b6a324841)

src/SystemContainer.hpp
arjo129: [d6462a2](https://github.com/gazebosim/gz-sim/pull/2232/commits/d6462a2a953f62dafc0936838276f13b6a324841)

src/SystemManager.cc
arjo129: 24f539dae

src/systems/pose_publisher/PosePublisher.cc
arjo129: Yep can confirm without this change I get a crash in `PosePublisher`:
```
#11   Object "[0xffffffffffffffff]", at 0xffffffffffffffff, in 
#10   Object "/lib/x86_64-linux-gnu/libc.so.6", at 0x7fc6e332684f, in 
#9    Object "/lib/x86_64-linux-gnu/libc.so.6", at 0x7fc6e3294ac2, in 
#8    Object "/lib/x86_64-linux-gnu/libstdc++.so.6", at 0x7fc6df4dc252, in 
#7    Object "/home/arjoc/ws/gz/ionic/install/lib/libgz-sim9.so.9", at 0x7fc6de77203e, in 
#6    Object "/home/arjoc/ws/gz/ionic/install/lib/gz-sim-9/plugins/libgz-sim-pose-publisher-system.so", at 0x7fc6aabfa3b7, in 
#5    Object "/home/arjoc/ws/gz/ionic/install/lib/gz-sim-9/plugins/libgz-sim-pose-publisher-system.so", at 0x7fc6aabf9ef0, in 
#4    Object "/home/arjoc/ws/gz/ionic/install/lib/libgz-transport14.so.14", at 0x7fc6de8ed6b0, in gz::transport::v14::Node::Publisher::Publish(google::protobuf::Message const&)
#3    Object "/home/arjoc/ws/gz/ionic/install/lib/libgz-msgs11.so.11", at 0x7fc6dde8e156, in gz::msgs::Pose::ByteSizeLong() const
#2    Object "/home/arjoc/ws/gz/ionic/install/lib/libgz-msgs11.so.11", at 0x7fc6dddf11e8, in gz::msgs::Header::ByteSizeLong() const
#1    Object "/home/arjoc/ws/gz/ionic/install/lib/libgz-msgs11.so.11", at 0x7fc6dddeec5a, in gz::msgs::Header_Map::ByteSizeLong() const
#0    Object "/lib/x86_64-linux-gnu/libprotobuf.so.23", at 0x7fc6dd8f3922, in google::protobuf::RepeatedPtrField<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >::Get(int) const
Segmentation fault (Address not mapped to object [0x8])
```

azeey: The stack trace indicates that the PosePublisher actually tried to publish after it was removed. To me, this looks like we haven't cleared the `PostUpdate` thread queue. Somehow we're calling the `PostUpdate` ona  deleted object? 

The `PosePublisher` system is not doing anything special. It's simply publishing in the `PostUpdate` call, so I don't think we can justify requiring system authors to think this deeply about lifetimes. I would understand having this requirement if it had its own thread like the `Sensors` system. But even then, the destructor of the system should be called, which should ideally terminate any internal threads and join them.

arjo129: You're right. I'll go figure it out. It seems like not all the threads are terminating cleanly. I dont also call the destructors the threads.

scpeters: maybe just make sure the `transport::Node` and `transport::Publisher` objects are defined at the end of the object so they get destroyed first?

~~~
diff --git a/src/systems/pose_publisher/PosePublisher.cc b/src/systems/pose_publisher/PosePublisher.cc
index 58661d481..56ea3e52b 100644
--- a/src/systems/pose_publisher/PosePublisher.cc
+++ b/src/systems/pose_publisher/PosePublisher.cc
@@ -83,18 +83,9 @@ class gz::sim::systems::PosePublisherPrivate
       const msgs::Time &_stampMsg,
       transport::Node::Publisher &_publisher);
 
-  /// \brief Gazebo communication node.
-  public: transport::Node node;
-
-  /// \brief publisher for pose data
-  public: transport::Node::Publisher posePub;
-
   /// \brief True to publish static transforms to a separate topic
   public: bool staticPosePublisher = false;
 
-  /// \brief publisher for pose data
-  public: transport::Node::Publisher poseStaticPub;
-
   /// \brief Model interface
   public: Model model{kNullEntity};
 
@@ -171,6 +162,15 @@ class gz::sim::systems::PosePublisherPrivate
 
   /// \brief Whether cache variables have been initialized
   public: bool initialized{false};
+
+  /// \brief Gazebo communication node.
+  public: transport::Node node;
+
+  /// \brief publisher for pose data
+  public: transport::Node::Publisher posePub;
+
+  /// \brief publisher for pose data
+  public: transport::Node::Publisher poseStaticPub;
 };
 
~~~

arjo129: I think the issue is SystemContainer does not call the destructor in the first place.

scpeters: I was able to reproduce the crash with the following world, but my patch appears to fix it:

~~~
gz sim -v 4 "https://fuel.gazebosim.org/1.0/OpenRobotics/worlds/Tugbot in Warehouse"
~~~

arjo129: @azeey what do you think of @scpeter's solution?

Also do I need to call destructors upon deletion? Right now I just defer calling the destructor until the next instance of remove is called.

arjo129: Destructors are now called in [a764498](https://github.com/gazebosim/gz-sim/pull/2232/commits/a764498bfe23f5e468820aab75dcca5d5adba578).

arjo129: Based on print statements I can guarantee that the pose publisher is actually being terminated. @scpeters patch causes a crash still. I am more or less sure this has to do with abusing a ProtoBuf as a storage class. Specifically I ran the warehouse world. I despawned entity 16. The threads belonging to 16 cleanly exit:
Screenshot from 2024-06-20 09-02-48

The thread that crashes belongs to entity 216 (which incidentally has a pose publisher as well): 

Screenshot from 2024-06-20 08-53-18

azeey: So I added some debug messages in the azeey/debug_remove_sytems branch and ran the `Tugbot in Warehouse` example at a slow update rate:
```
gz sim -z 10 "https://fuel.gazebosim.org/1.0/OpenRobotics/worlds/Tugbot in Warehouse"
```

Before removing the tugbot, here's the print out (repeated) from the PosePublisher. The format is `<thread id>: <entity>`. There are 6 PosePublishers each attached to a unique entity.
```
140267759843008:166
140267776628416:148
140267751450304:175
140267768235712:157
140268139423424:16
140267743057600:216

```

I then removed the tugbot and this is the print out right before crashing:

```
140267759843008:166
140267768235712:157
140267751450304:175
140267776628416:216
140267743057600:216
```
You can see that thread `140267776628416` is now associated with entity 216 instead of 148. This tells me there's some memory corruption happening.

src/SystemManager.hh
```diff
@@ -41,6 +44,21 @@ namespace gz
     // Inline bracket to help doxygen filtering.
     inline namespace GZ_SIM_VERSION_NAMESPACE {
 
+    /// \brief Helper container to keep track of
+    /// system interfaces and their parents
+    template<typename IFace>
+    struct SystemIfaceWithParent {
```
azeey, line 50: Have you considered using a std::unordered_map to associate the parent entity and the system instead of a new data structure to hold the two pieces of information. A new data structure is usually the right thing to do, but this has added a lot of complexity with the changes in this file and the implementation of SystemContainer which can all be avoided by using a map.

src/systems/pose_publisher/PosePublisher.cc
arjo129: Thanks for catching this I'm investigating.

src/SystemManager.hh
arjo129: I'm trying to recall why I didn't do that. My guess was to minimize changing other data structures and the fact that `Iface*` itself is not hashable. I should probably try an `unordered_map`, but won't be able to work on this till next week.
With regards to `SystemContainer`, I initially implemented it because I needed to prevent interface pointers from changing. I believe that, if I were to insert a new item to an `unordered_map`, it might invalidate all references.

src/systems/pose_publisher/PosePublisher.cc
arjo129: 883a0aba5 solves this.ll

src/SimulationRunner.cc
```diff
@@ -567,10 +570,19 @@ void SimulationRunner::ProcessSystemQueue()
...
         {
-          system->PostUpdate(this->currentInfo, this->entityCompMgr);
+          auto terminate = this->threadsToTerminate.find(parentEntity);
+          if (terminate != this->threadsToTerminate.end()) {
+            gzdbg << "Terminating thread " << id << ", " << parentEntity <<"\n";
+            this->postUpdateStartBarrier->Drop();
+            this->postUpdateStopBarrier->Drop();
+            break;
+          }
+          this_thread_system->PostUpdate(this->currentInfo, this->entityCompMgr);
```
scpeters, line 580: the linter is complaining that this line is over 80 characters

nice job finding the bug!

src/SimulationRunner.cc
```diff
@@ -559,6 +560,8 @@ void SimulationRunner::ProcessSystemQueue()
 
     this->postUpdateThreads.push_back(std::thread([&, id]()
     {
+      auto parentEntity = system.parent;
+      auto this_thread_system = system.system;
```
scpeters, line 564: I think (but I'm not 100% sure) another option is to change the ranged for loop starting on line 557 to create the `system` variable as a copy instead of a reference by deleting `&` from that line, but this approach works. I'll be ready to approve once the linter is happy

arjo129: Yet another way would be to stop all systems upon removal, just like we do when we spawn systems. This would reduce a significant amount of code. I'll hack it out this week.

src/SimulationRunner.cc
```diff
@@ -567,10 +570,20 @@ void SimulationRunner::ProcessSystemQueue()
...
         {
-          system->PostUpdate(this->currentInfo, this->entityCompMgr);
+          auto terminate = this->threadsToTerminate.find(parentEntity);
+          if (terminate != this->threadsToTerminate.end()) {
+            gzdbg << "Terminating thread " << id << ", " << parentEntity <<"\n";
+            this->postUpdateStartBarrier->Drop();
+            this->postUpdateStopBarrier->Drop();
+            break;
+          }
+          thisThreadSystem->PostUpdate(this->currentInfo,
```
scpeters, line 580: I ran the test on macOS and reproduced a seg-fault at this line, while the main body of `INTEGRATION_reset_sensors` is at [reset_sensors.cc:305](https://github.com/gazebosim/gz-sim/blob/arjo/feat/remove_systems_with_entities/test/integration/reset_sensors.cc#L305)

I haven't tried to reproduce on 24.04 yet

arjo129: Thanks for spotting this. I'm currently sleuthing around.

src/SystemManager.hh
arjo129: Went for a middle ground. Sorry I had to force push.

src/SimulationRunner.cc
arjo129: Fixed. Sorry I had to force push. Just re-worked the whole PR. Its now much cleaner.

arjo129:
I've reworked and simplified things significantly. We don't use barriers any more. Instead just like during addition, during removal we simply destroy all threads in the `PostUpdate` thread pool. The end result is a lot less code.

arjo129:
Not sure why I'm getting do many dartsim related errors:
```
libgz-physics-dartsim-plugin.so: undefined symbol: _ZN3sdf3v158internal17SdfScopeDelimiterB5cxx11Ev
```

scpeters:
> Not sure why I'm getting do many dartsim related errors. Can't reproduce locally:
> 
> ```
> libgz-physics-dartsim-plugin.so: undefined symbol: _ZN3sdf3v158internal17SdfScopeDelimiterB5cxx11Ev
> ```

I think that's my fault; I rebuilt the sdformat9 nightlies after removing some deprecated symbols but didn't rebuild gz-physics; I'll take a look

scpeters (commented):
yes, this is much simpler! CI looks happy now. I have a few suggestions about comments and variable names and one suggestion about moving a line of code

src/SystemManager.hh
```diff
@@ -187,24 +216,29 @@ namespace gz
...
 
       /// \brief Systems implementing Update
-      private: std::vector<ISystemUpdate *> systemsUpdate;
+      private: std::vector<SystemIfaceWithParent<ISystemUpdate>> systemsUpdate;
 
       /// \brief Systems implementing PostUpdate
       private: std::vector<ISystemPostUpdate *> systemsPostupdate;
 
+      /// \brief Parents of post update system.
+      private: std::vector<Entity> postUpdateParent;
```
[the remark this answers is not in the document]
arjo129: 6542accbf

src/SystemManager.cc
```diff
@@ -409,3 +428,91 @@ void SystemManager::ProcessPendingEntitySystems()
...
+    if(_ecm.IsMarkedForRemoval(postUpdateParent[i]))
+    {
+      markedForRemoval.insert(this->systemsPostupdate[i]);
+    }
+  }
+
+  RemoveFromVectorIf(this->systemsPostupdate,
+    [&](const auto system) {
+      // If system with a PostUpdate is marked for removal, mark its thread for
+      // termination.
```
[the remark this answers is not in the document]
arjo129: a49d638e5

src/SimulationRunner.cc
```diff
@@ -567,10 +571,13 @@ void SimulationRunner::ProcessSystemQueue()
         this->postUpdateStartBarrier->Wait();
         if (this->postUpdateThreadsRunning)
         {
-          system->PostUpdate(this->currentInfo, this->entityCompMgr);
+          system->PostUpdate(this->currentInfo,
+            this->entityCompMgr);
         }
         this->postUpdateStopBarrier->Wait();
+
```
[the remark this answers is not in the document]
arjo129: 376247597

src/SystemManager.hh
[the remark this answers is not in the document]
arjo129: 6542accbf

src/SystemManager.cc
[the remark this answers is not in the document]
arjo129: a49d638e5

src/SimulationRunner.cc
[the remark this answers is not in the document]
arjo129: 376247597

src/SystemManager.cc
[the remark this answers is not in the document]
arjo129: a49d638e5

src/SimulationRunner.cc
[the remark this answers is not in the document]
arjo129: 376247597

src/SimulationRunner.hh
```diff
@@ -543,6 +543,9 @@ namespace gz
       /// at the appropriate time.
       private: std::unique_ptr<msgs::WorldControlState> newWorldControlState;
 
+      /// \brief Set if we need to remove systems due to entity removal
+      private: bool needsCleanUp;
```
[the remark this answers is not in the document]
arjo129: 67d138034

src/SimulationRunner.cc
```diff
@@ -533,12 +534,15 @@ void SimulationRunner::ProcessSystemQueue()
 {
   auto pending = this->systemMgr->PendingCount();
 
-  if (0 == pending)
+  if (0 == pending && !this->needsCleanUp)
     return;
 
-  // If additional systems are to be added, stop the existing threads.
+  // If additional systems are to be added or removed, stop the existing
+  // threads.
```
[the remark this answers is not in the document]
arjo129: 376247597

src/SystemManager.cc
```diff
@@ -409,3 +428,91 @@ void SystemManager::ProcessPendingEntitySystems()
...
+    [&](const SystemIfaceWithParent<ISystemUpdate>& system) {
+      return _ecm.IsMarkedForRemoval(system.parent);
+    });
+
+  std::unordered_set<ISystemPostUpdate *> markedForRemoval;
+  for (std::size_t i = 0; i < this->systemsPostupdate.size(); i++)
+  {
+    if(_ecm.IsMarkedForRemoval(postUpdateParent[i]))
+    {
+      markedForRemoval.insert(this->systemsPostupdate[i]);
```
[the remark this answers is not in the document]
arjo129: a49d638e5

src/SystemManager.cc
```diff
@@ -409,3 +428,91 @@ void SystemManager::ProcessPendingEntitySystems()
...
+  // Note: This function has  O(n) time when an entity is removed
+  // where n is number of systems. Ideally we would only iterate
+  // over entities marked for removal but that would involve having
+  // a key value map. This can be marked as a future improvement.
+  if (!_ecm.HasEntitiesMarkedForRemoval())
+  {
+    return;
+  }
+
+  _needsCleanUp = true;
```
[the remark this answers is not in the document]
arjo129: a49d638e5

src/SystemManager.hh
```diff
@@ -145,6 +168,12 @@ namespace gz
       /// \brief Process system messages and add systems to entities
       public: void ProcessPendingEntitySystems();
 
+      /// \brief Remove systems that are attached to removed entities
+      /// \param[in] _entityCompMgr - ECM with entities marked for removal
```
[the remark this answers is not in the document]
arjo129: ca92ff873

arjo129:
Thanks @scpeters, I've addressed your feedback. Would like the :heavy_check_mark: from @azeey as well, then we can go ahead and merge this. Thanks for everyone's help on this!

scpeters (approved):

azeey:
See #2473

azeey:
> I haven't had a chance to look into this deeply, but one thing to keep is how this interacts with the levels feature. If a model that contains a system is unloaded by the levels feature and gets loaded back again, what would be the behavior? Would the system get configured and start anew? If that works properly, that would cover most cases, but we might also want the ability for a system to stay loaded even when its parent entity is removed so that if the parent entity is loaded back in, the system might continue from where it left off.

I've confirmed that when a level that contains a system get unloaded and loaded back again, the associated systems are also unloaded and loaded properly 🎉 . Regarding the ability for a system to stay loaded and continue where it left off, I propose we open a feature request and not address it right now since it's probably not going to be widely used.

Since there are a number of bug reports around this issue and since levels will work fine with this, I'm in favor of backporting this to Harmonic and earlier.

azeey (approved):
Thanks for this much needed feature/bug fix and for your patience in iterating!

landersson:
Thanks everyone for your efforts, I will be making use of this ASAP.

arjo129:
Thanks everyone for your guidance! Seems like out homebrew build is blocking the merge.

scpeters:
> Thanks everyone for your guidance! Seems like our homebrew build is blocking the merge.

I can reproduce the test failures on my macOS machine. I'm investigating, but it's a bit strange.

The test failures may have been introduced by https://github.com/gazebosim/gz-sim/pull/2473 or https://github.com/gazebosim/gz-sim/pull/2475, since macOS CI did not run for either of those pull requests.

scpeters:
> > Thanks everyone for your guidance! Seems like our homebrew build is blocking the merge.
> 
> I can reproduce the test failures on my macOS machine. I'm investigating, but it's a bit strange.
> 
> The test failures may have been introduced by #2473 or #2475, since macOS CI did not run for either of those pull requests.

* I can reproduce the seg-faults when building normally and running tests with `ctest`
* I tried reverting https://github.com/gazebosim/gz-sim/pull/2232/commits/7026b5714f0f88764ad058920f766c07096f0fb7, and it has no effect
* The tests pass when run with `lldb`
* The tests pass when recompiled with `-DCMAKE_BUILD_TYPE=Debug`

scpeters:
actually the tests have been failing on `main` since https://build.osrfoundation.org/job/gz_sim-ci-main-homebrew-amd64/68/

scpeters:
rerunning homebrew CI now that https://github.com/gazebosim/gz-math/pull/606 and https://github.com/gazebosim/sdformat/pull/1458 have been reverted after seeing that the reverts fixed tests in https://github.com/gazebosim/gz-sim/pull/2482

PR #2311, opened 2024-02-08
Use message generation in downstream bazel builds

mjcarroll:


azeey:
@mjcarroll is this still an active PR?

azeey:
@mjcarroll can we close this?

PR #2315, opened 2024-02-15
Ackermann steering plugin's steering only mode shows opposite behaviour. Additionally added a gz PID for steering

### LINKED ISSUE #2314: Ackermann Steering plugin's steering only mode shows opposite behaviour than expected
## Environment
* OS Version: Ubuntu 22.04
* Source build of `gz-sim8`
  
## Description
Issue pertains to Ackermann Steering plugin's `steering_only` mode specifically [these lines](https://github.com/gazebosim/gz-sim/blob/633ce72171e27e83bf2e0292c9998e036d5da3fc/src/systems/ackermann_steering/AckermannSteering.cc#L969-L976).

* Expected behavior: The inner wheel should turn more than the outer wheel during turning
* Actual behavior: The opposite takes place. (Here I have commented out the chassis visual in the SDF for better visibility) 

Screenshot from 2024-02-15 14-27-11

## Steps to reproduce

1. Add the following tag in the [example SDF file](https://github.com/gazebosim/gz-sim/blob/gz-sim8/examples/worlds/ackermann_steering.sdf) of Ackermann Steering plugin.
```xml
  <steering_only>True</steering_only>
```
2. Build and source the workspace
3. Launch using the following command
```bash
  gz sim ackermann_steering.sdf
```
4. Publish the following
```bash
  gz topic -t "/model/vehicle_blue/steer_angle" -m gz.msgs.Double -p "data: 0.4"
```

Additionally comment out [chassis visual](https://github.com/gazebosim/gz-sim/blob/633ce72171e27e83bf2e0292c9998e036d5da3fc/examples/worlds/ackermann_steering.sdf#L102-L113) for better visibility

## Solution
Changing signs in [these equations](https://github.com/gazebosim/gz-sim/blob/633ce72171e27e83bf2e0292c9998e036d5da3fc/src/systems/ackermann_steering/AckermannSteering.cc#L969-L976) should suffice.

P.S. Raising a PR regarding the same

sauk2:
# 🦟 Bug fix

Fixes #2314

## Summary

Issue pertains to Ackermann Steering plugin's `steering_only` mode specifically [these lines](https://github.com/gazebosim/gz-sim/blob/633ce72171e27e83bf2e0292c9998e036d5da3fc/src/systems/ackermann_steering/AckermannSteering.cc#L969-L976).

* Expected behavior: The inner wheel should turn more than the outer wheel during turning
* Actual behavior: The opposite takes place. (Here I have commented out the chassis visual in the SDF for better visibility) 

Screenshot from 2024-02-15 14-27-11

### Steps to reproduce

1. Add the following tag in the [example SDF file](https://github.com/gazebosim/gz-sim/blob/gz-sim8/examples/worlds/ackermann_steering.sdf) of Ackermann Steering plugin.
```xml
  <steering_only>True</steering_only>
```
2. Build and source the workspace
3. Launch using the following command
```bash
  gz sim ackermann_steering.sdf
```
4. Publish the following
```bash
  gz topic -t "/model/vehicle_blue/steer_angle" -m gz.msgs.Double -p "data: 0.4"
```

Additionally comment out [chassis visual](https://github.com/gazebosim/gz-sim/blob/633ce72171e27e83bf2e0292c9998e036d5da3fc/examples/worlds/ackermann_steering.sdf#L102-L113) for better visibility

### Solution

Changing signs in [these equations](https://github.com/gazebosim/gz-sim/blob/633ce72171e27e83bf2e0292c9998e036d5da3fc/src/systems/ackermann_steering/AckermannSteering.cc#L969-L976) should suffice.

Check output after implementation
Screenshot from 2024-02-15 14-51-32

## Checklist
- [ ] Added tests

# 🎉 New feature

## Summary
* Added a programmable PID (`gz::math::PID`) to control the steering joints for both modes - velocity and steering only.
* Default gains and other parameters are defined in `Configure` which can be overwritten from the SDF using appropriate tags.
* A single PID is used to control both, left and right steering joints.

### For steering only mode

1. Add the following tag in the [example SDF file](https://github.com/gazebosim/gz-sim/blob/gz-sim8/examples/worlds/ackermann_steering.sdf) of Ackermann Steering plugin.
```xml
  <steering_only>True</steering_only>
```
2. Build and source the workspace
3. Launch using the following command
```bash
  gz sim ackermann_steering.sdf
```
4. Publish the following
```bash
  gz topic -t "/model/vehicle_blue/steer_angle" -m gz.msgs.Double -p "data: 0.4"
```
### For velocity mode
1. Build and source the workspace
2. Launch using the following command
```bash
  gz sim ackermann_steering.sdf
```
3. Publish the following
```bash
  gz topic -t "/model/vehicle_blue/cmd_vel" -m gz.msgs.Twist -p "linear: {x: 1.0}, angular: {z: 0.5}"
```
## Checklist
- [ ] Added tests


retinfai:
A little tangent but somewhat related. Is it possible with the current plugin to be able to drive the vehicle using steering angle and forward velocity?

If not, is it something that can be implemented within the scope of this plugin?

src/systems/ackermann_steering/AckermannSteering.cc
```diff
@@ -331,11 +331,40 @@ void AckermannSteering::Configure(const Entity &_entity,
...
+  if(_sdf->HasElement("steer_i_min"))
+    steerPosIMin = _sdf->Get<double>("steer_i_min");
+  if(_sdf->HasElement("steer_cmd_max"))
+    steerPosCmdMax = _sdf->Get<double>("steer_cmd_max");
+  if(_sdf->HasElement("steer_cmd_min"))
+    steerPosCmdMin = _sdf->Get<double>("steer_cmd_min");
+  if(_sdf->HasElement("steer_cmd_offset"))
+    steerPosCmdOffset = _sdf->Get<double>("steer_cmd_offset");
+  this->dataPtr->steerPosPid.Init(
+    steerPosPGain,
```
bperseghetti, line 360:
```suggestion
    steerPosPGain,
```

bperseghetti (commented):
Good catch, it looks like there is a trailing whitespace that needs fixed to pass CI.

bperseghetti:
> A little tangent but somewhat related. Is it possible with the current plugin to be able to drive the vehicle using steering angle and forward velocity?
> 
> If not, is it something that can be implemented within the scope of this plugin?

Hmmmm, I think that's more the role of JointController,
Here's how I use them in a model together: https://github.com/CogniPilot/b3rb_simulator/blob/31554eb3126d6ee306a295b2f9f5e8524907b940/b3rb_gz_resource/models/b3rb/model.sdf#L540-L562

Note that JointController uses the velocity rad/s field when using actuator_msg and steering_only in Ackermann with actuator_msg uses the position (rad).

sauk2:
@bperseghetti, on a completely unrelated note. Would it be a good idea to calculate wheel separation in case it is not provided by the user? 

We can use the world pose of the left and right steering joints and calculate that value. This calculation will not be performed if there is a value given in the SDF. 

Would like to hear your thoughts on this

bperseghetti:
> @bperseghetti, on a completely unrelated note. Would it be a good idea to calculate wheel separation in case it is not provided by the user?
> 
> We can use the world pose of the left and right steering joints and calculate that value. This calculation will not be performed if there is a value given in the SDF.
> 
> Would like to hear your thoughts on this

Hmmmm, I think that might be risky depending on the intended use case for it. IE using 4 wheel steering (like you see on a good amount of newer electric trucks). I think maybe you could calculate it though and then see if it does not match the provided value within a "reasonable error percentage (IE 15%) that it publishes a gz-warn to let you know what value it thinks it should be.

But depending on things like suspension etc the steering joint might actually be a relative reference to another joint and then that would be somewhat messy to follow that potential dependency tree. 

IE having joints for a suspension such as macpherson, wishbone, etc would make the steering joint a relative joint to those other suspension joints.

Therefore, my personal take on it might lean toward it "probably not being a good choice"?

bperseghetti:
@sauk2 if you merge this it should fix the code coverage issue and cleanly separate the PID steering_only SDF elements: 
https://github.com/sauk2/gz-sim/pull/1

sauk2:
> Hmmmm, I think that might be risky depending on the intended use case for it. IE using 4 wheel steering (like you see on a good amount of newer electric trucks). I think maybe you could calculate it though and then see if it does not match the provided value within a "reasonable error percentage" (IE 15%) that it publishes a gz-warn to let you know what value it thinks it should be.
> 
> But depending on things like suspension etc the steering joint might actually be a relative reference to another joint and then that would be somewhat messy to follow that potential dependency tree.
> 
> IE having joints for a suspension such as macpherson, wishbone, etc would make the steering joint a relative joint to those other suspension joints and that just starts to possibly become more risk than reward.
> 
> Therefore, my personal take on it might lean toward it "probably not being a good choice"?

Yes, that makes sense. Thanks!

bperseghetti (approved):

src/systems/ackermann_steering/AckermannSteering.cc
```diff
@@ -923,14 +955,17 @@ void AckermannSteeringPrivate::UpdateVelocity(
...
 
-  // Simple proportional control with a gain of 1
-  // Adding programmable PID values might be a future feature.
-  // Works as is for tested cases
-  this->leftSteeringJointSpeed = leftDelta;
-  this->rightSteeringJointSpeed = rightDelta;
+  // Simple PID control with settable gains.
+  this->leftSteeringJointSpeed = this->steerPosPid.Update(
+    leftDelta, _info.dt);
+  this->steerPosPid.Reset();
```
azeey, line 964: Is there a reason for resetting `steerPosPid` here? If I'm not mistaken, if we reset it, the `D` and `I` terms of the PID will not do anything.

sauk2: I was resetting the `steerPosPid` because I was using the same PID for both steering joints and it was causing random motions. I didn't realise it would render the `I` and `D` terms useless. 

After giving it more thought, I think this will be solved if I create 2 PID controllers in `Configure` - one for each steering joint.

Would like to hear your thoughts on this.

bperseghetti: I think we should get the fix in for the steering angle, but I don't think we should necessarily couple that with the PID changes. I think these should be separate PRs.

sauk2: Makes sense. So all the changes will have to be rolled back. The easiest way I can see is to reset this branch to commit id 633ce72171e27e83bf2e0292c9998e036d5da3fc and then resolve the bug. 

Should I go ahead with this or is there any alternate method that you have thought of?

bperseghetti: I guess I would just pull latest commit on branch and then make the fix ontop of it: https://github.com/gazebosim/gz-sim/commit/6cfcd07c346a72546963ed56ade48daa828c2072

bperseghetti: And then push your current branch to another branch like ackermann-pid so you could reference it later, and push the changes ontop the other to this current branch so you don't lose history in this PR.

bperseghetti:
> A little tangent but somewhat related. Is it possible with the current plugin to be able to drive the vehicle using steering angle and forward velocity?
> 
> If not, is it something that can be implemented within the scope of this plugin?

Yes, for that you just set wheel velocity with rad/s using the JointController plugin and also the steering angle all using the actuator msg, example is here:

https://github.com/CogniPilot/b3rb_simulator/blob/7830685f9b9c3aca2105ee9e149969d2db696350/b3rb_gz_resource/models/b3rb/model.sdf#L1090-L1112

sauk2:
@bperseghetti Created a seperate PR #2342 for the steering error fix! Request you to take a look.

PR #2326, opened 2024-03-01
Add entity validation to OdometryPublisher

Bogdanov-am:
# 🦟 Bug fix

Until SystemManager has the ability to unload system plugins, plugins require an explicit check of the validity of the entities used in the Update methods. Such a check was missing in OdometryPublisher, which led to non-critical but annoying errors in the console.

P.S. It's my first PR to open source project, sorry if i missing something

### DIFF
--- a/src/systems/odometry_publisher/OdometryPublisher.cc
+++ b/src/systems/odometry_publisher/OdometryPublisher.cc
@@ -316,7 +316,6 @@ void OdometryPublisher::PostUpdate(const UpdateInfo &_info,
     const EntityComponentManager &_ecm)
 {
   GZ_PROFILE("OdometryPublisher::PostUpdate");
-  gzwarn << "HElllo Anton" << std::endl;
 
   // \TODO(anyone) This is a temporary fix for
   // gazebosim/gz-sim#2165 until gazebosim/gz-sim#2217 is resolved.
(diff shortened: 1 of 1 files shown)


src/systems/odometry_publisher/OdometryPublisher.cc
```diff
@@ -278,6 +278,10 @@ void OdometryPublisher::PreUpdate(const gz::sim::UpdateInfo &_info,
 {
   GZ_PROFILE("OdometryPublisher::PreUpdate");
 
+  if (!this->dataPtr->model.Valid(_ecm)) {
+    return;
```
azeey, line 282: What do you think about warning the user once that the OdometryPublisher has stopped? (see https://github.com/gazebosim/gz-sim/blob/633ce72171e27e83bf2e0292c9998e036d5da3fc/src/gui/plugins/environment_loader/EnvironmentLoader.cc#L153-L159)

azeey (commented):
Thanks for the contribution! And welcome to open source development. I just have one comment.

src/systems/odometry_publisher/OdometryPublisher.cc
Bogdanov-am: Yep, I copied solution from JointPositionController.

azeey (approved):
Thanks! If you're interested in contributing more, which we'd love, please take a look at these [good first issues](https://github.com/search?q=org%3Agazebosim+is%3Aissue++label%3A%22good+first+issue%22++no%3Aassignee+&type=issues&state=open)

g-arjones:
@azeey Any idea when this would be released?

azeey:
We'll shoot for making a release this week.

g-arjones:
Awesome. Thanks for the heads up :+1:

PR #2336, opened 2024-03-17
Add magnetometer plugin parameter to publish units as tesla

### LINKED ISSUE #2312: Magnetometer system publishes field strength in incorrect units
## Environment
* OS Version:
```
$ lsb_release -a                 
LSB Version:	core-11.1.0ubuntu4-noarch:security-11.1.0ubuntu4-noarch
Distributor ID:	Ubuntu
Description:	Ubuntu 22.04.3 LTS
Release:	22.04
Codename:	jammy
---
$ uname -a                
Linux <> 6.5.0-17-generic #17~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Tue Jan 16 14:32:32 UTC 2 x86_64 x86_64 x86_64 GNU/Linux
```
* Source or binary build?
```
Binary
---
$ gz sim --version                                         
Gazebo Sim, version 8.0.0
Copyright (C) 2018 Open Source Robotics Foundation.
Released under the Apache 2.0 License.
---
$ ls /usr/include/gz
cmake3   fuel_tools9  launch7  msgs10    plugin2     sdformat14  sim8         utils2
common5  gui8         math7    physics7  rendering8  sensors8    transport13
```

### Expected behavior
According to the `gz::msgs::Magnetometer` [proto message](https://github.com/gazebosim/gz-msgs/blob/gz-msgs10/proto/gz/msgs/magnetometer.proto#L36), the `field_tesla` is reported in units of Tesla. 

However, consider the following minimal world file (see attached) using the Bay Area, California as the origin for the `spherical_coordinates`.

Using this website, I would expect the local magnetic field to be something like this:

| Horizontal Intensity | North Comp | East Comp | Vertical Comp | Total Field |
| --- | --- | --- | --- | --- |
| 23,071.9 nT | 22,485.8 nT | 5,167.1 nT | 41,506.4 nT | 47,487.8 nT |
| 2.3072×10^-5 T | 2.2486×10^-5 T | 5.1671×10^-6 T | 4.15064×10^-5 T | 4.74878×10^-5 T |

However, when I start up the simulation and echo the `/magnetometer` topic
```
gz sim simple_mag.world
gz topic -e -t /magnetometer
```
I get the following:
```
header {
  stamp {
    sec: 67
    nsec: 230000000
  }
  data {
    key: "frame_id"
    value: "sensors_box::link::magnetometer"
  }
  data {
    key: "seq"
    value: "86"
  }
}
field_tesla {
  x: -0.32109555007132251
  y: -0.14981328784703124
  z: 0.40261160054926087
}
```

which is a few orders of magnitude off from what I was expecting. 

I did some digging, and it looks like the tables that were used to estimate the field vector were added [to gz-sim in this PR](https://github.com/gazebosim/gz-sim/pull/1907) which actually use the field strength table from [the old PX4 Sim](https://github.com/gazebosim/gz-sim/pull/1907#issuecomment-1443844856).

However, I believe [this line](https://github.com/gazebosim/gz-sim/blob/633ce72171e27e83bf2e0292c9998e036d5da3fc/src/systems/magnetometer/Magnetometer.cc#L450) is incorrectly calculating the magnetic field in Tesla as I would expect the field to be measured in. Instead, it calculates it as a [gauss](https://en.wikipedia.org/wiki/Gauss_(unit)) and never converts it back to teslas.

I can demonstrate that this is the case using the previous message.

```
field_tesla {
  x: -0.32109555007132251
  y: -0.14981328784703124
  z: 0.40261160054926087 <--- Assume we are looking at the Z component only for this
}
---
Assume that it is actually gauss
0.40261160054926087 g = 4.0261160054926087×10^-5 T
4.0261160054926087×10^-5 T = 40261.160054926087 nT
40,261.2 nT ~= 41,506.4 nT (vertical component from NOAA website)
```

## Steps to reproduce
1. gz sim simple_mag.world
2. gz topic -e -t /magnetometer
3. Observe a *very* high magnetic field strength (assuming `field_tesla` is in Teslas)

Unless I am mistaken, I think this is the case. My recommendation would be to fix the conversion in `systems/magnetometer/Magnetometer.cc` since we assume that the noise in the SDF file is [measured in teslas](https://github.com/gazebosim/gz-sensors/blob/gz-sensors8/src/MagnetometerSensor.cc#L120). I think people would be okay with reporting it in tesla, not nT, even though that is what most sensors are reporting.

As an aside, I'm actually surprised this seems like it was incorrect all the way back to the PX4 days. Was it originally reporting the field as Gauss? It isn't clear [from the message signature](https://github.com/PX4/PX4-SITL_gazebo-classic/blob/main/msgs/MagneticField.proto#L8)

I can put up a PR if you'd all like and agree with the approach. Please let me know if I missed anything.

gilborty:
# 🦟 Bug fix

Fixes #2312

## Summary
As mentioned in #2312, the magnetic field message is incorrectly using Gauss as the units for the field strength, rather than the SI unit of teslas. 

This PR adds a flag, `use_tesla_for_magnetic_field`, to the `MagnetometerSystem` to enable this fix as to not break any existing functionality that assumes Gauss. It also adds the `Configure` step to the `MagnetometerSystem`.

While I was here, I noticed that the integration test doesn't actually update the `MagnetometerSystem` because the SphericalCoorindates are not set. I think this might be an oversight. I did not change that existing test in case it was like that on purpose.

## Checklist
- [x] Added tests

### DIFF
--- a/test/integration/magnetometer_system.cc
+++ b/test/integration/magnetometer_system.cc
@@ -141,3 +141,152 @@ TEST_F(MagnetometerTest, GZ_UTILS_TEST_DISABLED_ON_WIN32(RotatedMagnetometer))
       field.Z(), TOL);
   mutex.unlock();
 }
+
+
+/////////////////////////////////////////////////
+// Test to check to ensure that the fix for using tesla units works correctly
+// Note(gilbert): It seems the above test (RotatedMagnetometer)
+// does not actually run because the spherical coordinates are not set.
+// In order to not break the above tests, we create a new world and run
+// two simulations, one with gauss and the other with tesla units.
+// See https://github.com/gazebosim/gz-sim/issues/2312
+
+class MagnetometerMessageBuffer {
+public:
+  void add(const msgs::Magnetometer &_msg)
+  {
+    std::lock_guard<std::mutex> lock(mutex);
+    magnetometerMsgs.push_back(_msg);
+  }
+
+  msgs::Magnetometer at(const size_t index) const
+  {
+    std::lock_guard<std::mutex> lock(mutex);
+    return magnetometerMsgs.at(index);
+  }
+
+  msgs::Magnetometer latest() const
+  {
+    std::lock_guard<std::mutex> lock(mutex);
+    return magnetometerMsgs.back();
+  }
+
+  size_t size() const
+  {
+    std::lock_guard<std::mutex> lock(mutex);
+    return magnetometerMsgs.size();
+  }
+
+private:
+  mutable std::mutex mutex;
+  std::vector<msgs::Magnetometer> magnetometerMsgs;
+};
+
+MagnetometerMessageBuffer gaussMessages;
+MagnetometerMessageBuffer teslaMessages;
+
+/////////////////////////////////////////////////
+void gaussMagnetometerCb(const msgs::Magnetometer &_msg)
+{
+  gaussMessages.add(_msg);
+}
+
+/////////////////////////////////////////////////
+void teslaMagnetometerCb(const msgs::Magnetometer &_msg)
+{
+  teslaMessages.add(_msg);
+}
+
+TEST_F(MagnetometerTest,
+  GZ_UTILS_TEST_DISABLED_ON_WIN32(GaussToTeslaCorrection))
+{
+  // First, run the world without the fix enabled and record the magnetic field.
+  ServerConfig serverConfig;
+  const auto sdfFile = std::string(PROJECT_SOURCE_PATH) +
+    "/test/worlds/magnetometer_with_tesla.sdf";
+  serverConfig.SetSdfFile(sdfFile);
+
+  Server gaussServer(serverConfig);
+  EXPECT_FALSE(gaussServer.Running());
+  EXPECT_FALSE(*gaussServer.Running(0));
+
+  const auto topic = "world/magnetometer_sensor/model/magnetometer_model/"
+      "link/link/sensor/magnetometer_sensor/magnetometer";
+
+  // subscribe to magnetometer topic
+  std::unique_ptr<transport::Node> gaussNode =
+    std::make_unique<transport::Node>();
+  gaussNode->Subscribe(topic, &gaussMagnetometerCb);
+
+  // step world and verify magnetometer's detected field
+  // Run the gauss server
+  const size_t iters = 100u;
+  gaussServer.Run(true, iters, false);
+
+  // Now, we need to create a new server where the magnetometer
+  // is correctly publishing using tesla.
+  // Flip the use_tesla_for_magnetic_field field to true
+  sdf::SDFPtr teslaWorldSdf(new sdf::SDF());
+  sdf::init(teslaWorldSdf);
+  ASSERT_TRUE(sdf::readFile(sdfFile, teslaWorldSdf));
+  sdf::ElementPtr root = teslaWorldSdf->Root();
+  EXPECT_TRUE(root->HasElement("world"));
+
+  sdf::ElementPtr world = root->GetElement("world");
+  const std::string pluginName = "gz::sim::systems::Magnetometer";
+  const std::string useTeslaElementName = "use_tesla_for_magnetic_field";
+  bool elementSet = false;
+  for (sdf::ElementPtr plugin = world->GetElement("plugin");
+       plugin;
+       plugin = plugin->GetNextElement("plugin")) {
+        if (plugin->Get<std::string>("name") == pluginName) {
+            // Found the magnetometer plugin, now force the field to true
+            if (plugin->HasElement(useTeslaElementName)) {
+                sdf::ElementPtr teslaElement =
+                  plugin->GetElement(useTeslaElementName);
+                teslaElement->Set(true);
+                elementSet = true;
+                break;
+            }
+        }
+    }
+  EXPECT_TRUE(elementSet);
+
+  // Now, re-run the world but with the tesla units being published
+  EXPECT_TRUE(serverConfig.SetSdfString(teslaWorldSdf->ToString()));
+
+  Server teslaServer(serverConfig);
+  EXPECT_FALSE(teslaServer.Running());
+  EXPECT_FALSE(*teslaServer.Running(0));
+
+  // subscribe to tesla magnetometer topic and unsubscribe from the gauss topic
+  gaussNode = nullptr;
+  std::unique_ptr<transport::Node> teslaNode =
+    std::make_unique<transport::Node>();
+  teslaNode->Subscribe(topic, &teslaMagnetometerCb);
+
+  // step world and verify magnetometer's detected field
+  // Run the tesla server
+  teslaServer.Run(true, iters, false);
+
+  // Now compare the two sets of magnetic data
+  ASSERT_EQ(gaussMessages.size(), teslaMessages.size());
+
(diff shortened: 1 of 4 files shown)


test/integration/magnetometer_system.cc
```diff
@@ -141,3 +141,146 @@ TEST_F(MagnetometerTest, GZ_UTILS_TEST_DISABLED_ON_WIN32(RotatedMagnetometer))
...
+    magnetometerMsgs.push_back(_msg);
+  }
+
+  msgs::Magnetometer at(const size_t index) const
+  {
+    std::lock_guard<std::mutex> lock(mutex);
+    return magnetometerMsgs.at(index);
+  }
+
+  size_t size() const
```
ahcorde, line 168: include this at the top `#include <cstddef>`

test/integration/magnetometer_system.cc
```diff
@@ -141,3 +141,146 @@ TEST_F(MagnetometerTest, GZ_UTILS_TEST_DISABLED_ON_WIN32(RotatedMagnetometer))
...
+  {
+    std::lock_guard<std::mutex> lock(mutex);
+    return magnetometerMsgs.size();
+  }
+
+private:
+  mutable std::mutex mutex;
+  std::vector<msgs::Magnetometer> magnetometerMsgs;
```
ahcorde, line 176: include this at the top `#include <vector>`

test/integration/magnetometer_system.cc
```diff
@@ -141,3 +141,146 @@ TEST_F(MagnetometerTest, GZ_UTILS_TEST_DISABLED_ON_WIN32(RotatedMagnetometer))
...
+  }
+
+  size_t size() const
+  {
+    std::lock_guard<std::mutex> lock(mutex);
+    return magnetometerMsgs.size();
+  }
+
+private:
+  mutable std::mutex mutex;
```
ahcorde, line 175: include this at the top `#include <mutex>`

is it the `mutable` required?

test/integration/magnetometer_system.cc
```diff
@@ -141,3 +141,146 @@ TEST_F(MagnetometerTest, GZ_UTILS_TEST_DISABLED_ON_WIN32(RotatedMagnetometer))
...
+  teslaMessages.add(_msg);
+}
+
+TEST_F(MagnetometerTest,
+  GZ_UTILS_TEST_DISABLED_ON_WIN32(GaussToTeslaCorrection))
+{
+  // First, run the world without the fix enabled and record the magnetic field.
+  ServerConfig serverConfig;
+  const auto sdfFile = std::string(PROJECT_SOURCE_PATH) +
+    "/test/worlds/magnetometer_with_tesla.sdf";
```
ahcorde, line 200: please use `gz::common::joinPaths`

test/integration/magnetometer_system.cc
```diff
@@ -141,3 +141,146 @@ TEST_F(MagnetometerTest, GZ_UTILS_TEST_DISABLED_ON_WIN32(RotatedMagnetometer))
...
+
+  Server gaussServer(serverConfig);
+  EXPECT_FALSE(gaussServer.Running());
+  EXPECT_FALSE(*gaussServer.Running(0));
+
+  const auto topic = "world/magnetometer_sensor/model/magnetometer_model/"
+      "link/link/sensor/magnetometer_sensor/magnetometer";
+
+  // subscribe to magnetometer topic
+  std::unique_ptr<transport::Node> gaussNode =
```
ahcorde, line 211: `#include <memory>`

test/integration/magnetometer_system.cc
```diff
@@ -141,3 +141,146 @@ TEST_F(MagnetometerTest, GZ_UTILS_TEST_DISABLED_ON_WIN32(RotatedMagnetometer))
...
+            // Found the magnetometer plugin, now force the field to true
+            if (plugin->HasElement(useTeslaElementName)) {
+                sdf::ElementPtr teslaElement =
+                  plugin->GetElement(useTeslaElementName);
+                teslaElement->Set(true);
+                elementSet = true;
+                break;
+            }
+        }
+    }
```
ahcorde, line 246: indentation looks wierd

test/integration/magnetometer_system.cc
```diff
@@ -141,3 +141,146 @@ TEST_F(MagnetometerTest, GZ_UTILS_TEST_DISABLED_ON_WIN32(RotatedMagnetometer))
...
+  EXPECT_TRUE(root->HasElement("world"));
+
+  sdf::ElementPtr world = root->GetElement("world");
+  const std::string pluginName = "gz::sim::systems::Magnetometer";
+  const std::string useTeslaElementName = "use_tesla_for_magnetic_field";
+  bool elementSet = false;
+  for (sdf::ElementPtr plugin = world->GetElement("plugin");
+       plugin;
+       plugin = plugin->GetNextElement("plugin")) {
+        if (plugin->Get<std::string>("name") == pluginName) {
```
ahcorde, line 236:
```suggestion
  for (sdf::ElementPtr plugin = world->GetElement("plugin");
       plugin;
       plugin = plugin->GetNextElement("plugin"))
  {
     if (plugin->Get<std::string>("name") == pluginName) {
```

ahcorde (changes requested):

test/integration/magnetometer_system.cc
gilborty: You need to mark a mutex as mutable to allow it to be locked and unlocked in the const member functions.

test/integration/magnetometer_system.cc
gilborty: Is there an automatic code formatting script I am missing? I ran `make codecheck`, but do you all have pre-commit hook-esque scripts that do the formatting automatically?

test/integration/magnetometer_system.cc
gilborty: Great, thanks for the help. I updated the previous test as well.

gilborty:
Thank you for the review @ahcorde, I think I updated everything you asked for. Thank you for the helper function point out as well!

gilborty:
Are the `INTEGRATION_triggered_publisher` tests failing on CI for anyone else? Mine are passing locally, I don't think it is related to this commit.

```
[       OK ] TriggeredPublisherTest.StringMsgReqBooleanRep (1053 ms)
[----------] 23 tests from TriggeredPublisherTest (13137 ms total)

[----------] Global test environment tear-down
[==========] 23 tests from 1 test suite ran. (13137 ms total)
[  PASSED  ] 23 tests.


```

iche033:
with https://github.com/gazebosim/gz-sim/pull/2460, support for publishing units as tesla should now be available. It would be nice to add the integration test from this PR though.

azeey:
@gilborty thanks for the contribution! Per https://github.com/gazebosim/gz-sim/pull/2336#issuecomment-2274363375, would you be able to update this PR to only include the integration tests?

Roboticistprogrammer:
Hi, may I know the state of this PR? @ahcorde 
Looking at [this](https://github.com/PX4/PX4-gazebo-models/blob/main/models/x500_base/model.sdf) sdf model, I find:
>
          <magnetometer>
          
          
          
          
          <x>
            <noise type="gaussian">
              <stddev>0.0001</stddev>
            </noise>

I was wondering if there is a way around this in Harmonic, like can use plugins from garden for barometer and magnetometer thanks.

PR #2359, opened 2024-04-04
Prototype system for initializing model state

### LINKED ISSUE #2318: Proposal: system for setting initial position / velocity
## Desired behavior

I'd like to be able to specify initial model/link velocity and initial joint position/velocity. For joints, I believe this can be done by setting [JointPositionReset](https://github.com/gazebosim/gz-sim/blob/gz-sim8/include/gz/sim/components/JointPositionReset.hh) and [JointVelocityReset](https://github.com/gazebosim/gz-sim/blob/gz-sim8/include/gz/sim/components/JointVelocityReset.hh) components programmatically. It would be consistent with this pattern to add `LinearVelocityReset` / `AngularVelocityReset` and/or `WorldLinearVelocityReset` / `WorldAngularVelocityReset` components that could be used to set initial model / link velocities.

## Alternatives considered

I had started an alternative approach in https://github.com/gazebosim/gz-sim/pull/2265 based on using the existing `AngularVelocityCmd` and `LinearVelocityCmd` components with the `VelocityControl` system, but I think the `*Reset` components make more sense.

## Implementation suggestion

~~~
<model name="example_model">
  <link name="parent_link" />
  <link name="child_link" />
  <joint name="multiaxis_joint" type="universal">
    <axis>
      <xyz>0 0 1</xyz>
    </axis>
    <axis2>
      <xyz>1 0 0</xyz>
    </axis2>
  </joint>
  <plugin name="SetInitialComponents" ...>
    <!-- different syntax ideas to set initial velocity for this model -->
      <!-- specify //entity/@type="model" and //entity/@name="__model__" -->
      <entity type="model" name="__model__">
        <set_component_value type="WorldLinearVelocityReset">1.0 4.0 9.0</set_component_value>
      </entity>
      <!-- specify //model/@name="__model__" -->
      <model name="__model__">
        <set_component_value type="WorldLinearVelocityReset">1.0 4.0 9.0</set_component_value>
      </model>
      <!-- or just treat set_component_value tags at the root scope as referring to the model -->
      <set_component_value type="WorldLinearVelocityReset">1.0 4.0 9.0</set_component_value>
  
    <!-- different syntax ideas to set initial velocity for a link -->
      <!-- specify //entity/@type="link" and //entity/@name="parent_link" -->
      <entity type="link" name="parent_link">
        <set_component_value type="WorldLinearVelocityReset">1.0 4.0 9.0</set_component_value>
      </entity>
      <!-- specify //link/@name="parent_link" -->
      <link name="parent_link">
        <set_component_value type="WorldLinearVelocityReset">1.0 4.0 9.0</set_component_value>
      </link>

    <!-- different syntax ideas to set initial position / velocity for a joint -->
      <!-- specify //entity/@type="joint" and //entity/@name="multiaxis_joint" -->
      <entity type="joint" name="multiaxis_joint">
        <set_component_value type="JointPositionReset">0.1 0.4</set_component_value>
        <set_component_value type="JointVelocityReset">1.0 -0.5</set_component_value>
      </entity>
      <!-- specify //joint/@name="multiaxis_joint" -->
      <joint name="multiaxis_joint">
        <set_component_value type="JointPositionReset">0.1 0.4</set_component_value>
        <set_component_value type="JointVelocityReset">1.0 -0.5</set_component_value>
      </joint>
  
  </plugin>
~~~

Thoughts are welcome for the right syntax here. I think we would have to hard-code a relationship between the `//set_component_value/@type` string and a `gz::sim::components::` class since the component Factory doesn't allow constructing just from a `string`. It does have [namesById and runtimeNamesById](https://github.com/gazebosim/gz-sim/blob/gz-sim8/include/gz/sim/components/Factory.hh#L464-L470) data structures but not the inverse (`IdsByName`). Even if we did, I'm not sure how it could be implemented.

I suggested the system name of `SetInitialComponents` in this example to go along with the `<set_component_value>` elements, but we could also use names with more specific semantics (similar to the [InitialVelocityPlugin from gazebo-classic](https://github.com/gazebosim/gazebo-classic/blob/gazebo11/worlds/initial_velocity.world#L48-L51)).

scpeters:
# 🎉 New feature

Closes #2318

## Summary

This adds a system called `SetModelState` that accepts an xml configuration containing `<model_state>` tags intending to serve as a prototype for new syntax for the `<state>` tag in SDFormat 1.12. Currently, only `<joint_state>` tags are supported; support should also be added for `<link_state>`.

### Example `<model_state>` syntax from [examples/worlds/set_model_state.sdf](https://github.com/gazebosim/gz-sim/blob/b99b4d1f2d24d94ba3c6002997854c20630b19db/examples/worlds/set_model_state.sdf)

* Using degrees

~~~
        <model_state>
          <joint_state name="j1">
            <axis_state>
              <position degrees="true">60</position>
              <velocity degrees="true">-30</velocity>
            </axis_state>
          </joint_state>
        </model_state>
~~~

* Using radians

~~~
        <model_state>
          <joint_state name="j1">
            <axis_state>
              <position>1.047</position>
              <velocity>-0.523</velocity>
            </axis_state>
          </joint_state>
        </model_state>
~~~

## Test it

~~~
gz sim -v 4 examples/worlds/set_model_state.sdf
~~~

The rotors should start at a 60 degree angle and rotate clock-wise at 30 degrees / second.

## Checklist
- [ ] Added tests


scpeters:
For specifying link velocity, there has been a request to change the link_state specification for velocity to not use the `pose` type (see https://github.com/gazebosim/sdformat/issues/1366). As mentioned in that issue, an alternative is to specify velocity as two vector3 types since there is no `vector6` type in SDFormat and this avoids confusion about the order of the data.

~~~
<model_state>
  <link_state name="link1">
    <linear_velocity>{vx} {vy} {vz}</linear_velocity>
    <angular_velocity>{wx} {wy} {wz}</angular_velocity>
  </link_state>
</model_state>
~~~

scpeters:
> Can this be used to set the model state of a model in which no `gz-sim-set-model-state-system`, for example to permit to have a generic model for a robot, that is include in a `.world` and in which world-specific positions are specified?

Yes, I would like to push these syntax changes upstream to the the `<state>` specification in SDFormat 1.12. I'm treating this pull request and system as a prototype for those syntax changes.

scpeters:
One difference between this prototype and what we could achieve by updating the `//world/state` specification is that this plugin allows specifying the initial state of a model from within the `//model` element, whereas the `//world/state` element must be defined at the world level. I think it is convenient to set model state from within the model, so I would also propose adding a `//model/state` tag as well, which would add complexity and require defining an order or precedence, but I think it would be a valuable addition.

scpeters:
I've updated the world file to use SDFormat 1.11 and auto-inertials in https://github.com/gazebosim/gz-sim/pull/2359/commits/7cab1bb44ece6d20986c9832afee436b5c3a69d2

src/systems/set_model_state/SetModelState.cc
```diff
@@ -0,0 +1,323 @@
...
+#include "gz/sim/Model.hh"
+#include "gz/sim/Util.hh"
+
+using namespace gz;
+using namespace sim;
+using namespace systems;
+
+namespace
+{
+  bool parseScalarWithDegrees(math::Angle &_scalar, sdf::ElementConstPtr _elem)
```
scpeters, line 43: I'm considering adding this as a helper function in sdformat: https://github.com/gazebosim/sdformat/commit/769f61b224de4c0496948f168342dd6b52c62be0

PR #2374, opened 2024-04-17
Drop python3-distutils from Noble package list (onwards)

j-rivero:
# 🎉 New feature

## Summary
There is no `python3-distutils` package anymore in Ubuntu starting with Noble. Drop it and keep a copy of current packages for Jammy.

## Checklist
- [ ] Added tests


azeey:
AFAIK, the only place we need it is in https://github.com/gazebosim/gz-sim/blob/dc3fb1124fa58439abaec56f1301bb70a3ec7852/python/CMakeLists.txt#L9-L14 which is for cmake<3.12. Jammy has a much newer cmake, so we can drop it altogether.

.github/ci/packages-jammy.apt
```diff
@@ -0,0 +1,47 @@
...
+libogre-1.9-dev
+libogre-next-2.3-dev
+libprotobuf-dev
+libprotoc-dev
+libsdformat14-dev
+libtinyxml2-dev
+libxi-dev
+libxmu-dev
+libpython3-dev
+python3-distutils
```
scpeters, line 27: I believe packages from both this file and `packages.apt` will be installed (see [action-gz-ci/entrypoint.sh](https://github.com/gazebo-tooling/action-gz-ci/blob/focal/entrypoint.sh#L102)), so I think this file just needs this one line with `python3-distutils`

scpeters:
> AFAIK, the only place we need it is in
> 
> https://github.com/gazebosim/gz-sim/blob/dc3fb1124fa58439abaec56f1301bb70a3ec7852/python/CMakeLists.txt#L9-L14
> 
> which is for cmake<3.12. Jammy has a much newer cmake, so we can drop it altogether.

attempting to drop it in gz-math8 in https://github.com/gazebosim/gz-math/pull/587

j-rivero:
> AFAIK, the only place we need it is in

You are right, [aebfe55](https://github.com/gazebosim/gz-sim/pull/2374/commits/aebfe555633073420104c9fa8653c61bea7cf17f)

CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.10.2 FATAL_ERROR)
+cmake_minimum_required(VERSION 3.22.1 FATAL_ERROR)
```
azeey, line 1: I don't think it's a good idea to change this in a stable release. We're making this change in `main`.

python/CMakeLists.txt
```diff
@@ -7,18 +7,7 @@ endif()
...
-  from distutils import sysconfig as sc
-  print(sc.get_python_lib(plat_specific=True))"
-      OUTPUT_VARIABLE Python3_SITEARCH
-      OUTPUT_STRIP_TRAILING_WHITESPACE)
-  else()
-    # Get install variable from Python3 module
-    # Python3_SITEARCH is available from 3.12 on, workaround if needed:
-    find_package(Python3 COMPONENTS Interpreter)
-  endif()
+  find_package(Python3 COMPONENTS Interpreter)
```
azeey, line 10: I think we can leave the code as is. It's technically possible for someone to build this with cmake 3.10. But we know for sure our CI will have a newer version of cmake.

azeey (changes requested):

CMakeLists.txt
j-rivero: arg, sorry, lost track while reviewing Steve gz-math's changes. I leave all the code as it was before.

.github/ci/packages-jammy.apt
```diff
@@ -0,0 +1 @@
...
```
azeey, line 1: Are you opposed to removing `python3-distutils` from any of the `.apt` files? I think it would be fine since Jammy has a newer version of cmake.

scpeters: I think it would be fine to remove for Harmonic since 22.04 has cmake 3.22.1

azeey (approved):

Issue #2268, opened 2023-12-14
Improve determinism of PostUpdate callbacks using sensor data


scpeters:
## Desired behavior

Currently many sensors are implemented with systems that publish new sensor data over gz-transport during `PostUpdate`, for example:

* [AirPressure](https://github.com/gazebosim/gz-sim/blob/gz-sim8_8.0.0/src/systems/air_pressure/AirPressure.cc#L124-L171)
* [AirSpeed](https://github.com/gazebosim/gz-sim/blob/gz-sim8_8.0.0/src/systems/air_speed/AirSpeed.cc#L126-L173)
* [Altimeter](https://github.com/gazebosim/gz-sim/blob/gz-sim8_8.0.0/src/systems/altimeter/Altimeter.cc#L125-L172)
* [ForceTorque](https://github.com/gazebosim/gz-sim/blob/gz-sim8_8.0.0/src/systems/force_torque/ForceTorque.cc#L159-L205)
* [Imu](https://github.com/gazebosim/gz-sim/blob/gz-sim8_8.0.0/src/systems/imu/Imu.cc#L129-L176)

Since all `PostUpdate` code execution is performed in separate threads, a system that wanted to use the data from the current round of `PostUpdate` calls would have to do something like subscribing to the sensor's `gz-transport` topic and waiting for that topic to update or just use the most recently received value. In any case, this is not deterministic.

One good option would be to move some sensor updates from `PostUpdate` into the `Update` call, while ensuring that they are executed after the update of the Physics system. This ordering of update execution could be implemented using execution groups with numerical priority values that determine the order of execution (simiilar to the approach taken in rendering pipelines and init scripts). The sensor data should also then be written to components in the ECM.

scpeters:
> One good option would be to move some sensor updates from PostUpdate into the Update call, while ensuring that they are executed after the update of the Physics system. This ordering of update execution could be implemented using execution groups with numerical priority values that determine the order of execution (simiilar to the approach taken in rendering pipelines and init scripts). The sensor data should also then be written to components in the ECM.

I just discussed this with @azeey, who recommended that we try to keep the sensor updates in `PostUpdate` so that they will still be executed in parallel threads, but still executed in separate groups.

[...]

Issue #1815, opened 2022-11-28
New phase for handling user updates of world state


arjo129:
## Desired behavior
Currently gazebo has three states when running simulations. `PreUpdate`, `Update` and `PostUpdate`. In #1808, I ran into an issue where plugins have to calculate forces to be applied in the `PreUpdate` step and the user command system also spawns new vehicles in the `PreUpdate` phase. This leads to incorrect behavior sometimes as the order in which plugins in the same phase are run is not guaranteed. My current fix involves changing the logic of the plugin to make sure we minimize the effect of unexplained behavior. This however requires that System developers have deep understanding of gazebo. What I would like to propose is to have a `UserEdit` phase where users can interact with the world prior to having `PreUpdate` run. This way systems which modify the state of the world don't clash with systems which apply physical forces and new entities will always have forces applied to them before the first physical step takes place. So `UserCommandSystem` would always spawn objects during the `UserEdit` phase.

## Alternatives considered
* Keeping the status quo: The problem is we cannot guarantee physics behavior of items which have been moved by the user.
* Have a way of specifying the exact or partial order in which each plugin should run thus improving determinism: The cons of this are it would require specialist knowledge of Gazebo and leaves correctness to the end user. It would also likely be harder to implement although it would greatly improve the debugging experience. This option is not mutually exclusive with the proposed approach of a `UserEdit` phase.

scpeters:
> This leads to incorrect behavior sometimes as the order in which plugins in the same phase are run is not guaranteed.

There is a prototype in #2394 for controlling order of execution for system callbacks. It could be possible to address this by thoughtful adjustment of the system priority values

PR #2394, opened 2024-05-02
Prototype: specify order of execution for System::Update callbacks

scpeters:
# 🎉 New feature

Part of #2268 

## Summary

This is a prototype for specifying an integer priority value for a `System` using a `<gz:system_priority/>` tag in its XML content. The `SystemManager` parses this tag and stores vectors of system callbacks in a `std::map` indexed by this priority value. The prototype only implements controllable execution order for the `System::Update` callbacks, but should be extended to support `PreUpdate` and `PostUpdate` as well (I'm not sure it's relevant for other callbacks like `Configure`, `ConfigureParameters`, or `Reset` but it could be implemented for those as well).

TODO:

* [ ] Move definition of priority type, default priority, and priority element name to public `System.hh` header
* [ ] Apply changes to `PreUpdate` as well
* [ ] Retarget to Ionic and test with sensor systems
* [ ] Profile performance of sensors using `PostUpdate` / `Update` / `PreUpdate`

## Test it

Follow the instructions for the `priority_printer` example to run an example world with a plugin that illustrates this feature.

## Checklist
- [ ] Added tests

### DIFF
--- a/examples/plugin/priority_printer_plugin/CMakeLists.txt
+++ b/examples/plugin/priority_printer_plugin/CMakeLists.txt
@@ -1,0 +1,17 @@
+cmake_minimum_required(VERSION 3.10.2 FATAL_ERROR)
+
+find_package(gz-cmake3 REQUIRED)
+
+project(Priority_printer)
+
+gz_find_package(gz-plugin2 REQUIRED COMPONENTS register)
+set(GZ_PLUGIN_VER ${gz-plugin2_VERSION_MAJOR})
+
+gz_find_package(gz-sim8 REQUIRED)
+set(GZ_SIM_VER ${gz-sim8_VERSION_MAJOR})
+
+add_library(PriorityPrinter SHARED PriorityPrinter.cc)
+set_property(TARGET PriorityPrinter PROPERTY CXX_STANDARD 17)
+target_link_libraries(PriorityPrinter
+  PRIVATE gz-plugin${GZ_PLUGIN_VER}::gz-plugin${GZ_PLUGIN_VER}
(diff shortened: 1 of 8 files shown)


arjo129:
@scpeters I think this PR might have some minor conflicts with  #2232. It might be worthwhile discussing the `simulation_runners` internals.

scpeters:
> I think this PR might have some minor conflicts with #2232. It might be worthwhile discussing the `simulation_managers` and `simulation_runners` internals.

yes, these will conflict; we should figure out an integrated design for both controlling system execution order and removing systems

> On another note, is it really needed to extend this to `postUpdate`, given that postUpdate systems should not modify the world?

`PostUpdate` calls can have effects outside of the ECM through publishing messages, so it may be useful to group them in a specific order. It is more complicated though, so we can give it a second thought before implementing it.

> Also related: #1815

Yes, I think the changes in this PR could address the concerns of #1815 without adding a new phase to the World step

scpeters:
> > I think this PR might have some minor conflicts with #2232. It might be worthwhile discussing the `simulation_managers` and `simulation_runners` internals.
> 
> yes, these will conflict; we should figure out an integrated design for both controlling system execution order and removing systems

now that I've thought about it some, I think the conflicts should be resolvable. The `std::map` added in this pull request can store a vector of `SystemHolder` instead of Iface raw pointers.

xela-95:
CC @traversaro

scpeters:
this is targeting harmonic; I'm going to open a separate PR for ionic

scpeters:
we can consider back porting to harmonic eventually, but I've closed this for now

PR #2425, opened 2024-05-31
Optimize rendering sensor pose updates

iche033:
# 🦟 Bug fix

Related PR: https://github.com/gazebosim/gz-sensors/pull/439

## Summary
Currently we are updating the pose of all rendering sensors every iteration. This is not necessary as we currently do not expect sensor pose to change at runtime. This PR removes the pose updates every iteration and it now just sets the pose of these rendering sensors once they created. For camera based sensors, their pose transform is the combined result of `//sensor/pose` and `//sensor/camera/pose` (previously the `//sensor/camera/pose` sdf element was ignored and never used). 

Here is an example world for test: [camera_test.sdf](https://gist.github.com/iche033/91616a5620cc64c89a9571c2ea2db767). The world consists of 2 cameras sensors (rgb and depth) with pose offsets in `//sensor/pose` and `//sensor/camera/pose`. Together with https://github.com/gazebosim/gz-sensors/pull/439, the camera sensor pose should be set correctly and you should see images like below:

camera_pose_offset

## Checklist
- [ ] Added tests

### DIFF
--- a/src/rendering/RenderUtil.cc
+++ b/src/rendering/RenderUtil.cc
@@ -24,6 +24,7 @@
 #include <vector>
 
 #include <sdf/Actor.hh>
+#include <sdf/Camera.hh>
 #include <sdf/Collision.hh>
 #include <sdf/Element.hh>
 #include <sdf/Joint.hh>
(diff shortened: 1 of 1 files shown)


azeey:
> previously the //sensor/camera/pose sdf element was ignored and never used)

I imagine this would break some existing SDF files. I wonder if we should provide a flag to enable/disable this bugfix. We should add an entry in the migration guide regardless.

iche033:
> if we should provide a flag to enable/disable this bugfix

I am thinking that I can remove the camera pose change in this PR which is targeting a release branch so we don't break existing sdf files in Harmonic. I'll then create a separate PR targeting `main` with the camera pose change and a migration guide entry. How does that sound?

azeey:
That sounds great!

iche033:
removed camera pose changes in [a129d93](https://github.com/gazebosim/gz-sim/pull/2425/commits/a129d93aed22e68bb1500d34e8a5e46e3d27e52f). I'll merge this forward to `main` first and then apply the camera pose changes in a separate PR.

PR #2437, opened 2024-06-07
Mountain and fog models have been added

serkanMzlm:
# 🎉 New feature
## Summary
- Fog has been added to the mountainous environment
- The collision part is closed by default.
- This application is created using DEM(Digital Elevation Model).
- The collision part can be created with simpler shapes.

Screenshot from 2024-06-06 17-19-41

Screenshot from 2024-06-06 09-17-35

## Test it
```
# The first step is to download the GitHub repository
git clone git@github.com:serkanMzlm/gz-sim.git

# The second step is to add the 'examples' directory of the repository to GZ_SIM_RESOURCE_PATH
export GZ_SIM_RESOURCE_PATH=/home/${USER}/gz-sim/examples
cd gz-sim/examples/worlds/

# The third step is to run the file
gz sim -r mountain_fog.sdf 
```


azeey:
Yes, I would suggest uploading to Fuel unless we have a specific reason to add it to the source code. Do you mind doing that @serkanMzlm?

serkanMzlm:
@azeey The main reason might actually be to demonstrate DEM usage

azeey:
> @azeey The main reason might actually be to demonstrate DEM usage

We already have [dem_moon.sdf](https://github.com/gazebosim/gz-sim/blob/0cc0464e904bb0bc05026580b1fa945e7c4ce33b/examples/worlds/dem_moon.sdf) and [dem_volcano.sdf](https://github.com/gazebosim/gz-sim/blob/0cc0464e904bb0bc05026580b1fa945e7c4ce33b/examples/worlds/dem_volcano.sdf), so I don't think we need more examples.

PR #2473, opened 2024-07-11
Use parentEntity available in `systems` instead of storing it for each interface

azeey:
@arjo129 I think it's possible to further simplify #2232 by using the `parentEntity` already available in the `systems` vector. I haven't fully tested this yet, but I wanted get your feedback if there's a downside to this approach. 

cc @scpeters

### DIFF
--- a/src/SystemManager.cc
+++ b/src/SystemManager.cc
@@ -18,7 +18,6 @@
 #include <list>
 #include <mutex>
 #include <set>
-#include <unordered_set>
 
 #include <gz/common/StringUtils.hh>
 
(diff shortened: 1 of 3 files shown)


scpeters:
the overall changes of this branch against `main`: https://github.com/gazebosim/gz-sim/compare/main...azeey/suggestion_2232

src/SystemManager.cc
```diff
@@ -460,58 +452,41 @@ void SystemManager::ProcessRemovedEntities(
...
-  }
 
-  RemoveFromVectorIf(this->systemsPostupdate,
-    [&](const auto system) {
-      if (markedForRemoval.count(system)) {
-        return true;
-      }
-      return false;
-    });
+    removeFromVector(this->systemsPostupdate, system.postupdate);
```
arjo129, line 477: I tested this. It works just fine. The only thing I'd say is that we do some extra work by iterating over each system (`n`) by the number of entities removed (`m`)  (time complexity `O(mn)`), rather than just each system (which is what the original PR does) (time complexity `O(n)`). Its fine if we have only a few entities removed at a time. If you'd like I can fix that by building `unordered_sets` of `ISystemPreupdate*` or we can just leave it as is.

arjo129 (approved):

src/SystemManager.cc
```diff
@@ -460,58 +452,41 @@ void SystemManager::ProcessRemovedEntities(
...
+    removeFromVector(this->systemsConfigure, system.configure);
+    removeFromVector(this->systemsConfigureParameters,
+                     system.configureParameters);
+    RemoveFromVectorIf(this->systems,
```
arjo129, line 481: This can be moved out of the loop using: `_ecm.IsMarkedForRemoval(system.parentEntity);`

src/SystemManager.cc
arjo129: OK Opened a PR to this PR implementing the changes:
#2475

src/SystemManager.cc
arjo129: See #2475

scpeters (approved):

PR #2474, opened 2024-07-12
Check if any entity actually has a ContactSensorData component before…

shameekganguly:
… calling GetContactsFromLastStep

`GetContactsFromLastStep` can be an expensive call if there are a lot of contact points in the scene. Currently we call this method in the Physics System when the ECM has the `ContactSensorData` component type. However, this will still be true if the `ContactSensorData` component has been removed for all entities.

I have updated the Physics System to additionally check if there is at least one entity with the `ContactSensorData` component before calling `GetContactsFromLastStep`. This enables consuming Systems to optimize their computational footprint by only adding the `ContactSensorData` for relevant entities when needed and remove it when not needed.

### DIFF
--- a/src/systems/physics/Physics.cc
+++ b/src/systems/physics/Physics.cc
@@ -3797,8 +3797,8 @@ void PhysicsPrivate::UpdateCollisions(EntityComponentManager &_ecm)
   // Also check if any entity currently has a ContactSensorData component.
   bool need_contact_sensor_data = false;
   _ecm.Each<components::Collision, components::ContactSensorData>(
-      [&](const Entity &_collEntity1, components::Collision *,
-          components::ContactSensorData *_contacts) -> bool
+      [&](const Entity &/*unused*/, components::Collision *,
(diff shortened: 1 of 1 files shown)


src/systems/physics/Physics.cc
```diff
@@ -3794,6 +3794,17 @@ void PhysicsPrivate::UpdateCollisions(EntityComponentManager &_ecm)
   if (!_ecm.HasComponentType(components::ContactSensorData::typeId))
     return;
 
+  // Also check if any entity currently has a ContactSensorData component.
+  bool need_contact_sensor_data = false;
```
iche033, line 3798: style nit: use camelCase

iche033 (commented):
looks good to me. Just one minor style comment

src/systems/physics/Physics.cc
shameekganguly: Thanks, updated

iche033 (approved):

PR #2477, opened 2024-07-12
Merge 6 -> 7 

iche033:
# ➡️ Forward port

Port ign-gazebo6 to gz-sim7

Branch comparison: https://github.com/gazebosim/gz-sim/compare/ign-gazebo6...gz-sim7

### DIFF
--- a/.github/ci/after_make.sh
+++ b/.github/ci/after_make.sh
@@ -6,12 +6,3 @@ set -e
 # Install (needed for some tests)
 make install
 export LD_LIBRARY_PATH=${LD_LIBRARY_PATH}:/usr/local/lib
-
-# For gz-tools
-export GZ_CONFIG_PATH=/usr/local/share/gz
-
-# For rendering / window tests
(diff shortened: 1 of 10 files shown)


azeey (approved):
Looks like there are some CI failures on GitHub actions. I think what's going on is that these tests weren't enabled before since Xvfb was running after `cmake`. With the changes in this PR, these tests are now active but seem to fail on Github Actions. Since it's probably GPU related, I would propose we disable them in a follow-up PR by checking for the `GITHUB_ACTIONS` environment variable (see https://github.com/gazebosim/gz-sim/blob/90a7ba7e79e76d50b2cb97e4a20aad5ebba9b023/src/gui/Gui_clean_exit_TEST.cc#L87)

iche033:
Created a follow up PR in https://github.com/gazebosim/gz-sim/pull/2480 to disable these tests. Checking to make sure CI is green first before merging these two PRs.

mjcarroll (approved):

PR #2490, opened 2024-07-22
Fix ResourceSpawner

### LINKED ISSUE #2486: Regression in Resource Spawner in Harmonic
## Environment
* OS Version: Ubuntu 22.04
* Source or binary build? Binary, 8.5.0

## Description
* Expected behavior: Resource spawner to work
* Actual behavior: Clicking on any of the models to download them fails with the error `[GUI] [Wrn] [ResourceSpawner.cc:589] Download failed.  Try again.`. Trying to spawn the model fails with the error 
```
Error [SDF.cc:165] Tried to use callback in sdf::findFile(), but the callback is empty.  Did you call sdf::setFindCallback()?
[GUI] [Wrn] [Spawn.cc:381] Only model or light entities can be spawned at the moment.
```

I bisected gz-fuel-tools and found that the regression started in https://github.com/gazebosim/gz-fuel-tools/pull/376

## Steps to reproduce

1. `gz sim -v4 empty.sdf`
2. Add the `ResourceSpawner` gui plugin
3. Try to download or spawn any model

mjcarroll:
# 🦟 Bug fix

Fixes #2486 

## Summary

`id.UniqueName()` now returns without `https://` prefixed, but we want to keep track of the URIs that the resources are downloaded from.    These changes were added in https://github.com/gazebosim/gz-fuel-tools/pull/376 and should have been accommodated here.

## Checklist
- [ ] Added tests

**Note to maintainers**: Remember to use **Squash-Merge** and edit the commit message to match the pull request summary while retaining `Signed-off-by` messages


ahcorde (approved):

src/gui/plugins/resource_spawner/ResourceSpawner.cc
```diff
@@ -739,7 +739,7 @@ void ResourceSpawner::RunFetchResourceListThread(const std::string &_owner)
             resource.isFuel = true;
             resource.isDownloaded = false;
             resource.owner = id.Owner();
-            resource.sdfPath = id.UniqueName();
+            resource.sdfPath = "https://" + id.UniqueName();
```
azeey, line 742: Can we avoid hard coding `https://`? Maybe get it from `server` somehow?

azeey (commented):
Thanks for the fix. Just one issue

src/gui/plugins/resource_spawner/ResourceSpawner.cc
mjcarroll: Actually looking at this again, the `UniqueName` won't work on Windows (though gui plugins currently don't have any test coverage).  Let me re-work it.

mjcarroll: Okay, I ended up adding a `Url` method to the Identifier objects in fuel_tools.  This now depends on https://github.com/gazebosim/gz-fuel-tools/pull/429

logocar3:
Disclaimer - outsider needing this bug to be fixed.

There is Url property in ServerConfig object.

Change the line, where build fails:
  742 |             resource.sdfPath = id.Url().Str();
  
 Into
  742 |             resource.sdfPath = id.Server().Url().Str();
  
  Regards

mjcarroll:
> There is Url property in ServerConfig object.

That URL only stores the base server URL.  In this case, we need the full URL to the asset in question.

azeey:
@mjcarroll looks like this needs a gz-fuel-tool9 release.

usmanNoor5:
[ruby $(which gz) sim-2] Error [SDF.cc:165] Tried to use callback in sdf::findFile(), but the callback is empty.  Did you call sdf::setFindCallback()?


how to handle this problem

PR #2487, opened 2024-07-20
Specify System::PreUpdate, Update execution order

scpeters:
# 🎉 New feature

Part of https://github.com/gazebosim/gz-sim/issues/2268 and https://github.com/gazebosim/gz-sim/issues/2391, second attempt at https://github.com/gazebosim/gz-sim/pull/2394

## Summary

While the `PreUpdate`, `Update`, and `PostUpdate` phases of gz-sim systems allow some control over the order in which code is executed, there are cases in which more control is desired. For example, as noted in #2268 and https://github.com/gazebosim/gz-sim/issues/2391, sensor systems currently update their data during `PostUpdate` (after the `Update` step of the physics system) but can't write their data to the ECM since `PostUpdate` has read-only access to the ECM. Moving sensor updates to `Update` would allow writing to the ECM, but would require ensuring that the sensor systems execute their `Update` step after the physics `Update`. Alternatively, if one wanted to implement a hierarchical cascade of systems acting on data sequentially, rather than adding `PrePreUpdate` and `PrePrePreUpdate` phases, one could specify the system priority during `PreUpdate` for the relevant systems instead.

This feature is used by specifying an integer value in a `<gz:system_priority>` tag for a system that will control the order in which System `PreUpdate` and `Update` callbacks are executed, with smaller priority values executing first.

## Test it

Follow the README instructions to build and run the `PriorityPrinter` example plugin to illustrate the feature.

## Checklist
- [X] Signed all commits for DCO
- [ ] Added tests
- [X] Added example and/or tutorial
- [ ] Updated documentation (as needed)
- [ ] Updated migration guide (as needed)
- [ ] Consider updating Python bindings (if the library has them)
- [X] `codecheck` passed (See [contributing](https://gazebosim.org/docs/all/contributing#contributing-code))
- [X] All tests passed (See [test coverage](https://gazebosim.org/docs/all/contributing#test-coverage))
- [ ] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Agazebosim+archived%3Afalse+) to support the maintainers

**Note to maintainers**: Remember to use **Squash-Merge** and edit the commit message to match the pull request summary while retaining `Signed-off-by` messages.

PR #2494, opened 2024-07-25
ForceTorque system: write WrenchMeasured to ECM

scpeters:
# 🎉 New feature

Part of https://github.com/gazebosim/gz-sim/issues/2268 and https://github.com/gazebosim/gz-sim/issues/2391, requires https://github.com/gazebosim/gz-sensors/pull/449 ~~and targets the branch for https://github.com/gazebosim/gz-sim/pull/2487~~

## Summary

This adds a new `WrenchMeasured` component and writes to it on the force-torque sensor entities in the ECM. The ForceTorque sensor configuration is changed to call sensor updates even when the gz-transport topic has no subscribers, so that the latest wrench data will be in the ECM. This also adds `Model::ModelCount` and `Model::ModelByName` APIs to `gz/sim/Model.hh` to support the new test.

## Test it

Build from source and run `ctest -R INTEGRATION_force_torque_system`

## Checklist
- [X] Added tests


include/gz/sim/components/WrenchMeasured.hh
```diff
@@ -0,0 +1,57 @@
+/*
+ * Copyright (C) 2021 Open Source Robotics Foundation
```
ahcorde, line 2:
```suggestion
 * Copyright (C) 2024 Open Source Robotics Foundation
```

scpeters: https://github.com/gazebosim/gz-sim/pull/2494/commits/9a68307a5d1bf15a3081481a46bce4cebe78f402

scpeters:
https://github.com/gazebosim/gz-sensors/pull/449 has been merged and nightlies rebuilt, rerunning CI now

src/systems/force_torque/ForceTorque.cc
```diff
@@ -176,15 +180,13 @@ void ForceTorque::PostUpdate(const UpdateInfo &_info,
...
-    // and that sensor has subscribers.
+    // we only update if there is at least one sensor that needs data.
     // note: gz-sensors does its own throttling. Here the check is mainly
     // to avoid doing work in the ForceTorquePrivate::Update function
     bool needsUpdate = false;
     for (const auto &[sensorEntity, sensor] : this->dataPtr->entitySensorMap)
     {
-      if (sensor->NextDataUpdateTime() <= _info.simTime &&
-          sensor->HasConnections())
+      if (sensor->NextDataUpdateTime() <= _info.simTime)
```
iche033, line 189: does it make sense for users to indicate that they want data to be written to ECM by creating a `components::WrenchMeasured` component? here we would just check for the presence of that component, e.g. 

```cpp
    if (sensor->NextDataUpdateTime() <= _info.simTime &&
        (sensor->HasConnections() ||
        _ecm.Component<components::WrenchMeasured>(sensorEntity) != nullptr))
```

Probably not going to make much perf difference for this system since the sensor `Update` call should be relatively fast and same for writing force torque values to ECM. It would be nice to keep things consistent if we do decide to port this functionality to other sensors, e.g. contact sensor

scpeters: ooh, I like that. Let me give that a try

scpeters: https://github.com/gazebosim/gz-sim/pull/2494/commits/c6fedb12d30f41e593061d54bd549f94ff447787

scpeters:
I noticed a python test failing in #2500 and realized I didn't set priority on its test file, so I've done so in https://github.com/gazebosim/gz-sim/pull/2494/commits/ba5408ffd47152917cf8fa26447ede4f822c75b1

I'm guessing it will fail, since the python test implicitly assumes that its `on_pre_update_cb` will execute after the `ForceTorque::PreUpdate`, but that is now changed. I think the python test should be updated to account for this ordering, which shouldn't be too hard

scpeters:
> I noticed a python test failing in #2500 and realized I didn't set priority on its test file, so I've done so in [ba5408f](https://github.com/gazebosim/gz-sim/commit/ba5408ffd47152917cf8fa26447ede4f822c75b1)
> 
> I'm guessing it will fail, since the python test implicitly assumes that its `on_pre_update_cb` will execute after the `ForceTorque::PreUpdate`, but that is now changed. I think the python test should be updated to account for this ordering, which shouldn't be too hard

the test should be fixed by https://github.com/gazebosim/gz-sim/pull/2494/commits/1f76c6a487a6c47b88100bb582ad58d0cd74f5e0, though it's hard for me to run these tests on my machine, so I'll wait for CI to confirm

iche033:
looks good to me. 

I noticed some [windows tests](https://build.osrfoundation.org/job/gz_sim-pr-win/687/) were failing. I think they are flaky as I've seen a couple of them fail before. I just triggered another build to double check.

iche033 (approved):

PR #2500, opened 2024-07-26
Add System interface to set default priority

scpeters:
# 🎉 New feature

Builds on top of https://github.com/gazebosim/gz-sim/pull/2494

## Summary

This adds a new `SystemConfigurePriority` interface to `gz::sim::System` to support setting the default priority for a system. The motivation for this feature is #2494, in which functionality from the ForceTorque system is moved from `PostUpdate` to `Update`, but it should still execute after `Physics::Update`, so `<gz:system_priority/>` tags with a positive value were added to the worlds using the ForceTorque system in #2494, but it is easier with `SystemConfigurePriority` to set the default priority to a positive value. The system priority can still be overridden by the `<gz:system_priority/>` tags, but it's easier to use the system now without extra configuration.

## Test it

Build and run `INTEGRATION_force_torque_system`

## Checklist
- [x] Added tests


src/systems/force_torque/ForceTorque.cc
```diff
@@ -129,6 +129,12 @@ ForceTorque::ForceTorque()
 //////////////////////////////////////////////////
 ForceTorque::~ForceTorque() = default;
 
+//////////////////////////////////////////////////
+System::PriorityType ForceTorque::ConfigurePriority()
+{
+  return 10;
```
shameekganguly, line 135: Instead of hardcoding the priority here, can we add a header file with a handful of well defined priorities
e.g.
```
constexpr System::PriorityType kPrePhysicsPriority = 0;
constexpr System::PriorityType kPhysicsPriority = 10;
constexpr System::PriorityType kPostPhysicsPriority = 100;
```
and use that instead? I think that will convey the intent of setting the default priority here better. I'd also add a default priority for the PhysicsSystem to go with this change.

arjo129: We probably should so the same for the `UserCommandSystem`.

src/systems/force_torque/ForceTorque.hh
```diff
@@ -50,6 +51,9 @@ namespace systems
     /// \brief Destructor
     public: ~ForceTorque() override;
 
+    /// Documentation inherited
+    public: System::PriorityType ConfigurePriority() final;
```
arjo129, line 55: The only issue I see with this API is what if  we want to configure `PreUpdate` and `Update` priorities separately. Perhaps we should provide two interfaces?

arjo129 (commented):
I've got a few comments on this PR. I think its worth noting that we should be able to configure `PreUpdate` and `Update` priorities seperately. Additionally, we should also take into account the `UserCommand` system which is a source of lot of pain when writing systems.

src/systems/force_torque/ForceTorque.hh
scpeters: at one point I had thought about being able to specify separate priority values for `PreUpdate` and `Update`, but to start with I opted for a simpler approach using a shared priority for each system. Looking through `src/systems`, I only see two systems that implement both `ISystemPreUpdate` and `ISystemUpdate`:

* [LinearBatteryPlugin](https://github.com/gazebosim/gz-sim/blob/main/src/systems/battery_plugin/LinearBatteryPlugin.hh#L76-L77)
* [PythonSystemLoader](https://github.com/gazebosim/gz-sim/blob/main/src/systems/python_system_loader/PythonSystemLoader.hh#L71-L72)

The `PythonSystemLoader` implements both in order to allow python systems to implement either callback, and I haven't looked into the details of `LinearBatteryPlugin`, but given that only one of our systems implements both `PreUpdate` and `Update`, I think it's reasonable to share the priority value, though we should open a feature request to configure priority for these callbacks separately

src/systems/force_torque/ForceTorque.cc
scpeters: I've added SystemPriorityConstants.hh in https://github.com/gazebosim/gz-sim/pull/2500/commits/20542139b2037ae0f108864d031267e64108e6e8 and used those constants to configure priority of the Physics and ForceTorque systems.

I haven't touched `UserCommandSystem` yet; I want to make sure the `Physics` change works properly first

scpeters: that broke the windows build; I'll fix shortly...

src/systems/force_torque/ForceTorque.hh
azeey: I agree we should open a separate issue for this. I also mentioned to @scpeters at one point that it would be good to have a way to configure priorities separately, but upon further thought, it might make plugin configuration very hard. I think it's worth having the discussion, but I would not be in favor of adding it here.

src/systems/force_torque/ForceTorque.cc
scpeters: inline namespace removed in https://github.com/gazebosim/gz-sim/pull/2500/commits/d32c9d693bc2d56a9be88d7c64730f8bffe2b079

include/gz/sim/SystemPriorityConstants.hh
```diff
@@ -0,0 +1,39 @@
...
+ *
+ */
+#ifndef GZ_SIM_SYSTEMPRIORITYCONSTANTS_HH_
+#define GZ_SIM_SYSTEMPRIORITYCONSTANTS_HH_
+
+#include <gz/sim/System.hh>
+
+namespace gz::sim::systems
+{
+  /// \brief A suggested priprity value for a system that should execute before
```
shameekganguly, line 24:
```suggestion
  /// \brief A suggested priority value for a system that should execute before
```

include/gz/sim/SystemPriorityConstants.hh
```diff
@@ -0,0 +1,39 @@
...
+
+  /// \brief Default priority value for the Physics system, with a negative
+  /// value ensuring that it will run before systems with priority
+  /// System::kDefaultPriority.
+  constexpr System::PriorityType kPhysicsPriority = -64;
+
+  /// \brief A suggested priority value for sensor systems that should execute
+  /// after the Physics system but before the systems with priority
+  /// System::kDefaultPriority.
+  constexpr System::PriorityType kPostPhysicsSensorPriority = -32;
```
shameekganguly, line 36: nit: Do we actually need a kPostPhysicsSensorPriority? What is an example of a general pattern where a user would want to run their System's Update specifically after a sensor system as opposed to relying on the data to be present when their System's PostUpdate is called? Just trying to see if we can simplify this conceptually for users.

src/SystemManager.cc
```diff
@@ -112,6 +112,10 @@ size_t SystemManager::ActivatePendingSystems()
     this->systems.push_back(system);
 
     PriorityType p {System::kDefaultPriority};
+    if (system.configurePriority)
+    {
+      p = system.configurePriority->ConfigurePriority();
+    }
```
shameekganguly, line 118: nit: consider logging below if a system has default priority set via the interface but was overriden by the priority in sdf. This would help users to verify the source of truth for their System's priority quickly.

shameekganguly (changes requested):
Thanks Steve! I added a couple small suggestions

include/gz/sim/SystemPriorityConstants.hh
scpeters: you are correct that a user's system PostUpdate call would run after the sensor Update call, but if a user wanted to read the sensor data and write it to the ECM during its own `Update` call, then I think using `kPostPhysicsSensorPriority` for the sensor systems would allow user system `Update` callbacks to access the sensor data

src/SystemManager.cc
scpeters: added in https://github.com/gazebosim/gz-sim/pull/2500/commits/14118685fabbe95ab3922a524927af191be3a76c; it currently prints a `gzdbg` any time a `<gz:system_priority/>` is read

include/gz/sim/SystemPriorityConstants.hh
scpeters: https://github.com/gazebosim/gz-sim/pull/2500/commits/14118685fabbe95ab3922a524927af191be3a76c

scpeters:
I've replied to all comments. Should I modify the priority for `UserCommandSystem` in this pull request as well?

arjo129:
> I've replied to all comments. Should I modify the priority for `UserCommandSystem` in this pull request as well?

I think that'd be great.

src/systems/force_torque/ForceTorque.cc
scpeters: updated user commands system priority in https://github.com/gazebosim/gz-sim/pull/2500/commits/735dd7dcd9f8643126d5d72d7898fc34123b1a4c

scpeters:
> > I've replied to all comments. Should I modify the priority for `UserCommandSystem` in this pull request as well?
> 
> I think that'd be great.

done in https://github.com/gazebosim/gz-sim/pull/2500/commits/735dd7dcd9f8643126d5d72d7898fc34123b1a4c

include/gz/sim/SystemPriorityConstants.hh
```diff
@@ -0,0 +1,46 @@
...
+
+  /// \brief Default priority value for the UserCommands system, with a very
+  /// negative value to indicate that it should run before most systems in
+  /// order to modify the ECM in response to user commands received over
+  /// gz-transport. This is especially important for user commands that add
+  /// objects to the world.
+  constexpr System::PriorityType kUserCommandsPriority = -10000;
```
shameekganguly, line 43: nit: I'd consider adding priorities to this file in an ascending order for clarity

include/gz/sim/SystemPriorityConstants.hh
shameekganguly: Ok, sounds good.

shameekganguly (approved):

include/gz/sim/SystemPriorityConstants.hh
scpeters: https://github.com/gazebosim/gz-sim/pull/2500/commits/811f8ee0c5f5a0ccc29736a9bc87f3bfbac3e411

include/gz/sim/SystemPriorityConstants.hh
```diff
@@ -0,0 +1,46 @@
...
+  /// \brief Default priority value for the Physics system, with a negative
+  /// value ensuring that it will run before systems with priority
+  /// System::kDefaultPriority.
+  constexpr System::PriorityType kPhysicsPriority = -64;
+
+  /// \brief A suggested priority value for sensor systems that should execute
+  /// after the Physics system but before the systems with priority
+  /// System::kDefaultPriority.
+  constexpr System::PriorityType kPostPhysicsSensorPriority = -32;
+}
```
iche033, line 44: move `kDefaultPriority` from System.hh to this file so all constants can be found in the same place?

iche033 (approved):
just a minor suggestion but looks good without it.

iche033 (approved):

include/gz/sim/SystemPriorityConstants.hh
azeey: I'm adding a file called `Constants.hh` in https://github.com/gazebosim/gz-sim/pull/2497/files#diff-01974b0165a380ef2dd215a32374bd9192d833c34ca887784b82d89a4605735e. We could also put it in there. I don't have a strong opinion.

scpeters: should I define all these constants in `System.hh`?

scpeters: I moved all the constants to `System.hh` in https://github.com/gazebosim/gz-sim/pull/2500/commits/5a4f1701eef8b544bd20e3bb303ceb732737f406

arjo129 (approved):

PR #2501, opened 2024-07-26
Improve signal handling

azeey:
# 🦟 Bug fix

## Summary
This mainly deals with the problem where a signal is ignored if it occurs right before `ServerPrivate::Run` is called because we generally rely on the `running` variable to know when to stop, but that variable is set to true at the beginning of `ServerPrivate::Run`.

## Checklist
- [ ] Added tests

### DIFF
--- a/src/Server.cc
+++ b/src/Server.cc
@@ -214,6 +214,11 @@ Server::Server(const ServerConfig &_config)
     this->dataPtr->AddRecordPlugin(_config);
   }
 
+  // If we've received a signal before we create entities, the Stop event
+  // won't be propagated to them. Instead, we just quit early here.
+  if (this->dataPtr->signalReceived)
+    return;
+
   this->dataPtr->CreateEntities();
 
   // Set the desired update period, this will override the desired RTF given in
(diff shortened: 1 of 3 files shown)


iche033:
checking what's a good way to test this? Launch gazebo then verify ctrl + c signal is handled right away?

azeey:
This mainly improves signal handling when Gazebo starts up. So the best way to test it would be to start Gazebo in server mode (`gz sim -s empty.sdf`) and type ctrl+c in the first few seconds. Without this PR, I noticed the signal would sometimes be completely ignored.

iche033 (approved):
I did some manual testing and found that there is a specific point in time after launch when I hit ctrl + c, the signal is handled with these changes and ignored without these changes. 

I did run into a case that gz sim would still freeze if I ctrl + c right after it starts up:

```
iche@jammy2:~/code/gz_i_ws$ gz sim -v 4 -s shapes.sdf
[Msg] Gazebo Sim Server v9.0.0
[Msg] Loading SDF world file[/home/iche/code/gz_i_ws/install/share/gz/gz-sim9/worlds/shapes.sdf].
^C^C^C^C^C^C^C^C
```

That can be addressed separately.

azeey:
> That can be addressed separately.

I'm hoping some of the remaining issues are related to https://github.com/gazebosim/gz-common/pull/618, but there might be other instances of timing.

PR #2526, opened 2024-08-16
Force Qt to use xcb plugin on Wayland

iche033:
# 🦟 Bug fix

## Summary

More and more users are running into the same issue on Wayland. We recently updated the troubleshooting guide for Wayland in https://gazebosim.org/docs/latest/troubleshooting/#wayland-issues. This PR just does the step described in the guide for the users. It sets Qt to use the xcb plugin if a) wayland is detected and b) only if the user has not explicitly set the `QT_QPA_PLATFORM` env var.

This is more of a quality of life fix to improve user experience and save them from having to go through the troubleshooting process.

Similar fix was done in rviz: https://github.com/ros2/rviz/pull/1253

## Checklist
- [ ] Added tests

### DIFF
--- a/src/gui/Gui.cc
+++ b/src/gui/Gui.cc
@@ -221,6 +221,16 @@ std::unique_ptr<gz::gui::Application> createGui(
     qputenv("QT_AUTO_SCREEN_SCALE_FACTOR", "1");
   }
 
+  // check for wayland and force to use X for rendering
+  if (QString::fromLocal8Bit(qgetenv("XDG_SESSION_TYPE")) == "wayland")
+  {
+    if (QString::fromLocal8Bit(qgetenv("QT_QPA_PLATFORM")).isEmpty())
+    {
+      qputenv("QT_QPA_PLATFORM", "xcb");
+    }
+  }
+
+
   bool isPlayback = (nullptr != _guiConfig &&
       std::string(_guiConfig) == "_playback_");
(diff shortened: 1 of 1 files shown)


src/gui/Gui.cc
```diff
@@ -221,6 +221,16 @@ std::unique_ptr<gz::gui::Application> createGui(
     qputenv("QT_AUTO_SCREEN_SCALE_FACTOR", "1");
   }
 
+  // check for wayland and force to use X for rendering
+  if (QString::fromLocal8Bit(qgetenv("XDG_SESSION_TYPE")) == "wayland")
+  {
+    if (QString::fromLocal8Bit(qgetenv("QT_QPA_PLATFORM")).isEmpty())
+    {
```
arjo129, line 228: Perhaps log that we are automatically setting the variable to `XCB` here.

iche033: added log msg here. [e2d8d33](https://github.com/gazebosim/gz-sim/pull/2526/commits/e2d8d332fd2d086d7dc079c02d58b4050740052c)

azeey (approved):
I've heard this workaround also has poor performance, so we should definitely be thinking about a more long term solution. 

I think `X` is needed to create a dummy window, but I've never understood why we need a dummy window in the first place.

iche033:
one option to try is run the gui with ogre2 + EGL

PR #2545, opened 2024-08-26
Fix crash at exit due to a race condition with new signal handler

azeey:
# 🦟 Bug fix

Fixes #<NUMBER>

## Summary
Previously, the signal handler would interrupt the main thread (thread that calls `Server::Run`), so there was no chance for the main thread to resume before the signal handler was finished. With the new signal handler implementation in gz::common, the signal handler is now called from a different thread which means the main thread might resume before the signal handler is finished. This causes a race condition in which the `Server` class, and therefore the `ServerPrivate` class are being destructed while the signal handler is still active. Since `ServerPrivate` also contains the class that encapsulates the signal handler implementation, the causes a crash as the signal handler tries to access data that has been deleted.

The fix here is to ensure that the main thread does not exit until the signal handler is finished by synchronizing on a mutex.

## Checklist
- [ ] Added tests


iche033:
CI caught some new test failures. I was able to reproduce the `INTEGRATION_dvl*` test failures

azeey:
> CI caught some new test failures. I was able to reproduce the `INTEGRATION_dvl*` test failures locally

Apparently I was not running tests that need a display because I didn't have glxinfo available on the container I was using. I have pushed a change that fixed `INTEGRATION_dvl*` for me. I did have some other flaky tests locally, but they seem to be flaky on the `main` branch as well.

iche033 (approved):

PR #2578, opened 2024-09-02
Fixed typo in gz sim python api

ahcorde:
# 🦟 Bug fix

Fixes #<NUMBER>

## Checklist
- [ ] Added tests

### DIFF
--- a/python/src/gz/sim/Joint.cc
+++ b/python/src/gz/sim/Joint.cc
@@ -94,7 +94,7 @@ void defineSimJoint(py::object module)
       py::arg("ecm"),
       py::arg("limits"),
       "Set the effort limits on a joint axis.")
-  .def("set_position_imits", &gz::sim::Joint::SetPositionLimits,
+  .def("set_position_limits", &gz::sim::Joint::SetPositionLimits,
       py::arg("ecm"),
       py::arg("limits"),
       "Set the position limits on a joint axis.")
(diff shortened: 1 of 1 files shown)


python/src/gz/sim/Joint.cc
```diff
@@ -94,7 +94,7 @@ void defineSimJoint(py::object module)
       py::arg("ecm"),
       py::arg("limits"),
       "Set the effort limits on a joint axis.")
-  .def("set_position_imits", &gz::sim::Joint::SetPositionLimits,
+  .def("set_position_limits", &gz::sim::Joint::SetPositionLimits,
```
arjo129, line 97: Shouldn't we be deprecating this instead of removing it.

arjo129 (changes requested):
I'm a bit torn by this change. On one hand given that we have released this, wouldn't this change break the API? Shouldn't we do a deprecation first?

On the other given how long its taken for this bug to be reported, maybe its fine to just go ahead and change the name.

python/src/gz/sim/Joint.cc
azeey: Yeah, I think we should deprecate the old function and add the new one.

ahcorde: https://github.com/gazebosim/gz-sim/pull/2578/commits/27a46daeeef890f949514b88b086daf3eb8312ea

azeey (approved):
LGTM!

python/src/gz/sim/Joint.cc
ahcorde: @azeey this is wrong in older versions, what should we do?

PR #2596, opened 2024-09-05
Improve load times by skipping serialization of entities when unecessary.

arjo129:
# 🦟 Bug fix

Fixes #<NUMBER>

## Summary
I was investigating
https://github.com/gazebosim/gazebo_test_cases/issues/1576 , in my investigation it came to my notice that `sdf::Root` takes a long time to be constructed. While I am submitting changes upstream to reduce the impact of the creation of `sdf::Root`, in the event that we don't serialize a model, I'm proposing we just send an empty string accross. This should minimize both network traffic and make load times more manageable. 

## Checklist
- [ ] Added tests


include/gz/sim/components/Model.hh
```diff
@@ -91,11 +92,17 @@ namespace serializers
     {
       sdf::Root root;
       std::string sdf(std::istreambuf_iterator<char>(_in), {});
+      if (sdf.find("model") == std::string::npos)
+      {
+        gzwarn << "No model was sent\n";
```
bperseghetti, line 97:
```suggestion
        gzwarn << "No model was sent." << std::endl;
```

azeey:
I remember we added this serializing of `sdf::Model` to support scaling on the GUI, but that was never implemented. Maybe @caguero remembers more. I'm not sure if we need it for anything else.

caguero:
> I remember we added this serializing of `sdf::Model` to support scaling on the GUI, but that was never implemented. Maybe @caguero remembers more. I'm not sure if we need it for anything else.

All that I remembered was that it started motivated to support scaling. I'm not sure if after that any other piece is using it.

arjo129:
Should we remove the serialization all together then?  The speed up from removing it is quite a bit. `3k_worlds.sdf` take well over 300s (5minutes) to load with serialization, if I remove it we can load in 22seconds an almost 13x improvement.

Each model instance takes about 0.1s to deserialize (which is not very good).

I do think we should eventually address the root cause which I've documented here: https://github.com/gazebosim/sdformat/issues/1478

azeey:
@arjo129 having long load times is not great, but since this has been an issue for a long time and only affects loading large files, I would not consider it a critical bug. I'd rather not make this change while we are in code freeze.

arjo129:
Sure lets revisit this after the release. That being said, I would say that 3k worlds only loads on the largest computer I have access to (which happens to be a cloud instance). It fails to load on any of my local machines.

include/gz/sim/components/Model.hh
```diff
@@ -75,10 +75,18 @@ namespace serializers
         }
       }
 
+      if (!skip)
+      {
       _out << "<?xml version=\"1.0\" ?>"
            << "<sdf version='" << SDF_PROTOCOL_VERSION << "'>"
            << (skip ? std::string() : modelElem->ToString(""))
            << "</sdf>";
```
iche033, line 83: add indentation, and since `skip` will always be false here, we can simplify the logic?

```suggestion
        _out << "<?xml version=\"1.0\" ?>"
             << "<sdf version='" << SDF_PROTOCOL_VERSION << "'>"
             << modelElem->ToString("")
             << "</sdf>";
```

iche033:
the Examples test failure is unrelated, and should be fixed by https://github.com/gazebosim/gz-sim/pull/2649

iche033 (approved):

scpeters:
this branch is based on `main` but targeted at `gz-sim9`. It should either be retargeted to `main` or rebased on `gz-sim9`

arjo129:
> this branch is based on `main` (see changes to README) but targeted at `gz-sim9`. It should either be retargeted to `main` or rebased on `gz-sim9`

:facepalm:  Thanks for spotting that!

arjo129:
Not sure why CI is still :x: on this. Seems unrelated.

scpeters:

arjo129:
https://github.com/Mergifyio backport gz-sim9 gz-sim8 ign-gazebo6

PR #2611, opened 2024-09-10
Fix log playback GUI display

azeey:
# 🦟 Bug fix

Fixes https://github.com/gazebosim/gazebo_test_cases/issues/1640

Alternative to: https://github.com/gazebosim/gz-sim/pull/2590

## Summary
Loading the logging playback plugin before `SceneBroadcaster` prevents playback from being displayed on the GUI.  This order was changed in #2452 which caused the regression. 

I haven't had time to delve into why the order matters, but it can easily be verified by running `test/worlds/log_playback.sdf` with `<playback_path>` set to the log directory. If the plugin comes before `SceneBroadcaster`, the playback is not displayed on the GUI.

## Checklist
- [ ] Signed all commits for DCO
- [ ] Added tests
- [ ] Updated documentation (as needed)
- [ ] Updated migration guide (as needed)
- [ ] Consider updating Python bindings (if the library has them)
- [ ] `codecheck` passed (See [contributing](https://gazebosim.org/docs/all/contributing#contributing-code))
- [ ] All tests passed (See [test coverage](https://gazebosim.org/docs/all/contributing#test-coverage))
- [ ] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Agazebosim+archived%3Afalse+) to support the maintainers

**Note to maintainers**: Remember to use **Squash-Merge** and edit the commit message to match the pull request summary while retaining `Signed-off-by` messages.

PR #2619, opened 2024-09-12
Adds a regression test for logplayback fix in #2611

arjo129:
Related to:

https://github.com/gazebosim/gazebo_test_cases/issues/1640

This commit adds a really simple regression check to make sure we actually publish on the right topic.


test/integration/log_system.cc
```diff
@@ -774,6 +774,18 @@ TEST_F(LogSystemTest, GZ_UTILS_TEST_DISABLED_ON_WIN32(RecordAndPlayback))
   // Start server
   Server playServer(playServerConfig);
 
+  // Simulate a client
+  gz::transport::Node node;
+  std::size_t numMsgs = 0;
```
azeey, line 779:
```suggestion
  std::atomic<std::size_t> numMsgs = 0;
```

Since a callback is invoked from a separate thread.

azeey (approved):
Thanks @arjo129 ! Just one minor comment.

test/integration/log_system.cc
arjo129: [c603ad6](https://github.com/gazebosim/gz-sim/pull/2619/commits/c603ad65a0b37f8a235ac37c56af7af435596fd3)

arjo129:
https://github.com/Mergifyio backport gz-sim8

PR #2546, opened 2024-08-26
Restore behavior change by reverting #2527

iche033:
# 🦟 Bug fix

## Summary

https://github.com/gazebosim/gz-sim/pull/2527 reverted the changes introduced in https://github.com/gazebosim/gz-sim/pull/2452 because it's a behavior breaking change. 

This PR restores the behavior change for `main` by undoing the changes in https://github.com/gazebosim/gz-sim/pull/2527. 

## Checklist
- [x] Signed all commits for DCO
- [ ] Added tests
- [ ] Updated documentation (as needed)
- [ ] Updated migration guide (as needed)
- [ ] Consider updating Python bindings (if the library has them)
- [ ] `codecheck` passed (See [contributing](https://gazebosim.org/docs/all/contributing#contributing-code))
- [ ] All tests passed (See [test coverage](https://gazebosim.org/docs/all/contributing#test-coverage))
- [ ] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Agazebosim+archived%3Afalse+) to support the maintainers

**Note to maintainers**: Remember to use **Squash-Merge** and edit the commit message to match the pull request summary while retaining `Signed-off-by` messages.

PR #2629, opened 2024-09-16
Joint Tutorial needs topic_name in SDF examples.

bperseghetti:
# 🦟 Bug fix

## Summary
This is my bad on not pushing up the final commit on adding `topic_name` in a few places so that the trajectory control portion works properly. 🤦🏻‍♂️ 

This also fixes the indents so they match when copying and pasting through tutorial.

## Checklist
- [ ] Added tests

### DIFF
--- a/tutorials/joint_controller.md
+++ b/tutorials/joint_controller.md
@@ -166,10 +166,10 @@ This is how the model will look:
 
 ```xml
 <plugin
- filename="gz-sim-joint-controller-system"
- name="gz::sim::systems::JointController">
- <joint_name>j1</joint_name>
- <initial_velocity>1.0</initial_velocity>
+  filename="gz-sim-joint-controller-system"
+  name="gz::sim::systems::JointController">
+  <joint_name>j1</joint_name>
+  <initial_velocity>1.0</initial_velocity>
 </plugin>
 ```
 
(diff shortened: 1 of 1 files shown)


tutorials/joint_controller.md
```diff
@@ -591,28 +592,28 @@ This is how the model will look:
...
+  <joint_name>RR_position_control_joint1</joint_name>
+  <initial_position>0.7854</initial_position>
+  <position_p_gain>20</position_p_gain>
+  <position_i_gain>0.4</position_i_gain>
+  <position_d_gain>1.0</position_d_gain>
+  <position_i_min>-1</position_i_min>
+  <position_i_max>1</position_i_max>
+  <position_cmd_min>-20</position_cmd_min>
+  <position_cmd_max>20</position_cmd_max>
+  <topic>topic_name</topic>
```
azeey, line 607: So what's the behavior without `<topic>` specified? Can we make sure that scenario works as well?

azeey: I guess my question is, has the default topic changed? I ask because https://github.com/gazebosim/gz-sim/pull/2546 changed some default topics for some systems and those systems had to be updated.

bperseghetti: No, the default topic I don't believe has changed, the tutorial shows to publish this to control it:
https://github.com/gazebosim/gz-sim/blob/76305dfe9065db3b17dfa562c2ee8e6bafa7b1a3/tutorials/joint_controller.md?plain=1#L622

Note that you need to have the topic name match that topic or it won't work.

And then it says to echo:
https://github.com/gazebosim/gz-sim/blob/76305dfe9065db3b17dfa562c2ee8e6bafa7b1a3/tutorials/joint_controller.md?plain=1#L670

Which once again if you change it to listen to published topic won't work. So one or the other needs to be changed/fixed to work.

bperseghetti: So @azeey do you want an example with only the default topic, only the "assigned topic" or one with both?

caguero: @azeey , any final decision here?

Issue #2506, opened 2024-08-01
TrackedVehicle system does unexpected movement on low turn commands.


ValeriiKoshelev:
## Environment
* OS Version: Ubuntu 22.04
* Source or binary build? Source, gz-sim7.
  
  

## Description
If you send command with zero linear component and small (<0.1) angular component, vehicle behaves in unexpected way.

Expected behavior: no movement or slow turn in place.
Observed behavior: slow linear motion.

Probably something wrong with edge case here: https://github.com/gazebosim/gz-sim/blame/007aa87cd0007c2bbe2887e79b3679070d7bc806/src/systems/tracked_vehicle/TrackedVehicle.cc#L689
Code looks same in gz-sim8, so probably reproducible in newer versions. 

## Steps to reproduce
Commands executed:
```
gz sim -r tracked_vehicle_simple.sdf
gz topic -t /model/simple_tracked/cmd_vel -m gz.msgs.Twist -p "linear: {x: 0.0}, angular: {z: 0.05}"
```

azeey:
Yeah, it seems like it could be an issue with how that condition is handled on line 689.  Would you be able to work on a fix? 

cc @peci1

peci1:
Yeah, it seems to be caused by the hardcoded threshold. The threshold is there to prevent division by zero when computing turning radius. But I don't see any reason why it couldn't have a lower value. Not sure if it makes sense to turn the thresholds into configurable values. Probably just changing the hardcoded values to something like 1e-4 would suffice. This would, of course, be a breaking change, but I don't think it would really affect anyone.

@azeey not sure what the current policy is, but feel free to add me as a codeowner for the files added in the tracked vehicles PR.

peci1:
You definitely can do that. I'm away this week.

iandresolares:
Hello, 
has there been any updates on this topic? I am using Gazebo Sim, version 8.6.0, and faced this same issue. Would lowering this threshold to a smaller value like 0.0001 will do it? Or it would brake something else. I am thinking on installing from source and modifying the plugin for now.

ntfshard:
Hi. I have a branch, with [fix](https://github.com/ntfshard/gz-sim/commit/e7b020f9aedff19f7f986648eb43ccf3434f7b3a), but I didn't check it properly yet. And I also want to think about this question, but maybe it can be done in a separate way

iandresolares:
[...]

PR #2643, opened 2024-10-09
Fix docker nightly build

ntfshard:
# 🦟 Bug fix

Fixes #<NUMBER>

## Summary
Clean build by instruction in `docker/README.md` not working. Base image builds fine, but the next step is failing.

Ubuntu Focal (20.04) is too old to build `gz-sim9` branch. It doesn't contain required GZ packages.
```
E: Package 'libgz-cmake4-dev' has no installation candidate
E: Unable to locate package libgz-common6-dev
E: Unable to locate package libgz-fuel-tools10-dev
E: Unable to locate package libgz-math8-eigen3-dev
E: Unable to locate package libgz-plugin3-dev
E: Unable to locate package libgz-physics8-dev
E: Unable to locate package libgz-rendering9-dev
E: Unable to locate package libgz-transport14-dev
E: Unable to locate package libgz-gui9-dev
E: Unable to locate package libgz-msgs11-dev
E: Unable to locate package libgz-sensors9-dev
E: Unable to locate package libsdformat15-dev
```
GCC8 is a very old compiler, I believe installation and replacing a default compiler was done for Ubuntu 18 (with default GCC7). Switched to GCC11 (default for Ubuntu 22). IMHO I'd prefer not to bind to compiler version in this case, but I made it as close as possible
Minor improvements in Readme and image
AFAIU `docker/run.bash` script not working well on modern systems, but it's out of scope of this question

Slightly out of scope, but couple of months ago tried same for a `gz-sim8` branch and had problem with 2 missing gz libraries. If this patch will be accepted, I'll do similar changes on that branch too

## Checklist
- [ ] Added tests


mjcarroll (approved):

mjcarroll:
Would it additionally make sense to introduce a CI job to make sure that this docker file can build on, for instance, a nightly basis?

ntfshard:
> Would it additionally make sense to introduce a CI job to make sure that this docker file can build on, for instance, a nightly basis?

Not a big master of GitHub actions, could try I suppose

ntfshard:
> > Would it additionally make sense to introduce a CI job to make sure that this docker file can build on, for instance, a nightly basis?
> 
> Not a big master of GitHub actions, could try I suppose

Maybe in another PR? It not fits very well widely spread pipelines AFAIS. Actually I planned to make fix for annoying bug #2506 which affects us. (and we use older version in which I'd like to propagate fix too).
[In other words, propagate this changes on previous branches: gz-sim8, gz-sim7 and made fix for mentioned bug above on this branch and propagate it on a same branches too]

mjcarroll:
> Maybe in another PR?

Absolutely, would just be a nice addition to keep these dockerfiles fresh and not regressing.

> It sounds reasonable?

Good with me.

ntfshard:
Ok. The last mystery, what happened in Jenkins. Looks like something goes wrong with couple of python tests. I can't try restart build there, don't have a permissions.

iche033:
> Ok. The last mystery, what happened in Jenkins. Looks like something goes wrong with couple of python tests. I can't try restart build there, don't have a permissions.

the failing python tests are likely due to https://github.com/osrf/homebrew-simulation/issues/2834 which is currently affecting all gz packages. So not the fault of this PR

ntfshard:
Thank you for your explanation. What is the next step for this PR?
And what is the right way to propagate this changes to other branches? Cherry-pick(commit after merge) or just implement same changes due to it still will be a merge conflict.

iche033 (approved):

ntfshard:
Ok, will try) Thank you!
But right now while CI is red, it can't be merged.

docker/Dockerfile.base
```diff
@@ -1,4 +1,4 @@
-FROM ubuntu:focal
+FROM ubuntu:jammy
```
azeey, line 1: The officially supported platform for gz-sim9 is Ubuntu:noble. https://gazebosim.org/docs/latest/install/#supported-platforms

docker/Dockerfile.nightly
```diff
@@ -10,7 +10,7 @@ COPY docker/scripts/enable_nightly.sh scripts/enable_nightly.sh
 RUN scripts/enable_nightly.sh
 
 RUN apt-get update \
- && apt-get install -y \
+ && apt-get install --no-install-recommends -y \
```
azeey, line 13: Have you verified this works? My recollection is that some of our packages, such as `gz-*-python` and `gz-*-cli` are suggested/recommended packages and my not install if we add `--no-install-recommends`.

docker/scripts/install_common_deps.sh
```diff
@@ -17,16 +17,17 @@ sudo apt-get install --no-install-recommends -y \
   cppcheck \
   curl \
   git \
-  g++-8 \
-  pkg-config \
+  g++11 \
```
azeey, line 20: We can just install the default version
```suggestion
  g++ \
```

docker/README.md
```diff
@@ -73,21 +75,21 @@ distribution using debians.
    image of Gazebo Garden:
 
     ```
-    ./build.bash gz-garden ./Dockerfile.gz
+    ./build.bash gz ./Dockerfile.gz
```
azeey, line 78: The instruction above says the first argument must be the name of the Gazebo distribution. Should this change be reverted?

ntfshard: Guess I used auto-replacement

I'll check everything and send update

docker/run.bash
```diff
@@ -40,7 +31,6 @@ docker run -it \
   -v "/etc/localtime:/etc/localtime:ro" \
   -v "/dev/input:/dev/input" \
   --rm \
-  --gpus all \
```
ntfshard, line 43: after this changes this script started working fine on a Ubuntu 22.04 host

docker/README.md
```diff
@@ -46,13 +48,13 @@ use the Gazebo code found in the current source tree.
...
     ```
-    docker run -it gz-sim:nightly /bin/bash
+    docker run --rm -it gz-sim:nightly /bin/bash
     ```
 
 4. Alternatively, you can directly run Gazebo using
 
     ```
-    ./docker/run.bash gz-sim:nightly gz-sim-server -v 4
+    ./docker/run.bash gz-sim:nightly gz sim -v 4
```
ntfshard, line 57: I tested changes on this line

docker/run.bash
azeey: I think this is needed to get GPU acceleration when using NVIDIA GPUs.  Maybe leave it as is and add a statement below to remove it if not using a GPU?

docker/run.bash
```diff
@@ -18,17 +18,8 @@ ARGS=("$@")
 # This is necessary so Gazebo can create a context for OpenGL rendering
 # (even headless).
 XAUTH=/tmp/.docker.xauth
-if [ ! -f $XAUTH ]
```
azeey, line 21: Was this change necessary?

docker/README.md
```diff
@@ -75,7 +75,7 @@ distribution using debians.
    image of Gazebo Garden:
 
     ```
-    ./build.bash gz ./Dockerfile.gz
+    ./build.bash gz-ionic ./Dockerfile.gz
```
azeey, line 78: Okay, but now we need to update the instruction right above to say "Gazebo Ionic"

azeey (commented):
Thanks for the changes. I have a few more comments.

docker/run.bash
ntfshard: Yes, previous code not working, here is example on a system with empty `/tmp`:
```
xauth:  /tmp/.docker.xauth not writable, changes will be ignored
xauth: (argv):1:  unable to read any entries from file "(stdin)"
chmod: changing permissions of '/tmp/.docker.xauth': Operation not permitted
Authorization required, but no authorization protocol specified
```

and previous code somehow creates `/tmp/.docker.xauth` as a directory

docker/run.bash
ntfshard: I'll roll-back. Not sure how to dispatch automatically it for a GPU/noGPU cases. In my environment (VirtualBox) it leads to error:
`docker: Error response from daemon: could not select device driver "" with capabilities: [[gpu]].`

docker/README.md
ntfshard: I'll roll-back it here to garden and lines below too, due to `build.bash` not working properly with Ionic AFAIU. It executes `Dockerfile.gz`, which is based on `nvidia/opengl:1.2-glvnd-devel-ubuntu20.04` image.

docker/run.bash
ntfshard: It still quite simple to verify. But results can be false positive (it will be mentioned above error, but gui will work) if  `xhost +local` was executed before (not sure, maybe it should be local:root, it's still part of ancient magic)

ntfshard:
> Would it additionally make sense to introduce a CI job to make sure that this docker file can build on, for instance, a nightly basis?

Added, thanks to @skorobogatydmitry for help

ntfshard:
Folks, can we wrap-up this PR? This PR fixes nightly docker build, which seems almost nobody using.

@iche033 @mjcarroll @azeey

docker/scripts/install_common_deps.sh
```diff
@@ -17,16 +17,17 @@ sudo apt-get install --no-install-recommends -y \
...
-  clang-tidy-6.0 \
-  python-yaml \
-  libclang-6.0-dev
+  clang-tidy \
+  python3-yaml
 
 sudo apt-get install --no-install-recommends -y \
   libbenchmark-dev \
-  libbenchmark1
+  libbenchmark1.8.3
```
azeey, line 30: I think we can get rid of `libbenchmark1.8.3`. It should be automatically installed with `libbenchmark-dev`

docker/run.bash
azeey: I just tried this locally and it didn't work for me. Running
```
XAUTH=/tmp/.docker.xauth
xauth nlist $DISPLAY | sed -e 's/^..../ffff/' | xauth -f $XAUTH nmerge -
```
I get 
```
xauth:  file /tmp/.docker.xauth does not exist
```

azeey (changes requested):

docker/run.bash
ntfshard: I double checked and both (old and new) implementations are working fine. AFAIR old version initially didn't work. Maybe xhost +x has some permanent effect(still applied after reboot)
Tried your experiment, just run a command outside of the script and got the same error message. 

TBH, it's not my area of expertise, so I'll rollback it with minor change from :0 to $DISPLAY

azeey (approved):

docker/scripts/install_common_deps.sh
azeey: `libbenchmark` is an optional dependency. There are tests that use it, but they're disabled if `libbenchmark` is not installed https://github.com/gazebosim/gz-sim/blob/c298007066eb3a0a4b260884fe9a1148903e7bc3/test/benchmark/CMakeLists.txt#L4

So I think we should keep it.

ntfshard: Reverted, commend added

kscottz:
@mjcarroll @azeey Can we get this merged? The user has already bumped us twice and has been waiting patiently. The only thing missing AFAICT is approval from @mjcarroll.

azeey:
CI failures seem unrelated to this PR. Merging!

PR #1342, opened 2022-02-16
Support disabling pose publisher from publishing top level model pose

iche033:
Signed-off-by: Ian Chen <ichen@osrfoundation.org>

# 🦟 Bug fix

## Summary

The pose-publisher system uses only one parameter, `publish_nested_model_pose`, to determine whether or not top level model **and** nested model poses should be published. This PR adds another parameter, `publish_model_pose`, to let users decouple pose publishing behavior between nested models and top level models.

The changes should preserve existing behavior. That is, if `publish_nested_model_pose` is set to true, the top level model pose  should still be published unless the user explicitly disables this by setting `publish_model_pose` to false. We should consider changing the behavior in garden.

A typical use case is to disable publishing top level model's poses (ground truth data) but still allow nested model poses to be available to the user.

## Checklist
- [x] Signed all commits for DCO
- [x] Added tests
- [ ] Updated documentation (as needed)
- [ ] Updated migration guide (as needed)
- [ ] `codecheck` passed (See [contributing](https://ignitionrobotics.org/docs/all/contributing#contributing-code))
- [ ] All tests passed (See [test coverage](https://ignitionrobotics.org/docs/all/contributing#test-coverage))
- [ ] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Aignitionrobotics+repo%3Aosrf%2Fsdformat+archived%3Afalse+) to support the maintainers

**Note to maintainers**: Remember to use **Squash-Merge** and edit the commit message to match the pull request summary while retaining `Signed-off-by` messages.

Issue #2321, opened 2024-02-21
PosePublisher's "publish_nested_model_pose" parameter is confusing


azeey:
The `PosePublisher` system has a parameter for publishing the pose of the model that contains the plugin. This parameter is called `publish_nested_model_pose`: https://github.com/gazebosim/gz-sim/blob/633ce72171e27e83bf2e0292c9998e036d5da3fc/src/systems/pose_publisher/PosePublisher.hh#L46-L49

However, anyone familiar with nested models in SDFormat would expect that the parameter is referring to publishing the pose of child (nested) models inside the model that contains the `PosePublisher` plugin.

I actually think this feature was a bug, but it has been used in many places, e.g. SubT, to provide ground truth. I propose we add a new parameter that allows just publishing the model's absolute pose (as opposed to relative to another parent model if it's nested) since that seems something a lot of users need (https://github.com/gazebosim/ros_gz/issues/172, https://github.com/gazebosim/ros_gz/issues/410).

Extra context:
https://github.com/gazebosim/gz-sim/pull/1342 added `<publish_model_pose>` that would allow users to disable publishing the pose of the top level model. i.e, if `publish_model_pose=false` and `publish_nested_model_pose=true`, the PosePublisher only publishes the child (nested) model poses.

PR #2697, opened 2024-12-10
Fix publishing only top level model pose in pose publisher

### LINKED ISSUE #2690: Bug: Unable to Get Model Pose Info When publish_model_pose Is Set to True in Gazebo Harmonic
## Environment
* OS Version: Ubuntu 22.04
* Source or binary build: Binary build (Gazebo Harmonic)
  
  

## Description
* Expected behavior:
I am trying to get the model position info with respect to the Gazebo world frame. Based on my understanding, the pose publisher system should be used for this purpose. However, when I call the system with the publish_model_pose parameter set to true, I am unable to get the pose information. I checked it using the command:
gz topic -e -t "topic name"
After examining the source code of the system, I noticed that the publish_model_pose parameter is not being used. Instead, the publish_nested_model parameter is being checked, and it is set to false by default. When I changed the publish_nested_model parameter to true, I was successfully able to get the model pose information.
It seems that the nested_model_pose is used when there is no model_pose parameter. However, there doesn't appear to be a conditional check for this in the code. I believe this condition may be missing and could be causing the issue.

* Actual behavior: 
The model pose is not publishing unless you set True for publish_nested_model_pose
## Steps to reproduce

1. Use pose publisher system in a basic gazebo world.
2. Set the publish_model_pose parameter to true.
3. Try to get the model pose information via the appropriate topic.
4. Observe that no pose information is received.
5. Change the publish_nested_model parameter to true to successfully retrieve the model pose info.

## Output
Using pose publisher system with correct parameters
Screenshot from 2024-12-02 16-32-39

The missing if condition part in code
Screenshot from 2024-12-02 16-28-05

iche033:
# 🦟 Bug fix

Fixes #2690

## Summary

The pose publisher system should now publish top level model pose with these settings:

```xml
        <publish_model_pose>true</publish_model_pose>
        <publish_nested_model_pose>false</publish_nested_model_pose>
```

This PR preserves the behavior and backward compatibility logic that were added in #1342, i.e.

1. The system should only publish nested model pose with these params:

    ```xml
            <publish_model_pose>false</publish_model_pose>
            <publish_nested_model_pose>true</publish_nested_model_pose>
    ```

2. Backward behavior - if `<publish_model_pose>` is not specified, it defaults to the same value as `<publish_nested_model_pose>` - we should remove this behavior in gz-sim10 / jetty. I added a todo note for this.

## Checklist
- [ ] Added tests

### DIFF
--- a/src/systems/pose_publisher/PosePublisher.cc
+++ b/src/systems/pose_publisher/PosePublisher.cc
@@ -204,6 +204,8 @@ void PosePublisher::Configure(const Entity &_entity,
 
   // for backward compatibility, publish_model_pose will be set to the
   // same value as publish_nested_model_pose if it is not specified.
+  // todo(iche033) Remove backward compatibility and decouple model and
+  // nested model pose parameter value in gz-sim10
   this->dataPtr->publishModelPose =
     _sdf->Get<bool>("publish_model_pose",
         this->dataPtr->publishNestedModelPose).first;
@@ -394,9 +396,6 @@ void PosePublisherPrivate::InitializeEntitiesToPublish(
         (collision && this->publishCollisionPose) ||
         (sensor && this->publishSensorPose);
 
-    // for backward compatibility, top level model pose will be published
-    // if publishNestedModelPose is set to true unless the user explicity
-    // disables this by setting publishModelPose to false
     if (isModel)
     {
       if (parent)
@@ -404,10 +403,8 @@ void PosePublisherPrivate::InitializeEntitiesToPublish(
         auto nestedModel = _ecm.Component<components::Model>(parent->Data());
         if (nestedModel)
           fillPose = this->publishNestedModelPose;
-      }
-      if (!fillPose)
-      {
-        fillPose = this->publishNestedModelPose && this->publishModelPose;
+        else
+          fillPose = this->publishModelPose;
       }
     }
 
--- a/test/integration/pose_publisher_system.cc
+++ b/test/integration/pose_publisher_system.cc
@@ -778,8 +778,7 @@ TEST_F(PosePublisherTest,
 
   poseMsgs.clear();
 
-  // test that the pose publisher plugin can be loaded by a nested model
-  // by subscribe to the its topic
+  // subscribe to the pose publisher
   transport::Node node;
   node.Subscribe(std::string("/model/test_publish_only_model_pose/pose"),
                  &poseCb);
--- a/test/integration/pose_publisher_system.cc
+++ b/test/integration/pose_publisher_system.cc
@@ -761,5 +761,52 @@ TEST_F(PosePublisherTest,
     auto p = msgs::Convert(poseMsgs[0]);
     EXPECT_EQ(expectedEntityPose, p);
   }
+}
 
+/////////////////////////////////////////////////
+TEST_F(PosePublisherTest,
+       GZ_UTILS_TEST_DISABLED_ON_WIN32(ModelPoseOnly))
+{
+  // Start server
+  ServerConfig serverConfig;
+  serverConfig.SetSdfFile(std::string(PROJECT_SOURCE_PATH) +
+      "/test/worlds/pose_publisher.sdf");
+
+  Server server(serverConfig);
+  EXPECT_FALSE(server.Running());
+  EXPECT_FALSE(*server.Running(0));
+
+  poseMsgs.clear();
+
+  // test that the pose publisher plugin can be loaded by a nested model
+  // by subscribe to the its topic
+  transport::Node node;
+  node.Subscribe(std::string("/model/test_publish_only_model_pose/pose"),
+                 &poseCb);
+
+  // Run server
+  unsigned int iters = 1000u;
+  server.Run(true, iters, false);
+
+  // Wait for messages to be received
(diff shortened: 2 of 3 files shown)


azeey:
This is related to https://github.com/gazebosim/gz-sim/issues/2321. There, I proposed adding a new parameter to publish absolute poses. Any thoughts on that?

iche033:
> This is related to #2321. There, I proposed adding a new parameter to publish absolute poses. Any thoughts on that?

Setting `<publish_model_pose>true</publish_model_pose>` should now publish absolute poses for the top level model. I remember that was also the intension in #1342 (in addition to letting users disable publishing top model poses) but I didn't add a test for this and so there was this bug. 

The confusing part currently is how `publish_nested_model_pose` can still be used to determine if the top level model pose is published or not (if `publish_model_pose` is not specified). This was so that it does not break applications like subt. I propose to decouple `publish_model_pose` and `publish_nested_model_pose` completely in gz-sim10, i.e. `publish_nested_model_pose` should not affect whether or not absolute poses are published - it should only affect nested models and ignored if nested models are not present.

azeey:
> Setting <publish_model_pose>true</publish_model_pose> should now publish absolute poses for the top level model.

This is true, but if the plugin is inside a nested model, it would publish relative poses right? I think having a separate flag to publish the absolute pose regardless of the hierarchy of the model would be good to have. I think it would be less confusing. 

But I see now that this PR is fixing a bug and I don't want to request a new feature here. I'll keep #2321 open, but won't block this PR.


> I propose to decouple publish_model_pose and publish_nested_model_pose completely in gz-sim10, i.e. publish_nested_model_pose should not affect whether or not absolute poses are published - it should only affect nested models and ignored if nested models are not present.

Agreed!

azeey:
I'm confused about #2690 though. It's saying
```xml
  <publish_model_pose>true</publish_model_pose>
  <publish_nested_model_pose>true</publish_nested_model_pose>
```

doesn't work, but looking at the code, I would expect it to work. @Kametor do you have an example SDF file we can try?

iche033:
> I'm confused about #2690 though. It's saying
> 
> ```
>   <publish_model_pose>true</publish_model_pose>
>   <publish_nested_model_pose>true</publish_nested_model_pose>
> ```
> 
> doesn't work, but looking at the code, I would expect it to work. @Kametor do you have an example SDF file we can try?

This line prevented the (non-nested) model pose to be published:

https://github.com/gazebosim/gz-sim/blob/f0cde0574b357fba3cb9ee3f22df0108bda44db1/src/systems/pose_publisher/PosePublisher.cc#L410

the logic requires both `publish_nested_model_pose` and `publish_model_pose` and  to be true in order to publish the model pose. 

I used the [pose_publisher.sdf](https://github.com/gazebosim/gz-sim/blob/gz-sim9/examples/worlds/pose_publisher.sdf) example world for manual testing.

iche033:
>  I think having a separate flag to publish the absolute pose regardless of the hierarchy of the model would be good to have.

ah ok I see. Yes that sounds good to me.

azeey:
I see. For some reason, I misread it as "it's not working when both parameters are set to true".

src/systems/pose_publisher/PosePublisher.cc
```diff
@@ -394,20 +396,15 @@ void PosePublisherPrivate::InitializeEntitiesToPublish(
...
       if (parent)
       {
         auto nestedModel = _ecm.Component<components::Model>(parent->Data());
         if (nestedModel)
           fillPose = this->publishNestedModelPose;
-      }
-      if (!fillPose)
-      {
-        fillPose = this->publishNestedModelPose && this->publishModelPose;
+        else
```
azeey, line 406: So, if I understand this correctly, the truth table is:
   | `publish_model_pose` | `publish_nested_model_pose` | Top level model is published |
  |-- | -- | -- | 
  | T | T | T |
  | T | F | T |
  | T | X | F |
  | F | T | F |
  | F | F | F |
  | F | X | F |
  | X | T | T (This is what we want to change in gz-sim10?)|
  | X | F | F |
  | X | X | F |
  


   whereas before, it was

   | `publish_model_pose` | `publish_nested_model_pose` | Top level model is published |
  |-- | -- | -- | 
  | T | T | T |
   | T | F | F (This is the only difference) |
  | T | X | F |
  | F | T | F |
  | F | F | F |
  | F | X | F |
  | X | T | T |
  | X | F | F |
  | X | X | F |

azeey (approved):

src/systems/pose_publisher/PosePublisher.cc
iche033: yes my interpretation is the same as what's described in the tables above.  

The code change in gz-sim10 would be:

```diff

   this->dataPtr->publishModelPose =
     _sdf->Get<bool>("publish_model_pose",
-        this->dataPtr->publishNestedModelPose).first;
+        this->dataPtr->publishdModelPose).first;
```

and the entry in that table would become:

  | `publish_model_pose` | `publish_nested_model_pose` | Top level model is published |
  |-- | -- | --|
  |X | T | F|

iche033:
@mergify backport gz-sim8

PR #2725, opened 2025-01-22
LinearBatteryPlugin stability improvement

ntfshard:
# 🦟 Bug fix

Fixes #<NUMBER>

## Summary
This PR contain 3 different changes
* LinearBatteryPlugin: `startDraining` should be atomic due to we are changing it in callback which can be called from other thread (valgrind --tool=drd or helgrind)
* LinearBatteryPlugin: In a C++ each type has own [alignment](https://en.cppreference.com/w/cpp/language/object#Alignment) and rule of thumb: sort fields in order from bigger to smaller to avoid creation padding between fields. For structures we can add compiler option [-Wpadded](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wpadded) to get warnings, but we can't do it for a project due to reports not only padding between fields but also in case of inheritance (and we have some case with quite base types). So I propose to reorder some members to avoid such padding
* battery_plugin: in this test we had a logical race condition: when we publishing message, propagation can took some time. And if we are starting simulation immediately in some cases(e.g. VM or overloaded CPU) we can get couple of iterations of simulation until we receive the message. So it's more durable to send message and spin a little bit to be sure of delivery

It's quite simple to add some keys to compiler just by defining some [env.variables](https://cmake.org/cmake/help/latest/envvar/CXXFLAGS.html) before a clean build execution

## Checklist
- [ ] Added tests

### DIFF
--- a/test/integration/battery_plugin.cc
+++ b/test/integration/battery_plugin.cc
@@ -400,22 +400,25 @@ TEST_F(BatteryPluginTest, GZ_UTILS_TEST_DISABLED_ON_WIN32(PowerDrainTopic))
   // Send a message on one of the <power_draining_topic> topics, which will
   // start the battery draining when the server starts again.
   gz::transport::Node node;
-  auto pub = node.Advertise<msgs::StringMsg>("/battery/discharge");
+  auto discharge_pub = node.Advertise<msgs::StringMsg>("/battery/discharge");
   msgs::StringMsg msg;
-  pub.Publish(msg);
+  discharge_pub.Publish(msg);
 
   // Run the server again.
   server.Run(true, 100, false);
 
-  // The state of charge should be <1, since the battery has started
-  // draining.
-  double stateOfCharge = batComp->Data();
-  EXPECT_LT(batComp->Data(), 1.0);
-
   // Send a message on one of the <stop_power_draining_topic> topics, which
   // will stop the battery draining when the server starts again.
-  auto pub2 = node.Advertise<msgs::StringMsg>("/battery/stop_discharge");
-  pub2.Publish(msg);
+  auto stop_pub = node.Advertise<msgs::StringMsg>("/battery/stop_discharge");
+  stop_pub.Publish(msg);
+
+  // Run the server a little bit to allow msg be propagated inside plugin
+  server.Run(true, 50, false);
+
+  // The state of charge should be <1, since the battery has started
+  // draining.
+  const double stateOfCharge = batComp->Data();
+  EXPECT_LT(batComp->Data(), 1.0);
 
   // Run the server again.
   server.Run(true, 100, false);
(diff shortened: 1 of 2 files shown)


test/integration/battery_plugin.cc
```diff
@@ -400,23 +400,26 @@ TEST_F(BatteryPluginTest, GZ_UTILS_TEST_DISABLED_ON_WIN32(PowerDrainTopic))
   // Send a message on one of the <power_draining_topic> topics, which will
   // start the battery draining when the server starts again.
   gz::transport::Node node;
-  auto pub = node.Advertise<msgs::StringMsg>("/battery/discharge");
+  auto discharge_pub = node.Advertise<msgs::StringMsg>("/battery/discharge");
```
caguero, line 403: Nit: Do you mind to use camelCase here?

test/integration/battery_plugin.cc
```diff
@@ -400,23 +400,26 @@ TEST_F(BatteryPluginTest, GZ_UTILS_TEST_DISABLED_ON_WIN32(PowerDrainTopic))
...
   msgs::StringMsg msg;
-  pub.Publish(msg);
+  discharge_pub.Publish(msg);
 
   // Run the server again.
   server.Run(true, 100, false);
 
+  // Send a message on one of the <stop_power_draining_topic> topics, which
+  // will stop the battery draining when the server starts again.
+  auto stop_pub = node.Advertise<msgs::StringMsg>("/battery/stop_discharge");
```
caguero, line 412: Nit: Do you mind to use camelCase here?

caguero (changes requested):

caguero (approved):
Thanks for the patch! Agreed on your 1) and 3) points. About 2), not sure about the impact of that optimization because there are other aspects to be considered as well (are members used together, etc.). We normally don't consider that type of optimization when programming. In any case, it's absolutely fine your proposal of rearranging the member variables as I think is harmless, just saying that without seeing numbers I'm not 100% convinced about its impact :)

In any case, this topic is super interesting and we usually don't pay much attention to it. If you're interested in these kind of optimizations we could try to explore them in more depth trying to consider alignments with cache lines, etc. and measure the impact of the changes.

ntfshard:
> alignments with cache lines, etc. and measure the impact of the changes.

It's not about alignment with cache lines (for this we have [alignas](https://en.cppreference.com/w/cpp/language/alignas) and it's more about [false-sharing](https://en.wikipedia.org/wiki/False_sharing))

It's more about memory throughput. For a big structures or if we have just few instances it's maybe has lower impact.
In this particular case, it was _456_ and became _432_ bytes. So it's -5% (measured with [hack ](https://gist.github.com/Bktero/3fb41a617588b914317af6d678e08c54)).

Still optimization is a very big topic, I done it here maybe almost automatically, and TBH, I more interested in more convenient and productive usage of simulator, but my resources as a contributor are very limited

iche033 (approved):

PR #2707, opened 2024-12-26
Visualze Frustum

BA-Utkarsh:
# 🎉 New feature

## Summary
This PR mainly adds the visualization of Frustum.
We could see it was present in gazebo classic and from gazebo garden onwards the plugin/feature is not available.

## Test it
$ Build gazebo from source.
$ . install/setup.sh
$ gz sim examples/worlds/visualize_frustum.sdf

## Test Ref images,
1. Play the simulation.
frustum-1

2. Select the topic from scroll down.
frustum-2

3. Refresh it to get the "logical_camera/frustum" topic.
frustum-3

4. Subcriibed to "logical_camera/frustum".
frustum-4

5. Final output
frustum-6

## Checklist
- [x] Signed all commits for DCO
- [x] Added tests
- [ ] Added example and/or tutorial
- [ ] Updated documentation (as needed)
- [ ] Updated migration guide (as needed)
- [ ] Consider updating Python bindings (if the library has them)
- [ ] `codecheck` passed
- [ ] All tests passed
- [ ] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Agazebosim+archived%3Afalse+) to support the maintainers

## Supporting PRs,
- https://github.com/gazebosim/gz-sensors/pull/491
- https://github.com/gazebosim/gz-rendering/pull/1095

PR #2739, opened 2025-01-28
Improvement for Visualize lidar plugin

ntfshard:
# 🦟 Bug fix

Fixes #<NUMBER>

## Summary
As was discovered in PR #2707, this code a potential for improvement/optimization. And to avoid copying suboptimal code to another plugins I propose this changes:
 * For `VisualizeLidarPrivate` reorder members to avoid padding between it. Enabling `-Wpadded` for a whole project is not possible right now, due to warnings in some high-level headers
 * `VisualizeLidarPrivate` class actually do not need a `msgs::LaserScan msg` member. Everything, including copying points, are happening in callback
 * Avoiding redundant copying with const references
 * Typos and alignment fix
 * Minor C++ fixes, like removed redundant default arguments or used better api

## Checklist
- [ ] Added tests


ahcorde (approved):

src/gui/plugins/visualize_lidar/VisualizeLidar.cc
```diff
@@ -453,35 +448,26 @@ void VisualizeLidar::OnScan(const msgs::LaserScan &_msg)
   std::lock_guard<std::mutex> lock(this->dataPtr->serviceMutex);
   if (this->dataPtr->initialized)
   {
-    this->dataPtr->msg = std::move(_msg);
```
ntfshard, line 456: Also here is quite interesting moment I suppose. `_msg` is const reference. In this case move can't work, it will be copy 
operation https://godbolt.org/z/fjdrKcn9a

Not sure about shenanigans inside `msgs::*` but if it's a fair deep copy it looks like a quite wasteful operation, especially together with `...SetPoints(std::vector<double>(_msg.ranges().begin()....` where we have copy from msg to vector, which will be passed by const reference to SetPoint which also should involve copying data.

I can be wrong, feel free to correct me

azeey: It's a typical pattern to copy/move messages to avoid blocking the service/topic callback. However, since the mutex is held at the top of the function, this wasn't really accomplishing that. The service would have still been blocked. For the typical pattern to work, we'd also need to be processing the copied message in another thread. You're also right that there woudn't have been a move since it's a const ref.

Overall, the change you've made just to use `_msg` everywhere makes sense.

src/gui/plugins/visualize_lidar/VisualizeLidar.cc
```diff
@@ -109,6 +100,10 @@ inline namespace GZ_SIM_VERSION_NAMESPACE
     /// maxVisualRange
     public: std::mutex serviceMutex;
 
+    /// \brief Visual type for lidar visual
+    public: rendering::LidarVisualType visualType{
```
azeey, line 104: I'm not sure if recording this is necessary. Yes, there's probably padding added, but I don't think that would make a perceptible impact on performance. I'd rather avoid making the change to reduce the chances of merge conflicts down the road. If you think there's indeed a significant performance impact, I'm open to hear your reasoning.

src/gui/plugins/visualize_lidar/VisualizeLidar.cc
```diff
@@ -150,18 +145,18 @@ void VisualizeLidar::LoadLidar()
     return;
 
   // assume there is only one engine loaded
-  auto engineName = loadedEngNames[0];
+  const auto& engineName = loadedEngNames[0];
```
azeey, line 148: nit:
```suggestion
  const auto &engineName = loadedEngNames[0];
```

src/gui/plugins/visualize_lidar/VisualizeLidar.cc
iche033: +1, yeah we usually copy the msg and does the processing in another thread to avoid blocking the callback. In this case, it's already processing the msg contents so we shouldn't need to store a copy of the msg in `this->dataPtr->msg` since it's not used anywhere else.

ntfshard: Maybe we should add PROFILE macro to callbacks? But maybe it took a fraction of second and can't be too visible on a live top-down view. Not sure

src/gui/plugins/visualize_lidar/VisualizeLidar.cc
ntfshard: With this [gist](https://gist.github.com/Bktero/3fb41a617588b914317af6d678e08c54) we can check size of class before and after reordering. It will be 224 -> 216 bytes. Around 4% difference doesn't looks big, especially for a couple of instances of structure per program.
But from other side, typical modern CPU has a 64 bytes cache line. And padding in 7 bytes (compiler reporting this value in Wpadded warning) feels like a wasting quite significant piece of very limited cache memory.
Surely, it's quite controversial question and structure in both cases requires the same amount of cache lines and it's not a hot path of execution. But I don't like to waste resources even little by little

About possible merge conflicts, this plugins seems has a very low frequency of changes. I suppose it shouldn't be a problem with it in future.

src/gui/plugins/visualize_lidar/VisualizeLidar.cc
```diff
@@ -150,18 +145,18 @@ void VisualizeLidar::LoadLidar()
     return;
 
   // assume there is only one engine loaded
```
ntfshard, line 147: I realized that code around can be replaced by call `rendering::sceneFromFirstRenderEngine`

src/gui/plugins/visualize_lidar/VisualizeLidar.cc
azeey: Okay. it feels like premature optimization, but I won't block on this. You're right that this file probably won't be updated too often.

PR #2745, opened 2025-01-29
Update install link

EmmanuelMess:
This changes the "install" link in the README, to link to the docs, instead of the api, as it has better information, such as a link to "ROS/Gazebo installation" which is important for people using ROS.

In the old link I couldn't find the table (which is at  "ROS/Gazebo installation") and I also couldn't find where the "api" instalation page is located to add the link. So I updated the link in the readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,7 +78,7 @@ introspection and control.
 
 # Install
 
-See the [installation tutorial](https://gazebosim.org/api/sim/9/install.html).
+See the [installation tutorial](https://gazebosim.org/docs/latest/getstarted/).
 
 # Usage
 
--- a/README.md
+++ b/README.md
@@ -78,7 +78,7 @@ introspection and control.
 
 # Install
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -78,7 +78,7 @@ introspection and control.
 
 # Install
 
-See the [installation tutorial](https://gazebosim.org/api/sim/9/install.html).
+See the [installation tutorial](https://gazebosim.org/docs/latest/getstarted/).
```
azeey, line 81: I'd rather not completely get rid of the other link since it might be useful for folks who are using libgz-sim as a library.  Maybe we can say, 
```suggestion
For installing Gazebo, see the [getting started guide](https://gazebosim.org/docs/latest/getstarted/). If you want to use `libgz-sim` as a library, see the [installation tutorial](https://gazebosim.org/api/sim/9/install.html)
```

azeey (approved):

PR #2638, opened 2024-10-02
Break out server constructor

nkoenig:
# 🎉 New feature

## Summary

Moves a section of the `Server` constructor code to a separate function. The primary reason is to make the [eventual download models in parallel PR](https://github.com/gazebosim/gz-sim/compare/gz-sim8...download_parallel_8) easier to review. There is a side benefit of breaking up a large function into two smaller functions.

I also replaced the static `DefaultWorld` with a default constructed `sdf::World` object. This removes a `static` object, and will make sure the default world uses the latest sdf version.

Commits:
1. Move the code to a new function
2. Patches the moved code so that it builds.
3. Replace `DefaultWorld` with a default constructed `sdf::World` object

## Test it

Tests should pass, and everything should work as normal.

## Checklist
- [x] Signed all commits for DCO
- [ ] Added tests
- [ ] Added example and/or tutorial
- [x] Updated documentation (as needed)
- [ ] Updated migration guide (as needed)
- [ ] Consider updating Python bindings (if the library has them)
- [x] `codecheck` passed (See [contributing](https://gazebosim.org/docs/all/contributing#contributing-code))
- [x] All tests passed (See [test coverage](https://gazebosim.org/docs/all/contributing#test-coverage))
- [ ] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Agazebosim+archived%3Afalse+) to support the maintainers

**Note to maintainers**: Remember to use **Squash-Merge** and edit the commit message to match the pull request summary while retaining `Signed-off-by` messages.

PR #2061, opened 2023-07-28
Added Mesh Moment of Inertia Calculator 

jasmeet0915:
# 🎉 New feature

Based on [this](http://sdformat.org/tutorials?tut=auto_inertial_params_proposal&cat=pose_semantics_docs&branch=jasmeet%2Fauto_inertial_params&repo=https%3A%2F%2Fgithub.com%2Fjasmeet0915%2Fsdf_tutorials) proposal.

## Summary
- The PR implements a Moment of Inertia Calculator based on the method described in [this](https://www.geometrictools.com/Documentation/PolyhedralMassProperties.pdf) document.
- The calculator works using the callback-based API from the sdformat made in [this](https://github.com/gazebosim/sdformat/pull/1304) PR.

## Demo & Usage Example

**Demo 1:** This demo shows the automatic inertia calculation feature on a rubber ducky model which is a non-convex mesh. On the left, we have the rubber ducky mesh with automatic calculations enabled and on the right, the mesh uses the default values.
> **Note:** The inertial values are due to the scale of the mesh. You can see the banana for scale in between the 2 ducks. A density value for the duck was used which was calculated by using the mass and volume data of the duck found online.

> **Note:** The `voxel_size` inertia param given in the snippet below is just for showing how the `<auto_inertia_params>` element could be used. This is not actually used by the calculator since we are not using a voxelization-based calculator.
> 
<details>
  <summary>SDF snippet for the duck mesh with auto inertial</summary>

  ```xml
<?xml version="1.0" ?>
<sdf version="1.6">
  <model name="duck">
    <link name="duck_link">
      <pose>0 0 0 0 0 0</pose>
      <inertial auto="true" />
      <collision name="duck_collision">
      	<pose>0 0 0 0 0 0</pose>
        <density>111.8</density>
        <auto_inertia_params>
          <gz:voxel_size>0.01</gz:voxel_size>
        </auto_inertia_params>
        <geometry>
          <mesh>
            <uri>meshes/duck_collider.dae</uri>
          </mesh>
        </geometry>
      </collision>
      <visual name="duck_visual">
        <pose>0 0 0 0 0 0</pose>
        <geometry>
          <mesh>
            <uri>meshes/duck.dae</uri>
          </mesh>
        </geometry>
      </visual>
    </link>
    <static>true</static>
  </model>
</sdf>
  ```
</details>

duck_mesh_inertia

[...]

PR #2754, opened 2025-01-31
Register MeshInertialCalculator when loading sim from an SDF string

iche033:
# 🎉 New feature

## Summary
MeshInertialCalculator was added in #2061. It was only set for sim loaded from SDF file. This PR turns it on for sim loaded from SDF string.

Note this PR is targeted at `gz-sim8` because it looks like there are a lot of changes between `gz-sim8` and `gz-sim9` due to https://github.com/gazebosim/gz-sim/pull/2638. After this is merged, we should merge `gz-sim8` forward to `gz-sim9`

## Checklist
- [ ] Added tests

### DIFF
--- a/src/ServerPrivate.cc
+++ b/src/ServerPrivate.cc
@@ -584,6 +584,13 @@ sdf::Errors ServerPrivate::LoadSdfRootHelper(const ServerConfig &_config)
 {
   sdf::Errors errors;
 
+  sdf::ParserConfig sdfParserConfig = sdf::ParserConfig::GlobalConfig();
+  sdfParserConfig.SetStoreResolvedURIs(true);
+  sdfParserConfig.SetCalculateInertialConfiguration(
+    sdf::ConfigureResolveAutoInertials::SKIP_CALCULATION_IN_LOAD);
+  MeshInertiaCalculator meshInertiaCalculator;
+  sdfParserConfig.RegisterCustomInertiaCalc(meshInertiaCalculator);
+
   switch (_config.Source())
   {
     // Load a world if specified. Check SDF string first, then SDF file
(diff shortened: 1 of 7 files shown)


iche033:
@azeey - just want to double check with you if there's any reason that MeshInertialCalculator was only enabled for sim loaded from an SDF file and not SDF string.

azeey:
It's probably an oversight. Looking at the code, it looks like there's some code duplication that led to it. I'd recommend moving the creation of `sdfParserConfig` out of the case statement so that the two branches can share the same `sdfParserConfig` object.

iche033:
> It's probably an oversight. Looking at the code, it looks like there's some code duplication that led to it. I'd recommend moving the creation of `sdfParserConfig` out of the case statement so that the two branches can share the same `sdfParserConfig` object.

done in [4af3ca6](https://github.com/gazebosim/gz-sim/pull/2754/commits/4af3ca62168c3e6c47735032c38236ae97ea97a3)

azeey (commented):
Mind adding a test for the string case?

iche033:
> Mind adding a test for the string case?

refactored `mesh_inertia_calculation.cc` in [bb85422](https://github.com/gazebosim/gz-sim/pull/2754/commits/bb85422a04f457d5a4309967446478bedd488934). Each test should now run with a server loaded from sdf file and a server loaded from sdf string.

azeey (approved):
LGTM!

src/ServerPrivate.cc
```diff
@@ -584,6 +584,13 @@ sdf::Errors ServerPrivate::LoadSdfRootHelper(const ServerConfig &_config)
 {
   sdf::Errors errors;
 
+  sdf::ParserConfig sdfParserConfig = sdf::ParserConfig::GlobalConfig();
+  sdfParserConfig.SetStoreResolvedURIs(true);
+  sdfParserConfig.SetCalculateInertialConfiguration(
+    sdf::ConfigureResolveAutoInertials::SKIP_CALCULATION_IN_LOAD);
```
scpeters, line 590: should this be `SAVE_CALCULATION` instead of `SKIP_CALCULATION_IN_LOAD`?  @iche033

scpeters: actually, maybe not for meshes

iche033: hmm yeah looks like mesh inertia should not be resolved in load.

Issue #2737, opened 2025-01-27
Decoupling the server from the client libraries


j-rivero:
While working on https://github.com/gazebo-release/gz-sim8-release/issues/8 I found some difficulties into splitting the packaging from client/server in gz-sim. Note that we mainly want to split the GUI libraries (specially those that needs Qt) into a separate package.

Reviewing the situation: there are two libraries that links against Qt:
```
/usr/lib/x86_64-linux-gnu/libgz-sim8-gui.so.8.6.0
	libQt5Widgets.so.5 => /lib/x86_64-linux-gnu/libQt5Widgets.so.5 (0x000074453e000000)
	libQt5Quick.so.5 => /lib/x86_64-linux-gnu/libQt5Quick.so.5 (0x000074453da00000)
	libQt5Qml.so.5 => /lib/x86_64-linux-gnu/libQt5Qml.so.5 (0x000074453d400000)
	libQt5Gui.so.5 => /lib/x86_64-linux-gnu/libQt5Gui.so.5 (0x000074453cc00000)
	libQt5Core.so.5 => /lib/x86_64-linux-gnu/libQt5Core.so.5 (0x000074453c600000)
	libQt5QmlModels.so.5 => /lib/x86_64-linux-gnu/libQt5QmlModels.so.5 (0x000074453d892000)
	libQt5Network.so.5 => /lib/x86_64-linux-gnu/libQt5Network.so.5 (0x000074453b454000)
...
/usr/lib/x86_64-linux-gnu/libgz-sim8-gz.so.8.6.0
	libQt5Core.so.5 => /lib/x86_64-linux-gnu/libQt5Core.so.5 (0x0000792777000000)
	libQt5Widgets.so.5 => /lib/x86_64-linux-gnu/libQt5Widgets.so.5 (0x0000792776000000)
	libQt5Quick.so.5 => /lib/x86_64-linux-gnu/libQt5Quick.so.5 (0x0000792775a00000)
	libQt5Qml.so.5 => /lib/x86_64-linux-gnu/libQt5Qml.so.5 (0x0000792775400000)
	libQt5Gui.so.5 => /lib/x86_64-linux-gnu/libQt5Gui.so.5 (0x0000792774c00000)
	libQt5QmlModels.so.5 => /lib/x86_64-linux-gnu/libQt5QmlModels.so.5 (0x0000792776aba000)
[...]

PR #2724, opened 2025-01-21
Standalone `gz model` executable

sauk2:
# 🎉 New feature

Part of https://github.com/gazebosim/gz-tools/issues/7

## Summary

This PR introduces a standalone `gz model` executable, following a similar approach to [`gz-transport`](https://github.com/gazebosim/gz-transport/pull/216). The new executable is integrated into `cmdmodel.rb.in`, replacing its existing functionality.

**The `gz sim` implementation remains unchanged in this PR and will be addressed in a subsequent update**. Additionally, the command-line tests (gz_TEST.cc and ModelCommandAPI_TEST.cc) are now configured to build after the `gz model` executable is created.

## Test it

Check for `UNIT_gz_TEST` and `UNIT_ModelCommandAPI_TEST`

## Checklist
- [x] Signed all commits for DCO
- [x] Added tests
- [x] Added example and/or tutorial
- [x] Updated documentation (as needed)
- [ ] Updated migration guide (as needed)
- [ ] Consider updating Python bindings (if the library has them)
- [x] `codecheck` passed (See [contributing](https://gazebosim.org/docs/all/contributing#contributing-code))
- [x] All tests passed (See [test coverage](https://gazebosim.org/docs/all/contributing#test-coverage))
- [ ] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Agazebosim+archived%3Afalse+) to support the maintainers

**Note to maintainers**: Remember to use **Squash-Merge** and edit the commit message to match the pull request summary while retaining `Signed-off-by` messages.

PR #2809, opened 2025-03-05
Standalone `gz sim` executable

sauk2:
# 🎉 New feature

Related to https://github.com/gazebosim/gz-tools/issues/7 and https://github.com/gazebosim/gz-sim/issues/2737

## Summary

This PR introduces a standalone executable for `gz sim` by following a similar approach to [`gz model`](https://github.com/gazebosim/gz-sim/pull/2724). The new executable is integrated into `cmdsim.rb.in`, replacing the existing functionality. It is also a step in the direction of decoupling the server from GUI components, allowing us to build `gz-sim` library without GUI support.

Two executables are created - `gz-sim-main` and `gz-sim-gui`. The `gz-sim-main` executable contains the code to launch the server in the same process and invoke `gz-sim-gui` in a separate process. For, Unix-based systems, `std::system()` is used, and for Windows, `CreateProcess()` from `windows.h` is used.

The majority of the behavior remains consistent, including:
- Launching the Server using `-s`
- Launching the GUI using `-g`
- Launching the Quickstart menu when no argument is provided

### Changes to the launch behaviour

- Launching with `-s` runs the server in the main thread.
- Launching with `-g` runs the GUI as a separate process.
- Running without `-s` or `-g` launches both the Server and GUI in separate threads. The Server instance is intentionally created non-blocking so its lifecycle can be controlled. A `SIGINT` closes both the server and GUI but in case the GUI is closed from the screen (through the window `x`), the server is shutdown manually. 

### Enabling/disabling GUI support

An option `ENABLE_GUI` is provided to enable/disable building with GUI. It is `ON` by default, and turning it `OFF` builds `libgz-sim<version>.so` without any link to `gz-gui` by not building `libgz-sim<version>-gui.so`. The executable `gz-sim-main` is also built without any link to `libgz-sim<version>-gui.so`, and `gz-sim-gui` executable is not built at all.

Additionally, a new job has been added CI to build `gz-sim` without GUI enabled.

## Test it

To test the executables, check for `UNIT_gz_TEST` and `INTEGRATION_log_system`.

To build `gz-sim` without GUI support, the library must be built with `-DENABLE_GUI=OFF` as a CMake argument. Check linkages using `ldd` for `libgz-sim<version>.so` and `gz-sim-main` to find no linkages to any GUI related libraries.

## Checklist
- [x] Added tests


j-rivero:
Before looking into the code: `gz-sim-sim` seems a weird name to my eyes :) although it can serve as a transitioning step to other naming or model.

Given the client  / server architecture I would expect to find something like: `gz-sim-client` and `gz-sim-server` separate standalone applications that can be run @sauk2 what do you think? We would need some separate that does not like to Qt libraries for packaging the server-only and get that into the official docker hub.

traversaro:
Thanks a lot @sauk2 ! 

I put together an hack (or let's call it a "proof of concept", to be more fancy) to understand if this new standalone executable could be used to be able to get a working `gz sim` on Windows and macOS (i.e. fixing https://github.com/gazebosim/gz-sim/issues/168 and https://github.com/gazebosim/gz-sim/issues/44), and it seems that indeed it is working fine, at least on Windows, while macOS is untested. For anyone curious the hack/poc is at https://github.com/traversaro/ign-gazebo/tree/sim-standalone-exe-win .

traversaro:
> For anyone curious the hack/poc is at https://github.com/traversaro/ign-gazebo/tree/sim-standalone-exe-win .

For clarity, there I used `tinyprocesslib` because it is a quick copy&paste from https://github.com/gazebosim/gazebo-classic/blob/b22c6e15e52299865b31093b8feebc9ca19e26e8/gazebo/gazebo_main.cc#L259 . However, probably it make more sense to use `gz::utils::Subprocess`, see https://gazebosim.org/api/utils/2/classgz_1_1utils_1_1Subprocess.html and https://github.com/gazebosim/gz-utils/pull/127 . By the way, I wonder if it could make sense to use two different process for server and client also on Linux, so that we avoid having different code paths for Linux vs macOS/Windows, or there is some advantage of using a single process for `gz sim` on Linux?

j-rivero:
> By the way, I wonder if it could make sense to use two different process for server and client also on Linux, so that we avoid having different code paths for Linux vs macOS/Windows, or there is some advantage of using a single process for gz sim on Linux?

That is a good point. Ideally we should go with one code path for all unless we find a good reason or a problem in a particular arch. Maybe wrong but I have always assumed that the ruby code currently spawns [two different processes via `Process.fork`  for the GUI and the server](https://github.com/gazebosim/gz-sim/blob/gz-sim9/src/cmd/cmdsim.rb.in#L526-L555) so we maintain the two separate processes although we change the way that they are coordinated and orchestratred.

sauk2:
@j-rivero and @traversaro Thanks a lot for all your feedback! Sorry for the delay in making changes and responding. I've moved this PR back to draft since I believe further discussion is needed before it's ready for merge.

Based on your comments, it made sense to handle the server and GUI as separate processes rather than separate threads. To that end, I've split them into two distinct CLI executables - `gz-sim-server` and `gz-sim-gui`, each responsible for its own flags. These are now invoked by a main CLI executable, `gz-sim-main`, which forwards the appropriate flags to the Server and GUI executables while launching them as separate subprocesses using the [`gz::utils::Subprocess`](https://github.com/gazebosim/gz-utils/blob/gz-utils3/include/gz/utils/Subprocess.hh) class. This executable is now available to `gz-tools` and the `cmdsim.rb` script. By maintaining a single outward-facing executable, process management is handled at the executable level rather than at the Ruby level, as it has been until now.

Handling the `-s` and `-g` flags to launch the server and GUI separately is straightforward, but the real challenge lies in orchestrating the full Server+GUI launch. This scenario can terminate in multiple ways - a `SIGINT` will cleanly shut down both processes but issues arise when the user manually closes the GUI as this does not signal the Server to shut down. Currently, I’m using the [`Alive()`](https://github.com/gazebosim/gz-utils/blob/fd0a40c09be4fd7b2a40d877bb56d2fa1c7607e3/include/gz/utils/Subprocess.hh#L178-L184) function to monitor both processes and if the GUI exits due to user input, the server is explicitly terminated as well.

Another issue is that logs are not streamed to the terminal in the current implementation. The [`Stdout()`](https://github.com/gazebosim/gz-utils/blob/gz-utils3/include/gz/utils/Subprocess.hh#L145-L159) function in `gz::utils::Subprocess` only provides output after the process completes, which is not ideal. One potential solution is exposing [`subprocess_read_stdout`](https://github.com/gazebosim/gz-utils/blob/fd0a40c09be4fd7b2a40d877bb56d2fa1c7607e3/include/gz/utils/detail/subprocess.h#L181-L193) within `gz::utils::Subprocess` to enable real-time output streaming. I’ve put together a [mock implementation](https://github.com/sauk2/gz-sim/blob/4013fb6577cc3cfe5aa0a684ccad87daa827c8a3/src/cmd/gz.hh#L104-L125) for testing and I’d appreciate any thoughts on how best to integrate this into the utility class.

Additionally, the server and GUI flags are currently defined within their respective executables. I’m exploring ways to extract these flags so they can be included in the main help message - one idea is to invoke each executable with `--help` and capture their stdout.

I’d appreciate any suggestions or feedback on these points. Thanks again!

traversaro:
Thanks a lot for the great work!

> Currently, I’m using the [`Alive()`](https://github.com/gazebosim/gz-utils/blob/fd0a40c09be4fd7b2a40d877bb56d2fa1c7607e3/include/gz/utils/Subprocess.hh#L178-L184) function to monitor both processes and if the GUI exits due to user input, the server is explicitly terminated as well.

We used a similar solution in Gazebo Classic on Windows, and it worked fine.

> Another issue is that logs are not streamed to the terminal in the current implementation. The [Stdout()](https://github.com/gazebosim/gz-utils/blob/gz-utils3/include/gz/utils/Subprocess.hh#L145-L159) function in gz::utils::Subprocess only provides output after the process completes, which is not ideal. One potential solution is exposing [subprocess_read_stdout](https://github.com/gazebosim/gz-utils/blob/fd0a40c09be4fd7b2a40d877bb56d2fa1c7607e3/include/gz/utils/detail/subprocess.h#L181-L193) within gz::utils::Subprocess to enable real-time output streaming. I’ve put together a [mock implementation](https://github.com/sauk2/gz-sim/blob/4013fb6577cc3cfe5aa0a684ccad87daa827c8a3/src/cmd/gz.hh#L104-L125) for testing and I’d appreciate any thoughts on how best to integrate this into the utility class.

Interesting. I do not have experience with subprocess.h, but with reproc and tiny-process-lib the subprocess output is automatically printed in the terminal that launched the main process. Perhaps there is some way to achieve this, without the need to manually process the stdout and stderr from the child processes?

> Additionally, the server and GUI flags are currently defined within their respective executables. I’m exploring ways to extract these flags so they can be included in the main help message - one idea is to invoke each executable with --help and capture their stdout.

Good point. Could it make sense to move the logic of those flags to a simple header-only library, and call it from both the main executable and server/client ones?

sauk2:
@traversaro and @j-rivero, Sorry for the delay in getting back to you, and thanks for the feedback!

I initially tried using `gz::utils::Subprocess`, but it was challenging to retrieve logs and manage termination properly. Instead, I switched to using the system() function which was much simpler to work with and appears to be cross-platform.

I've made a few changes that seem to work well. Instead of maintaining separate executables for `gz-sim-main` and `gz-sim-server`, I’ve merged them into a single executable while `gz-sim-gui` remains separate. The separation between the server and GUI is still maintained.

Now, the server process runs directly from the main executable and the GUI is launched as a separate process using `system()`. One of the challenges with having separate server and GUI executables was passing the correct flags to each. The `allow_extras()` function of `CLI11` allows additional arguments, but in many cases, it causes the positional argument to be misrecognized. To resolve this, I now launch the server directly from the main executable while constructing and forwarding the correct arguments to the GUI.

I still need to figure out how to make this work on Windows.

I’d be happy to get your feedback and do let me know if this approach works for you!

traversaro:
> I still need to figure out how to make this work on Windows, so it would be great to get some advice on this!

Sorry, I arrived late and I see that you already worked on this. To be honest I always used [tiny-process-lib](https://gitlab.com/eidheim/tiny-process-library) and [reproc](https://github.com/DaanDeMeyer/reproc), so even without using them directly you can check for their implementations to understand more.

src/cmd/CMakeLists.txt
```diff
@@ -8,10 +8,39 @@ target_link_libraries(${model_executable}
...
+target_link_libraries(gz
+  ${PROJECT_LIBRARY_TARGET_NAME}
+  gz-common${GZ_COMMON_VER}::gz-common${GZ_COMMON_VER}
+  gz-utils${GZ_UTILS_VER}::cli
+  gz-sim${PROJECT_VERSION_MAJOR}
+  gz-sim${PROJECT_VERSION_MAJOR}-gui
+)
+
+add_executable(gz-sim-gui gui_main.cc)
+target_link_libraries(gz-sim-gui gz)
```
j-rivero, line 25:
```suggestion
target_link_libraries(gz-sim-gui PRIVATE gz)
```

src/cmd/CMakeLists.txt
```diff
@@ -8,10 +8,39 @@ target_link_libraries(${model_executable}
...
+
+set(sim_executable gz-sim-main)
+add_executable(${sim_executable} sim_main.cc)
+target_link_libraries(${sim_executable} PUBLIC gz)
```
j-rivero, line 29:
```suggestion
target_link_libraries(${sim_executable} PRIVATE gz)
```

j-rivero (commented):
Going slowly through the changes:

Blackbox testing a little detected some differences with current gz behaviour:
 * Running launching a simulation and running `gz-gui-main` after it displays the screen to select default worlds. If I run current  `gz sim -g` this does not happen and the gui appears with the world that it is being run by the server.
 * If I use `gz-gui-main` from the command line, select a world, and close the GUI window (using the window manager X) , the command line is still running, not finished. 
 
We can address this problem later no need to do it in this PR but: the PR is an step in the right direction but the current approach does not isolate the gui/server in a way that we can package them separately.

src/cmd/sim_main.cc
```diff
@@ -0,0 +1,515 @@
...
+    "The default verbosity level is 1. Use -v\n"
+    "without arguments for level 3.\n")
+    ->expected(0, 1)
+    ->default_val(3);
+
+  // Dummy flags handled by gz-tools
+  app.add_flag("--force-version", "Use a particular library version.");
+  app.add_flag("--versions", "Show the available versions.");
+
+  app.add_flag("-g", opt->launchGui, "Run and manage Gazebo GUI");
```
j-rivero, line 410:
```suggestion
  app.add_flag("-g", opt->launchGui, "Run and manage only the Gazebo GUI");
```

src/cmd/sim_main.cc
```diff
@@ -0,0 +1,515 @@
...
+
+  app.add_flag_callback("-s",
+    [opt]{
+      opt->launchServer = true;
+      opt->launchGui = false;
+      opt->waitGui = 0;
+    },
+    "Run and manage Gazebo Server");
```
j-rivero, line 418:
```suggestion
    "Run and manage only the Gazebo Server (no GUI)");
```

src/cmd/sim_main.cc
```diff
@@ -0,0 +1,515 @@
...
+      }
+      _opt->networkSecondaries = _networkSecondaries;
+    },
+    "Number of secondary participants expected\n"
+    "to join a distributed simulation environment.")
+    ->needs(networkRoleOpt);
+
+  _app.add_flag("--record", _opt->record,
+                "Use logging system to record states and console\n"
+                "to default location, ~/.gz/sim/log.");
```
j-rivero, line 237:
```suggestion
                "messages to the default location, ~/.gz/sim/log.");
```

j-rivero:
Gave it a try on Windows, it builds nicely.

There is a problem with the initial screen probably unrelated to this PR.
image

gz-sim-exe runs fine for launch both and only the server. Same problem than in Linux, it does not tear down the server if I close the gui. gz-sim-gui works fine.

sauk2:
> Sorry, I arrived late and I see that you already worked on this. To be honest I always used [tiny-process-lib](https://gitlab.com/eidheim/tiny-process-library) and [reproc](https://github.com/DaanDeMeyer/reproc), so even without using them directly you can check for their implementations to understand more.

@traversaro thank you for pointing me in the right direction! I have moved to using the [`CreateProcess()`](https://learn.microsoft.com/en-us/windows/win32/procthread/creating-processes) function from `windows.h` to launch the GUI as a separate process.

> We can address this problem later no need to do it in this PR but: the PR is an step in the right direction but the current approach does not isolate the gui/server in a way that we can package them separately.

@j-rivero Thanks for your feedback and for trying out the functionality! 

> If I use gz-gui-main from the command line, select a world, and close the GUI window (using the window manager X) , the command line is still running, not finished.

I made some changes to try to fix this issue by running a non-blocking server and manually controlling its lifecycle. Now, when the GUI is closed, the server is shut down manually without having to worry about handling signals.  Do give it a look and let me know if you are still getting that problem.

> We can address this problem later no need to do it in this PR but: the PR is an step in the right direction but the current approach does not isolate the gui/server in a way that we can package them separately.

I have started working on this now by wrapping the GUI components within conditional statements. Hoping to push some commits on this soon!

traversaro:
> @traversaro thank you for pointing me in the right direction! I have moved to using the [`CreateProcess()`](https://learn.microsoft.com/en-us/windows/win32/procthread/creating-processes) function from `windows.h` to launch the GUI as a separate process. Let me know what you think of this approach!

That is great, thanks!

j-rivero:
> > We can address this problem later no need to do it in this PR but: the PR is an step in the right direction but the current approach does not isolate the gui/server in a way that we can package them separately.
> 
> I have started working on this now by wrapping the GUI components within conditional statements. Hoping to push some commits on this soon!

Thanks ! Let me clarify one detail: we do want to have the Gazebo server independent of using (and/or linking) any GUI component. This can be achieve by building Gazebo without the GUI (like the option you are adding) but we also should be able to get it done without disabling the GUI and providing independent binaries that we can package in different packages (i.e: gz-sim-server.deb / gz-sim-gui.deb) from a single build. Installing the gz-sim-server should bring a server only experience without all the GUI dependencies.

>  I made some changes to try to fix this issue by running a non-blocking server and manually controlling its lifecycle. Now, when the GUI is closed, the server is shut down manually without having to worry about handling signals. Do give it a look and let me know if you are still getting that problem.

Yes ! testing now works as expected on Linux. The only caveat is that invoking the GUI display the initial world selection window that I think make no sense to have it there.

j-rivero:
After speaking with the PMC yesterday we agree that the change in this PR is a bit risky to deploy it directly on a released distribution like Ionic, so we probably want it to go into a development version so it gets well testing during the release cycle before we ship it to the users. @sauk2sorry for the late decision, I can take care of migrating it the Jetty branch.

The other point to move this forward is to list what is still missing? From my personal testing:
  * Manual testing in Mac [blocker]
  * gz-sim-guii displays the world initial selection screen [non-merge-blocker]
  * gz-sim-gui input parameters referenced SDF files [non-merge-blocker]
  * Parameters are duplicated between gz-sim-gui / gz-sim-main [non-merge-blocker]

sauk2:
Thank you for the feedback! I agree that it would be best to include this in the next version. I appreciate your offering to change the target branch to `main`; please feel free to proceed with that. I'll address the review comments once that's done.

j-rivero:
Follow up in #2849

PR #2818, opened 2025-03-09
Add velocity limits component and method to Joint class (#2664)

### LINKED ISSUE #2664: Add API for getting Joint limits
## Desired behavior
Add methods Joint::VelocityLimits, Joint::EffortLimits, Joint::PositionLimits for getting limits, corresponding to the existing setter methods.

## Alternatives considered
I looked at getting these values from the ECM using components::JointVelocityLimitsCmd, components::JointEffortLimitsCmd, and components::JointPositionLimitsCmd. However, their headers all contain comments such as "It is expected that the physics plugin reads this component and sets the limit to the dynamics engine. After setting it, the data of this component will be cleared (i.e. the vector will have length zero)." So it is expected that these components will not contain the desired data.

## Additional context
I'm using gz-sim8 on Ubuntu 24.04.

mo-zaghloul:
# 🎉 New feature

Closes #2664  

## Summary  

This PR introduces the `JointVelocityLimits` component and starts integrating it into the `Joint` class and physics system. However, I need guidance on completing the physics implementation.  

### **Changes in this PR**  
- Added `JointVelocityLimits.hh` component to store joint velocity limits.  
- Integrated the component into `Joint.cc`.  
- Started integrating it into `Physics.cc` (incomplete, need guidance).  

## Test it  
The component has been added, but since the physics integration is incomplete, full testing isn't possible yet. Once I receive guidance on the correct approach, I will add appropriate tests.  

## Guidance Needed  
I am currently stuck on integrating velocity limits correctly into the physics system. Specifically, I would appreciate guidance on:  
- The best approach to retrieve and apply velocity limits within the physics update loop.  
- Ensuring that the component interacts correctly with the existing joint properties in simulation.  

## Checklist  
- [x] Added new component `JointVelocityLimits`  
- [ ] Completed physics system integration 
- [ ] Added tests (Will finalize after physics integration)  
- [ ] Updated documentation (Will finalize after implementation)  

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,5 @@
+{
+    "files.associations": {
+        "optional": "cpp"
+    }
+}
--- a/include/gz/sim/components/JointVelocityLimits.hh
+++ b/include/gz/sim/components/JointVelocityLimits.hh
@@ -1,0 +1,55 @@
+/*
+ * Copyright (C) 2021 Open Source Robotics Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+#ifndef GZ_SIM_COMPONENTS_JOINTVELOCITYLIMITS_HH_
+#define GZ_SIM_COMPONENTS_JOINTVELOCITYLIMITS_HH_
+
+#include <vector>
+#include <gz/math/Vector2.hh>
+
+#include <gz/sim/components/Component.hh>
+#include <gz/sim/components/Factory.hh>
+#include <gz/sim/components/Serialization.hh>
+#include <gz/sim/config.hh>
+#include <gz/sim/Export.hh>
+
+namespace gz
+{
+namespace sim
+{
+// Inline bracket to help doxygen filtering.
+inline namespace GZ_SIM_VERSION_NAMESPACE {
+
+namespace components
+{
+/// \brief Component for recording the current velocity limits of a joint.
+/// Data are a vector with a Vector2 for each DOF. The X() component of the
+/// Vector2 specifies the minimum velocity limit, the Y() component stands
+/// for maximum limit.
+using JointVelocityLimits = Component<
+  std::vector<gz::math::Vector2d>,
+  class JointVelocityLimitsTag,
+  serializers::VectorSerializer<gz::math::Vector2d>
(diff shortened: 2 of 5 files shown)


.vscode/settings.json
```diff
@@ -0,0 +1,5 @@
+{
```
ahcorde, line 1: remove this file

include/gz/sim/components/JointVelocityLimits.hh
```diff
@@ -0,0 +1,55 @@
+/*
+ * Copyright (C) 2021 Open Source Robotics Foundation
```
ahcorde, line 2:
```suggestion
 * Copyright (C) 2025 Open Source Robotics Foundation
```

ahcorde (changes requested):

mo-zaghloul:
Hi @ahcorde,

Thanks for the review! I’ve updated the copyright year as suggested.

I also noticed check failures in my PR, and I was wondering if I should be targeting gz-sim9 instead of gz-sim8. Since the issue originally appeared in gz-sim8, I based my PR on that branch, but I’d appreciate any guidance on whether I should rebase it to gz-sim9 instead.

Looking forward to your feedback!

Thanks!

src/Joint.cc
```diff
@@ -354,6 +355,14 @@ std::optional<std::vector<double>> Joint::Position(
       this->dataPtr->id);
 }
 
+//////////////////////////////////////////////////
+std::optional<std::vector<gz::math::Vector2d>> Joint::VelocityLimits(
+    const EntityComponentManager &_ecm) const
+{
+  return _ecm.ComponentData<components::JointVelocityLimits>(
+      this->dataPtr->id);
```
iche033, line 363: I think you can just get this information from the `JointAxis` component, e.g.

```cpp
auto jointAxis = _ecm.Component<components::JointAxis>(this->dataPtr->id);
if (jointAxis)
{
  return jointAxis->Data().MaxVelocity();
}
```

For joint velocity, there is no lower or upper limit, but only just one [velocity limit field](http://sdformat.org/spec?ver=1.12&elem=joint#limit_velocity)

ahcorde (approved):
@iche033 can you review it one more time?

src/Joint.cc
```diff
@@ -354,6 +355,19 @@ std::optional<std::vector<double>> Joint::Position(
       this->dataPtr->id);
 }
 
+//////////////////////////////////////////////////
+std::optional<gz::math::Vector2d> Joint::VelocityLimits(
```
iche033, line 359: there is a compile error here. The declaration in the header is `std::optional<std::vector<gz::math::Vector2d>` but it is `std::optional<gz::math::Vector2d>` here. 

I think it should be `std::optional<std::vector<gz::math::Vector2d>` so that it is consistent with the `JointVelocityLimitsCmd` component. 

On the other hand, the function returns `jointAxis->Data().MaxVelocity()` which is just a `double`. So we'll need to change it to return a vector of `gz::math::Vector2d`, something like (untested):

```cpp
return {gz::math::Vector2d(-jointAxis->Data().MaxVelocity(), 
        jointAxis->Data().MaxVelocity())};
```

azeey:
Hi @mo-zaghloul, any progress on this?

PR #2822, opened 2025-03-13
Upgrades OdometryPublisher system plugin

### LINKED ISSUE #2803: Allow OdometryPublisher to work with non-fixed link
Allow the OdometryPublisher to publish odometry for a non-fixed link of the robot model. Introduce the possibility to publish the odometry of a point after a non-fixed joint. 

Solves StackExchange question:
https://robotics.stackexchange.com/questions/112266/how-to-get-the-global-position-of-a-link-in-gazebo-sim

Morgantiz:
# 🎉 New feature

Closes #2803

## Summary
Allows OdometryPublisher to publish odometry for a non-fixed link of the robot model. Introduces the possibility to publish the odometry of a point after a non-fixed joint.  This commit does not change the default behavior of the plugin.

### DIFF
--- a/src/systems/odometry_publisher/OdometryPublisher.cc
+++ b/src/systems/odometry_publisher/OdometryPublisher.cc
@@ -40,6 +40,7 @@
 #include "gz/sim/components/Pose.hh"
 #include "gz/sim/components/JointPosition.hh"
 #include "gz/sim/Model.hh"
+#include "gz/sim/Link.hh"
 #include "gz/sim/Util.hh"
 
 using namespace gz;


iche033:
there are compile issues. On the other hand, see https://github.com/gazebosim/gz-sim/issues/2803#issuecomment-2725903486 for an alternative solution to that issue.

Morgantiz:
The solution proposed in https://github.com/gazebosim/gz-sim/issues/2803#issuecomment-2725903486 does not allow to use the OdometryPublisher for a link other than the first frame of the robot model. 
However, now there is only the possibility to use the xyz and rpy fixed offesets to publish with respect to another fixed position relative to the robot base frame. 

The new feature that i propose uses the link name written inside the OdometryPublisher plugin and introduces a new parameter for the message parent frame.  

In the last commit i solved the compile errors.

Morgantiz:
There should no longer be any compilation errors.

Morgantiz:
are there any updates?

azeey:
Hi @Morgantiz, we are currently in the middle of Code Freeze in preparation for the Gazebo Jetty release, so we might not be able to get to this until Jetty is released. Thanks for your patience.

PR #2776, opened 2025-02-14
Attempts to print a stacktrace on segfault in our integration tests

arjo129:
🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸

# 🎉 New feature

Closes #<NUMBER>

~~Depends on: https://github.com/gazebo-tooling/action-gz-ci/pull/81~~

## Summary
Our integration tests often segfault in CI leaving us clueless as to whats happening. This commit is an attempt to vendor backward-cpp and at the same time allows us to get a stacktrace of our segfaulting tests in CI. The current status of this PR is draft.

Other options considered:

We vendor backward-cpp in `gz-launch` and in `gz-utils`. Sadly the ubuntu package is not very useful as no cmake targets are provided. I could try to vendor it in gz-cmake or write a bespoke `FindBackward` package. However, our goal is to use it in tests. This is a very specific use case and I feel maybe the simplest option is to vendor it here and link it against the tests. That way we do not introduce more dependencies via 3rd party package managers and the user has no need to have anything else installed.

## Test it

[This commit](https://github.com/gazebosim/gz-sim/pull/2776/commits/256b4cae0df6e0098bad7967b4b6b7ed39cb2b89) has an example of a test I purposely introduced a segfault in. You can take a look at the github runners and see the stacktrace, Unfortunately, for Jenkins, there may need to be some minor changes on the buildfarm side. In particular we need to install `libdwarf-dev`, `libdw-dev` and `binutils-dev`. We will also need to build with `RelWithDebInfo`.

## Checklist
- [x] Signed all commits for DCO
- [ ] Added tests
- [ ] Added example and/or tutorial
- [ ] Updated documentation (as needed)
- [ ] Updated migration guide (as needed)
- [ ] Consider updating Python bindings (if the library has them)
- [x] `codecheck` passed (See [contributing](https://gazebosim.org/docs/all/contributing#contributing-code))
- [ ] All tests passed (See [test coverage](https://gazebosim.org/docs/all/contributing#test-coverage))
- [ ] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Agazebosim+archived%3Afalse+) to support the maintainers

[...]

PR #2838, opened 2025-03-28
Don't install vendored backward files

scpeters:
# 🦟 Bug fix

Follow-up to #2776

## Summary

We vendored backward-cpp in #2776 to use in integration tests, but I noticed while investigating an unstable nightly build that some vendored files are being installed with gz-sim. I don't think these should be installed with our packages, so I am modifying the cmake logic to prevent their installation here.

## Checklist
- [ ] Added tests


iche033 (approved):

j-rivero:
The patch works although it is fragile if at some point we update the version of the backward vendor copy.  Since the effects of installing the files are not severe I'm inclined to to merge the change.

An alternative would be to deal with file removal in packaging but the problem is not solve for source compilations and we scale the size of the problem to many packaging platforms.

j-rivero (approved):

Issue #2889, opened 2025-04-28
standalone_gtest_setup example fails to build with cmake 4.0


scpeters:
## Environment
* OS Version: macos 13.6.3-x86_64
* Source or binary build? source of gz-sim9 since 3bdadad3edabef15a03e32ae822ce42e2a82cb04

## Description
* Expected behavior: the `standalone_gtest_setup` example should build on macOS, as it [builds successfully on Ubuntu](https://build.osrfoundation.org/job/gz_sim-ci-gz-sim9-noble-amd64/85/testReport/(root)/Plugins_ExamplesBuild/Build_standalone_gtest_setup/) was [building successfully on macOS](https://build.osrfoundation.org/job/gz_sim-ci-gz-sim9-homebrew-amd64/85/) until recently. Since there were no related changes in the [first build in which it began to fail](https://build.osrfoundation.org/job/gz_sim-ci-gz-sim9-homebrew-amd64/86/testReport/junit/(root)/Plugins_ExamplesBuild/Build_standalone_gtest_setup/), I looked for additional differences in the console log, and the most relevant difference is that cmake updated from 3.31.6 to 4.0.0 when the test started failing
* Actual behavior: the example fails to build with a cmake error in `gtest_discover_tests_impl`

## Steps to reproduce

1. Install `gz-sim9` on macOS
2. Attempt to build `standalone_gtest_setup` example
3.

## Output

~~~
135: [ 41%] Building CXX object CMakeFiles/command_TEST.dir/command_TEST.cc.o
135: [ 50%] Linking CXX executable command_TEST
135: dyld[22146]: Symbol not found: __ZN2gz3sim2v910components7Factory8InstanceEv
135:   Referenced from: <0854A3DC-E052-3C00-91F5-8D9FCC53E1B2> /private/var/folders/ng/v8gtqrqs1sj3dzz5pvh558mh0000gp/T/gz/temp_dirlCZu4K/command_TEST
135:   Expected in:     <no uuid> unknown
135: CMake Error at /usr/local/share/cmake/Modules/GoogleTestAddTests.cmake:132 (message):
135:   Error running test executable.
135: 
135:     Path: '/private/var/folders/ng/v8gtqrqs1sj3dzz5pvh558mh0000gp/T/gz/temp_dirlCZu4K/command_TEST'
135:     Working directory: '/private/var/folders/ng/v8gtqrqs1sj3dzz5pvh558mh0000gp/T/gz/temp_dirlCZu4K'
[...]

PR #2892, opened 2025-04-29
Disable standalone_gtest_setup example test on mac

iche033:
# 🦟 Bug fix

Fixes https://github.com/gazebosim/gz-sim/issues/2889

## Summary

The test was also failing locally on my mac but with a slightly different error msg from the output in #2889:

```
dyld[22115]: Library not loaded: libgz-sim.10.dylib
```

Setting `DISCOVERY_MODE` to `PRE_TEST` in `gtest_discover_tests` makes gtest delay test discovery until before test execution. By default it's done at build time with `POST_BUILD`.

From [cmake doc](https://cmake.org/cmake/help/latest/module/GoogleTest.html#command:gtest_discover_tests) on `PRE_TEST`:

> This way test discovery occurs in the target environment where the test has a better chance at finding appropriate runtime dependencies.

I was able to build and run the test locally with these changes.

## Checklist
- [ ] Added tests

### DIFF
--- a/examples/standalone/gtest_setup/CMakeLists.txt
+++ b/examples/standalone/gtest_setup/CMakeLists.txt
@@ -31,6 +31,11 @@ foreach(TEST_TARGET
     gtest_main
     gz-sim::gz-sim
   )
+
   include(GoogleTest)
-  gtest_discover_tests(${TEST_TARGET})
+  if (APPLE)
+    gtest_discover_tests(${TEST_TARGET} DISCOVERY_MODE PRE_TEST)
+  else()
+    gtest_discover_tests(${TEST_TARGET})
+  endif()
 endforeach()


examples/standalone/gtest_setup/CMakeLists.txt
```diff
@@ -31,6 +31,11 @@ foreach(TEST_TARGET
     gtest_main
     gz-sim::gz-sim
   )
+
   include(GoogleTest)
-  gtest_discover_tests(${TEST_TARGET})
+  if (APPLE)
```
scpeters, line 36: Do you think this issue will occur on other platforms with cmake 4? If so, I would check the cmake version instead of APPLE

I'm not sure though

iche033: I did some tests:
* Ubuntu + cmake 4.0.1: built fine
* Mac + cmake 3.31.7: same error
* Mac + cmake 3.31.6: same error

hmm maybe something else has changed

scpeters: weird

I've tried disabling ccache and that seems to fix it for me?

testing without ccache using https://github.com/gazebo-tooling/release-tools/commit/b7cd4e3681d81628613e6bdd3051c58f13cfee0a:

* [Build Status](https://build.osrfoundation.org/job/gz_sim-ci-gz-sim9-homebrew-amd64/97/) https://build.osrfoundation.org/job/gz_sim-ci-gz-sim9-homebrew-amd64/97/

scpeters: that didn't help

scpeters: now I can't reproduce the failure on my personal machine at all, though it still fails in CI

iche033: hmm I can still reproduce it. I don't have ccache setup.

PR #2919, opened 2025-05-22
backwards-cpp support for gz-sim-main and gz-sim-gui

### LINKED ISSUE #2906: Restore Stack Traces in jetty
## Environment
* OS Version: Ubuntu 24.04
* Source: main

## Description
We used to generate stack traces in older versions of gazebo using `backward.cpp`. These are crucial for debugging end user problems and save us developers a lot of time. In jetty it seems like we are not linking against backward correctly as I am getting a segfault without stack traces.

## Steps to reproduce

1. Run any bug that will segfault gazebo.

j-rivero:
# 🎉 New feature

Closes #2906 

## Summary

The PR adds support for backwards-cpp in gz-sim-main and gz-sim-gui. To do so:

 * Bump backwards vendor to 1.6 5913812bb6d382fa2c8e124f1a9aca9009221fe8
 * Relocate the backwards vendor directory outside of the root to be used by CLI 965d46370d963dcb59ccee282377aad9a3b6dbc5
 * Build binaries with backwards support fd0edc98262394385bd9a4440022f350a81c1136

## Checklist
- [ ] Added tests


arjo129 (changes requested):
I tried this branch and introduced a segfault, however the stack traces are not being printed. I'll go through and check more thoroughly.

vendor/backward-cpp/CMakeLists.txt
```diff
@@ -77,58 +65,75 @@ if (CMAKE_CXX_COMPILER_ID STREQUAL "Clang" OR CMAKE_COMPILER_IS_GNUCXX)
...
-# (Note that this target is not exported, since CMake currently does not allow
-# exporting an OBJECT library.)
+# BACKWARD OBJECT
 ###############################################################################
 
 add_library(backward_object OBJECT backward.cpp)
-set_target_properties(backward_object PROPERTIES EXPORT_NAME Object)
-target_link_libraries(backward_object PUBLIC Backward::Interface)
-add_library(Backward::Object ALIAS backward_object)
+target_compile_definitions(backward_object PRIVATE ${BACKWARD_DEFINITIONS})
```
arjo129, line 72: I think we do need to link against this object for some reason `Backward::Backward` is not enough

arjo129:
So after reverting the version of backward to the current one, I managed to link it to `backward_object` and it works correctly in this branch: https://github.com/gazebosim/gz-sim/tree/arjo/properly_segfaults_w_stacktrace

I don't think the above branch has the ideal solution however. TBH the CMake-fu going on in the library exceeds my domain of knowledge. One option is to get rid of all the cmake stuff and only include the header. we can manually instantiate `BackwardHandle`

j-rivero:
> I tried this branch and introduced a segfault, however the stack traces are not being printed.

Makes sense. The build log does not display any modifcation of code being done by the linker. re-reading the backwards-cpp lib instructions we can use either the header file or the .cpp file. I think that you are using the later when linking against `backwards_object` target. I would prefer to keep the installation more standard by using the header and the Backwards::Backwards CMake target.

Said that, I've brought the backwards-cpp code into the main/gui in [42383a5](https://github.com/gazebosim/gz-sim/pull/2919/commits/42383a5400b7c83a01549a2082f392277fe8fd02) by including the headers and the signal handler. Segfaulting on propose triggers clear stacktraces on my system after using `RelWithDebInfo`.

CMakeLists.txt
```diff
@@ -32,10 +32,17 @@ else()
...
+#--------------------------------------
+# Find Backward-cpp for stack trace support
+# See https://github.com/gazebosim/gz-cmake/issues/477 before copy this code
+find_package(Backward REQUIRED
+  HINTS ${CMAKE_CURRENT_SOURCE_DIR}/vendor/backward-cpp
+  NO_DEFAULT_PATH)
+
 if (UNIX AND NOT APPLE)
-  set (EXTRA_TEST_LIB_DEPS stdc++fs backward_object)
+  set (EXTRA_TEST_LIB_DEPS stdc++fs Backward::Backward)
```
arjo129, line 43: Sadly, this won't work for the tests, perhaps `Backward::Object`?

arjo129 (changes requested):
This works great for the binaries but introduces a regression on the tests where segfaulting tests no longer print stack traces.

j-rivero:
> This works great for the binaries but introduces a regression on the tests where segfaulting tests no longer print stack traces.

You are right. I've reverted the usage of the header-only files, I think that it is more intrusive for the code [255dd43](https://github.com/gazebosim/gz-sim/pull/2919/commits/255dd435e5d3cb106f5894b03782d4260f4a1a45).

For using the linking approach and the "object" CMake target we need to stop using the find_package and use the add_subdirectory as detailed in the backwards-cpp README. Changing that and using `Backwards::Object` was good enough for the binaries.

For the special case of `EXTRA_TEST_LIB_DEPS` I think that `CMake` is not resolving properly that CMake `Backwards::Object` so I used the `backward_object` instead.

Seems good enough for me after tesing tests and binaries.

arjo129 (approved):
Thanks! LGTM! There was a minor style issue with regards to spacing, I've pushed a commit that fixes it.

vendor/backward-cpp/CMakeLists.txt
```diff
@@ -115,16 +115,58 @@ set_target_properties(backward PROPERTIES EXPORT_NAME Backward)
...
 # check if Backward is being used as a top-level project or included as a subproject
 if(CMAKE_SOURCE_DIR STREQUAL PROJECT_SOURCE_DIR)
-    install(
-        FILES "backward.hpp"
-        DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}
-    )
-    install(
-        FILES "BackwardConfig.cmake"
-        DESTINATION "${CMAKE_INSTALL_LIBDIR}/cmake/${PROJECT_NAME}"
-    )
```
scpeters, line 127: this reverted #2838, and @j-rivero predicted this fragility in https://github.com/gazebosim/gz-sim/pull/2838#issuecomment-2763239445

for now the debbuild is unstable due to these installed files since BackwardConfig.cmake is not included in any package install rules

* [Build Status](https://build.osrfoundation.org/job/gz-sim10-debbuilder/247/) https://build.osrfoundation.org/job/gz-sim10-debbuilder/247/

azeey:
@j-rivero I think we want to avoid installing `backward.hpp`. Currently, `libgz-sim10-dev` wants to install it in `/usr/include` which conflicts with `libbackward-cpp-dev`.

```bash
$ dpkg -L libgz-sim10-dev | rg backward
/usr/include/backward.hpp
```

j-rivero:
> @j-rivero I think we want to avoid installing `backward.hpp`. Currently, `libgz-sim10-dev` wants to install it in `/usr/include` which conflicts with `libbackward-cpp-dev`.

There are different workarounds for this including the modification of the files imported that we did in #2838 or the customization of our deb metadata. The first is fragile as time has demonstrated, the second is only affecting .deb installations and leaving other packagers/builds with the problem.

There is an open PR with changes close to what Steve was proposing https://github.com/bombela/backward-cpp/pull/338. I'm include to resurrect the code changes and document it a bit better for gaining visibility on the changes. I'll open a PR.

PR #2932, opened 2025-06-03
Creating a standalone server executable

sauk2:
# 🎉 New feature

Related to https://github.com/gazebosim/gz-tools/issues/7, https://github.com/gazebo-release/gz-sim8-release/issues/8 and https://github.com/gazebosim/gz-sim/issues/2737

## Summary

This PR extends on the recent migration of `gz sim` to use a standalone executable, `gz-sim-main`. This executable also launches a GUI executable when built using `-DENABLE_GUI=ON`. This change updates the CMake configuration to allow the `gz-sim-main` executable to be reused for building a standalone server executable, `gz-sim-server` irrespective of the `-DENABLE_GUI=ON` argument. Additionally, GUI-related CLI flags are now conditionally compiled using WITH_GUI, ensuring that GUI-specific options do not appear in `gz-sim-server` or in `gz-sim-main` when built without GUI support.

## Test it

After compilation, check for a new executable named `gz-sim-server-client` inside `libexec/gz/sim10`. Calling it should launch the server without any GUI.

## Checklist
- [ ] Added tests


j-rivero:
> to be reused for building a standalone server executable, `gz-sim-server-client`

I think that we can simply call it `gz-sim-server`

Before continue reviewing the code, note that the validation step that we need for the official docker images coming from our .deb packages is that a single build WITH_GUI produces a server binary that does not use any X / Qt / other Xorg related library. We are linking the new server against the Qt libraries:

```bash
gz-sim/build on  standalone-server-exe [?] via C v13.3.0-gcc via △ v3.28.3 ❯ ldd ./bin/gz-sim-server-client  | grep Qt
	libQt6Core.so.6 => /lib/x86_64-linux-gnu/libQt6Core.so.6 (0x000074d147e00000)
	libQt6Quick.so.6 => /lib/x86_64-linux-gnu/libQt6Quick.so.6 (0x000074d146e00000)
	libQt6Qml.so.6 => /lib/x86_64-linux-gnu/libQt6Qml.so.6 (0x000074d146800000)
	libQt6Widgets.so.6 => /lib/x86_64-linux-gnu/libQt6Widgets.so.6 (0x000074d146000000)
	libQt6Gui.so.6 => /lib/x86_64-linux-gnu/libQt6Gui.so.6 (0x000074d145800000)
	libQt6QmlModels.so.6 => /lib/x86_64-linux-gnu/libQt6QmlModels.so.6 (0x000074d146d40000)
	libQt6Network.so.6 => /lib/x86_64-linux-gnu/libQt6Network.so.6 (0x000074d144869000)
	libQt6OpenGL.so.6 => /lib/x86_64-linux-gnu/libQt6OpenGL.so.6 (0x000074d1474a4000)
	libQt6DBus.so.6 => /lib/x86_64-linux-gnu/libQt6DBus.so.6 (0x000074d145f4b000)
``` 

This probably comes from the fact of llnking it to gz.

j-rivero (changes requested):
Let's try to fix the X / Qt issue

sauk2:
> Let's try to fix the X / Qt issue

Thanks for catching this! I have made changes in CMake to not link the `gz` library to `gz-sim-server`. I think that should solve this issue.

j-rivero:
Sorry for the lack of response, I'm blocked in https://github.com/gazebosim/gz-sim/issues/2945. The testing on Windows is no going well but I need to first build the ´main´ branch again. If you can give it a try on Windows @sauk2 , would be great.

j-rivero:
There is something that we need to define behaviourly in the parameters. We have the specialized versions of ` --render-engine` and ` --render-engine-api-backend` that affects only to the server that are really nice to have in `gz-sim-server`  (same applies to the client BTW).

```
          --render-engine-server TEXT=ogre2
                              Gazebo Rendering engine plugin to load for the Server.
                              Gazebo will use OGRE2 by default. Make sure custom
                              plugins are in GZ_SIM_RENDER_ENGINE_PATH.

          --render-engine-server-api-backend TEXT:{opengl,vulkan,metal}
                              Same as --render-engine-api-backend but only
                              for the Server.

          --render-engine TEXT=ogre2
                              Gazebo Rendering engine plugin to load for both the Server
                              and the GUI. Gazebo will use OGRE2 by default.
                              Make sure custom plugins are inside
                              GZ_SIM_RENDER_ENGINE_PATH.

          --render-engine-api-backend TEXT:{opengl,vulkan,metal}
                              API to use for both Server and GUI.
                              Possible values for ogre2:
                              - opengl (default)
                              - vulkan (beta)
                              - metal (Apple only. Default for Apple)
                              Note: If Vulkan is being in the GUI and gz-gui was
                              built against Qt < 5.15.2, it may be very slow
```

My doubt is if we want to keep supporting the general parameters  ` --render-engine` and ` --render-engine-api-backend` together with specialized server/client parameters for compatibility reasons in the calls to the different executables or if we want to remove them since the behaviour of them would change and it will only affect one part although the description says otherwise and the previous behaviour is also different (since them affect client and server and now would only affect one of the parts).

j-rivero:
The approach of the PR is to reuse the gz common part to create the new `gz-sim-server`executable. The gz source file is shared with `gz-sim-main`. I've initially imaging that the new gz-sim-server executable was going to launched and orchestrated in `sim_main.cc` like the `GZ_SIM_GUI_EXE` is being launched in a thread and start/stopped from there.

We are not implementing this in the PR, I think it is fine but the long term goal would be to encapsulate server logic and client logic into their own source files and executables and orchestrate them from the main unit.

Said that, this PR unblock the decoupling and the docker images :)

j-rivero:
The testing on Windows for new `gz-sim-server` is good, also `gz-sim-main` called with -s. I can not start the GUI due to problems with https://github.com/gazebosim/gz-sim/issues/2945 but as far as I can reach, all works as expected.

sauk2:
> The approach of the PR is to reuse the gz common part to create the new `gz-sim-server`executable. The gz source file is shared with `gz-sim-main`. I've initially imaging that the new gz-sim-server executable was going to launched and orchestrated in `sim_main.cc` like the `GZ_SIM_GUI_EXE` is being launched in a thread and start/stopped from there.

Thanks for all the feedback! I agree this is the ideal long-term architecture we want to achieve. I initially explored this approach in https://github.com/gazebosim/gz-sim/pull/2809 as well, aiming to launch the Server and GUI as individual subprocesses from a shared orchestrator executable.

However, I ran into issues with signal handling in this setup. Each executable internally creates its own signal handlers, so when a single SIGINT is received, only one handler captures it, shutting down that process while the other continues running. As a result, it required two separate CTRL+C commands to fully terminate both processes cleanly.

Some changes may be required in the [SignalHandler](https://github.com/gazebosim/gz-common/blob/gz-common6/src/SignalHandler.cc) class, along with adjustments to the behavior for [Server](https://github.com/gazebosim/gz-sim/blob/2e1679e851a8e0b14ad8cdb69a832c5ec792511f/src/ServerPrivate.cc#L118-L128) and [GUI](https://github.com/gazebosim/gz-sim/blob/2e1679e851a8e0b14ad8cdb69a832c5ec792511f/src/gui/Gui.cc#L325-L330) launches, to support this cleanly in the future.

PR #2960, opened 2025-06-30
DetachableJoint: allow physics constraint properties to be set

srmainwaring:
# 🎉 New feature

Allow joint constraint physics parameters to be set for detachable joints.

Depends on:

- https://github.com/gazebosim/gz-physics/pull/754

## Summary

- Add a <physics> parameter block with <cfm> and <erp> sub elements.
- Update the `Physics` system to use the feature `physics::SetDynamicJointConstraintPropertiesFeature`
- Add optional fields to the detachable joint component.

## Test it

Add a SDF parameter block to a model using a detachable joint:

```xml
    <plugin name="gz::sim::systems::DetachableJoint"
      filename="libgz-sim-detachable-joint-system.so">
      <parent_link>blade_rod_1a_link</parent_link>
      <child_link>blade_rod_1b_link</child_link>
      <child_model>rotor_head_cw</child_model>
      <joint_type>fixed</joint_type>
      <physics>
        <cfm>1.0e-6</cfm>
        <erp>0.8</erp>
      </physics>
    </plugin>
``` 

Inspect the console when running with verbose output:

```bash
(2025-06-30 15:16:29.373) [debug] [Physics.cc:1942] Creating detachable joint [318]
(2025-06-30 15:16:29.373) [debug] [Physics.cc:1956] SetConstraintForceMixing: 1e-06
(2025-06-30 15:16:29.373) [debug] [JointFeatures.cc:746] SetConstraintForceMixing: 1e-06
(2025-06-30 15:16:29.373) [debug] [Physics.cc:1964] SetErrorReductionParameter: 0.8
(2025-06-30 15:16:29.373) [debug] [JointFeatures.cc:760] SetErrorReductionParameter: 0.8
```

The default values of `cfm` and `erp` may result in simulation that is unable to constrain joints in a closed kinematic chain - resulting in the model flying apart (often followed by a segmentation fault). Increasing these parameters may result in a stable simulation. 

## Checklist
- [ ] Added tests

### DIFF
--- a/src/systems/detachable_joint/DetachableJoint.cc
+++ b/src/systems/detachable_joint/DetachableJoint.cc
@@ -99,7 +99,7 @@ void DetachableJoint::Configure(const Entity &_entity,
   // Optional physics parameters
   if (_sdf->HasElement("physics"))
   {
-    // Clone as GetElement() requries non-const access.
+    // Clone as GetElement() requires non-const access.
     auto sdfClone = _sdf->Clone();
     auto physicsSdf = sdfClone->GetElement("physics");
     if (physicsSdf->HasElement("cfm"))


ntfshard:
Proposal: keep sdf structure similar to sdf joint definition http://sdformat.org/spec?ver=1.9&elem=joint
e.g. physics -> ode -> (new fields)
And also possible to use defaults from sdf instead of optional for consistency

include/gz/sim/components/DetachableJoint.hh
```diff
@@ -39,14 +39,20 @@ namespace components
     Entity parentLink;
     /// \brief Entity of the echild link
     Entity childLink;
-    // \brief Type of joint. Only the "fixed" joint type is currently supported.
+    /// \brief Type of joint. Only the "fixed" joint type is currently supported.
     std::string jointType = {"fixed"};
+    /// \brief Optional constraint force mixing parameter
+    std::optional<double> cfm;
```
ahcorde, line 46: include `<optional>`

src/systems/detachable_joint/DetachableJoint.cc
```diff
@@ -95,6 +95,22 @@ void DetachableJoint::Configure(const Entity &_entity,
...
 
+  // Optional physics parameters
+  if (_sdf->HasElement("physics"))
+  {
+    // Clone as GetElement() requries non-const access.
+    auto sdfClone = _sdf->Clone();
+    auto physicsSdf = sdfClone->GetElement("physics");
+    if (physicsSdf->HasElement("cfm"))
+    {
+      this->cfm = std::optional<double>(physicsSdf->Get<double>("cfm"));
```
ahcorde, line 107: include `<optional>`

src/systems/detachable_joint/DetachableJoint.hh
```diff
@@ -140,6 +140,12 @@ namespace systems
     /// \brief Whether child entity is attached
     private: std::atomic<bool> isAttached{false};
 
+    /// \brief Optional constraint force mixing parameter
+    private: std::optional<double> cfm;
```
ahcorde, line 145: include `<optional>`

srmainwaring:
> Proposal: keep sdf structure similar to sdf joint definition http://sdformat.org/spec?ver=1.9&elem=joint
> e.g. physics -> ode -> (new fields)

I thought about this, but decided it was a little overkill adding another 2 or 3 layers of nesting. It's also a little confusing as the engine in use is `dartsim` not `ode`, and here and in some other places fields from one engine are used in another. Keeping the fields flat is more in line with the reworked physics libraries strategy of using typelists of features - if a library supports a feature it uses it, otherwise it is ignored.

> And also possible to use defaults from sdf instead of optional for consistency

The question then is what defaults to use in SDF, and how to ensure that they are the same as what the physics engine is using? Using optional leaves the definition of defaults in one place - the physics engine, rather than placing them across the system.

include/gz/sim/components/DetachableJoint.hh
srmainwaring: Fixed in https://github.com/gazebosim/gz-sim/pull/2960/commits/4ae1e9c418526aee1f8d590436f4ba30637eda26

src/systems/detachable_joint/DetachableJoint.cc
srmainwaring: Fixed in https://github.com/gazebosim/gz-sim/pull/2960/commits/4ae1e9c418526aee1f8d590436f4ba30637eda26

src/systems/detachable_joint/DetachableJoint.hh
srmainwaring: Fixed in https://github.com/gazebosim/gz-sim/pull/2960/commits/4ae1e9c418526aee1f8d590436f4ba30637eda26

ntfshard:
> > Proposal: keep sdf structure similar to sdf joint definition http://sdformat.org/spec?ver=1.9&elem=joint
> > e.g. physics -> ode -> (new fields)
> 
> I thought about this, but decided it was a little overkill adding another 2 or 3 layers of nesting. It's also a little confusing as the engine in use is `dartsim` not `ode`, and here and in some other places fields from one engine are used in another. Keeping the fields flat is more in line with the reworked physics libraries strategy of using typelists of features - if a library supports a feature it uses it, otherwise it is ignored.
> 
> > And also possible to use defaults from sdf instead of optional for consistency
> 
> The question then is what defaults to use in SDF, and how to ensure that they are the same as what the physics engine is using? Using optional leaves the definition of defaults in one place - the physics engine, rather than placing them across the system.

I guess in a perfect way it should be as same as original joint so using the same structure can allow to switch to existing parser(at least potentially). In future I want to add other joints types to this plugin tbh

iche033:
> I thought about this, but decided it was a little overkill adding another 2 or 3 layers of nesting. It's also a little confusing as the engine in use is dartsim not ode, and here and in some other places fields from one engine are used in another. Keeping the fields flat is more in line with the reworked physics libraries strategy of using typelists of features - if a library supports a feature it uses it, otherwise it is ignored.

there was a proposal related to this in https://github.com/gazebosim/sdformat/issues/508, which proposes that we move away from engine-specific tags.

PR #2966, opened 2025-07-08
[bazel] Add support for Systems

shameekganguly:
# 🎉 New feature

## Summary
Adds bazel targets for Systems under `src/systems` so that downstream bazel users can use System plugins in their binaries/ tests.

The following changes are required to support this feature:
1. Added headers under `include/gz/sim/physics` to the `gz-sim` library target since they are required for the `Physics` System.
2. Added a bazel macro `gz_sim_system_libraries` to generate 
    - a `cc_binary` shared library target for a given System plugin that can be loaded dynamically, and
    - a `cc_library` statically linked library target for the System plugin that can be loaded from the static plugin registry.
3. The macro requires new `bazel_dep`s for `rules_cc` and `bazel_skylib` in MODULE.bazel. I also updated BUILD.bazel and test/BUILD.bazel to use `cc_*` rules from `rules_cc` instead of native rules to be consistent (this is required anyway for bazel 9).
4. Added `gz_sim_system_libraries` targets for all Systems except `camera_video_recorder`, `dvl`, `elevator` and `python_system_loader`. These either have missing dependencies or fail to build with bazel.
    - Few missing headers had to be included to build with bazel.
    - Some uses of `protoMsg->GetTypeName()` had to be updated since the bazel build uses protobuf v30, which changed the return type for `GetTypeName()` to be `std::string_view` (see [related gz-msgs PR](https://github.com/gazebosim/gz-msgs/pull/521) for more details).
    - The methods defined in `src/systems/environmental_sensor_system/TransformTypes.hh` had to marked `inline` to avoid linker ODR issues in bazel.
5. Enabled some more unit tests in bazel, see exclusions under the `test_sources` list in BUILD.bazel for details.
    - Updated `test/helpers/EnvTestFixture.hh` to use `common::testing::TempPath` to set up a temp home dir in tests. The  `common::testing::*` methods are compatible with both cmake and bazel (with the `GZ_BAZEL` env variable added to test targets). This requires updating the cmake build as well so that test targets depend on `gz-common${GZ_COMMON_VER}::testing`.
    - Also updated some sdf file paths in `src/SdfGenerator_TEST.cc` to be generated using `common::testing::TestFile()` instead of using the `PROJECT_SOURCE_PATH` variable which works well for cmake but not for bazel (returns relative path instead of absolute path, which some tests are sensitive to).
    - Some tests and Systems expect to find runtime data such as plugins or server.config files at locations indicated by env variables. Unfortunately, this doesn't work well in bazel since (i) predefined make variables only provide paths to the actual data artifact whereas the methods in `common::SystemPaths` expect the directory containing the artifact to be set in the env var, and (ii) the canonical way of navigating the runfiles system is using the `rules_cc//runfiles` library, but this cannot be used in code that can be cross compiled with cmake. To circumvent this issue, I added the required data artifacts to well-known paths under test runfiles with a few copy `genrule`s. This allows us to set the path env vars to well-known paths for tests.

## Test it
```
$ bazel test ...
$ bazel build --action_env=CC=/usr/bin/clang  ...
```

## Checklist
- [x] Added tests

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -1,7 +1,10 @@
 load("@gz-msgs//tools:gz_msgs_generate.bzl", "gz_proto_library")
+load("@rules_cc//cc:cc_library.bzl", "cc_library")
+load("@rules_cc//cc:cc_test.bzl", "cc_test")
 load("@rules_gazebo//gazebo:headers.bzl", "gz_configure_header", "gz_export_header")
 load("@rules_license//rules:license.bzl", "license")
 load("@rules_proto//proto:defs.bzl", "proto_library")
+load("//bazel:gz_sim_system_libraries.bzl", "gz_sim_system_libraries")
 
 package(
     default_applicable_licenses = [":license"],
@@ -80,6 +83,7 @@ public_headers_no_gen = glob(
         "include/gz/sim/comms/*.hh",
         "include/gz/sim/components/*.hh",
         "include/gz/sim/detail/*.hh",
+        "include/gz/sim/physics/*.hh",
         "include/gz/sim/rendering/*.hh",
     ],
 )
@@ -136,6 +140,7 @@ cc_library(
         ":RenderingExport",
         ":gzsim_cc_proto",
         "@com_google_protobuf//:protobuf_lite",
+        "@eigen",
         "@gz-common",
         "@gz-common//events",
         "@gz-common//geospatial",
@@ -146,9 +151,9 @@ cc_library(
         "@gz-math",
         "@gz-msgs",
         "@gz-msgs//:gzmsgs_cc_proto",
+        "@gz-physics",
         "@gz-plugin//:core",
         "@gz-plugin//:loader",
-        "@gz-plugin//:register",
         "@gz-rendering",
         "@gz-transport",
         "@gz-transport//parameters",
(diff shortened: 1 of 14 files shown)


BUILD.bazel
```diff
@@ -207,3 +280,1507 @@ test_sources = glob(
         "@tinyxml2",
     ],
 ) for src in test_sources]
+
+# Following systems are not yet supported in the Bazel BUILD:
+# - src/systems/camera_video_recorder (av sub-package in gz-common is not yet available in Bazel build)
+# - src/systems/dvl (av sub-package in gz-common is not yet available in Bazel build)
```
iche033:
```suggestion
# - src/systems/dvl (dvl and rendering sub-packages in gz-sensors are not yet available in Bazel build)
```

src/systems/environmental_sensor_system/TransformTypes.hh
```diff
@@ -45,7 +45,7 @@ enum TransformType {
 /// \brief Given a string return the type of transform
 /// \param[in] _str - input string
 /// \return std::nullopt if string invalid, else corresponding transform
-std::optional<TransformType> getTransformType(const std::string &_str)
+inline std::optional<TransformType> getTransformType(const std::string &_str)
```
iche033, line 48: can you add a brief comment here explaining the need for `inline`

iche033 (commented):
looks good, just minor comments.

BUILD.bazel
shameekganguly: Actually I was thrown off by `#include <gz/common/VideoEncoder.hh>` in src/systems/dvl/DopplerVelocityLogSystem.cc and led to believe this needs to depend on the missing `@gz-common//av` subpackage. But looking more closely, the header is unused. So I removed the include and added the dvl system in the bazel build: https://github.com/gazebosim/gz-sim/pull/2966/commits/821fdd06d5fe1dd803c30441a760cf3056d5ca3e

src/systems/environmental_sensor_system/TransformTypes.hh
shameekganguly: I don't know if we actually need a comment here - it is best practice to `inline` methods defined within a header file anyway (or mark them `constexpr` where possible). As to why it is needed specifically for the bazel build, I have documented that in this PR description.

IMO, adding a comment "inline this method to avoid ODR linker errors" is redundant since that's what the `inline` keyword indicates anyway?

arjo129: Yeah I think I wrote these in a hurry, it makes perfect sense to have them either `inline` or figure out the necessary mechanism to have a separate implementation. This has the added issue of needing a separate library target to be built by the cmake files, which is totally not unreasonable and probably the right way, but completely out of the scope of this PR.

iche033 (approved):

PR #2971, opened 2025-07-10
JointController: allow braking to be disabled in force mode

srmainwaring:
# 🎉 New feature

## Summary

Allow braking to be disabled when operating the joint controller in force mode. What this means is that a force will only be applied to bring up a joint to it's target velocity (accounting for sign). If a joint is rotating at a velocity higher than the commanded velocity the joint will freewheel and spin down subject to damping and friction only.

The optional parameter <disable_braking> is set to true to enable the feature. The default is to use force braking (current behaviour).

## Test it

Edit the `joint_controller_demo_2` model in the example `joint_controller.sdf` adding damping to the joint and the `<disable_braking>` element:

```xml
      <joint name="j1" type="revolute">
        <pose>0 0 -0.5 0 0 0</pose>
        <parent>base_link</parent>
        <child>rotor2</child>
        <axis>
          <xyz>0 0 1</xyz>
          <dynamics>
            <damping>0.0001</damping>
          </dynamics>
        </axis>
      </joint>

      <!-- Force mode -->
      <plugin
        filename="gz-sim-joint-controller-system"
        name="gz::sim::systems::JointController">
        <joint_name>j1</joint_name>
        <use_force_commands>true</use_force_commands>
        <disable_braking>true</disable_braking>
        <p_gain>0.2</p_gain>
        <i_gain>0.01</i_gain>
      </plugin>
```

Spin up the joint

```bash
gz topic -t "/model/joint_controller_demo_2/joint/j1/cmd_vel" -m gz.msgs.Double -p "data: 10.0"
```

then lower the velocity set point to >= 0.0:

```bash
gz topic -t "/model/joint_controller_demo_2/joint/j1/cmd_vel" -m gz.msgs.Double -p "data: 0.0" 
```

The joint will not stop abruptly, but freewheel with damping.

## Checklist
- [ ] Added tests

### DIFF
--- a/src/systems/joint_controller/JointController.cc
+++ b/src/systems/joint_controller/JointController.cc
@@ -77,6 +77,9 @@ class gz::sim::systems::JointControllerPrivate
   /// velocity.
   public: bool useForceCommands{false};
 
+  /// \brief True if braking is disabled.
+  public: bool disableBraking{false};
+
   /// \brief Velocity PID controller.
   public: math::PID velPid;
 };
@@ -146,6 +149,11 @@ void JointController::Configure(const Entity &_entity,
 
     this->dataPtr->velPid.Init(p, i, d, iMax, iMin, cmdMax, cmdMin, cmdOffset);
 
+    if (_sdf->HasElement("disable_braking"))
+    {
+      this->dataPtr->disableBraking = _sdf->Get<bool>("disable_braking");
+    }
+
     gzdbg << "[JointController] Force mode with parameters:" << std::endl;
     gzdbg << "p_gain: ["     << p         << "]"             << std::endl;
     gzdbg << "i_gain: ["     << i         << "]"             << std::endl;
@@ -155,6 +163,8 @@ void JointController::Configure(const Entity &_entity,
     gzdbg << "cmd_max: ["    << cmdMax    << "]"             << std::endl;
     gzdbg << "cmd_min: ["    << cmdMin    << "]"             << std::endl;
     gzdbg << "cmd_offset: [" << cmdOffset << "]"             << std::endl;
+    gzdbg << "disable_braking: [" << this->dataPtr->disableBraking
+          << "]" << std::endl;
   }
   else
   {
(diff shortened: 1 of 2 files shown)


src/systems/joint_controller/JointController.cc
```diff
@@ -316,9 +326,19 @@ void JointController::PreUpdate(const UpdateInfo &_info,
 
   double error = jointVelComp->Data().at(0) - targetVel;
 
+  // If braking is disabled only apply force if the actual velocity is below
+  // the target - i.e. allow joint to freewheel.
+  bool applyForce = true;
+  if (this->dataPtr->disableBraking)
+  {
+      double sgn = math::signum(targetVel);
+      applyForce = sgn && sgn * error <= 0.0;
```
iche033: I get a minor compile warning on my machine:

```sh
/home/iche/code/gz_i_ws_noble2/src/gz-sim/src/systems/joint_controller/JointController.cc:335:42: warning: comparing floating-point with ‘==’ or ‘!=’ is unsafe [-Wfloat-equal]
  335 |       applyForce = sgn && sgn * error <= 0.0;
      |                    ^~~
```
This should fix it

```suggestion
      applyForce = (sgn > 0) && sgn * error <= 0.0;
```

iche033 (commented):
works for me. Just one comment on compile warning

src/systems/joint_controller/JointController.cc
srmainwaring: We want to set `applyForce` false if `sgn == 0`, this is so that setting `targetVel = 0` will not brake when the joint has either a positive or negative velocity. So maybe `!(sgn == 0)` instead of just `(sgn)` - we are doing an 'is a float zero operation' - so maybe there's an existing util in `gz-math` we can use to make it robust?

iche033: something like [math::equal](https://github.com/gazebosim/gz-math/blob/c35a71a2cafd910144090e13b27c8e93cd6507d1/include/gz/math/Helpers.hh#L383C15-L383C20), e.g. `!math::equal(sgn, 0.0)` ?

srmainwaring: Updated in https://github.com/gazebosim/gz-sim/pull/2971/commits/3215156f530b34fbc5eb3456fc204f2e4f9bd69e

iche033 (approved):

PR #2984, opened 2025-07-19
Add param in physics system to enforce fixed constraint

iche033:
# 🎉 New feature

Depends on https://github.com/gazebosim/gz-physics/pull/759

## Summary

Added a `<enforce_fixed_constraint>` param in the physics system. If set to `true`, it uses the `SetFixedJointWeldChildToParentFeature` feature to enforce the relative pose of the parent and child links in the detachable joint., see https://github.com/gazebosim/gz-physics/pull/759 for more info on this gz-physics feature. By default, this param is set to `false`.

## Test it

This feature currently only works with bullet-featherstone physics plugin, so launch the world like so:

```sh
gz sim -v 4 -r detachable_joint.sdf --physics-engine libgz-physics-bullet-featherstone-plugin
```

The difference with the original detachable joint behavior should be very subtle (quite unnoticeable). The fixed joint connecting the child model `B1` to the parent blue vehicle chassis should now be `stiff` as if they are welded together. The detachable joint should still continue to work, e.g. this makes the green `B1` model drop:

```sh
gz topic -t "/B1/detach" -m gz.msgs.Empty -p "unused: true"
```

## Checklist
- [ ] Added tests


src/systems/detachable_joint/DetachableJoint.hh
```diff
@@ -143,6 +143,10 @@ namespace systems
     /// \brief Whether all parameters are valid and the system can proceed
     private: bool validConfig{false};
 
+    /// \brief Set whether the fixed joint should weld the child entity to the
+    /// parent entity.
+    /// No-op if the underlying physics engine does not support the feature.
+    private: bool weldChildToParent{false};
```
azeey: Please add documentation to the "System parameters" section above. Also mention that this is needed for the bullet physics engine and add some explanation as to why and when it would be needed.

src/systems/detachable_joint/DetachableJoint.cc
```diff
@@ -249,10 +254,11 @@ void DetachableJoint::PreUpdate(
         // We do this by creating a detachable joint entity.
         this->detachableJointEntity = _ecm.CreateEntity();
 
+        std::string type = (this->weldChildToParent) ? "weld" : "fixed";
```
azeey: Can you update the documentation in the `DetachableJoint` component that "weld" is now another option?

azeey (commented):
Overall, I'm not a fan of the term `weld` being used as though it's a different type of joint than `fixed`. In the DART physics engine, the class that implements fixed joints is called [WeldJoint](https://github.com/dartsim/dart/blob/main/dart/dynamics/WeldJoint.hpp). So, the term could lead to confusion. I also think that this is an implementation detail/limitation of the bullet physics engine, so it would be better if the plugin parameter indicated that somehow. e.g. `enforce_constraints_manually`. It's wordy, but that's what I can think of right now.

src/systems/detachable_joint/DetachableJoint.cc
iche033: updated doc. [aaeb155](https://github.com/gazebosim/gz-sim/pull/2984/commits/aaeb1550237b554dea54b1b2e3832804638d61a6)

src/systems/detachable_joint/DetachableJoint.hh
iche033: added doc for param. [aaeb155](https://github.com/gazebosim/gz-sim/pull/2984/commits/aaeb1550237b554dea54b1b2e3832804638d61a6)

iche033:
> Overall, I'm not a fan of the term `weld` being used as though it's a different type of joint than `fixed`. In the DART physics engine, the class that implements fixed joints is called [WeldJoint](https://github.com/dartsim/dart/blob/main/dart/dynamics/WeldJoint.hpp). So, the term could lead to confusion. I also think that this is an implementation detail/limitation of the bullet physics engine, so it would be better if the plugin parameter indicated that somehow. e.g. `enforce_constraints_manually`. It's wordy, but that's what I can think of right now.

I changed the term `weld` to `enforce_fixed_contraint` for both the system parameter and the joint type. I chose to be more specific and convey that we only enforce 'fixed' constraints for now (as opposed to any type of constraints). We can add `_manually` to the end (or `manual_enforce_fixed_constraint`) to be more explicit but then I left it out to keep the name short. No strong opinion here.

iche033:
> I still feel like the implementation detail is leaking too much. What do you think of making this a parameter of the Physics system. Then the changes will be limited to `Physics.cc`. The `DetachableJoint` component will not need to be changed.

ok I limited the changes to the physics system. [f4be659](https://github.com/gazebosim/gz-sim/pull/2984/commits/f4be6590a257b43e8d9691c96155ebf98f501294) The resulting side-effect is that as opposed to the previous approach that lets users configure this setting for each individual detachable joint,  setting the `<enforce_fixed_constraint>` param to true in the physics system will now affect all fixed constraints.

azeey (approved):

PR #2950, opened 2025-06-13
Add support for loading System plugin from static plugin registry

shameekganguly:
# 🎉 New feature

## Summary
Gz Plugin added support for loading plugins from statically linked plugin libraries in https://github.com/gazebosim/gz-plugin/commit/a49b9f45366662663cd789b0efcfdfddb2526412.
This PR builds on top of this feature to enable System plugins to be loaded from the static plugin registry.

To indicate that a plugin should be loaded from the static registry, users can specify the sdf `plugin@filename` attribute as `static://<fully qualified plugin class name>` or `static://<registered plugin alias>`. (The `plugin@name` attribute is unused in this case and can conventionally use the same plugin name or alias stem as the `plugin@filename` attribute).

I updated the `SystemLoader::InstantiateSystemPlugin` method to check if the plugin filename conforms to the above pattern and attempt to instantiate the plugin from the static registry in this case. Loading fails if the plugin either cannot be found in the static registry or it does not implement the `System` interface.

I also added an integration test in `test/integration/load_system_static_registry.cc` to verify this feature (runs both in cmake and bazel build). This test uses the `TestModelSystem` test System class that I registered with the static plugin registry in `test/plugins/TestStaticModelSystem.cc`. Note that the `TestStaticModelSystem` static library needs to be linked with special flags into the test binary target in cmake to ensure that the plugin class symbols are not stripped out by the linker (similar to this [test target in gz plugin](https://github.com/gazebosim/gz-plugin/blob/gz-plugin3/test/integration/CMakeLists.txt#L47)).

## Test it
bazel:
```
$ bazel test test:INTEGRATION_load_system_static_registry
```
cmake:
```
$ colcon test --merge-install --event-handlers console_direct+ --packages-select gz-sim9 --ctest-args -R INTEGRATION_load*
```

## Checklist
- [X] Signed all commits for DCO
- [X] Added tests
- [ ] Added example and/or tutorial
- [ ] Updated documentation (as needed)
- [ ] Updated migration guide (as needed)
- [ ] Consider updating Python bindings (if the library has them)
- [X] `codecheck` passed (See [contributing](https://gazebosim.org/docs/all/contributing#contributing-code))
- [] All tests passed (See [test coverage](https://gazebosim.org/docs/all/contributing#test-coverage))
- [ ] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Agazebosim+archived%3Afalse+) to support the maintainers
[...]

PR #2991, opened 2025-07-23
Add support for loading physics engine plugins from static plugin registry

iche033:
# 🎉 New feature

Depends on https://github.com/gazebosim/gz-physics/pull/761

## Summary

https://github.com/gazebosim/gz-physics/pull/761 added support for registering physics engine plugins in the static plugin registry. This PR extends the physics system to load these physics engine plugins from the static plugin registry. 

To indicate that a physics engine plugin is from the static plugin registry, specify the filename with the prefix `static://`, e.g.

```xml
        <plugin
          filename="static://gz::sim::systems::Physics"
          name="gz::sim::systems::Physics">
          <engine>
            <filename>static://gz::physics::dartsim::Plugin</filename>
          </engine>
        </plugin>
```

This follows the same convention for loading static gz-sim systems introduced in https://github.com/gazebosim/gz-sim/pull/2950

## Test it

Note: to run the test below, you'll need https://github.com/gazebosim/gz-physics/pull/761. 

A `INTEGRATION_load_physics_system_static_registry` test is added in the bazel build. To run the test, update your `MODULE.bazel` to use a local copy of gz-physics with the changes in https://github.com/gazebosim/gz-physics/pull/761, e.g.

```diff
+local_path_override(
     module_name = "gz-physics",
-    strip_prefix = "gz-physics-gz-physics8",
-    urls = ["https://github.com/gazebosim/gz-physics/archive/refs/heads/gz-physics8.tar.gz"],
+    path = "<path_to_your_bzl_workspace>/gz-physics",
 )
 
```

then run the test:

```sh
bazel test -c opt //test:INTEGRATION_load_physics_system_static_registry --action_env=CC=/usr/bin/clang-19  --dynamic_mode=off --test_output=streamed --cache_test_results=no
```

### DIFF
--- a/include/gz/sim/Util.hh
+++ b/include/gz/sim/Util.hh
@@ -342,6 +342,15 @@ namespace gz
     GZ_SIM_VISIBLE std::optional<math::AxisAlignedBox> meshAxisAlignedBox(
       const sdf::Mesh &_sdfMesh);
 
+    /// \brief Get the static plugin prefix
+    /// \return The static plugin prefix
+    GZ_SIM_VISIBLE const std::string &staticPluginPrefixStr();
+
+    /// \brief Check if input filename of a library is a static plugin or not.
+    /// \param _filename_ Library filename to check
+    /// \return True if input filename has a static plugin string format.
+    GZ_SIM_VISIBLE bool isStaticPlugin(const std::string &_filename);
+
     /// \brief Environment variable holding resource paths.
     const std::string kResourcePathEnv{"GZ_SIM_RESOURCE_PATH"};
 
@@ -355,6 +364,9 @@ namespace gz
     /// \brief Environment variable holding paths to custom rendering engine
     /// plugins.
     const std::string kRenderPluginPathEnv{"GZ_SIM_RENDER_ENGINE_PATH"};
+
+    /// \brief Static plugin filename prefix string.
+    const std::string kStaticPluginFilenamePrefix{"static://"};
     }
   }
 }
--- a/src/systems/physics/Physics.cc
+++ b/src/systems/physics/Physics.cc
@@ -925,9 +925,6 @@ void Physics::Configure(const Entity &_entity,
         continue;
       }
 
-      //this->dataPtr->engine =
-      //    this->dataPtr->RequestPhysicsEngineFromPlugin(plugin);
-
       this->dataPtr->engine = physics::RequestEngine<
         physics::FeaturePolicy3d,
         PhysicsPrivate::MinimumFeatureList>::From(plugin);
--- a/src/systems/physics/Physics.cc
+++ b/src/systems/physics/Physics.cc
@@ -358,14 +358,6 @@ class gz::sim::systems::PhysicsPrivate
   /// \param[in] _ecm The entity component manager.
   public: void UpdateModelsBoundingBoxes(EntityComponentManager &_ecm);
 
-  /// \brief Get physics engine from plugin and verify it has all
-  /// requested features.
-  /// \param[in] _plugin Physics engine plugin to check
-  /// \return The physics engine pointer or null if the plugin
-  /// does not contain a physics engine that has the requested features.
-  // public: EngineTypePtr RequestPhysicsEngineFromPlugin(
-  //     const plugin::PluginPtr &_plugin);
-
   /// \brief Cache the top-level model for each entity.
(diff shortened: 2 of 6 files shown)


include/gz/sim/Util.hh
```diff
@@ -342,6 +342,15 @@ namespace gz
     GZ_SIM_VISIBLE std::optional<math::AxisAlignedBox> meshAxisAlignedBox(
       const sdf::Mesh &_sdfMesh);
 
+    /// \brief Get the static plugin prefix
+    /// \return The static plugin prefix
+    GZ_SIM_VISIBLE const std::string &staticPluginPrefixStr();
```
shameekganguly, line 347: Please replace https://github.com/gazebosim/gz-sim/blob/gz-sim9/src/SystemLoader.cc#L44 with this call so as to not redefine the string constant

iche033: replaced. [b023298](https://github.com/gazebosim/gz-sim/pull/2991/commits/b02329832c75389e252ad10c890079bac892ff66)

src/systems/physics/Physics.cc
```diff
@@ -852,87 +852,110 @@ void Physics::Configure(const Entity &_entity,
...
-    << "]. Have you checked the " << this->dataPtr->pluginPathEnv
-    << " environment variable?" << std::endl;
+    const size_t prefixLen = staticPluginPrefixStr().size();
+    const std::string pluginToInstantiate =
+        pluginLib.substr(prefixLen);
+    auto plugin = pluginLoader.Instantiate(pluginToInstantiate);
 
-    return;
-  }
+    this->dataPtr->engine = physics::RequestEngine<
```
shameekganguly, line 871: nit: might be good to check if plugin is null here before instantiating the engine ptr. That way, we can disambiguate between the plugin not being in the static registry vs not satisfying `MinimumFeatureList`.

src/systems/physics/Physics.cc
```diff
@@ -852,87 +852,110 @@ void Physics::Configure(const Entity &_entity,
...
-          << "library [" << pathToLib << "]." << std::endl;
-    return;
+    if (!this->dataPtr->engine)
+    {
+      gzerr << "Failed to load physics engine plugin: "
+            << "(Reason: static plugin registry does not contain the requested "
+               "plugin)\n"
+            << "- Requested plugin name: [" << pluginLib << "]\n";
+      return;
+    }
```
shameekganguly, line 882:
```suggestion
    }
    gzdbg << "Loaded [" << pluginLib <<"] from the static registry" << std::endl;;
```

test/integration/load_physics_system_static_registry.cc
```diff
@@ -0,0 +1,75 @@
...
+  // Verify that the physics system is loading the static dartsim plugin.
+  auto mockSystem = std::make_shared<MockSystem>();
+  mockSystem->postUpdateCallback =
+      [](const sim::UpdateInfo &,
+         const sim::EntityComponentManager &_ecm)
+      {
+        auto plugin = _ecm.ComponentData<components::PhysicsEnginePlugin>(
+            worldEntity(_ecm));
+        ASSERT_TRUE(plugin.has_value());
+        EXPECT_EQ("static://gz::physics::dartsim::Plugin", plugin.value());
```
shameekganguly: I think this test-case passes even if the physics engine is not successfully initialized from the static registry (you can verify by commenting it out from the deps for the test target). To improve coverage, can we add an object in the world and verify that it has a non-zero acceleration after 1 step?

iche033: extended test to verify that a model falls because of the physics engine. [71d5958](https://github.com/gazebosim/gz-sim/pull/2991/commits/71d595873ef30b60682d7490da9aa65eab7a8a94)

src/systems/physics/Physics.cc
iche033: [71d5958](https://github.com/gazebosim/gz-sim/pull/2991/commits/71d595873ef30b60682d7490da9aa65eab7a8a94)

src/systems/physics/Physics.cc
iche033: added separate checks for plugin and engine, [71d5958](https://github.com/gazebosim/gz-sim/pull/2991/commits/71d595873ef30b60682d7490da9aa65eab7a8a94)

Issue #1260, opened 2021-12-20
Download Fuel models on the background


chapulina:
When opening a world with models that have not been downloaded, there can be a long delay before the GUI opens.
It's also not possible to close gazebo while it's doing the initial download of models.

## Desired behavior

While models are loading, the user should be able to interact with the window and close it if wanted.

## Alternatives considered

I guess we could not open the window until the models are loaded, but that may be more confusing for users.

## Implementation suggestion

* Download models in a separate thread
* Display models as they're downloaded
* Cancel loading if a signal is received
* Display status to user of what models are being loaded and what are missing
* Once we have a splash screen (https://github.com/ignitionrobotics/ign-gui/issues/336), we can use it to display this information

ahcorde:
I have been working in this issue but I have some blockers. This is the branch of the current implementation https://github.com/ignitionrobotics/ign-gazebo/tree/ahcorde/download_background

- When I call `Root::Load()`

https://github.com/ignitionrobotics/ign-gazebo/blob/0e255a076c9063356c301e9d70118263051b3307/src/Server.cc#L168

I need to "fail" this call when there is a `include` tag because I launch a thread for each one of the `include` tag.

https://github.com/ignitionrobotics/ign-gazebo/blob/0e255a076c9063356c301e9d70118263051b3307/src/ServerPrivate.cc#L558-L561

The callback method will return an empty string making the `Load` method to fail, but when the thread has downloaded the resources are included. This is not problematic at all (we can filter the error in the load method), but when a include fails then I loose the `pose` and `name`, which are inside the `include` tag. But the model is not included because I force to fail to create the thread.

**First question** I don't know how to recover this data, is there any way to get this values ? the only idea that came to my mund is reloading the file again

This gif shows how models are downloaded in the background but without the right position and naming.

![download_background](https://user-images.githubusercontent.com/1933907/162485982-36a22a06-8119-4601-8663-444fa66003a2.gif)

-------- 

The other issue is about SDFs that define a geometry with an URI, this is still blocking the GUI. Like this one

```
 <model name="Radio">
      <pose>3 -1.5 0 0 0 0</pose>
      <static>true</static>
      <link name="link">
        <collision name="collision">
          <geometry>
            <mesh>
              <uri>https://fuel.ignitionrobotics.org/1.0/OpenRobotics/models/Radio/4/files/meshes/Radio.dae</uri>
            </mesh>
          </geometry>
        </collision>
        <visual name="visual">
          <geometry>
[...]

PR #2992, opened 2025-07-23
Parallel asset download

nkoenig:
Fixes: https://github.com/gazebosim/gz-sim/issues/1260

# 🎉 New feature

## Summary

This will download simulation assets in background threads by default. You can force simulation to wait for assets using the `--wait-for-assets` CLI option. I'm defaulting to async download because:

1. It's a better user experience to have some instant GUI feedback.
2. If you're using only the simulation server (no GUI), then you're in the experienced user category and can choose whether to wait for assets or not.

## Test it

1. Clear(or stash) your ~/.gz/fuel directory, and load a world with many assets asynchronously: 
```
gz sim -v4 tunnel.sdf
```

1. Try the same thing, but with blocking asset download:
```
gz sim -v4 --wait-for-assets tunnel.sdf
```

4. Try the same thing, but specify a certain number of iterations:
```
gz sim -v4 --iterations 10 tunnel.sdf
```

## Checklist
- [ ] Added tests

### DIFF
--- a/src/SimulationRunner.hh
+++ b/src/SimulationRunner.hh
@@ -572,7 +572,6 @@ namespace gz
       /// at the appropriate time.
       private: std::unique_ptr<msgs::WorldControlState> newWorldControlState;
 
-
       /// \brief Set if we need to remove systems due to entity removal
       private: bool threadsNeedCleanUp{false};
 


src/Server.cc
```diff
@@ -77,42 +82,62 @@ Server::Server(const ServerConfig &_config)
...
 
+  // This will create the simulation runners, but not entities.
+  this->dataPtr->CreateSimulationRunners(sdfRoot);
+
+  // Storing the sdf root. The ServerPrivate.hh header file mentions
+  // that other classes may keep pointers to child nodes of the root,
+  // so we need to keep this object around.
+  // However, everything seems to work fine without storing this.
+  // \todo(nkoenig): Look into removing the sdfRoot member variable.
+  this->dataPtr->sdfRoot = sdfRoot;
```
arjo129, line 119: Knit: I think its perfectly safe to remove this. It seems we no longer use this inside `ServerPrivate` and this is the only mention in `Server.cc`. While the copy constructor for `SdfRoot` has been optimized recently, this is still an unessecary operation.

src/SimulationRunner.hh
```diff
@@ -550,10 +572,21 @@ namespace gz
       /// at the appropriate time.
       private: std::unique_ptr<msgs::WorldControlState> newWorldControlState;
 
+
```
arjo129: Knit: Remove space

src/Server.cc
azeey: We mainly want to keep the `sdf::Element` shared pointers valid since they will be used by systems to configure themselves. Some of these plugins might traverse to their parent elements via a weak pointer. If the `sdf::Root` object is destroyed, it could cause these shared pointers to become invalid.

I do think we should use a pointer here to avoid the extra copy, but that could be a followup PR.

azeey:
I think the homebrew test failures are due to a type mismatch on component IDs between gz-sim and gz-msg. See https://github.com/gazebosim/gz-msgs/pull/529

src/SimulationRunner.hh
azeey: Fixed in https://github.com/gazebosim/gz-sim/pull/2992/commits/17834d5e6f4d03b9e8e2c5ba83c891ad58c45208

azeey (commented):
Haven't done a full review, but can confirm it works as described.

PR #3039, opened 2025-08-27
Update references of main to gz-sim10

iche033:
Part of https://github.com/gazebosim/gz-jetty/issues/33


iche033:
replaced by #3040

PR #1073, opened 2021-09-29
Fix guirunner timing

nkoenig:
# 🦟 Bug fix

## Summary

The GuiRunner fails to get entities spawned from the GazeboFactory plugin, in ign-launch, when simulation is started in the running state.

This issue was found when porting SubT to Fortress. The following subt launch command will cause the GUI to miss robot creation:

```
ign launch -v 4 cloudsim_sim.ign worldName:=simple_ca_01 circuit:=cave robotName1:=X1 robotConfig1:=X1_SENSOR_CONFIG_1
```

Here is what's happening under the hood:

1. Factory message is received by the gazebo server.
2. The GuiRunner get the initial state and then subscribes to state topic.
3. The ECM creates the new robot.
4. The GuiRunner's subscription is connected (missing the robot creation event).

## Checklist
- [x] Signed all commits for DCO
- [ ] Added tests
- [ ] Updated documentation (as needed)
- [ ] Updated migration guide (as needed)
- [x] `codecheck` passed (See [contributing](https://ignitionrobotics.org/docs/all/contributing#contributing-code))
- [ ] All tests passed (See [test coverage](https://ignitionrobotics.org/docs/all/contributing#test-coverage))
- [ ] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Aignitionrobotics+repo%3Aosrf%2Fsdformat+archived%3Afalse+) to support the maintainers

**Note to maintainers**: Remember to use **Squash-Merge**

PR #3044, opened 2025-08-29
Update GUI Runner initial state request timing

iche033:
# 🦟 Bug fix

Related issue: https://github.com/gazebosim/jetty_demo/issues/3

## Summary

While testing the [Jetty demo](https://github.com/gazebosim/jetty_demo/) world, it was found that Gazebo took a while to load because there is a large number of entities in the world. The slow down is mainly on the GUI side. One cause of the long load time was found to be that the GuiRunner setting its ECM initial state twice: Once from the [async state request](https://github.com/gazebosim/gz-sim/blob/7267e3648a9c9a0406dfd59ad55819947986f06d/src/gui/GuiRunner.cc#L255) and the other from the [state topic subscription callback](https://github.com/gazebosim/gz-sim/blob/7267e3648a9c9a0406dfd59ad55819947986f06d/src/gui/GuiRunner.cc#L251) (sync version). 

More info:

There was a PR (https://github.com/gazebosim/gz-sim/pull/1073) to only start the processing the states from topic subscription callback (sync version) only after the async version is done, which is tracked using the `this->dataPtr->receivedInitialState` bool variable. That worked fine until this [merge forward change](https://github.com/gazebosim/gz-sim/commit/63d4983790a4c177c101250eac678b4e7fcb10c9#diff-6ce9681832237a23588ca3966cc842bba8f76e2283d06f020de3136e9b442408L174) that changed the processing of the state request in the aysnc callback to a `QueuedConnection`. As the result, the state is not processed immediately but `this->dataPtr->receivedInitialState` is incorrectly marked as `true`. This causes the sync version to start processing states right away.

The fix is to properly set `this->dataPtr->receivedInitialState` to `true` in the state callback function after the states are actually processed.

This patch reduces the load time by ~15 seconds (from ~1min 10sec to ~55 sec) on my machine.

## Checklist
- [ ] Added tests

### DIFF
--- a/src/gui/GuiRunner.cc
+++ b/src/gui/GuiRunner.cc
@@ -74,7 +74,7 @@ class gz::sim::GuiRunner::Implementation
   public: std::thread updateThread;
 
   /// \brief True if the initial state has been received and processed.
-  public: bool receivedInitialState{false};
+  public: std::atomic<bool> receivedInitialState{false};
 
   /// \brief Name of WorldControl service
   public: std::string controlService;
@@ -264,7 +264,6 @@ void GuiRunner::OnStateAsyncService(const msgs::SerializedStepMap &_res)
   // variables.
   QMetaObject::invokeMethod(this, "OnStateQt", Qt::QueuedConnection,
                             Q_ARG(gz::msgs::SerializedStepMap, _res));
-  this->dataPtr->receivedInitialState = true;
 
   // todo(anyone) store reqSrv string in a member variable and use it here
   // and in RequestState()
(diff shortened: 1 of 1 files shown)


arjo129 (approved):
Thanks for catching this. I think at some point someone should get a full flamegraph of the startup process using a tool like `perf`.

ahcorde (approved):

iche033:
> Thanks for catching this. I think at some point someone should get a full flamegraph of the startup process using a tool like `perf`.

+1

iche033:
the homebrew test failure looks unrelated

azeey:
@iche033 Can you backport this to ionic and harmonic? Also, we should probably make a prerelease with this.

iche033:
@mergifyio backport gz-sim9 gz-sim8

iche033:
> @iche033 Can you backport this to ionic and harmonic? Also, we should probably make a prerelease with this.

yep, I'll do a little more profiling and see if it's possible to further improve GUI startup time. I'll do another pre-release afterwards

PR #3075, opened 2025-09-05
CAD to SDFormat Tutorial

sanjanasrinivasan:
## Checklist
- [ ] Added documentation to the Tutorial section.


azeey:
> This tutorial should live in https://github.com/gazebosim/docs. What do you think @azeey ?

I thought about this before recommending this repo to @sanjanasrinivasan. I think in the long run, yes, I think this belongs in gazebosim/docs, but I would consider this as a "How to guide" similar to https://docs.ros.org/en/kilted/How-To-Guides.html rather than part of a tutorial series that currently exists in gazebosim/docs. And since we don't have that section there and there's some precedent to put blender tutorials in here, I suggested for her to create the PR here. I'm hoping to create the how-to guide section in the next couple months and move all of the blender tutorials there.

azeey (commented):
Thanks for the contribution @sanjanasrinivasan!

I haven't had a chance to do a full review yet, but I wanted to mention that you'll need to add an entry in https://github.com/gazebosim/gz-sim/blob/gz-sim10/tutorials.md.in so that this tutorial will show up on https://gazebosim.org/api/sim/10/tutorials.html once it's merged.

PR #3070, opened 2025-09-05
Skip Model SDF serialization

iche033:
# 🦟 Bug fix

Related https://github.com/gazebosim/sdformat/issues/1478

## Summary

Skip model sdf serialization. The main problem is actually in the perf of deserialization (`root.LoadSdfString`), see https://github.com/gazebosim/sdformat/issues/1478. This change returns an empty string in the `Serialize` function which will cause the `Deserialize` function to just skip trying to load the sdf string. 

Testing with the Jetty demo world, the reduced load time from 55s to 42s and resuming from paused state (ie hit Play button) is now  almost instantaneous.

## Checklist
- [x] Signed all commits for DCO
- [ ] Added tests
- [ ] Updated documentation (as needed)
- [ ] Updated migration guide (as needed)
- [ ] Consider updating Python bindings (if the library has them)
- [ ] `codecheck` passed (See [contributing](https://gazebosim.org/docs/all/contributing#contributing-code))
- [ ] All tests passed (See [test coverage](https://gazebosim.org/docs/all/contributing#test-coverage))
- [ ] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Agazebosim+archived%3Afalse+) to support the maintainers
- [ ] Was GenAI used to generate this PR? If so, make sure to add "Generated-by" to your commits. (See [this policy](https://osralliance.org/wp-content/uploads/2025/05/OSRF-Policy-on-the-Use-of-Generative-Tools-Generative-AI-in-Contributions.pdf) for more info.)

Generated-by: Remove this if GenAI was not used.

**Note to maintainers**: Remember to use **Squash-Merge** and edit the commit message to match the pull request summary while retaining `Signed-off-by` and `Generated-by` messages.

PR #3082, opened 2025-09-09
Improve Model Sdf deserialization performance

iche033:
# 🦟 Bug fix

## Summary

Alternative to https://github.com/gazebosim/gz-sim/pull/3070. Instead of completely skipping the model serialization / materialization step, this PR uses a workaround to speed up the deserialization process. This is done by implementing a modified version of [sdf::Root::LoadSdfString](https://github.com/gazebosim/sdformat/blob/067c73bc5328c9687da1b0e43588f324722e5e7a/src/Root.cc#L217). It uses a static `sdf::SDFPtr` object to avoid multiple expensive `sdf::init` calls. Note that this workaround is similar to the one done in [sdf/src/pasrer.cc](https://github.com/gazebosim/sdformat/blob/067c73bc5328c9687da1b0e43588f324722e5e7a/src/parser.cc#L1741-L1745). 

We could also implement this workaround directly in sdformat but this I chose to implement it here to reduce the impact of this change during the code freeze period.

For the [Jetty demo world ](https://github.com/gazebosim/jetty_demo), the load times are:

* Before any changes: 55s
* Skip model sdf serialization (#3070): 42s
* this pull request: 48s

## Checklist
- [ ] Added tests


iche033:
Testing this with the Jetty demo world and https://github.com/gazebosim/sdformat/pull/1589,  I see that this PR still improves performance.

The initial load time dropped from 33 seconds to 22 seconds on my machine, and resuming to play state from paused state reduced from 17 to 3 seconds.

iche033:
changed from draft to ready status because I think this PR still helps with performance, especially when changing from paused to play state in the jetty demo world.

arjo129:
Thanks Ian! Just to clarify Im assuming we are sure that serialization and deserialization only ever happen in a single threaded context

iche033:
> Thanks Ian! Just to clarify Im assuming we are sure that serialization and deserialization only ever happen in a single threaded context

Not sure about serialization but this changes only the deserialization part. In the current gz-sim use case, I believe deserialization only happens in a single thread. Deserialization happens when the gui receives a state from the server, i.e. in `GuiRunner::OnStateQt`
https://github.com/gazebosim/gz-sim/blob/fefd16bd3e1f2fa515b3b2363df4c80eb96f7493/src/gui/GuiRunner.cc#L299 which then calls `Deserialize`
https://github.com/gazebosim/gz-sim/blob/fefd16bd3e1f2fa515b3b2363df4c80eb96f7493/src/EntityComponentManager.cc#L1981

The state callback function does not happen in multiple threads.

arjo129 (approved):

include/gz/sim/components/Model.hh
```diff
@@ -103,12 +104,30 @@ namespace serializers
         return _in;
       }
 
-      // Its super expensive to create an SDFElement for some reason
+      // Its super expensive to create an sdf::SDFPtr object.
+      // Workaround this by making it a static object so we only initialize it
+      // once.
+      // https://github.com/gazebosim/sdformat/issues/1478
+      sdf::Errors errors;
+      static sdf::SDFPtr sdfParsed;
```
arjo129:
```suggestion
      static sdf::SDFPtr sdfParsed;
```

So there are definitely places where this seems to introduce regressions:

https://build.osrfoundation.org/job/gz_sim-ci-pr_any-homebrew-amd64/2111/consoleFull#-140582033379332c8d-713a-4d09-92ce-eb424edb41b6

image

These tests will need to be fixed before we merge.

arjo129 (changes requested):
Our tests do rely on multi-threading and this patch would introduce a flake in our tests.

include/gz/sim/components/Model.hh
```diff
@@ -104,27 +105,31 @@ namespace serializers
...
-      if (!sdfParsed)
-      {
-        sdfParsed.reset(new sdf::SDF());
-        sdf::init(sdfParsed);
-      }
-      auto config = sdf::ParserConfig::GlobalConfig();
-      // Read an SDF string, and store the result in sdfParsed.
-      if (!sdf::readString(sdf, config, sdfParsed, errors))
+      sdf::Root root;
+      std::mutex mutex;
```
azeey: This would also have to be `static` as well since, otherwise, each instance of the component will have it's own mutex.
```suggestion
      static std::mutex mutex;
```

include/gz/sim/components/Model.hh
```diff
@@ -103,12 +106,34 @@ namespace serializers
...
       sdf::Root root;
-      sdf::Errors errors = root.LoadSdfString(sdf);
+      {
+        // It's super expensive to create an sdf::SDFPtr object.
+        // Workaround this by making it a static object so we only initialize it
+        // once.
+        // https://github.com/gazebosim/sdformat/issues/1478
+        sdf::Errors errors;
+        static std::mutex mutex;
+        static sdf::SDFPtr sdfParsed;
```
arjo129, line 117: Simply adding static to a mutex may or maynot work. I can see failures on mac OS where the mutex is throwing an exception in our tests. Apparently according to this [issue on another project](https://github.com/facebook/componentkit/commit/37bcfdb0f153b2fb4f06bcdf24451d6c24e59532), this may have to do with destruction. I dont think itll impact UX but it'll make our tests more flaky.

In hind-sight perhaps #3070 may be a better option for future releases of gazebo ( GZ K and up)

iche033: I saw the test failure and haven't been able to resolve it yet. I wasn't able to reproduce the crash locally on my mac. I'm also testing different workarounds and rerunning homebrew builds and will see if I have any luck. 

Yeah, we can retarget #3070 to `main` as it causes a behavior change.

arjo129:
Since https://github.com/gazebosim/sdformat/pull/1610 is merged, Im closing this.

PR #3102, opened 2025-10-09
Fix reset for rendering sensors

azeey:
# 🦟 Bug fix

## Summary
In the case where objects are spawned after simulation has started and a Reset occurs, the Sensors system was failing to remove the spawned objects. Subsequent attempts to respawn the same objects also fail because they are likely to be assigned the same Entity IDs.

The thing to note here is that the `EachNew` and `EachRemoved` function of the ECM have to be called from the Reset handler, since the newly created/removed entities are cleared by the next PreUpdate.

TODO:
- [ ] Suppress the "already exists" warning if a Reset has occured. 

## Checklist
- [ ] Added tests

### DIFF
--- a/src/systems/sensors/Sensors.cc
+++ b/src/systems/sensors/Sensors.cc
@@ -19,6 +19,7 @@
 
 #include <atomic>
 #include <chrono>
+#include <mutex>
 #include <set>
 #include <unordered_map>
 #include <unordered_set>
@@ -867,7 +868,7 @@ void Sensors::Configure(const Entity &/*_id*/,
 }
 
 //////////////////////////////////////////////////
-void Sensors::Reset(const UpdateInfo &_info, EntityComponentManager &)
+void Sensors::Reset(const UpdateInfo &_info, EntityComponentManager &_ecm)
 {
   GZ_PROFILE("Sensors::Reset");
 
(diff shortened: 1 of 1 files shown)


arjo129:
Any chance we can get a regression test for this?

src/systems/sensors/Sensors.cc
```diff
@@ -899,6 +900,17 @@ void Sensors::Reset(const UpdateInfo &_info, EntityComponentManager &)
...
+
+  {
+    std::unique_lock<std::mutex> lock(this->dataPtr->renderUtilMutex);
+    this->dataPtr->updateTimeCv.wait(lock, [this]()
+    {
+      return !this->dataPtr->updateAvailable ||
+              (this->dataPtr->updateTimeToApply ==
+              this->dataPtr->updateTimeApplied);
+    });
+    this->dataPtr->renderUtil.UpdateFromECM(_info, _ecm);
```
iche033, line 912: I arrived at a very similar fix. I had this call inside the mutex lock block a few lines above, i.e.

```cpp
    std::unique_lock<std::mutex> lock2(this->dataPtr->renderUtilMutex);
    this->dataPtr->updateTime =  _info.simTime;
    this->dataPtr->updateTimeToApply =  _info.simTime;
    this->dataPtr->updateTimeApplied =  _info.simTime;
    this->dataPtr->renderUtil.UpdateFromECM(_info, _ecm);
```

but want to check, does this need to be in a separate block with the wait condition?

azeey: I'm not 100% sure. Did it work for you? I saw the comment
https://github.com/gazebosim/gz-sim/blob/7374a6c1197ce13c9e0317a280ed8379a66c7489/src/systems/sensors/Sensors.cc#L974-L980

which made me think it was necessary.

iche033: yeah it worked for me. 

My thinking is that we could just override the state in renderUtil without waiting for any pending updates (updates from before the reset) to be applied because it's going to be reset to the initial state anyway.

PR #3138, opened 2025-10-20
EnvironmentPreload visualise static environments

srmainwaring:
# 🦟 Bug fix

## Summary

This PR fixes an issue where a static environment loaded by the EnvironmentPreload system would not display using the Environment Visualization Resolution and Point Cloud GUI widgets.

There is also a fix that prevents segmentation faults in the VisualizationTool caused by copying a DataFrame which may invalidate the session iterators.

## Details

Before this change, if the `<ignore_time>` element is added to the EnvironmentPreload system xml it is ignored when visualising the point cloud data.

```xml
    <plugin
      filename="gz-sim-environment-preload-system"
      name="gz::sim::systems::EnvironmentPreload">
      <data>environmental_data.csv</data>
      <dimensions>
        <ignore_time>1</ignore_time>
        <time>timestamp</time>
        <space>
          <x>x</x>
          <y>y</y>
          <z>z</z>
        </space>
      </dimensions>
    </plugin>
``` 

This means that if only the time zero data is provided, for example:

```bash
# environmental_data.csv - modified to only contain data for timestamp = 0
timestamp,humidity,x,y,z
0,80,-1,-1,-1
0,80,-1,-1, 1
0,80,-1, 1,-1
0,80,-1, 1, 1
0,90, 1,-1,-1
0,90, 1,-1, 1
0,90, 1, 1,-1
0,90, 1, 1, 1
```

the VisualizationTool would never display the point cloud data.

A workaround for this is to set a very large timestamp for a second copy of data. This is unsatisfactory because it doubles the size of the environment data required (which may be significant for large vector fields), and also requires a temporal interpolation when none is required, which results in a performance penalty.

The other fix is to pay attention to unintended copies when using the auto keyword to obtain aliases to the data frame objects. The main change is

```diff
- auto frame = _data->frame[this->pubs.begin()->first];
+ const auto &frame = _data->frame[this->pubs.begin()->first];
```

The problem is that the session iterators will in general be invalid for the copied frame, and this may result in a segmentation fault when they are dereferenced later in the code. In this particular case the check that the iterator was in range in the interpolation code failed resulting in a out of range access violation a few lines later. This type of error can be difficult to trace.

With this change, and some other related fixes to the environmental data system, large static vector fields can be loaded and viewed in the GUI. Below is an example for a wind field containing 64000 rows of data.
 
05-1-gz-wind-pc-x-axis

## Checklist
- [ ] Added tests

### DIFF
--- a/examples/worlds/environmental_sensor.sdf
+++ b/examples/worlds/environmental_sensor.sdf
@@ -36,6 +36,7 @@
       name="gz::sim::systems::EnvironmentPreload">
       <data>environmental_data.csv</data>
       <dimensions>
+        <ignore_time>0</ignore_time>
         <time>timestamp</time>
         <space>
           <x>x</x>
--- a/src/systems/environment_preload/EnvironmentPreload.cc
+++ b/src/systems/environment_preload/EnvironmentPreload.cc
@@ -349,7 +349,7 @@ void EnvironmentPreload::PreUpdate(
   if (this->dataPtr->visualize)
   {
     std::lock_guard<std::mutex> lock(this->dataPtr->mtx);
-    auto samples = this->dataPtr->samples;
+    const auto &samples = this->dataPtr->samples;
     this->dataPtr->visualizationPtr->Step(_info, _ecm, this->dataPtr->envData,
       samples.X(), samples.Y(), samples.Z());
   }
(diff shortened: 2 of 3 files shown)


arjo129 (approved):
Thanks for the changes. I tested them and they seem to be working. I realize I have to revisit #2608.

arjo129:
@mergifyio backport gz-sim9 gz-sim8 ign-gazebo6

arjo129:
@mergifyio backport main

PR #3009, opened 2025-08-01
Populate default material for model

ntfshard:
# 🦟 Bug fix

Fixes #<NUMBER>

## Summary
Adding model with Entity Tree gui plugin, or in sdf w/o material won't creating a material entity in ECM. Object will looks flat and it's not possible to change material properties in gui. Populating default material will allow to configure it in gui.

Behaviour changing: before color of model was white and after this change it will be black. According to SDF specification http://sdformat.org/spec?elem=material&ver=1.9 it's correct behaviour. So technically it's a fix

**Before (no Material submenu):**
Screenshot 2025-08-02 032129

**After:**
Screenshot 2025-08-02 033135
image

I checked this change on gz-sim8 branch (I already had all libraries compiled); I guess it should be the same on gz-sim9

## Checklist
- [ ] Signed all commits for DCO
- [ ] Added tests
- [ ] Updated documentation (as needed)
- [ ] Updated migration guide (as needed)
- [ ] Consider updating Python bindings (if the library has them)
- [ ] `codecheck` passed (See [contributing](https://gazebosim.org/docs/all/contributing#contributing-code))
- [ ] All tests passed (See [test coverage](https://gazebosim.org/docs/all/contributing#test-coverage))
- [ ] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Agazebosim+archived%3Afalse+) to support the maintainers
- [ ] Was GenAI used to generate this PR? If so, make sure to add "Generated-by" to your commits. (See [this policy](https://osralliance.org/wp-content/uploads/2025/05/OSRF-Policy-on-the-Use-of-Generative-Tools-Generative-AI-in-Contributions.pdf) for more info.)

**Note to maintainers**: Remember to use **Squash-Merge** and edit the commit message to match the pull request summary while retaining `Signed-off-by` and `Generated-by` messages.

PR #3139, opened 2025-10-20
Populate default material for model

ntfshard:
# 🦟 Bug fix

Fixes #<NUMBER>

## Summary
Adding model with Entity Tree gui plugin, or in sdf w/o material won't creating a material entity in ECM. Object will looks flat and it's not possible to change material properties in gui. Populating default material will allow to configure it in gui.

Behaviour changing: before color of model was white and after this change it will be black. According to SDF specification http://sdformat.org/spec?elem=material&ver=1.9 it's correct behaviour. So technically it's a fix

(This is a port of changes from #3009 to main branch; and screenshots are available in initial PR)

## Checklist
- [ ] Added tests

### DIFF
--- a/test/integration/mesh_uri.cc
+++ b/test/integration/mesh_uri.cc
@@ -60,19 +60,20 @@ std::string meshModelStr(bool _static = false)
   // it should be able to load this as a mesh.
   // Similarly the user can create a common::Mesh object and add that
   // to the MeshManager by calling MeshManager::Instance()->AddMesh;
-  return std::string("<?xml version=\"1.0\" ?>") +
-      "<sdf version='1.6'>" +
-      "<model name='spawned_model'>" +
-      "<link name='link'>" +
-      "<visual name='visual'>" +
-      "<geometry><mesh><uri>name://unit_box</uri></mesh></geometry>" +
-      "</visual>" +
-      "<collision name='visual'>" +
-      "<geometry><mesh><uri>name://unit_box</uri></mesh></geometry>" +
-      "</collision>" +
-      "</link>" +
-      "<static>" + std::to_string(_static) + "</static>" +
-      "</model>" +
+  return std::string("<?xml version=\"1.0\" ?>"
+      "<sdf version='1.6'>"
+      "<model name='spawned_model'>"
+      "<link name='link'>"
+      "<visual name='visual'>"
+      "<geometry><mesh><uri>name://unit_box</uri></mesh></geometry>"
+      "<material><emissive>1 1 1 1</emissive></material>"
+      "</visual>"
+      "<collision name='collision'>"
+      "<geometry><mesh><uri>name://unit_box</uri></mesh></geometry>"
+      "</collision>"
+      "</link>"
+      "<static>") + std::to_string(_static) + "</static>"
+      "</model>"
       "</sdf>";
 }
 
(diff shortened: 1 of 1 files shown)


arjo129:
I'm hesitant to back-port this. I know @azeey suggested back-porting, but it would mean some end users get a bit of a shock when they wake up to find their simulations change color. Perhaps the less jarring fix would be to make the default white in previous gz-sim versions and then keep things black on the main branch.

ntfshard:
> I'm hesitant to back-port this. I know @azeey suggested back-porting, but it would mean some end users get a bit of a shock when they wake up to find their simulations change color. Perhaps the less jarring fix would be to make the default white in previous gz-sim versions and then keep things black on the main branch.

I guess "their simulations changed color" is a very big generalization in this context. I guess there are no users who are using models w/o materials due to you can't see any edges and faces of a model. It's just a uniformly colored silhouette of the model (first screenshot in initial pull request). It looks strange and not convenient (in context of 3d graphics).

Keeping different behaviour will just increase complexity of codebase support imo. More merge conflicts

scpeters:
I think there are two concerns that this pull request addresses:

1. "it's not possible to change material properties in gui. Populating default material will allow to configure it in gui."
2. The default material properties don't match the SDFormat specification (white instead of black).

I would recommend separating these two concerns into separate pull requests:

1. Add a material (white by default) to ECM so that it can be configured through the GUI.
2. Change the default material properties from white to something else.

I think it would not be controversial to merge and backport the bugfix for the first concern. The second concern is more complex, but I would rather have that discussion in a way that doesn't block the bugfix.

ntfshard:
Initial problem was: I can't see shape of added mesh object (I guess I spawned it from plugin). And I even can't fix it in gui.

For alignment, here is how basic shapes and mesh looks if user not provided any material:

Screenshot 2025-10-22 at 21 26 30

[world_example.zip](https://github.com/user-attachments/files/23053577/world_example.zip)
(Took this picture on not-modified Harmonic version.)

If we specify default style in `SdfEntityCreator::CreateEntities(const sdf::Visual *_visual)` I guess it will impact everything with visual. (all basic shapes w/o material became not black in this case in other words)

test/integration/mesh_uri.cc
```diff
@@ -60,19 +60,20 @@ std::string meshModelStr(bool _static = false)
...
-      "</link>" +
-      "<static>" + std::to_string(_static) + "</static>" +
-      "</model>" +
+  return std::string("<?xml version=\"1.0\" ?>"
+      "<sdf version='1.6'>"
+      "<model name='spawned_model'>"
+      "<link name='link'>"
+      "<visual name='visual'>"
+      "<geometry><mesh><uri>name://unit_box</uri></mesh></geometry>"
+      "<material><emissive>1 1 1 1</emissive></material>"
```
ntfshard, line 69: I pushed commit to make cube explicitly white.
White was initial color of it. Previous version relied on default behaviour.

ntfshard:
I found why it shouldn't be merged: if we have a material in ECM, even if it's default, material information from mesh file will be ignored. 
Still, having `visual` w/o material in ECM is not very convenient in general, but it should be or clear priority between ecm record and what is inside file or some explicit override methods.
This problem has several approaches, but seems not very high priority.

Feel free to close this PR

PR #3244, opened 2025-12-27
ParticleEmitter can use topic from SDF

ntfshard:
# 🦟 Bug fix

Fixes #<NUMBER>

## Summary
Seems particle_emitter not using topic from [SDF](https://sdformat.org/spec/1.11/link/#particle_emitter_topic). It looks like old workaround for reading topic from msg.header is not working in a new (codebase src/systems/particle_emitter/ParticleEmitter.cc:159-168)
Also it seems rest of old logic still here (same file, below comment `// Set the topic information if it was not set via SDF.`)
(not sure, should it be here, or both loops should be removed.

Attached simple reproducer, based on example. Topic can be checked with `gz topic -l`
(not sure how to make a good simple test for this)
[emitter.sdf.txt](https://github.com/user-attachments/files/24356174/emitter.sdf.txt)

## Checklist
- [ ] Added tests

simple reproducer:
[emitter.sdf.txt](https://github.com/user-attachments/files/24356174/emitter.sdf.txt)

### DIFF
--- a/.bcr/metadata.template.json
+++ b/.bcr/metadata.template.json
@@ -1,0 +1,22 @@
+{
+    "homepage": "https://github.com/gazebosim/gz-sim",
+    "maintainers": [
+        {
+            "email": "iche@intrinsic.ai",
+            "github": "iche033",
+            "github_user_id": 4000684,
+            "name": "Ian Chen"
+        },
+        {
+            "email": "shameek@intrinsic.ai",
+            "github": "shameekganguly",
+            "github_user_id": 2412842,
+            "name": "Shameek Ganguly"
+        }
+    ],
+    "repository": [
+        "github:gazebosim/gz-sim"
+    ],
+    "versions": [],
+    "yanked_versions": {}
+}
--- a/.bcr/presubmit.yml
+++ b/.bcr/presubmit.yml
@@ -1,0 +1,20 @@
+matrix:
+  platform:
+  - ubuntu2004
+  - macos
+  # Disabled due to https://github.com/bazel-contrib/rules_foreign_cc/issues/1305.
+  # rules_foreign_cc is transitively pulled in through libzmq.
+  # - macos_arm64
+  bazel:
+  - 8.x
+  - 7.x
+tasks:
+  verify_targets:
+    name: Verify build targets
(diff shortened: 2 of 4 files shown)


src/systems/particle_emitter/ParticleEmitter.cc
```diff
@@ -152,11 +152,11 @@ void ParticleEmitter::PreUpdate(const gz::sim::UpdateInfo &_info,
           const components::ParentEntity *_parent,
           const components::Pose *_pose)->bool
         {
-          std::string topic;
+          std::string topic = _emitter->Data().topic().data();
 
           // Get the topic information from the header, which is currently a
           // hack to avoid breaking the particle_emitter.proto message.
-          if (_emitter->Data().has_header())
+          if (topic.empty() && _emitter->Data().has_header())
```
iche033, line 159: I think we can just remove the logic here that gets the topic from the header. That was changed in https://github.com/gazebosim/gz-sim/pull/2461

ntfshard: Had worry about users, who can rely on it in their plugins. Ok, will remove it

And what about this part in this file, just curious
Screenshot 2026-01-07 at 12 27 23

iche033: Given that this  PR is targeted at `main`, I think it's fine to remove. Same for this part of the code.

ntfshard: Hm... But we need this fix in current LTS release too. What is the better approach in this case?

Upd: maybe we can merge and port this PR as is and I can make cleanup for main branch as a separate PR?

iche033: hmm ok, let's merge this without removing the header stuff so I can backport these changes to release versions. We can remove the logic here later in a separate PR targeted only at `main`

iche033:

src/systems/particle_emitter/ParticleEmitter.cc
ntfshard: https://github.com/gazebosim/gz-sim/pull/3260

PR #3267, opened 2026-01-15
Set parent world entity before loading plugins for spawned entities

iche033:
# 🦟 Bug fix

## Summary

This PR fixes the Pose Publisher's topic name for models spawned into the world. It should now be in the format `/model/<model_name>/pose` (instead of `/model_name/pose`).

More info:

When a model is spawned, calling `scopedName` (and similarly `topicFromScopedName`) in the system's `Configure` function does not generate a scoped name that includes the world entity.  As an example, `scopedName` returns:

```sh
# expected:
/world/empty/model/box

# actual:
/model/box
```

This is because the plugin is loaded before the parent world entity is set for the model. This PR fixes the issue in `UserCommands` by loading the entities and plugins in the same order as it is done by `SdfEntityCreator` (see [code here](https://github.com/gazebosim/gz-sim/blob/9a6478a35b29674e25b065facf3ab4f3dcbc8541/src/SdfEntityCreator.cc#L335-L337)).  I had to make two private functions `public` in `SdfEntityCreator` so I can replicate the logic in `UserCommands`.

## Checklist
- [x] Added tests


src/systems/user_commands/UserCommands.cc
```diff
@@ -1026,7 +1026,7 @@ bool CreateCommand::CreateFromMsg(const msgs::EntityFactory &_createMsg)
   {
     auto model = *root.Model();
     model.SetName(desiredName);
-    entity = this->iface->creator->CreateEntities(&model);
+    entity = this->iface->creator->CreateEntities(&model, false);
```
azeey: Is this used here to prevent plugins from being loaded? If so, I'd rather add a new API to make this explicit rather than relying on a parameter called `_staticParent` to implicitly keep plugins from being loaded.

iche033: added a new function called `CreateEntitiesWithoutLoadingPlugins` in [e50a3cc](https://github.com/gazebosim/gz-sim/pull/3267/commits/e50a3cca38378e170fc30fb281282641f52e66a4)

PR #3268, opened 2026-01-16
DetachableJoint: add initial_attach SDF parameter

### LINKED ISSUE #3021: Add an option to specify the initial attachment state in DetachableJoint-plugin
## Desired behavior
Currently the DetachableJoint-plugin attaches the child joints to the parent entity when the plugin is initially loaded. In the DetachableJoint.hh we can see the initial values that dictate this:

`private: std::atomic<bool> attachRequested{true};`
`private: std::atomic<bool> detachRequested{false};`
`private: std::atomic<bool> isAttached{false};  `

In the PreUpdate method of the plugin, the check for attachRequested is always true on initial load and thus the joint is attached to the parent. It would be nice to change this initial behaviour if desired.

## Alternatives considered
None

## Implementation suggestion
Add a sdf tag called **initial_attach** to the plugin which can be set to true or false by the user. When the plugin loads, the existence and value of this tag is checked and the value for the attachRequested is changed accordingly. The default value could stay as it is. Adding this tag would not break systems that utilize the plugin as it would not be required.

## Additional context
[DetachableJoint.hh](https://github.com/gazebosim/gz-sim/blob/cdd1a76274eb3dbf8936537d60230960d276e0fb/src/systems/detachable_joint/DetachableJoint.hh#L138)

[DetachableJoint.cc](https://github.com/gazebosim/gz-sim/blob/cdd1a76274eb3dbf8936537d60230960d276e0fb/src/systems/detachable_joint/DetachableJoint.cc#L225)

SurafelAnshebo:
# 🎉 New feature

Closes #3021

## Summary
This PR adds an `initial_attach` SDF parameter to the `DetachableJoint` system,
allowing users to control whether a joint starts attached or detached on
simulation startup.

Previously, joints were always initialized as attached, requiring to be detached later. This change enables declarative control of the initial joint state directly from SDF, which is useful for scripted assembly/disassembly setup.

The default behavior is preserved (`initial_attach` defaults to `true`) to
maintain backward compatibility.

## Test it
1. Build `gz-sim` from source
2. Load a world using the `DetachableJoint` system with:
   ```xml
   <plugin
     filename="gz-sim-detachable-joint-system"
     name="gz::sim::systems::DetachableJoint">
     <initial_attach>false</initial_attach>
   </plugin>

### Demo
Short simulation video showing `initial_attach=false` behavior:
https://github.com/user-attachments/assets/742f042d-10b2-42d6-9986-44e4f0f7f593

### DIFF
--- a/src/systems/detachable_joint/DetachableJoint.hh
+++ b/src/systems/detachable_joint/DetachableJoint.hh
@@ -50,6 +50,11 @@ namespace systems
   /// - `<child_link>`: Name of the link in the child model to be used in
   /// creating a fixed joint with a link in the parent model.
   ///
+  /// - `<initial_attach>` (optional): If true, the joint starts attached when
+  /// the system loads. If false, the joint starts detached (no initial joint is
+  /// created until an attach request is received). Defaults to true to preserve
+  /// backward compatibility.
+  ///
   /// - `<topic>` (optional): Topic name to be used for detaching connections.
   /// Using <detach_topic> is preferred.
   ///
--- a/test/integration/detachable_joint.cc
+++ b/test/integration/detachable_joint.cc
@@ -435,3 +435,81 @@ TEST_F(DetachableJointTest,
    // should be close.
    EXPECT_TRUE(abs(distTraveledB1 - distTraveledVehicle) < 0.01);
  }
+
+/////////////////////////////////////////////////
+// Test that initial_attach parameter works correctly when set to false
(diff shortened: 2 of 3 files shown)


SurafelAnshebo:
@azeey I've added the documentation and test as requested:

I have added `<initial_attach>` parameter description to DetachableJoint.hh
I have also added InitialAttachFalse gtest that checks the joint starts detached and can be re-attached later. All tests passed. (attached a screen grab)

I believe its ready for review!
Integration_test

PR #3272, opened 2026-01-19
Add getters for joint velocity, effort, and position limits

### LINKED ISSUE #2664: Add API for getting Joint limits

naitikpahwa18:
# 🎉 New feature

Closes #2664

## Summary
This PR adds getter APIs for retrieving joint velocity, effort, and position limits.

gz::sim::Joint already provides setter methods for velocity limits, but
no corresponding getter. This change adds Joint::VelocityLimits(), which
reads the limit from the existing components::JointAxis state
(sdf::JointAxis::MaxVelocity), including support for multi-axis joints.

In addition, this PR also introduces getter methods for effort and position
limits for completeness and API symmetry with the existing setters.

## Test it
ctest -R Joint_TEST

## Checklist
- [x] Added tests

### DIFF
--- a/include/gz/sim/Joint.hh
+++ b/include/gz/sim/Joint.hh
@@ -276,14 +276,15 @@ namespace gz
           const EntityComponentManager &_ecm) const;
 
       /// \brief Get joint velocity limits.
-      /// \details For each axis, returns the maximum absolute velocity limit. 
-      /// The size of the returned vector corresponds to the joint's degrees of freedom.
-      /// For multi-axis joints, limits are returned for each axis in order (for
-      /// example, including both JointAxis and JointAxis2 when applicable).
+      /// \details For each axis, returns the maximum absolute velocity limit.
+      /// The size of the returned vector corresponds to the joint's degrees
+      /// of freedom. For multi-axis joints, limits are returned for each axis 
+      /// in order (for example, including both JointAxis and JointAxis2 
+      /// when applicable).
       /// \param[in] _ecm Entity component manager.
       /// \return Velocity limits if available.
-      public: std::optional<std::vector<double>>
-      MaxVelocityLimits(const EntityComponentManager &_ecm) const;
+      public: std::optional<std::vector<double>> MaxVelocityLimits(
+          const EntityComponentManager &_ecm) const;
 
       /// \brief Private data pointer.
       GZ_UTILS_IMPL_PTR(dataPtr)
--- a/include/gz/sim/Joint.hh
+++ b/include/gz/sim/Joint.hh
@@ -276,14 +276,14 @@ namespace gz
           const EntityComponentManager &_ecm) const;
 
       /// \brief Get joint velocity limits.
-      /// \details For each axis, returns the maximum velocity limit. The size of
-      /// the returned vector corresponds to the joint's degrees of freedom. For
-      /// multi-axis joints, limits are returned for each axis in order (for
+      /// \details For each axis, returns the maximum absolute velocity limit. 
+      /// The size of the returned vector corresponds to the joint's degrees of freedom.
+      /// For multi-axis joints, limits are returned for each axis in order (for
       /// example, including both JointAxis and JointAxis2 when applicable).
       /// \param[in] _ecm Entity component manager.
       /// \return Velocity limits if available.
       public: std::optional<std::vector<double>>
-      VelocityLimits(const EntityComponentManager &_ecm) const;
+      MaxVelocityLimits(const EntityComponentManager &_ecm) const;
 
       /// \brief Private data pointer.
       GZ_UTILS_IMPL_PTR(dataPtr)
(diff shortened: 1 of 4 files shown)


include/gz/sim/Joint.hh
```diff
@@ -275,6 +275,18 @@ namespace gz
...
+      /// \brief Get joint velocity limits.
+      /// \details For each axis, the X() component of the gz::math::Vector2d
+      /// specifies the minimum velocity limit, and the Y() component specifies
+      /// the maximum velocity limit. The size of the returned vector corresponds
+      /// to the joint's degrees of freedom. For multi-axis joints, limits are
+      /// returned for each axis in order (for example, including both JointAxis
+      /// and JointAxis2 when applicable).
+      /// \param[in] _ecm Entity component manager.
+      /// \return Velocity limits if available.
+      public: std::optional<std::vector<gz::math::Vector2d>>
```
arjo129: Why not justy return a `std::vector<double>` since the  max/min velocity is the same?

arjo129 (changes requested):
Thanks for iterating and re-opening a new PR. I think we should change this to just return a `std::vector<double>`. In the future if SDF supports asymetric max and min values we can add a function for retrieving min velocity.

include/gz/sim/Joint.hh
naitikpahwa18: Agreed since SDF currently exposes only maximum velocity, returning a std::vector<double> is a better fit. I’ll update the API and tests.

include/gz/sim/Joint.hh
```diff
@@ -275,6 +275,16 @@ namespace gz
...
 
+      /// \brief Get joint velocity limits.
+      /// \details For each axis, returns the maximum velocity limit. The size of
+      /// the returned vector corresponds to the joint's degrees of freedom. For
+      /// multi-axis joints, limits are returned for each axis in order (for
+      /// example, including both JointAxis and JointAxis2 when applicable).
+      /// \param[in] _ecm Entity component manager.
+      /// \return Velocity limits if available.
+      public: std::optional<std::vector<double>>
+      VelocityLimits(const EntityComponentManager &_ecm) const;
```
arjo129: Knit: Since we are returning Max velocity update the function name as well.

include/gz/sim/Joint.hh
```diff
@@ -275,6 +275,16 @@ namespace gz
       public: std::optional<Model> ParentModel(
           const EntityComponentManager &_ecm) const;
 
+      /// \brief Get joint velocity limits.
+      /// \details For each axis, returns the maximum velocity limit. The size of
```
arjo129:
```suggestion
      /// \details For each axis, returns the maximum absolute velocity limit. The size of
```

arjo129 (changes requested):
Thanks for the iteration, one last small knit. Lets put `Max` somewhere in the function name.

arjo129:
CI is complaining about line length, also would be worth updating the python bindings here: https://github.com/gazebosim/gz-sim/blob/gz-sim10/python/src/gz/sim/Joint.cc

If you cant update the python bindings open a ticket on the issue tracker so we remember to do it at some point.

naitikpahwa18:
I’ve applied the requested changes (API rename to MaxVelocityLimits, updated Python bindings and fix line length).

naitikpahwa18:
@arjo129 I’m seeing multiple CI failures related to Joint_TEST. Could you please advise if you’d like any changes from my side?

arjo129:
Can you fix the style complaints?
https://github.com/gazebosim/gz-sim/actions/runs/21163230348/job/60866147875#step:5:1

I'm guessing the issues arise from bazel as those are the runners that are failing to compile.

naitikpahwa18:
Fixed style issues and added missing JointAxis include in Joint_TEST. CI is re-running.

naitikpahwa18:
Hi @arjo129, just a quick reminder to check this out when you get a chance. Thanks!

arjo129 (approved):
Thanks for the iteration

naitikpahwa18:
Hi @arjo129, I’ve added effort and position limit getters following the `MaxVelocityLimits` pattern, with unit tests included. This should complete the current issue. Let me know if any changes are needed.

arjo129 (approved):
Thanks for the iteration on this PR.

naitikpahwa18:
Hi @arjo129, just a quick reminder to check this out when you get a chance. Thanks!

azeey:
@arjo129 can this be backported?

arjo129:

arjo129:
Sorry for missing this. Once my thesis defense is done I intend on generating a script for tracking which prs are pending backports.

PR #3288, opened 2026-01-26
Make Model::Name return std::optional and update call sites

### LINKED ISSUE #987: Model.hh Name() member function type not consinstent with other entities
Apparently, in the [model header file](https://github.com/ignitionrobotics/ign-gazebo/blob/0c401c17c1d4b748233c61921c47da5725ce9d5c/include/ignition/gazebo/Model.hh#L99) the member function `Name` seems to be returning a different type (std::string) than the one present eg in the [World one](https://github.com/ignitionrobotics/ign-gazebo/blob/0c401c17c1d4b748233c61921c47da5725ce9d5c/include/ignition/gazebo/World.hh#L100) 
(at least in the gazebo-5 version of the library). 

Apparently also `Link.hh` is consistent with  `World`. 
I think it's accidental, but maybe there is a precise choice behind this.

naitikpahwa18:
# 🦟 Bug fix

Fixes #987 

## Summary
This updates gz::sim::Model::Name to return std::optional<std::string>,
making it consistent with World and Link APIs.

All call sites have been updated to safely handle the optional return,
using entity ID fallbacks where a stable identifier is required.

## Local validation:
- Built and tested on Ubuntu 24.04 (Docker)
- All unit and GUI tests passed (GUI tests via xvfb)

## Checklist

### DIFF
--- a/include/gz/sim/Model.hh
+++ b/include/gz/sim/Model.hh
@@ -18,6 +18,7 @@
 #define GZ_SIM_MODEL_HH_
 
 #include <memory>
+#include <optional>
 #include <string>
 #include <vector>
 
@@ -54,7 +55,7 @@ namespace gz
     /// a simpler function call:
     ///
     ///    Model model(entity);
-    ///    std::string name = model.Name(ecm);
+    ///    std::string name = model.Name(ecm).value_or("");
     ///
     /// \todo(louise) Store the ecm instead of passing it at every API call.
     class GZ_SIM_VISIBLE Model {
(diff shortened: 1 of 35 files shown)


arjo129:
@naitikpahwa18 Thank you for the significant effort here. I see you've updated all the call sites and the implementation is consistent.

However, after reviewing the underlying architecture, I’m going to have to close this PR. I apologize for the churn, but I realized that making this optional actually conflicts with our SDF guarantees.

Since libsdformat strictly enforces that every <model> must have a name (Gazebo will crash/fail to load otherwise), our C++ API should reflect that guarantee. Returning `std::optional` implies to downstream users that a model might be nameless, forcing them to write error-handling code for a case that is theoretically impossible in a valid simulation.

For future reference: Before starting a large refactor or API change like this, please open an issue (or start a discussion) first. We'd love to discuss the design beforehand to ensure you don't end up investing this much time on a direction that might conflict with system invariants.

Sorry again for the wasted time on this specific patch. We do appreciate the contribution and hope to see you on other issues.

Also if you're wondering about the inconsistency between `std::optional` and no `std::optional` in the APIs, this is a result of the fact that std::optional was added in C++17, Gazebo used to use older C++ variants back in the day. APIs with `optional` were added after our move to C++17.

naitikpahwa18:
Hi @arjo129, thanks a lot for the detailed explanation and review. I understand now how returning std::optional would conflict with the guarantees enforced by libsdformat. I appreciate the background on how the API evolved, and the guidance on discussing larger API changes upfront. Thanks again for the feedback.

Issue #3285, opened 2026-01-23
Server does not exit when ServerConfig::SdfErrorBehavior::EXIT_IMMEDIATELY is set for bad SDF file/ string


shameekganguly:
## Environment
* OS Version: Any (tested on MacOS 15.7.3)
* Source or binary build? Any (tested with binary build from homebrew)

When the `ServerConfig::SdfErrorBehavior::EXIT_IMMEDIATELY` setting is applied (set by default), the behavior on Ionic is that the server exits immediately if a bad sdf file is supplied.

However, on Jetty, the server simply runs with a default empty world and does not exit.

e.g. file
```
<?xml version="1.0"?>
<sdf version="1.9">
  <world name="test">
    <model name="bad_model_no_link" />
  </world>
</sdf>
```

Run with

```
gz sim -v 4 -s -r bad_model.sdf
```

## Description
* Expected behavior: Server should print errors and exit
* Actual behavior: Server prints errors, then loads a default world and continues to run

## Steps to reproduce
1. Save the above sdf file snippet to bad_model.sdf
2. Run `gz sim -v 4 -s -r bad_model.sdf`

## Output
```
$ gz sim -v 4 -s -r bad_model.sdf 
(2026-01-23 15:44:59.846) [info] [gz.cc:407] Gazebo Sim Server v10.0.0
(2026-01-23 15:44:59.872) [info] [ServerPrivate.cc:697] Loading SDF world file[/tmp/bad_model.sdf].
...
(2026-01-23 15:45:01.897) [error] [ServerPrivate.cc:794] Error Code 17: Msg: A model must have at least one link.
(2026-01-23 15:45:01.897) [error] [ServerPrivate.cc:794] Error Code 17: Msg: A model must have at least one link.
(2026-01-23 15:45:01.897) [error] [ServerPrivate.cc:794] Error Code 9: Msg: Failed to load a world.
(2026-01-23 15:45:01.897) [error] [ServerPrivate.cc:804] Unable to find world with name[test]. Downloaded models may not appear.
(2026-01-23 15:45:01.898) [info] [LevelManager.cc:646] Loaded level [default]
(2026-01-23 15:45:01.898) [debug] [ServerConfig.cc:1042] Loading (3) plugins from file [/Users/shameek/.gz/sim/10/server.config]
(2026-01-23 15:45:01.899) [info] [SimulationRunner.cc:1697] No systems loaded from SDF, loading defaults
...
```

[...]

PR #3290, opened 2026-01-27
Fix regression in Server behavior when it is run with a bad sdf file

shameekganguly:
# 🎉 New feature

Fixes [#3285](https://github.com/gazebosim/gz-sim/issues/3285)

## Summary
Adds a `GetStatus` call to `Server` class, which returns one of `EXITED` or `STOPPED` or `RUNNING`:
- `EXITED`: Server was configured with `SdfErrorBehavior::EXIT_IMMEDIATELY` and encountered an error when loading the SDF file/ string.
- `RUNNING`: Server is currently running (i.e. `Server::IsRunning()` would return true)
- `STOPPED`: Server is currently not running

If server is in `EXITED` state, calls to `Run` and `RunOnce` will return false immediately. Also, `world_control` service calls will return false and have no effect on the server.

I fixed a regression introduced with the asset download optimization that loading an SDF file with errors would cause the server to start up and run even if `SdfErrorBehavior::EXIT_IMMEDIATELY` was configured (default value): https://github.com/gazebosim/gz-sim/issues/3285.

Added tests.

## Test it
Run `gz sim -s` with the following sdf file which contains errors:
```
<?xml version="1.0"?>
<sdf version="1.9">
  <world name="test">
    <model name="bad_model_no_link" />
  </world>
</sdf>
```

The server should print the sdf errors and exit. Repeat with the `--wait-for-assets` flag.

## Checklist
- [x] Added tests

Generated-by: Gemini 3 Pro (new tests)

### DIFF
--- a/test/benchmark/CMakeLists.txt
+++ b/test/benchmark/CMakeLists.txt
@@ -1,4 +1,5 @@
 set(TEST_TYPE "BENCHMARK")
+find_package(benchmark)
 include(GzBenchmark OPTIONAL RESULT_VARIABLE GzBenchmark_FOUND)
 
 if (GzBenchmark_FOUND)
@@ -8,4 +9,9 @@ if (GzBenchmark_FOUND)
   )
 
   gz_add_benchmarks(SOURCES ${tests})
+
+  if (benchmark_VERSION VERSION_LESS "1.9.5")
+    add_compile_definitions(BENCHMARK_INTERNAL)
+  endif()
+
 endif()
--- a/test/benchmark/each.cc
+++ b/test/benchmark/each.cc
@@ -30,6 +30,11 @@
 #include "gz/sim/components/Pose.hh"
 #include "gz/sim/components/World.hh"
 
+#ifdef BENCHMARK_INTERNAL
+using Benchmark = ::benchmark::internal::Benchmark;
+#else
+using Benchmark = ::benchmark::Benchmark;
+#endif
 
 using namespace gz;
 using namespace sim;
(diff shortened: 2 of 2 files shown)


src/Server_TEST.cc
```diff
@@ -1286,6 +1287,102 @@ TEST_P(ServerFixture, Stop)
...
+  serverConfig.SetSdfFile(common::joinPaths(PROJECT_SOURCE_PATH,
+      "test", "worlds", "shapes.sdf"));
+
+  sim::Server server(serverConfig);
+
+  // Initial state should be STOPPED
+  EXPECT_EQ(Server::Status::STOPPED, server.GetStatus());
+
+  // Run non-blocking
+  server.Run(false, 0, false);
```
arjo129, line 1302: Im always very wary of putting sleep in tests as (1) it unnecessarily increases CI time and (2) it adds potential for flakiness. Is there a good reason for us to run it non-blocking?

iche033 (commented):
apart from Arjo's comments, the code changes look good to me

src/Server_TEST.cc
shameekganguly: Yeah, good point. We need the server to run non-blocking so that we can check the Server status, which should be `RUNNING`. So I updated the test to use an infinite wait until the server has stepped at least once. This way, the test is less sensitive to test server loads.

I also removed the second `GZ_SLEEP_MS` call, which isn't needed - once `Server::Stop` returns, the status should be `STOPPED`.

shameekganguly:
> apart from Arjo's comments, the code changes look good to me

Sounds good, I addressed Arjo's concern about using sleep in the test in https://github.com/gazebosim/gz-sim/pull/3290/commits/bed2d5d7c8aa097c18707ad9996aba0a4191a02d

arjo129 (approved):
Thanks for the fix!

shameekganguly:
@mergifyio backport main

PR #3309, opened 2026-02-04
Add Inspect action to entity context menu

### LINKED ISSUE #177: Show selected entity when opening Component Inspector
The GUI Component Inspector plugin always starts with the World properties shown. If I have an entity selected, then I would like the Component Inspector to open with the properties of the selected entity displayed.

naitikpahwa18:
# 🎉 New feature

Closes #177 

## Summary
Adds an **Inspect** action to the entity context menu. Selecting it opens the
Component Inspector for the clicked entity, or updates the already-open
Component Inspector to show the selected entity.

## Output

https://github.com/user-attachments/assets/617e4ce6-82e9-4a54-b778-fbf2b6d8990d

## Checklist
- [ ] Added tests


src/gui/plugins/entity_context_menu/EntityContextMenuPlugin.cc
```diff
@@ -197,7 +203,20 @@ void EntityContextMenuHandler::HandleMouseContextMenu(
...
+      entityId = std::get<uint64_t>(visual->UserData("gazebo-entity"));
+    }
+    catch(const std::bad_variant_access &)
+    {
+      // It's ok to get here
+    }
+
+    emit ContextMenuRequested(
+      visual->Name().c_str(),
+      QString::number(static_cast<uint64_t>(entityId)),
```
arjo129, line 214: Why are we converting to a string here can't we directly write an `std::optional<int>` ?

src/gui/plugins/modules/EntityContextMenu.cc
```diff
@@ -345,3 +351,77 @@ void EntityContextMenu::OnRequest(const QString &_request, const QString &_data)
...
+  for (const auto plugin : plugins)
+  {
+    if (plugin == nullptr)
+      continue;
+
+    const char *classNameC = plugin->metaObject()->className();
+    if (classNameC == nullptr)
+      continue;
+
+    const std::string className(classNameC);
```
arjo129, line 386: I understand we might be doing this to maintain backward compatibility. Why aren't we specifying the exact string?

arjo129 (changes requested):

src/gui/plugins/entity_context_menu/EntityContextMenuPlugin.cc
naitikpahwa18: `entityId` is consumed by QML, where QString is the existing API type. Switching to `std::optional<int>` would break the C++ to QML interface, since QML doesn’t natively support `std::optional`, so the ID is passed as a string for compatibility.

src/gui/plugins/modules/EntityContextMenu.cc
naitikpahwa18: `metaObject()->className()` returns the Qt-reported class name, which can include namespaces and may change across refactors or builds. Converting it to `std::string` and matching by suffix avoids relying on a single hard-coded fully qualified name while still reliably identifying the ComponentInspector plugin.

src/gui/plugins/entity_context_menu/EntityContextMenuPlugin.cc
arjo129: Thanks for clarifying.

src/gui/plugins/modules/EntityContextMenu.cc
arjo129: I understand where youre coming from. I dont think we will be changing names anytime soon. The name change was done due to legal reasons once. I think there may be some affect of the gazebo version on the namespaces name. In all honesty, Id be inclined to not backport this if it helps avoid confusion and simplify code.

naitikpahwa18: I double-checked across `main`, `gz-sim7` through `gz-sim10`:
`ComponentInspector` is declared inside namespace `gz::sim` in all branches and registered via `GZ_ADD_PLUGIN(ComponentInspector, gz::gui::Plugin)`.
Since `metaObject()->className()` may include namespaces (e.g. `gz::sim::ComponentInspector)`, matching only the exact string would miss valid cases.
Keeping the `::ComponentInspector` suffix check ensures correctness across builds and branches and doesn’t impact backporting.

src/gui/plugins/modules/EntityContextMenu.cc
```diff
@@ -196,7 +202,7 @@ void EntityContextMenu::OnRemove(
 {
   if (this->dataPtr->worldName.empty())
   {
-    auto runners = gui::App()->findChildren<GuiRunner *>();
+    auto runners = gz::gui::App()->findChildren<GuiRunner *>();
```
arjo129, line 205: Knit: unessecary change

arjo129 (approved):

src/gui/plugins/modules/EntityContextMenu.cc
naitikpahwa18: Updated

naitikpahwa18: I reverted it to gui::App(), but CI failed due to namespace resolution (`no member named 'App' in namespace 'gz::sim::gui';`).

naitikpahwa18: `gui::App()` breaks because `#include <gz/sim/gui/GuiEvents.hh>` makes the compiler find `gz::sim::gui` first when searching from the enclosing `gz::sim` namespace. `gz::gui::App()` uses a full path so it's unambiguous regardless of what's included, keeping `gz::gui::App()` explicit is safer as the codebase evolves.

arjo129 (approved):

azeey:
@naitikpahwa18 can you fix the build failures on CI?

naitikpahwa18:
I've pushed a fix addressing the CI build failure

src/gui/plugins/entity_context_menu/EntityContextMenuPlugin.cc
```diff
@@ -197,7 +203,20 @@ void EntityContextMenuHandler::HandleMouseContextMenu(
...
 
-    emit ContextMenuRequested(visual->Name().c_str());
+    gz::sim::Entity entityId = gz::sim::kNullEntity;
+    try
+    {
+      entityId = std::get<uint64_t>(visual->UserData("gazebo-entity"));
+    }
+    catch(const std::bad_variant_access &)
+    {
+      // It's ok to get here
```
ntfshard: It's not a good practice to use exceptions for execution control flow([*](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#e3-use-exceptions-for-error-handling-only)).
Seems UserData returning std::variant, so we can use
https://en.cppreference.com/cpp/utility/variant/get_if  or   https://en.cppreference.com/cpp/utility/variant/holds_alternative

src/gui/plugins/modules/EntityContextMenu.cc
```diff
@@ -345,3 +351,77 @@ void EntityContextMenu::OnRequest(const QString &_request, const QString &_data)
...
+  if (mainWindow == nullptr)
+  {
+    gzerr << "Internal error: no MainWindow found." << std::endl;
+    return;
+  }
+
+  bool inspectorOpen{false};
+  bool inspectorJustLoaded{false};
+  auto plugins = mainWindow->findChildren<gz::gui::Plugin *>();
+  for (const auto plugin : plugins)
```
ntfshard: if it's container of raw pointers, seems better to constrain auto to pointer type

```suggestion
  for (const auto * plugin : plugins)
```

src/gui/plugins/modules/EntityContextMenu.cc
```diff
@@ -345,3 +351,77 @@ void EntityContextMenu::OnRequest(const QString &_request, const QString &_data)
...
+    if (classNameC == nullptr)
+      continue;
+
+    const std::string className(classNameC);
+    const std::string suffix("::ComponentInspector");
+    const bool matchesExact = (className == "ComponentInspector");
+    const bool matchesSuffix = (className.size() >= suffix.size() &&
+      className.compare(className.size() - suffix.size(),
+        suffix.size(), suffix) == 0);
+    if (matchesExact || matchesSuffix)
```
ntfshard: imo, it's unnecessary complication (premature optimization I suppose).
It's harder to read, change and verify I believe.

This strings are not very long, regular [string::find](https://en.cppreference.com/cpp/string/basic_string/find) should have almost same execution time I suppose. Also you constructing copy of `classNameC` for each plugin. I hope compiler could optimize constructing `suffix` string.

I'd suggest to write simple code here. This code do not called billion times in a row

src/gui/plugins/modules/EntityContextMenu.cc
```diff
@@ -345,3 +351,77 @@ void EntityContextMenu::OnRequest(const QString &_request, const QString &_data)
     gzwarn << "Unknown request [" << request << "]" << std::endl;
   }
 }
+
+/////////////////////////////////////////////////
+void EntityContextMenu::OnInspect(const QString &_entityId)
+{
+  bool ok{false};
+  auto id = _entityId.toULongLong(&ok);
```
ntfshard:
```suggestion
  const auto id = _entityId.toULongLong(&ok);
```

src/gui/plugins/modules/EntityContextMenu.cc
```diff
@@ -345,3 +351,77 @@ void EntityContextMenu::OnRequest(const QString &_request, const QString &_data)
...
+      return;
+    }
+    else
+    {
+      inspectorJustLoaded = true;
+    }
+  }
+
+  std::vector<Entity> entitySet{static_cast<Entity>(id)};
+  auto sendSelection = [mainWindow, entitySet]()
```
ntfshard: Seems right now you creating vector and copying it inside lambda to match lifetime.
You can put construction of vector inside lambda, and pass `id` by copy (which is POD), not std::vector.

ntfshard (commented):
some suggestions

naitikpahwa18:
Hi @arjo129 , I’ve addressed the feedback. Could you review this again when you get a chance? Thanks!

PR #3334, opened 2026-02-13
Auto-load missing SceneBroadcaster and seed scene graph for late-load…

### LINKED ISSUE #2962: Consider spawning a scene broadcaster automatically once the gui is launched
## Description:
When new users add custom plugins or modify the default configuration in Gazebo, they may unintentionally end up in a state where the `SceneBroadcaster` is not running. As a result, the minimalscene view appears completely empty. This leads to a poor first-time experience, with many users concluding that "Gazebo is not working" when in fact the issue is a missing system plugin.

## Steps to Reproduce:

Create a custom Gazebo world with modified system plugins.
Launch the world without including `SceneBroadcaster`.
Open minimalscene or a remote viewer.

## Expected Behavior:
A warning or fallback should indicate that `SceneBroadcaster` is missing, or it should be enabled by default to avoid confusion.

## Actual Behavior:
The viewer opens with an empty scene, with no feedback explaining the issue.

## Impact:
High for new users — this can lead to significant confusion and the false impression that Gazebo is broken or improperly installed.

## Suggested Fix:
Consider:
- Including `SceneBroadcaster` by default in example and minimal configurations.
- Emitting a clear warning if the scene cannot be visualized due to missing broadcasters. 
- Possibly even automatically spawn the scene broadcaster plugin via the add plugin service from the Minimal Scene GUI client.

naitikpahwa18:
# 🦟 Bug fix

Fixes #2962 

## Summary
This PR improves robustness when a world is loaded without the `SceneBroadcaster` system by automatically injecting it from the GUI if it is missing and seeding the scene graph when it is loaded after world creation, ensuring that existing entities are immediately visualized and preventing blank or non-responsive GUI behavior.

## Checklist
- [ ] Added tests

### DIFF
--- a/src/gui/Gui.cc
+++ b/src/gui/Gui.cc
@@ -25,6 +25,8 @@
 
 #include <QScreen>
 
+#include <memory>
+#include <mutex>
 #include <unordered_set>
 
 #include <gz/common/Console.hh>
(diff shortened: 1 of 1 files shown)


naitikpahwa18:
Hi @arjo129, sharing a quick summary of my changes:

* Added late-load support to SceneBroadcaster by reconstructing the scene graph from existing entities.
* Made scene graph access thread-safe.
* Ensured scene components (ambient/background) are preserved.
* Extended seeding to sensors, particle emitters, and projectors.
* Improved GUI-side injection logic and added an integration test for runtime addition.

It works, but adds ~500 LOC, would appreciate feedback on whether this is the right architectural direction or if there’s a cleaner approach.

arjo129:
@naitikpahwa18 this is an awesome attempt. My advice would be to split the improvements into two PRs. The first would be the `SceneBroadcaster` changes that targets main. The second PR would include the rest of the gui.

arjo129:
@naitikpahwa18 sorry I took so long to re-open the pull request. Was swamped with other work.

naitikpahwa18:
> @naitikpahwa18 this is an awesome attempt. My advice would be to split the improvements into two PRs. The first would be the `SceneBroadcaster` changes that targets main. The second PR would include the rest of the gui.

I've split the work into two PRs, #3361 for the Scene Broadcaster changes targeting main and #3334 for the GUI side targeting gz-sim8

src/gui/Gui.cc
```diff
@@ -243,6 +249,73 @@ std::unique_ptr<gz::gui::Application> createGui(
...
+  auto ensureSceneBroadcaster = [&](const std::string &worldName)
+  {
+    const unsigned int timeout = 5000;
+    bool sceneBroadcasterFound = false;
+    bool sceneResult = false;
+    msgs::Empty sceneReq;
+    msgs::Scene sceneRes;
+    auto sceneService = transport::TopicUtils::AsValidTopic(
+      "/world/" + worldName + "/scene/info");
+    for (int attempt = 0; attempt < 2 && !sceneBroadcasterFound; ++attempt)
```
XINJIANGMO: @naitikpahwa18  the behavior indeed changes , but I think we should not use this way to check if SceneBroadcaster exists , it will block 10s if not exists. It affects performance a lot.

XINJIANGMO: we should use a new thread to check the topic if exsits. 
For me , in a new thread , we should use a polling of 100ms × 10. If the service exists, it might be detected and immediately exit in the first or second loop (within 0.1~0.2 seconds); if it doesn't exist, it will at most wait silently in the background for 1 second before initiating the injection request.

naitikpahwa18: Moved the check to a background thread with async injection, so it no longer blocks the GUI.

XINJIANGMO (approved):
Works for me!

PR #3336, opened 2026-02-17
Add entity field to RaycastResultInfo for identifying hit collision s…

### LINKED ISSUE #3314: Add hit entity information to RaycastResultInfo
## Desired behavior

RaycastResultInfo should include the entity that was hit by the ray. Currently it only returns point, normal, and fraction.
In Gazebo Classic, Entity::GetNearestEntityBelow() used RayShape::GetIntersection() which returned both the hit distance and the entity name via the physics engine (ODE/Bullet). There is no equivalent in gz-sim.

## Alternatives considered

Contact sensors: Require physical contact, add overhead, and are unreliable (intermittent contact, lag behind actual position).
RaycastData + ECM iteration: Cast a ray, get the hit point, then iterate all collision entities to guess which one was hit. Unreliable and hacky.
Zone-based lookup: Define terrain regions manually in YAML and match by XY position. Works but bypasses the physics engine entirely and doesn't scale.
None of these replace the simple, reliable behavior of Gazebo Classic's GetIntersection().

## Implementation suggestion

Add an entity field to RaycastResultInfo:
```cpp
struct RaycastResultInfo
{
  gz::math::Vector3d point;
  gz::math::Vector3d normal;
  double fraction;
  gz::sim::Entity entity;  // NEW: the entity that was hit
};
```
The DART physics plugin (gz-physics-dart) already has access to dart::collision::CollisionObject, as seen in GzCollisionDetector where contact.collisionObject1 / contact.collisionObject2 are used for contact processing. The same mapping from CollisionObject to  gz::sim::Entity can be applied to raycast results via FCL's ray query output.

## Additional context

Use case: Terramechanics simulation, identifying terrain type under each rover wheel to apply different soil parameters (Bekker-Wong model). Each terrain patch has different soil properties (clay, sand, lunar soil, etc.), and the plugin needs to know which terrain is below each wheel at every time step.
This is a common need for any ground vehicle simulation that interacts with heterogeneous terrain.

khalidbourr:
# 🎉 New feature

Closes #3314 

## Summary

Add `entity` field to `RaycastResultInfo` to identify which collision shape was hit by a raycast. The physics shape ID from gz-physics is resolved to a gz-sim Entity via `entityCollisionMap.GetByPhysicsId()`.

Depends on: https://github.com/gazebosim/gz-physics/pull/879

## Changes

- `include/gz/sim/components/RaycastData.hh` I added `Entity entity{kNullEntity}` to `RaycastResultInfo`
- `src/systems/physics/Physics.cc` I resolved `collisionShapeId` to gz-sim Entity in `UpdateRayIntersections()`

## Test it

1. Create a world with Bullet collision detector and multiple named collision shapes
2. Attach a `RaycastData` component to an entity
3. Cast a ray that hits a collision shape
4. Check `result.entity` it should return the hit collision Entity (not `kNullEntity`)

## Checklist
- [x] Added tests

## Usage example

Here's how I use it to identify entity collision  type under each wheel:
```cpp
auto rayComp = _ecm.Component<gz::sim::components::RaycastData>(
    wheels[wheel_idx].rayEntity);
if (rayComp && !rayComp->Data().results.empty())
{
  const auto &result = rayComp->Data().results[0];
  if (result.entity != gz::sim::kNullEntity)
  {
    auto *name = _ecm.Component<gz::sim::components::Name>(result.entity);
    if (name)
      return name->Data(); 
  }
}
```

image


arjo129:
Thanks for this PR. Can you fix the DCO issue.

Instructions are here: https://github.com/gazebosim/gz-sim/pull/3336/checks?check_run_id=63886608246

Also please try to target main for new features. This is indeed a useful feature.

If you could work one or two tests in it'd be great.

khalidbourr:
> Thanks for this PR. Can you fix the DCO issue.
>  
> Instructions are here: https://github.com/gazebosim/gz-sim/pull/3336/checks?check_run_id=63886608246

I think for this is done 


> Also please try to target main for new features. This is indeed a useful feature.
> If you could work one or two tests in it'd be great.

I will try to do that asap

arjo129:
Yep, the DCO is correctly done. But, I think you should target main both in gz-physics and here.

khalidbourr:
> Yep, the DCO is correctly done. But, I think you should target main both in gz-physics and here.

I targeted the main in both. The build here won't proceed if the gz-physics PR is not approved

arjo129:
That's what the "needs upstream release" tag is for.

Writing a test would not be a bad idea for this in the mean time.

khalidbourr:
> That's what the "needs upstream release" tag is for.

Ah yep, I didn't notice that.
> Writing a test would not be a bad idea for this in the mean time.

Actually in the last push I've added simple tests to the already existing RayIntersections test, checking that hit rays return a valid entity and missed rays return kNullEntity. I’m available to add more test coverage if needed.

test/integration/physics_system.cc
```diff
@@ -3136,6 +3137,7 @@ TEST_F(PhysicsSystemFixture, GZ_UTILS_TEST_DISABLED_ON_WIN32(RayIntersections))
             math::eigen3::convert(results2[i].normal).array().isNaN().all());
           ASSERT_TRUE(
             std::isnan(results2[i].fraction));
```
arjo129: Can we have some test with real intersections with things other than kNullEntity?

arjo129 (changes requested):

test/integration/physics_system.cc
khalidbourr: sure !

khalidbourr: I’ve updated the test so it now checks that hit rays return the actual ground plane collision entity, instead of just confirming the result isn’t null.

test/integration/physics_system.cc
```diff
@@ -3113,13 +3113,23 @@ TEST_F(PhysicsSystemFixture, GZ_UTILS_TEST_DISABLED_ON_WIN32(RayIntersections))
...
 
+        Entity groundCollision{kNullEntity};
+        _ecm.Each<components::Collision, components::ParentEntity>(
+            [&](const Entity &_entity, const components::Collision *,
+                const components::ParentEntity *) -> bool
+            {
+              groundCollision = _entity;
+              return true;
+            });
+        ASSERT_NE(groundCollision, kNullEntity);
```
azeey: Could you use ECM::EntityByComponents instead?

src/systems/physics/Physics.cc
```diff
@@ -4375,6 +4375,15 @@ void PhysicsPrivate::UpdateRayIntersections(EntityComponentManager &_ecm)
           const math::Vector3d normal =
             math::eigen3::convert(rayIntersectionResult.normal);
           result.normal = entityWorldPose.Rot().RotateVectorReverse(normal);
+
+          // Map physics shape ID to gz-sim Entity
+          auto *extraData = rayIntersectionResult.template
```
azeey: `rayIntersection` is the `Composite` type
```suggestion
          auto *extraData = rayIntersection.template
```

include/gz/sim/components/RaycastData.hh
```diff
@@ -57,6 +58,9 @@ struct RaycastResultInfo
 
   /// \brief The normal at the hit point in entity frame
   gz::math::Vector3d normal;
+
+  /// \brief The entity at the hit point
+  Entity entity{kNullEntity};
```
azeey, line 63: This breaks ABI, so unfortnately, this means this cannot be backported.

src/systems/physics/Physics.cc
```diff
@@ -4375,6 +4375,15 @@ void PhysicsPrivate::UpdateRayIntersections(EntityComponentManager &_ecm)
...
+              Query<ExtraRayIntersectionData>();
```
azeey: You'll need to define `ExtraRayIntersectionData` somewhere.

```c++
  using ExtraRayIntersectionData =
      physics::GetRayIntersectionFromLastStepFeature::ExtraRayIntersectionDataT<Policy>;
```

azeey (commented):
Thanks for the contribution!! There's a compile error that should be fixed after applying my suggestion. Otherwise, this looks great!

The PR changes the `RaycastResultInfo`, which breaks ABI, so we will not be backporting this to older versions.

src/systems/physics/Physics.cc
azeey: https://github.com/gazebosim/gz-sim/pull/3336/commits/47da1a83b71c2d79dff86cedc6ad8527e65cd1e7

src/systems/physics/Physics.cc
azeey: https://github.com/gazebosim/gz-sim/pull/3336/commits/47da1a83b71c2d79dff86cedc6ad8527e65cd1e7

test/integration/physics_system.cc
khalidbourr: Thank you for the suggestion, 
I remove all that part and substituted with : 

auto groundCollision = _ecm.EntityByComponents(
            components::Collision());

include/gz/sim/components/RaycastData.hh
khalidbourr: Understood, thanks for the note.

src/systems/physics/Physics.cc
khalidbourr: I appreciate the fixes !

azeey:
This is waiting for the [Rotary](https://github.com/gazebo-tooling/release-tools/issues/1446) release to be completed.

khalidbourr:
> This is waiting for the [Rotary](https://github.com/gazebo-tooling/release-tools/issues/1446) release to be completed.

Got it, thanks for the update. Let me know if anything is needed from my side in the meantime.

khalidbourr:
Seems docker hub reached its rate limit, not sure if i can re-trigger the CI or someone else can do it!

khalidbourr:
Test fails: hit rays return entity=0 instead of 7. Not sure if extraData is null or the ID lookup is wrong. Could someone please take a look?

PR #3343, opened 2026-02-19
Add CpuLidar system plugin with optimized physics raycasting

apojomovsky:
# 🎉 New feature

This PR is 3 out of 3 that implement: https://github.com/gazebosim/gz-sensors/issues/26

Depends on:
- [x] https://github.com/gazebosim/gz-sensors/pull/593
- [x] https://github.com/gazebosim/gz-physics/pull/880

## Summary

Add the `CpuLidar` system plugin that bridges `CpuLidarSensor` (gz-sensors) with the physics raycasting pipeline. The end result is a close match in performance to that found on Gazebo Classic, while tested with a 128-beam × 80-ring lidar scene).

### CpuLidar system plugin

A new system plugin following the `Imu` system pattern (`ISystemConfigure` + `ISystemPreUpdate` + `ISystemPostUpdate`):

- **Configure**: stores the `SensorFactory` reference.
- **PreUpdate**: discovers new `<sensor type="lidar">` entities via   `EachNew<CpuLidar>`, creates `CpuLidarSensor` instances, calls `GenerateRays()` to populate the `RaycastData` ECM component, and sets the `needsRaycast` flag when a scan is due.
- **PostUpdate**: reads `RaycastData` results (populated by the Physics system), feeds them to `CpuLidarSensor::SetRaycastResults()`, and calls `sensor->Update()` to trigger publishing.

### Performance optimization 1: Batch raycasting in Physics.cc

`UpdateRayIntersections()` now tries the batch API first. When available (dartsim + Bullet), all rays are submitted in a single `GetBatchRayIntersectionFromLastStep()` call. The single-ray loop is retained as a fallback.

### Performance optimization 2: Frequency-gated raycasting

`UpdateRayIntersections()` was called every physics step (1 kHz default), but the sensor only needs results at its own rate (e.g. 10 Hz). A `needsRaycast` flag in `RaycastDataInfo` is set by `CpuLidar::PreUpdate` only when a scan is due and
there are active subscribers. The Physics system skips entities where the flag is `false`. Both the batch and fallback paths are gated.

| Stage | RTF |
|---|---|
| Baseline (DART `raycast()` at physics rate) | ~7% |
| + Batch API with `btWorld::rayTest()` | ~13% |
| + `needsRaycast` frequency gating | **~95%** |

## Test it

```bash
colcon build --merge-install --packages-up-to gz-sim \
  --cmake-args '-DBUILD_TESTING=ON'
colcon test --packages-select gz-sim \
  --ctest-args -R INTEGRATION_cpu_lidar
```

Manual test:

```bash
gz sim examples/worlds/cpu_lidar_sensor.sdf

# In another terminal:
gz topic -e -t /lidar
gz topic -e -t /lidar/points
```

## Checklist
- [x] Added tests
- [x] Updated Bazel files (if adding new files). Created an issue otherwise.

Generated-by: Claude Opus 4.6

Disclaimer: The code here was reviewed, tested, and profiled by hand by the author.

https://github.com/user-attachments/assets/c1c9e9d7-50dd-45d9-9c65-4fe0b977e951


src/SdfEntityCreator.cc
```diff
@@ -1125,11 +1126,11 @@ Entity SdfEntityCreator::CreateEntities(const sdf::Sensor *_sensor)
...
-    // \todo(anyone) Implement CPU-based lidar
-    // this->dataPtr->ecm->CreateComponent(sensorEntity,
-    //     components::Lidar(*_sensor));
-    gzwarn << "Sensor type LIDAR not supported yet. Try using"
-      << "a GPU LIDAR instead." << std::endl;
+    this->dataPtr->ecm->CreateComponent(sensorEntity,
+        components::CpuLidar(*_sensor));
+
+    this->dataPtr->ecm->CreateComponent(sensorEntity,
+        components::WorldPose(math::Pose3d::Zero));
```
Zyrin, line 1143: Why is creating this Zero world pose necessary?

apojomovsky: The "neighboring" sensors above and below this definition (Camera, GPU_LIDAR, DepthCamera, etc.) are rendering-based, so they get their world pose from the rendering scene graph for free.
Our CpuLidar doesn't use the rendering pipeline (see other examples: IMU, Magnetometer, and AirPressure), so it reads WorldPose directly from the ECM and needs the component created here for the Physics system to populate.

test/integration/cpu_lidar_system.cc
```diff
@@ -0,0 +1,359 @@
...
+  EXPECT_FALSE(*server.Running(0));
+
+  bool resultsFound = false;
+
+  // Subscribe to the sensor topic so HasConnections() returns true
+  // and the system requests raycasts from Physics.
+  transport::Node node;
+  std::function<void(const msgs::LaserScan &)> cb =
+    [](const msgs::LaserScan &){};
+  node.Subscribe("/test/cpu_lidar", cb);
```
Zyrin:
```suggestion
  node.Subscribe("/test/cpu_lidar", [](const msgs::LaserScan &){});
```

azeey:
As mentioned in https://github.com/gazebosim/gz-sensors/issues/26, here's the manually triggered CI job that builds all of your branches on gz-sensors, gz-physics and gz-sim.

[Build Status](https://build.osrfoundation.org/job/_azeey_ci__colcon_gpu_any-manual_ubuntu_noble_amd64/11/)

test/integration/cpu_lidar_system.cc
Zyrin, line 200:
```suggestion
  node.Subscribe("/test/cpu_lidar", [](const msgs::LaserScan &){});
```
To make it consistent with e.g. line 205.
And there are some more lambdas further down where this can be applied, too.

apojomovsky: Addressed in https://github.com/gazebosim/gz-sim/pull/3343/commits/895ca91ce97675d86074fa4d3e87f49a6cc1325f

include/gz/sim/components/RaycastData.hh
```diff
@@ -67,6 +67,9 @@ struct RaycastDataInfo
 
   /// @brief The results of the raycasting.
   std::vector<RaycastResultInfo> results;
+
+  /// @brief Whether the sensor needs fresh raycast results this step.
+  bool needsRaycast = true;
```
azeey: This breaks ABI. Is there a way we can avoid adding a field here so we can backport this PR to older versions?

src/systems/cpu_lidar/CpuLidar.hh
```diff
@@ -0,0 +1,64 @@
...
+    /// Documentation inherited
+    public: void PreUpdate(const UpdateInfo &_info,
+                           EntityComponentManager &_ecm) final;
+
+    /// Documentation inherited
+    public: void PostUpdate(const UpdateInfo &_info,
+                            const EntityComponentManager &_ecm) final;
+
+    /// \brief Private data pointer.
+    private: std::unique_ptr<CpuLidarPrivate> dataPtr;
```
azeey: This is not necessary since systems (plugins) don't need to maintain ABI stability.

src/systems/physics/Physics.cc
```diff
@@ -4307,6 +4314,75 @@ void PhysicsPrivate::UpdateRayIntersections(EntityComponentManager &_ecm)
...
+    using BatchedRayIntersectionData = BatchWorld::BatchedRayIntersectionData;
+
+    _ecm.Each<components::RaycastData,
+              components::Pose>(
+        [&](const Entity &_entity,
+            components::RaycastData *_raycastData,
+            components::Pose */*_pose*/) -> bool
+        {
+          if (!_raycastData->Data().needsRaycast)
+            return true;
```
azeey: The `needsRaycast` flag can be a separate private component defined in Physics.cc as opposed to adding it to the RayCastData component.

src/systems/physics/Physics.cc
```diff
@@ -4307,6 +4314,75 @@ void PhysicsPrivate::UpdateRayIntersections(EntityComponentManager &_ecm)
...
+
+          BatchedRayIntersectionData batchOutput;
+          worldBatchRayFeature->GetBatchRayIntersectionFromLastStep(
+            batchInput, batchOutput);
+
+          for (const auto &hit :
+              batchOutput.Get<std::vector<RayIntersection>>())
+          {
+            results.emplace_back();
+            auto &result = results.back();
```
azeey: nit:
```suggestion
            auto &result = results.emplace_back();
```

src/systems/physics/Physics.cc
```diff
@@ -4307,6 +4314,75 @@ void PhysicsPrivate::UpdateRayIntersections(EntityComponentManager &_ecm)
...
+            q.origin = math::eigen3::convert(
+              entityWorldPose.Pos() +
+              entityWorldPose.Rot().RotateVector(ray.start));
+            q.target = math::eigen3::convert(
+              entityWorldPose.Pos() +
+              entityWorldPose.Rot().RotateVector(ray.end));
+            batchInput.push_back(q);
+          }
+
+          BatchedRayIntersectionData batchOutput;
```
azeey: I wonder if `BatchedRayIntersectionData` can be stored in a private component attached to the sensor entity. This would avoid memory allocations in every time step.

src/systems/physics/Physics.cc
```diff
@@ -4307,6 +4314,75 @@ void PhysicsPrivate::UpdateRayIntersections(EntityComponentManager &_ecm)
...
+          _raycastData->Data().needsRaycast = false;
+
+          const auto &rays = _raycastData->Data().rays;
+          auto &results = _raycastData->Data().results;
+          results.clear();
+          results.reserve(rays.size());
+
+          const auto &entityWorldPose = worldPose(_entity, _ecm);
+
+          std::vector<RayQuery> batchInput;
```
azeey: This can also be stored as a private component.

src/systems/physics/Physics.cc
```diff
@@ -4307,6 +4314,75 @@ void PhysicsPrivate::UpdateRayIntersections(EntityComponentManager &_ecm)
...
+
+  if (worldBatchRayFeature)
+  {
+    using BatchWorld = physics::World3d<BatchRayIntersectionFeatureList>;
+    using RayQuery = BatchWorld::RayQuery;
+    using RayIntersection = BatchWorld::RayIntersection;
+    using BatchedRayIntersectionData = BatchWorld::BatchedRayIntersectionData;
+
+    _ecm.Each<components::RaycastData,
+              components::Pose>(
```
azeey: `Pose` can be dropped if not used.

src/SdfEntityCreator.cc
azeey: Okay, but it doesn't look like `WorldPose` is being used for ray generation in Physics.cc

test/integration/physics_system.cc
```diff
@@ -3135,7 +3135,7 @@ TEST_F(PhysicsSystemFixture, GZ_UTILS_TEST_DISABLED_ON_WIN32(RayIntersections))
           ASSERT_TRUE(
             math::eigen3::convert(results2[i].normal).array().isNaN().all());
           ASSERT_TRUE(
-            std::isnan(results2[i].fraction));
+            std::isinf(results2[i].fraction));
```
azeey, line 3145: Can you extract this into it's own PR? gz-sim CI is failing since https://github.com/gazebosim/gz-physics/pull/880 was merged.

src/systems/cpu_lidar/CpuLidar.cc
```diff
@@ -0,0 +1,334 @@
...
+  const components::CpuLidar *_cpuLidar,
+  const components::ParentEntity *_parent)
+{
+  std::string sensorScopedName =
+      removeParentScope(scopedName(_entity, _ecm, "::", false), "::");
+  sdf::Sensor data = _cpuLidar->Data();
+  data.SetName(sensorScopedName);
+  if (data.Topic().empty())
+  {
+    std::string topic = scopedName(_entity, _ecm) + "/cpu_lidar";
```
azeey: We changed the default topic to `/lidar`

src/systems/cpu_lidar/CpuLidar.hh
```diff
@@ -0,0 +1,64 @@
...
+  /// \class CpuLidar CpuLidar.hh gz/sim/systems/CpuLidar.hh
+  /// \brief This system manages all CPU-based lidar sensors in simulation.
+  /// Each sensor publishes lidar scan data over Gazebo Transport.
+  class CpuLidar:
+    public System,
+    public ISystemPreUpdate,
+    public ISystemPostUpdate
+  {
+    /// \brief Constructor
+    public: explicit CpuLidar();
```
azeey:
```suggestion
    public: CpuLidar();
```

test/integration/physics_system.cc
apojomovsky: Extracted here: https://github.com/gazebosim/gz-sim/pull/3554

include/gz/sim/components/RaycastData.hh
apojomovsky: Done, split into a separate `NeedsRaycast` component (`Component<bool, class NeedsRaycastTag>`) defined alongside `RaycastData` in `RaycastData.hh`. `RaycastDataInfo` now only holds `rays` and `results`.

src/systems/cpu_lidar/CpuLidar.hh
apojomovsky: Done. Removed the pimpl entirely, now all members and private methods are now directly on `CpuLidar`.

src/systems/physics/Physics.cc
apojomovsky: Good point. Moved it to a separate `NeedsRaycast` component defined in `RaycastData.hh` (public, since both `CpuLidar.cc` and `Physics.cc` need it). If you'd prefer it private to Physics, I can move it to a .cc-local type, but that would require `CpuLidar` to set it via a different mechanism.

src/systems/physics/Physics.cc
apojomovsky: Done, fixed in both the batched and single-ray paths.

src/systems/physics/Physics.cc
apojomovsky: Good idea. I went with a `std::unordered_map<Entity, BatchRayCacheEntry>` in `PhysicsPrivate` instead of an ECM private component, since `BatchedRayIntersectionData` is a physics-engine template type and pulling it into a component header would require physics headers outside Physics.cc. The map achieves the same goal: both the input vector and the output buffer persist across steps so their allocated capacity is reused.

src/systems/physics/Physics.cc
apojomovsky: Same reasoning as above: the input vector holds `BatchWorld::RayQuery` values, which are also a physics-engine template type. Both input and output are cached together in the `PhysicsPrivate` map to keep physics types out of any component header.

src/systems/physics/Physics.cc
apojomovsky: Done. Removed from both `Each<>` calls in `UpdateRayIntersections`.

src/systems/cpu_lidar/CpuLidar.cc
apojomovsky: Fixed the fallback topic to use `/lidar` to match.

src/systems/cpu_lidar/CpuLidar.hh
apojomovsky: Done

src/systems/physics/Physics.cc
```diff
@@ -4307,6 +4324,75 @@ void PhysicsPrivate::UpdateRayIntersections(EntityComponentManager &_ecm)
...
+          if (!_needsRaycast->Data())
+            return true;
+          _needsRaycast->Data() = false;
+
+          const auto &rays = _raycastData->Data().rays;
+          auto &results = _raycastData->Data().results;
+          results.clear();
+          results.reserve(rays.size());
+
+          const auto &entityWorldPose = worldPose(_entity, _ecm);
```
azeey: I noticed you removed the `WorldPose` component. However, it might be more efficient to use that component rather than calling `worldPose` multiple times for the same entity.

apojomovsky: Sorry about that, my bad. I had missed that `UpdateSim` already keeps `WorldPose` current for sensor entities every step, so the component read is O(1) with no extra work needed.

Reverted: `SdfEntityCreator` is back to creating `WorldPose(Zero)` at spawn, and both `CpuLidar::Update` and `UpdateRayIntersections` are back to reading from the component directly.

src/SdfEntityCreator.cc
apojomovsky: It is now, fixed some commits ago (forgot to answer in this thread). Both paths in `UpdateRayIntersections` include `WorldPose` in the `Each<>` query and use it to transform rays into world space.

test/integration/cpu_lidar_system.cc
```diff
@@ -0,0 +1,359 @@
...
+  transport::Node node;
+  std::function<void(const msgs::LaserScan &)> cb =
+    [&](const msgs::LaserScan &_msg)
+    {
+      std::lock_guard<std::mutex> lock(mutex);
+      received.push_back(_msg);
+      messageReceived = true;
+      cv.notify_one();
+    };
+  node.Subscribe("/test/cpu_lidar", cb);
```
azeey: You can use the `Subscription` class from https://github.com/gazebosim/gz-sim/blob/daabfc8ebbf283fa14862385dd109c71e2c1715a/test/helpers/Subscription.hh#L39

test/integration/cpu_lidar_system.cc
```diff
@@ -0,0 +1,359 @@
...
+  auto &scan = received.back();
+  EXPECT_EQ(640, scan.count());
+  EXPECT_DOUBLE_EQ(0.08, scan.range_min());
+  EXPECT_DOUBLE_EQ(10.0, scan.range_max());
+
+  // Verify all range values are valid (not inf or nan and within bounds)
+  bool anyHit = false;
+  for (int i = 0; i < scan.ranges_size(); ++i)
+  {
+    if (!std::isinf(scan.ranges(i)) && !std::isnan(scan.ranges(i)))
```
azeey: nit: use `std::isfinite`

azeey (approved):
Just a few minor comments. Otherwise, this looks great! Thanks for the wonderful contribution.

test/integration/cpu_lidar_system.cc
apojomovsky: Done, switched both `LaserScanPublished` and `PointCloudPublished` to use `Subscription<T>`. Removed the manual mutex/condition_variable/callback boilerplate in both tests.

test/integration/cpu_lidar_system.cc
apojomovsky: Fixed, using `std::isfinite` now.

azeey (approved):

PR #3362, opened 2026-03-02
[Performance] - Remove / Optimize EntitiesByComponents calls

luca-della-vedova:
# 🎉 New feature

Make things faster.
And code smaller! Yay for negative diffs!
I split this into two commits, one that really just makes things faster https://github.com/gazebosim/gz-sim/commit/516332f34f0d968418b3a865e3dbd6209fbc36b6 and one that generally applies the same pattern throughout the code but without adding big performance gains. I'm happy to revert the second if we want a smaller diff.

## Summary
I ran some rudimentary profiling (aka Ctrl-C in GDB) on `3k_shapes.sdf` and noticed a lot of times I ended up with this stack trace:

```
0x00007ffff7ab7003 in std::_Hashtable<unsigned long, std::pair<unsigned long const, unsigned long>, std::allocator<std::pair<unsigned long const, unsigned long> >, std::__detail::_Select1st, std::equal_to<unsigned long>, std::hash<unsigned long>, std::__detail::_Mod_range_hashing, std::__detail::_Default_ranged_hash, std::__detail::_Prime_rehash_policy, std::__detail::_Hashtable_traits<false, false, true> >::find (this=0x55556086d970, __k=<optimized out>) at /usr/include/c++/13/bits/hashtable_policy.h:528
528	    { return __num % __den; }
(gdb) where
#0  0x00007ffff7ab7003 in std::_Hashtable<unsigned long, std::pair<unsigned long const, unsigned long>, std::allocator<std::pair<unsigned long const, unsigned long> >, std::__detail::_Select1st, std::equal_to<unsigned long>, std::hash<unsigned long>, std::__detail::_Mod_range_hashing, std::__detail::_Default_ranged_hash, std::__detail::_Prime_rehash_policy, std::__detail::_Hashtable_traits<false, false, true> >::find (this=0x55556086d970, __k=<optimized out>) at /usr/include/c++/13/bits/hashtable_policy.h:528
#1  0x00007ffff7aba6fc in std::unordered_map<unsigned long, unsigned long, std::hash<unsigned long>, std::equal_to<unsigned long>, std::allocator<std::pair<unsigned long const, unsigned long> > >::find (__x=<synthetic pointer>: <optimized out>, this=<optimized out>)
    at /usr/include/c++/13/bits/unordered_map.h:876
#2  gz::sim::v10::EntityComponentManager::ComponentImplementation (this=this@entry=0x555555964418, _entity=_entity@entry=8221, _type=50322453549405034) at /home/lucadv/ws_jetty/src/gz-sim/src/EntityComponentManager.cc:1253
#3  0x00007ffff7b2484f in gz::sim::v10::EntityComponentManager::Component<gz::sim::v10::components::Component<unsigned long, gz::sim::v10::components::ParentEntityTag, gz::sim::v10::serializers::DefaultSerializer<unsigned long> > > (_entity=8221, this=0x555555964418)
    at /home/lucadv/ws_jetty/src/gz-sim/include/gz/sim/detail/EntityComponentManager.hh:116
#4  gz::sim::v10::EntityComponentManager::EntitiesByComponents<gz::sim::v10::components::Component<unsigned long, gz::sim::v10::components::ParentEntityTag, gz::sim::v10::serializers::DefaultSerializer<unsigned long> >, gz::sim::v10::components::Component<std::add_lvalue_reference<void>, gz::sim::v10::components::LinkTag, gz::sim::v10::serializers::DefaultSerializer<std::add_lvalue_reference<void> > > >(gz::sim::v10::components::Component<unsigned long, gz::sim::v10::components::ParentEntityTag, gz::sim::v10::serializers::DefaultSerializer<unsigned long> > const&, gz::sim::v10::components::Component<std::add_lvalue_reference<void>, gz::sim::v10::components::LinkTag, gz::sim::v10::serializers::DefaultSerializer<std::add_lvalue_reference<void> > > const&) const::{lambda(auto:1 const&)#1}::operator()<gz::sim::v10::components::Component<unsigned long, gz::sim::v10::components::ParentEntityTag, gz::sim::v10::serializers::DefaultSerializer<unsigned long> > >(gz::sim::v10::components::Component<unsigned long, gz::sim::v10::components::ParentEntityTag, gz::sim::v10::serializers::DefaultSerializer<unsigned long> > const&) const (_desiredComponent=<optimized out>, __closure=<synthetic pointer>) at /home/lucadv/ws_jetty/src/gz-sim/include/gz/sim/detail/EntityComponentManager.hh:235
#5  gz::sim::v10::EntityComponentManager::ForEach<gz::sim::v10::EntityComponentManager::EntitiesByComponents<gz::sim::v10::components::Component<unsigned long, gz::sim::v10::components::ParentEntityTag, gz::sim::v10::serializers::DefaultSerializer<unsigned long> >, gz::sim::v10::components::Component<std::add_lvalue_reference<void>, gz::sim::v10::components::LinkTag, gz::sim::v10::serializers::DefaultSerializer<std::add_lvalue_reference<void> > > >(gz::sim::v10::components::Component<unsigned long, gz::sim::v10::components::ParentEntityTag, gz::sim::v10::serializers::DefaultSerializer<unsigned long> > const&, gz::sim::v10::components::Component<std::add_lvalue_reference<void>, gz::sim::v10::components::LinkTag, gz::sim::v10::serializers::DefaultSerializer<std::add_lvalue_reference<void> > > const&) const::{lambda(auto:1 const&)#1}, gz::sim::v10::components::Component<unsigned long, gz::sim::v10::components::ParentEntityTag, gz::sim::v10::serializers::DefaultSerializer<unsigned long> >, gz::sim::v10::components::Component<std::add_lvalue_reference<void>, gz::sim::v10::components::LinkTag, gz::sim::v10::serializers::DefaultSerializer<std::add_lvalue_reference<void> > > >(gz::sim::v10::EntityComponentManager::EntitiesByComponents<gz::sim::v10::components::Component<unsigned long, gz::sim::v10::components::ParentEntityTag, gz::sim::v10::serializers::DefaultSerializer<unsigned long> >, gz::sim::v10::components::Component<std::add_lvalue_reference<void>, gz::sim::v10::components::LinkTag, gz::sim::v10::serializers::DefaultSerializer<std::add_lvalue_reference<void> > > >(gz::sim::v10::components::Component<unsigned long, gz::sim::v10::components::ParentEntityTag, gz::sim::v10::serializers::DefaultSerializer<unsigned long> > const&, gz::sim::v10::components::Component<std::add_lvalue_reference<void>, gz::sim::v10::components::LinkTag, gz::sim::v10::serializers::DefaultSerializer<std::add_lvalue_reference<void> > > const&) const::{lambda(auto:1 const&)#1}, gz::sim::v10::components::Component<unsigned long, gz::sim::v10::components::ParentEntityTag, gz::sim::v10::serializers::DefaultSerializer<unsigned long> > const&, gz::sim::v10::components::Component<std::add_lvalue_reference<void>, gz::sim::v10::components::LinkTag, gz::sim::v10::serializers::DefaultSerializer<std::add_lvalue_reference<void> > > const&) (_f=...) at /home/lucadv/ws_jetty/src/gz-sim/include/gz/sim/detail/EntityComponentManager.hh:438
#6  gz::sim::v10::EntityComponentManager::EntitiesByComponents<gz::sim::v10::components::Component<unsigned long, gz::sim::v10::components::ParentEntityTag, gz::sim::v10::serializers::DefaultSerializer<unsigned long> >, gz::sim::v10::components::Component<std::add_lvalue_reference<void>, gz::sim::v10::components::LinkTag, gz::sim::v10::serializers::DefaultSerializer<std::add_lvalue_reference<void> > > > (this=0x555555964418) at /home/lucadv/ws_jetty/src/gz-sim/include/gz/sim/detail/EntityComponentManager.hh:233
#7  gz::sim::v10::Model::Links (this=this@entry=0x7fffffffc1c0, _ecm=...) at /home/lucadv/ws_jetty/src/gz-sim/src/Model.cc:174
#8  0x00007fffe80f706f in gz::sim::v10::systems::PhysicsPrivate::UpdateModelPose (this=this@entry=0x55556166e500, _model=<optimized out>, _canonicalLink=<optimized out>, _ecm=..., _linkFrameData=std::map with 468 elements = {...})
    at /home/lucadv/ws_jetty/src/gz-sim/src/systems/physics/Physics.cc:3348
#9  0x00007fffe80f7718 in gz::sim::v10::systems::PhysicsPrivate::UpdateSim (this=0x55556166e500, _ecm=..., _linkFrameData=std::map with 468 elements = {...}) at /home/lucadv/ws_jetty/src/gz-sim/src/systems/physics/Physics.cc:3481
#10 0x00007fffe80f910b in gz::sim::v10::systems::Physics::Update (this=0x55556166e4c0, _info=<optimized out>, _ecm=...) at /usr/include/c++/13/bits/unique_ptr.h:199
#11 0x00007ffff7b88e13 in gz::sim::v10::SimulationRunner::UpdateSystems (this=0x5555559643c0) at /home/lucadv/ws_jetty/src/gz-sim/src/SimulationRunner.cc:648
#12 0x00007ffff7b9155e in gz::sim::v10::SimulationRunner::Step (this=this@entry=0x5555559643c0, _info=...) at /home/lucadv/ws_jetty/src/gz-sim/src/SimulationRunner.cc:999
#13 0x00007ffff7b91e17 in gz::sim::v10::SimulationRunner::Run (this=0x5555559643c0, _iterations=<optimized out>) at /home/lucadv/ws_jetty/src/gz-sim/src/SimulationRunner.cc:888
#14 0x00007ffff7b79223 in gz::sim::v10::ServerPrivate::Run (this=0x5555556b8140, _iterations=<optimized out>, _iterations@entry=0, _cond=Python Exception <class 'gdb.error'>: value has been optimized out
...) at /usr/include/c++/13/bits/unique_ptr.h:199
#15 0x00007ffff7b5b7c5 in gz::sim::v10::Server::Run (this=this@entry=0x7fffffffd180, _blocking=_blocking@entry=true, _iterations=0, _paused=<optimized out>) at /home/lucadv/ws_jetty/src/gz-sim/src/Server.cc:185
#16 0x0000555555565b5c in main (argc=3, argv=0x7fffffffd8a8) at /home/lucadv/ws_jetty/src/gz-sim/src/cmd/sim_main.cc:578
```

What's very interesting is stack number 7, the Links call in the `Model` API.
This call under the hood uses `EntitiesByComponents`.
`EntitiesByComponents` is **expensive**, what it does under the hood is that it iterates throughout the _whole ECM_ and checks that we can find an entity with the matched components.
I generally feel this API should be very much discouraged and its usage scrutinized because it's really expensive. But anyway what I did in this PR is:

#### Avoid EntitiesByComponents and use ChildrenByComponents instead

In the vast majority of cases (including the bottleneck in the stacktrace above), calls to `EntitiesByComponents` look like this:

```
ecm.EntitiesByComponents(components::ParentEntity(id), [some components]);
```

I simply changed them to:

```
ecm.ChildrenByComponents(id, [some components]);
```

#### Improve efficiency of `ChildrenByComponents`

The above wouldn't do anything in the previous implementation of `EntitiesByComponents` that iterated throughout all the entities in the view (that might be a lot) and checked if any of them was the child of the right entity.
I switched the iteration, now it iterates through the children and makes sure they are in the view. This should be much faster in cases in which the the size of the view is >> the number of direct children of a specific entity (that should be most cases I guess)?

## Test it

Run 3k_shapes.sdf with and without this branch:

```
gz sim 3k_shapes.sdf -r
```

On my machine I get about 0.30% RTF without the PR and about 0.95% with the PR, so about 3 times faster.
I'm not sure what sort of speedups you can observe but hopefully similar.

## Followups

My rudimentary profiling now shows the bulk of the time spent in physics which is what I would hope to see. I still see some non-negligible amount of time spent in an [EachNoCache](https://github.com/gazebosim/gz-sim/blob/5049587241f63826dca7fd9a4b501d1ba91a0a73/src/SimulationRunner.cc#L1435-L1444) call in SimulationRunner that shouldn't actually contain any entity in most cases. I might look into it if I have time.

There is also an `EntityByComponents` API that returns a single entity rather than a vector, that's a more obvious candidate for optimization in a followup.

## Checklist
- [ ] Added tests


luca-della-vedova:
CI failure looks real (and a bit scary), will dig further

arjo129:
I think its just an issue of timeouts/gz-transport/gz-fuel inside our unit test . I suspect the test is flaky now with the faster computation time.

luca-della-vedova:
> I think its just an issue of timeouts/gz-transport/gz-fuel inside our unit test . I suspect the test is flaky now with the faster computation time.

No the `ModelCommandAPI` test is failing locally as well where the entities returned are different.
I suspect the issue is that we have two sources of truth for parenthood information, `components::ParentEntity(parent)` and the `EntityGraph` node adjacency list. This PR switches from the first to the second but it seems they are not perfectly matching.
I'll try to get a minimal fix for this PR but I do wonder if we should get rid of the graph approach and keep everything in components instead (perhaps a parent can have a `Children` component, similar to how Bevy does it)

luca-della-vedova:
The ModelCommandAPI uses a [service call](https://github.com/gazebosim/gz-sim/blob/5049587241f63826dca7fd9a4b501d1ba91a0a73/src/cmd/ModelCommandAPI.cc#L542-L580) to get the state of the ECM.
Afterwards, it uses the [SetState API](https://github.com/gazebosim/gz-sim/blob/5049587241f63826dca7fd9a4b501d1ba91a0a73/src/EntityComponentManager.cc#L1813-L1908) to populate a brand new ECM that it then queries to find models, links.

I _think_ the issue is that we populate that ECM only with component values and as I mentioned above information about which parent has which children is not encoded in a component but in a property of the EntityGraph, hence this can't be serialized / deserialized.
The behavior of the `ChildrenByComponents` API has probably been broken when setting states from protobuf messages but there was no users of this until now.

What I can't figure out is how come it finds _some_ of the models but not all of them.

luca-della-vedova:
I couldn't figure out the ModelCommandAPI breakage so I reverted it in https://github.com/gazebosim/gz-sim/pull/3362/commits/0798e40a8cb6f1e514feb55fa8de96442ac42952.
I'm a bit worried that this is hiding some incorrect behavior in the `SetState` API call, perhaps not updating the hierarchy correctly, but I couldn't really get anywhere.
With that reverted CI is now green (other than an expected homebrew breakage)

include/gz/sim/detail/EntityComponentManager.hh
```diff
@@ -264,9 +264,9 @@ std::vector<Entity> EntityComponentManager::ChildrenByComponents(Entity _parent,
 
   // Iterate over entities
   std::vector<Entity> result;
-  for (const Entity entity : view->Entities())
+  for (const auto [entity, _] : children)
```
ntfshard: seems can be const reference `&`, not by copy
also I guess `_` can be treated as unused variable by compiler. As possible you can capture by reference a pair and get first element where you need it.

luca-della-vedova: The _idea_ behind it is that usually you don't need to pass by reference primitive types since copying a `uint64_t` will probably be cheaper than dereferencing a reference.
Now this won't actually be the case if the compiler behind the scenes still creates a copy for `_`, I admit that syntax was borrowed from Rust and I don't actually know if a `_` variable will be instanced and copied or if it will (hopefully) just be optimized away.

ntfshard: > The _idea_ behind it is that usually you don't need to pass by reference primitive types since copying a `uint64_t` will probably be cheaper than dereferencing a reference.

Yes, usually don't need to pass by a reference/pointer [POD type](https://en.cppreference.com/w/cpp/named_req/PODType) variable. But it should be much cheaper than calling copy constructor for a second object (if it's not POD too)

> Now this won't actually be the case if the compiler behind the scenes still creates a copy for `_`, I admit that syntax was borrowed from Rust and I don't actually know if a `_` variable will be instanced and copied or if it will (hopefully) just be optimized away.

So, _g++15_ complaining on a similar construction: `warning: loop variable '<structured bindings>' creates a copy from type...` https://godbolt.org/z/fYa7Tvz6q
I believe compiler have to make copy due to it should keep a side-effects (discussable, I can try to find justification in standard. In current line it's my IMHO)
Also std::ignore is not helpful in this case afais

About the name, in a latest draft(basically can use any release, just site is convenient) of standard. We can refer to https://eel.is/c++draft/lex.name which contains `identifier` definition. Which should start from `nondigit` which is listed there and underscore `_` is listed. 
But [(3.2)](https://eel.is/c++draft/lex.name#3.2): 
`Each identifier that begins with an underscore is reserved to the implementation for use as a name in the global namespace.` So potentially it can be name collision which will trigger `Wshadow` warning.

And if I already started digging, as an idea, can you try to run a simulation w/o GUI.
Something like: `time gz sim -s -r --iterations 128 3k_shapes.sdf` to get a cleaner results. I'm not trusting GUI due to it involving rendering pipeline(and also ECM synchronization).
Still better to have something from a real profiler tool like `perf` or `valgring --tool=cachegrind` or VTune or gprof... Could be more informative I believe

luca-della-vedova: Thank you for the feedback and the godbolt link! I agree that saving a few characters is really not worth the risk of compiler warnings, variable shadowing and potentially having a copy behind the scenes, I just changed it to the classic approach of iterating by range and adding a `.first` access in https://github.com/gazebosim/gz-sim/pull/3362/commits/8f45fde5d183d43166c92d7927746ece8005e843 as mentioned in your original comment.

azeey:
@mergifyio backport gz-sim8 gz-sim9 ign-gazebo6 main

PR #3373, opened 2026-03-08
Enable advanced magnetic field navigation simulations with WMMHR 2025

bperseghetti:
### The built-in magnetometer system currently uses a low-resolution lookup table (10-degree sampling, 13x37 grid) that only covers latitudes -60 to +60 degrees. This means:

- No coverage at high latitudes — any simulation above 60N or below 60S gets clamped/zero values, making polar and sub-polar robotics simulation unreliable (e.g., Arctic AUVs, Antarctic rovers).
- Low spatial accuracy — the 10-degree grid (~1100 km spacing) uses bilinear interpolation, which can be off by hundreds of nanotesla compared to real measurements. Robots relying on magnetometer-based heading (compass navigation, magnetic anomaly detection) will see unrealistic behavior.
- No altitude dependence — the lookup table ignores elevation entirely. The field at sea level and at 100 km altitude returns the same value, which is physically wrong.
- Frozen in time — the table has no secular variation, so it doesn't reflect the year being simulated. Earth's magnetic field drifts measurably year over year.

### The WMMHR 2025 integration solves all of these:

- Global coverage including poles, with singularity-safe evaluation at exactly 90/-90 degrees latitude.
- Degree-133 spherical harmonic expansion (~150 km spatial resolution), matching the reference values from NOAA's official tool to within 5 nT.
- Altitude-dependent field computation using geodetic coordinates on the WGS84 ellipsoid.
- Secular variation support for any date in the 2025-2030 epoch.
- Per-sensor caching with configurable Haversine distance threshold so the expensive spherical harmonic evaluation doesn't run every simulation step.
- Zero breaking changes — the WMM is opt-in via <use_world_magnetic_model>true</use_world_magnetic_model>. Existing worlds using the lookup table are unaffected. The COF file auto-discovers from the install path, so minimal SDF configuration is needed.
- Unit and frame metadata in every published message, so downstream consumers (ROS bridges, logging, GUIs) know exactly what they're receiving without out-of-band assumptions.

Depends on:
https://github.com/gazebosim/gz-msgs/pull/568
https://github.com/gazebosim/gz-sensors/pull/595

## Checklist
- [x] Added tests


peci1:
For ROS, I've built magnetic_model package which uses Geographiclib for the computations: https://github.com/ctu-vras/compass/blob/9f1b4822659b731b20ee87125588e53fda0c5cbb/magnetic_model/src/magnetic_model.cpp#L174 . 

I think it would be also beneficial here to use GeographicLib and not reinvent the wheel. If you want to get serious about magnetic stuff, you basically need it (or maybe for gravity, too, in the future!).

Support for WMMHR2025 was added in version 2.5. However, Ubuntu Noble and RHEL 8+9 only ship 2.3. But I've checked the commit history and now I'm 75% sure even 2.3 would correctly compute WMMHR model. The math is the same, it just has more depth.

peci1:
Gazebo-implementation-wise, I'm not 100% sure specifying the reference time in the plugin properties is a good idea. I'd instead suggest just using the current sim time (if it is in the range of the model), or current system time. If you bake in the month and year into SDF, it would be difficult to share the model with community and have it reusable.

The same holds for specifying a concrete model (someone might want to use your model  5 years later). Here, it would maybe be good to just specify the desired precision of the model (low (Gazebo), medium (WMM), high (WMMHR)). Or maybe just a class (like Gazebo/WMM/WMMHR, but without year).

I know that most simulations start from time 0 which is not suitable for usage with these models. Even now, Gazebo has `----initial-sim-time` argument using which you can set the time. And I think it wouldn't be difficult to implement something like `--now` that would just start the sim with current sim time value (but then it'd run with its own pace).

There would, of course, have to be some default used by Gazebo when the sim starts with 0 time. I think using current system time wouldn't hurt anything. The difference between specifying `--now` and not specifying anthing would be that with `--now` you set the time of the whole simulation, while without it, sim starts in 0 and only magnetic model applies the current system time.

tutorials/magnetometer.md
```diff
@@ -0,0 +1,170 @@
+\page magnetometer Magnetometer
```
nkoenig, line 1: Can you add this tutorial to the `tutorials.md.in` file please?

src/systems/magnetometer/Magnetometer.cc
```diff
@@ -269,6 +317,82 @@ void Magnetometer::Configure(const Entity &/*_entity*/,
...
+      double decimalYear = -1.0;
+      if (_sdf->HasElement("wmm_date"))
+      {
+        decimalYear = _sdf->Get<double>("wmm_date");
+      }
+
+      // Recalculation distance
+      if (_sdf->HasElement("wmm_recalculation_distance"))
+      {
+        this->dataPtr->wmmRecalcDistance =
```
nkoenig, line 347: Do you want to check that this value is non-negative and non-zero?

src/systems/magnetometer/Magnetometer.cc
```diff
@@ -269,6 +317,82 @@ void Magnetometer::Configure(const Entity &/*_entity*/,
...
+      if (_sdf->HasElement("wmm_coefficient_file"))
+      {
+        cofFile = _sdf->Get<std::string>("wmm_coefficient_file");
+      }
+
+      // Decimal year (default: current date)
+      double decimalYear = -1.0;
+      if (_sdf->HasElement("wmm_date"))
+      {
+        decimalYear = _sdf->Get<double>("wmm_date");
```
nkoenig, line 341: I think there is no check that wmm_date falls within the models epoch. Is this something you want to check for?

nkoenig (changes requested):

nkoenig:
> Geographiclib

I like this idea. My recommendation would be to capture this as follow-on work unless you think it's a blocker for this contribution.

nkoenig:
@bperseghetti I think CI checks for this PR needs some love.

PR #3421, opened 2026-03-24
Introduce relative topic names to DiffDrive

C88-YQ:
# 🎉 New feature

Related to  #3420 

## Summary

This PR introduces support for relative topic names in the `DiffDrive` system.
* Topics starting with `/` are treated as absolute
* Otherwise, topics are treated as relative and automatically prefixed with the model name
```
/<model_name>/<topic>
```
This behavior is applied to: `cmd_vel`, `odometry`, and `tf`.

New tests are added, and existing tests are refactored to reduce duplication by extracting common logic.

## Test it
```
ctest -R diff
```

## Checklist
- [x] Added tests

### DIFF
--- a/src/systems/diff_drive/DiffDrive.cc
+++ b/src/systems/diff_drive/DiffDrive.cc
@@ -336,7 +336,12 @@ void DiffDrive::Configure(const Entity &_entity,
   std::vector<std::string> topics;
   if (_sdf->HasElement("topic"))
   {
-    topics.push_back(_sdf->Get<std::string>("topic"));
+    std::string topic = _sdf->Get<std::string>("topic");
+    if (topic.front() != '/')
+    {
+      topic = "/" + this->dataPtr->model.Name(_ecm) + "/" + topic;
+    }
+    topics.push_back(topic);
   }
   topics.push_back("/model/" + this->dataPtr->model.Name(_ecm) + "/cmd_vel");
   auto topic = validTopic(topics);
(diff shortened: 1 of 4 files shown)


arjo129 (changes requested):
This makes a lot of sense, but it does break convention. This may make it difficult to backport to older versions. One potential option is to only enable this behaviour via a flag. 

I would recommend updating [`Migration.md`](https://github.com/gazebosim/gz-sim/blob/main/Migration.md)

C88-YQ:
@arjo129 Thank you very much for the review!

I agree that this would be a breaking change, so we are also trying to find a more flexible approach that would be easier to backport.

After some discussion, we found that another possible direction may be to introduce a `namespace` field in `SDFormat` instead. I opened an issue for that discussion here: [Add namespace support to SDF elements for easier multi-robot simulation](https://github.com/gazebosim/sdformat/issues/1659)

I am not fully sure yet whether this PR is still needed if the SDFormat-side discussion reaches an agreement. One possible reason to keep this PR, or something similar to it, is that it could let users decide whether a specific topic should inherit the namespace or remain global. For example, some topics such as `/tf` may be better kept as a shared global topic in some workflows, otherwise the `ros_gz_bridge` configuration may become more complicated.

That said, if we end up keeping this behavior, I agree that we may need to introduce something like a `gz::policies` option to explicitly control whether this feature is enabled.

For now, I will mark this PR as a draft and wait until the overall approach becomes clearer. Depending on the result of the SDFormat discussion, I will either update this PR to match the final design or close it if it no longer fits the direction.

PR #3459, opened 2026-04-07
Fix EventManager RTTI failures across shared-library boundaries

### LINKED ISSUE #1992: EventManager use of typeid to index events may have undefined behaviour in plugins
## Environment
* OS Version: macOS
* Source or binary build? Source: branch: gz-sim7, commit: bb85d05a3

## Description
* Expected behavior: Callbacks bound to an event using `gz::sim::EventManager::Connect` should be called when the corresponding event is emitted. This should work in plugins as well as the `gz-sim` core.
* Actual behavior: The callback is not called.

## Steps to reproduce

1. Build and run the `gz-sim/examples/plugin/rendering_plugins` on macOS.
2. The `RenderingServerPlugin` does not update the scene's ambient lighting as expected.

## Output

The issue arises because use of:

```c++
private: using TypeInfoRef = std::reference_wrapper<const std::type_info>;
```

to index the `unordered_map` of event types is not generally portable across shared library boundaries.  There is a helpful discussion here http://cpptruths.blogspot.com/2018/11/non-colliding-efficient.html with some suggested resolutions.

On macOS the map is populated with multiple entries for the same event type, because there is no standard forcing different shared libraries to return the same hash for a reference to `typeid({EventType})`. That this works on Linux may be accidental. There is no guarantee it would work for plugins built using different compiler settings to the released libraries.

Using `RTLD_GLOBAL` for macOS in `gz::plugin::Loader::PrivateGetPluginDlHandlePtr` does work for the plugin example , but unfortunately does not work in general (and caused segmentation faults - probably due to other symbols conflicting).

taylorhoward92:
# 🦟 Bug fix

Fixes #1992

## Summary

When plugins are loaded with `RTLD_LOCAL` and built with hidden visibility (the default via gz-cmake), each dylib gets its own copy of the RTTI `type_info` for the same C++ type. This causes `dynamic_cast` to return `nullptr` in `EventManager::Connect`, `Emit`, and `ConnectionCount` even though the object really is the requested type. It also causes `typeid(E).hash_code()` and `operator==` on `type_info` references to disagree across dylib boundaries, creating duplicate map entries.

This is the root cause of the issues described in #1992.

### Changes

- Replace `dynamic_cast` with `static_cast` in `Connect()`, `Emit()`, and `ConnectionCount()`. The map key already guarantees the correct type by construction.
- Switch `Hasher` to name-based hashing via `std::hash<std::string>(typeid::name())`.
- Add a name-based fallback in `EqualTo` so that the same type loaded from different shared libraries compares equal.

## Checklist
- [ ] Added tests

Generated-by: Claude Code

### DIFF
--- a/include/gz/sim/EventManager.hh
+++ b/include/gz/sim/EventManager.hh
@@ -17,6 +17,7 @@
 #ifndef GZ_SIM_EVENTMANAGER_HH_
 #define GZ_SIM_EVENTMANAGER_HH_
 
+#include <cstring>
 #include <functional>
 #include <memory>
 #include <typeinfo>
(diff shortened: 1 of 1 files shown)


azeey (approved):
Makes sense and much simpler than the alternative approach in #1992. This fixes a crash-at-exit on macOS for me when running `gz sim -v4 gpu_lidar_sensor.sdf`. Thanks for the contribution!

azeey:

PR #3478, opened 2026-04-10
Use high resolution timer on Windows

johnwason:
## Summary

This PR fixes timing issues on Windows. Simulation ratios have been observed as low as 6.5% on Windows with an empty scene, and simulation rates often hover around 80% with busier scenes. The problem was caused by the loop rate stabilization strategy. By default the Windows kernel timing resolution is 15.6 ms, meaning that a normal sleep command can not wake before 15.6 ms. On older version of Windows, a system-wide command was used to modify the system timing resolution, meaning that if any process requested a higher timing resolution all timers would also have this resolution. Newer Windows updates have changed this behavior, and each high resolution timer must now use the `CREATE_WAITABLE_TIMER_HIGH_RESOLUTION` with the Win32 API function `CreateWaitableTimerEx()` to create a timer that has the higher resolution capability.

This PR modifies the `SimulationRunner` class to use a high resolution timer instead of `std::this_thread::sleep_until()` when on Windows.

We ran into a similar problem in the Robot Raconteur project: https://github.com/robotraconteur/robotraconteur/pull/239

## Checklist
- [ ] Added tests

@traversaro

### DIFF
--- a/src/SimulationRunner.cc
+++ b/src/SimulationRunner.cc
@@ -185,11 +185,6 @@ SimulationRunner::SimulationRunner(const sdf::World &_world,
 
 #ifdef _WIN32
   winPrecisionTimer = CreateWaitableTimerExA(NULL, NULL, CREATE_WAITABLE_TIMER_HIGH_RESOLUTION, TIMER_ALL_ACCESS);
-  if (winPrecisionTimer == NULL)
-  {
-    gzerr << "Could not initialize Windows precision timer" << std::endl;
-    return;
-  }
 #endif
 
   // World control
@@ -942,17 +937,24 @@ bool SimulationRunner::Run(const uint64_t _iterations)
 #ifndef _WIN32
           std::this_thread::sleep_until(sleepTarget);
 #else
-          auto sleepTargetDuration = std::chrono::duration_cast<std::chrono::microseconds>(sleepTarget - now);
-          LARGE_INTEGER due_time;
-          memset(&due_time, 0, sizeof(due_time));
-          due_time.QuadPart = -sleepTargetDuration.count() * 10;
-          if (SetWaitableTimer(winPrecisionTimer, &due_time, 0, NULL, NULL, FALSE) != TRUE)
+          if (winPrecisionTimer != NULL)
           {
-            gzerr << "Could not SetWaitableTimer" << std::endl;
+            auto sleepTargetDuration = std::chrono::duration_cast<std::chrono::microseconds>(sleepTarget - now);
+            LARGE_INTEGER due_time;
+            memset(&due_time, 0, sizeof(due_time));
+            due_time.QuadPart = -sleepTargetDuration.count() * 10;
+            if (SetWaitableTimer(winPrecisionTimer, &due_time, 0, NULL, NULL, FALSE) != TRUE)
+            {
+              gzerr << "Could not SetWaitableTimer" << std::endl;
+            }
+            else
+            {
+              WaitForSingleObject(winPrecisionTimer, INFINITE);
+            }
           }
           else
           {
-            WaitForSingleObject(winPrecisionTimer, INFINITE);
+            std::this_thread::sleep_until(sleepTarget);
           }
 #endif
         }
--- a/src/SimulationRunner.cc
+++ b/src/SimulationRunner.cc
@@ -67,6 +67,16 @@
 #include "LevelManager.hh"
 #include "SdfGenerator.hh"
 
+#ifdef _WIN32
+#ifndef NOMINMAX
+#define NOMINMAX
+#endif
+#ifndef WIN32_LEAN_AND_MEAN
+#define WIN32_LEAN_AND_MEAN
+#endif
+#include <windows.h>
+#endif
+
 using namespace gz;
 using namespace sim;
 
@@ -104,6 +114,31 @@ struct MaybeGilScopedRelease
 
 }
 
+#ifdef _WIN32
+namespace gz
+{
+namespace sim
+{
+inline namespace GZ_SIM_VERSION_NAMESPACE {
+class SimulationRunnerWinHandleStorage
+{
+  private: HANDLE handle_storage{NULL};
+
+  public: HANDLE handle() { return handle_storage; }
+
+  public: SimulationRunnerWinHandleStorage(HANDLE h) : handle_storage(h) {}
+
+  public: ~SimulationRunnerWinHandleStorage() {
+    if (handle_storage != NULL)
+    {
+      CloseHandle(handle_storage);
+    }
+  }
+};
+}
+}
+}
+#endif
 
 //////////////////////////////////////////////////
 SimulationRunner::SimulationRunner(const sdf::World &_world,
@@ -184,7 +219,11 @@ SimulationRunner::SimulationRunner(const sdf::World &_world,
   this->currentInfo.simTime = this->simTimeEpoch;
 
 #ifdef _WIN32
-  winPrecisionTimer = CreateWaitableTimerExA(NULL, NULL, CREATE_WAITABLE_TIMER_HIGH_RESOLUTION, TIMER_ALL_ACCESS);
+  HANDLE winPrecisionTimer_handle = CreateWaitableTimerExA(NULL, NULL, CREATE_WAITABLE_TIMER_HIGH_RESOLUTION, TIMER_ALL_ACCESS);
+  if (winPrecisionTimer_handle != NULL)
+  {
+    winPrecisionTimer = std::make_unique<SimulationRunnerWinHandleStorage>(winPrecisionTimer_handle);
+  }
 #endif
 
   // World control
@@ -328,13 +367,6 @@ SimulationRunner::SimulationRunner(const sdf::World &_world,
 SimulationRunner::~SimulationRunner()
 {
   this->StopWorkerThreads();
-#ifdef _WIN32
-  if (winPrecisionTimer != NULL)
-  {
-    CloseHandle(winPrecisionTimer);
-    winPrecisionTimer = NULL;
-  }
-#endif
 }
 
 /////////////////////////////////////////////////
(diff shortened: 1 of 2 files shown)


traversaro:
fyi @RiccardoGrieco, this may be useful to know for your work.

src/SimulationRunner.hh
```diff
@@ -589,6 +589,9 @@ namespace gz
       /// initialization and should exit immediately. See
       /// `SetExitedWithErrors()`.
       private: bool exitedWithErrors{false};
+#ifdef _WIN32
+      private: HANDLE winPrecisionTimer;
```
traversaro: This breaks the ABI, can't we put this in the `SimulationRunnerPrivate` class?

src/SimulationRunner.cc
```diff
@@ -183,6 +183,15 @@ SimulationRunner::SimulationRunner(const sdf::World &_world,
   );
   this->currentInfo.simTime = this->simTimeEpoch;
 
+#ifdef _WIN32
+  winPrecisionTimer = CreateWaitableTimerExA(NULL, NULL, CREATE_WAITABLE_TIMER_HIGH_RESOLUTION, TIMER_ALL_ACCESS);
```
traversaro: Per include-what-you-use, should't we include some headers that define this function and macro?

traversaro:
xref https://github.com/ros2/rclpy/issues/1565 and fyi @knmcguire as the most Windows-interested person I know in the community.

src/SimulationRunner.cc
```diff
@@ -183,6 +183,15 @@ SimulationRunner::SimulationRunner(const sdf::World &_world,
...
+  if (winPrecisionTimer == NULL)
+  {
+    gzerr << "Could not initialize Windows precision timer" << std::endl;
```
traversaro: If I understood correctly, this may happen in older Windows systems. Could it make sense to just go back to use `sleep_until` on older systems?

traversaro: A similar fallback is implemented in https://github.com/rust-lang/rust/pull/116461 .

src/SimulationRunner.hh
johnwason: Unless I am missing something I don't see an implementation for `SimulationRunnerPrivate`

src/SimulationRunner.cc
johnwason: It looks like Windows.h is included already somewhere. With Windows there isn't really a clean way to include anything other than the global Windows.h. There are some macros like `WIN32_LEAN_AND_MEAN` that affect what is included so I wouldn't recommend including again.

src/SimulationRunner.cc
johnwason: I deal with this problem using this strategy elsewhere:

https://github.com/robotraconteur/robotraconteur/blob/13412656a51a745c059fc7c34f537377c17a06a8/RobotRaconteurCore/src/Timer.cpp#L234-L247

I am not sure if it is better to `sleep_until` or this strategy as a fallback.

traversaro: As you prefer, as I think it is harder to find an old Windows box to test this on, `sleep_until` is probably safer (even with the performance problems you mentioned and that this PR is fixing).

src/SimulationRunner.cc
traversaro: Can't we just include synchapi.h ?

src/SimulationRunner.hh
traversaro: You are perfectly right, I was confused by the forward declaration of the class in this file. However, this is not ideal as it relies on `.hh` including the `Windows.h` (that as you mentioned, creates a lot of problems). Even if that is happening now transitively, I do not think we should do that. As `winPrecisionTimer` is only used in the `SimulationRunner::Run` method, can't we just define it as a local variable there? That would permit also to have a fix that can be backported to older gz-sim versions, without breaking the ABI.

johnwason: The timer is created in the constructor and deleted in the destructor. I think for older gz-sim versions we can declare it as a thread_local variable at the module level and that should be fine. It is a bit of a hack but with the way the class works in practice it should work. If it is a method level variable we would have to create and destroy the timer each cycle and that I don't recommend. I can create a wrapper class to store the `HANDLE` so that type only appears in the cpp file. A "smart handle" pattern is very common that cleans up when the wrapper is destroyed.

src/SimulationRunner.cc
johnwason: I added the fallback to `sleep_until`.

src/SimulationRunner.hh
johnwason: I created a wrapper class for the handle. Take a look and see if you agree with this design.

src/SimulationRunner.cc
johnwason: I moved the handle into the cpp file and added a conditional include to `windows.h` similar to how other files include it.

johnwason:
I think it would make sense to implement the `SimulationRunnerPrivate` class but this is probably not the place to do it.

knmcguire:
@traversaro thanks for letting me know! folks need for testing it out? 

I think the PR content itself is probably a little out of depth for me (although I learned about the LEAN_AND_MEAN variable which is cool!), but I can definitely build and run test it on my machine ones I have time

azeey:
This is awesome! The RTF not ever reaching the set value on Windows has always bugged me, but I never had a chance to look into it.

As far as API/ABI compatibility is concerned, SimulationRunner is a private class and the `SimulationRunner.hh` header is not installed, so we don't need to worry about it. I don't know why we have a forward declaration for `SimulationRunnerPrivate`. That class doesn't actually exist.

traversaro:
> This is awesome! The RTF not ever reaching the set value on Windows has always bugged me, but I never had a chance to look into it.
> 
> As far as API/ABI compatibility is concerned, SimulationRunner is a private class and the `SimulationRunner.hh` header is not installed, so we don't need to worry about it. I don't know why we have a forward declaration for `SimulationRunnerPrivate`. That class doesn't actually exist. 

Oh, this is great news!

src/SimulationRunner.cc
```diff
@@ -104,6 +114,31 @@ struct MaybeGilScopedRelease
 
 }
 
+#ifdef _WIN32
+namespace gz
+{
+namespace sim
+{
+inline namespace GZ_SIM_VERSION_NAMESPACE {
+class SimulationRunnerWinHandleStorage
```
azeey, line 126: Mind a line or two about what this class does?

src/SimulationRunner.cc
```diff
@@ -104,6 +114,31 @@ struct MaybeGilScopedRelease
...
+{
+  private: HANDLE handle_storage{NULL};
```
azeey: nit: we use `camelCase` for member variables.

src/SimulationRunner.cc
```diff
@@ -183,6 +218,14 @@ SimulationRunner::SimulationRunner(const sdf::World &_world,
   );
   this->currentInfo.simTime = this->simTimeEpoch;
 
+#ifdef _WIN32
+  HANDLE winPrecisionTimer_handle = CreateWaitableTimerExA(NULL, NULL, CREATE_WAITABLE_TIMER_HIGH_RESOLUTION, TIMER_ALL_ACCESS);
```
azeey: nit: camelCase

```suggestion
  HANDLE winPrecisionTimerHandle = CreateWaitableTimerExA(NULL, NULL, CREATE_WAITABLE_TIMER_HIGH_RESOLUTION, TIMER_ALL_ACCESS);
```

src/SimulationRunner.cc
```diff
@@ -914,14 +957,38 @@ bool SimulationRunner::Run(const uint64_t _iterations)
...
         {
+#ifndef _WIN32
           std::this_thread::sleep_until(sleepTarget);
+#else
+          if (winPrecisionTimer)
+          {
+            auto sleepTargetDuration = std::chrono::duration_cast<std::chrono::microseconds>(sleepTarget - now);
+            LARGE_INTEGER due_time;
+            memset(&due_time, 0, sizeof(due_time));
+            due_time.QuadPart = -sleepTargetDuration.count() * 10;
```
azeey, line 988: Consider adding a comment here as to why the duration negative and is multiplied by 10 since the API is a bit unusual.

azeey (approved):
LGTM! I haven't tested locally, but I've checked the usage against other people's [examples](https://blog.bearcats.nl/perfect-sleep-function/).  I left a few minor comments.

arjo129:
This is cool stuff. Currently it is blocked by minor style issues reported by CI:
```
  + make cpplint
  /usr/share/gz/gz-cmake/codecheck/cpplint.py:55: DeprecationWarning: module 'sre_compile' is deprecated
    import sre_compile
  /github/workspace/src/SimulationRunner.cc:222:  Lines should be <= 80 characters long  [whitespace/line_length] [2]
  /github/workspace/src/SimulationRunner.cc:225:  Lines should be <= 80 characters long  [whitespace/line_length] [2]
  /github/workspace/src/SimulationRunner.cc:974:  Lines should be <= 80 characters long  [whitespace/line_length] [2]
  /github/workspace/src/SimulationRunner.cc:978:  Lines should be <= 80 characters long  [whitespace/line_length] [2]
  /github/workspace/src/SimulationRunner.hh:596:  Lines should be <= 80 characters long  [whitespace/line_length] [2]
  Total errors found: 5
```

Do you mind fixing these then I will give it a :heavy_check_mark: and merge it.

traversaro:
> @traversaro thanks for letting me know! folks need for testing it out?
> 
> I think the PR content itself is probably a little out of depth for me (although I learned about the LEAN_AND_MEAN variable which is cool!), but I can definitely build and run test it on my machine ones I have time

I think you can test new builds on conda-forge once we merge https://github.com/conda-forge/gz-sim-feedstock/pull/137 .

azeey:
@traversaro or @knmcguire Could one of you could test this locally and give it a ✅? I think it's ready to merge otherwise.

johnwason:
The conda forge package now has this patch applied.

knmcguire:
ah sorry! Yes ofcourse I see to try it out

Stupid question, but how do I get the latest by conda forge? If I look here the last update is in February: https://anaconda.org/channels/conda-forge/packages/libgz-sim/overview

Or is that not the right package

knmcguire:
This is ionic with about 75 % RTF  with an NVidia GPU RTX 3060 (laptop) on windows 11 pro:

image


And this is the latest conda package (I hope) with now close to 100% RTF:

image

This is the pixi file I used to test it: https://gist.github.com/knmcguire/c5b14909cf76cc80593c98ddebef51c6

and I ran it like this: `pixi run -e rotary gz sim shapes.sdf`

Seems like quite a significant improvement!

azeey:
Thanks a bunch @knmcguire!

azeey:

PR #3447, opened 2026-03-30
ECM Implementation with Entt

luca-della-vedova:
# 🎉 New feature

This is a complex PR that effectively removes the in-house implementation of an ECM and instead uses a custom version with Entt. AI was used in this PR (Gemini 3)

# Why

Some numbers! I added some benchmarks to the performance tests under the [each benchmark](https://github.com/luca-della-vedova/gz-sim/blob/gz-sim10/test/benchmark/each.cc) (happy to clean them up and add them to another PR if we want to keep them there regardless). The results are... Brace yourselves...
[Gist](https://gist.github.com/luca-della-vedova/75ead15c556abf322a8be8694f2931f6) with gz-sim10 Gazebo, [gist with Entt rewrite](https://gist.github.com/luca-della-vedova/866b1b2d452747247ca0b9a30e5fe0a3). Line numbers match so comparing them shouldn't be too bad.

Some highlights:

`Each` where there are 1000 entities and they a match a query, from 1ms to pretty much below benchmark tolerance.

```
# Gazebo main
EntityComponentManagerFixture/EachCache/1000/1000        0.961 ms        0.960 ms          725
# Entt
EntityComponentManagerFixture/EachCache/1000/1000        0.001 ms        0.001 ms       557758
```

Each on 1 component then getting 9 conditionally (something we do a lot) is less dramatic but still noteworthy (20x speedup):

```
# Gazebo main
ManyComponentFixture/Each1ComponentGet9More/1000          91.5 ms         91.5 ms            8
# Entt
ManyComponentFixture/Each1ComponentGet9More/1000          4.32 ms         4.32 ms          162
```

EntitiesByComponents (or EntityByComponents) that is something we do _a lot_ in our high level APIs, let's say 1000 entities, ~70x speedup:

```
# Gazebo main
EntityByComponentsFixture/EntityByComponents/1000         10.3 ms         10.3 ms           68
# Entt
EntityByComponentsFixture/EntityByComponents/1000        0.146 ms        0.146 ms         4740
```

Now of course microbenchmarks are only useful if this is an actual bottleneck in real simulations. After recommendation @azeey I tried to run the 3k_shapes with bullet featherstone to have a world with many entities but not completely dominated by physics steps.

```
gz sim -v3 3k_shapes.sdf -r --physics-engine gz-physics-bullet-featherstone-plugin
```

On _my machine_ (cloud machine with software rendering but very high CPU power), at the steady state I get about 5% RTF on main, 20% with this PR (so 4x speedup!).
[...]

PR #3488, opened 2026-04-22
Make component typeIds constexpr

### LINKED ISSUE #3608: Deep dive into Gazebo performance issues
Gazebo M roadmap item

POC: @azeey

luca-della-vedova:
# 🎉 New feature

Part of #3447 

## Summary
This PR changes registration to be fully constexpr rather than made through a function.
This is a necessity for migration to Entt since it requires type IDs to be constexpr, but is generally a good practice. Our type IDs are indeed computed through a `constexpr` function so it should be possible to have it so.
Apart from the changes that are pretty much split up from the PR (first commit), I added another commit to deprecate the typeId field and instead recommend the `TypeId()` function (if we need a virtual function) or the `TypeIdStatic()` (if we need a static / constexpr function). 
The `typeId` field being public was quite a footgun and editing at runtime would likely cause very unpredictable behavior (i.e. any part of the code that relied on its previous type ID would stop working).
This could potentially be a minor performance gain as well, since every typeId variable access will become a `constexpr` access that can be computed at compile time.

Still, if the diff from the second commit is too large I'm happy to revert it and leave the public field unchanged.

## Test it
Just through CI

## Checklist
- [ ] Added tests

Generated-by: Gemini 3

### DIFF
--- a/include/gz/sim/components/Component.hh
+++ b/include/gz/sim/components/Component.hh
@@ -200,6 +200,16 @@ namespace serializers
 
 namespace components
 {
+  /// \brief Fallback function for ADL component registration.
+  /// This returns 0 for components that are not registered using the macro.
+  /// It is a template so it has lower priority than the non-template
+  /// overloads defined by the GZ_SIM_REGISTER_COMPONENT macro.
+  template <typename T>
+  constexpr ComponentTypeId componentTypeId(T*)
+  {
+    return 0;
+  }
+
   /// \brief Convenient type to be used by components that don't wrap any data.
   /// I.e. they act as tags and their presence is enough to infer something
   /// about the entity.
@@ -349,6 +359,25 @@ namespace components
     // Documentation inherited
     public: std::unique_ptr<BaseComponent> Clone() const override;
 
+    /// \brief Returns the unique ID for the component's type.
+    public: static constexpr ComponentTypeId typeIdStatic()
+    {
+      // This call triggers Argument Dependent Lookup (ADL) to find the
+      // `componentTypeId` function defined by the GZ_SIM_REGISTER_COMPONENT
+      // macro in the component's namespace.
+      //
+      // We pass a null pointer of the type `Component*` to trigger ADL. ADL
+      // searches the namespaces of the template arguments of `Component`,
+      // which includes the namespace where the tag type (and thus the
+      // component) is defined. This removes the constraint that all
+      // components must be defined in the `gz::sim::components` namespace.
+      //
+      // Passing `Component*` also ensures that components sharing the same
+      // tag type but having different data types generate unique function
+      // overloads, preventing ODR collisions.
+      return componentTypeId((Component*)nullptr);
+    }
+
     // Documentation inherited
     public: ComponentTypeId TypeId() const override;
 
@@ -383,7 +412,7 @@ namespace components
 
     /// \brief Unique ID for this component type. This is set through the
     /// Factory registration.
-    public: inline static ComponentTypeId typeId{0};
+    public: inline static ComponentTypeId typeId = typeIdStatic();
 
     /// \brief Unique name for this component type. This is set through the
     /// Factory registration.
@@ -401,6 +430,7 @@ namespace components
   template <typename Identifier, typename Serializer>
   class Component<NoData, Identifier, Serializer> : public BaseComponent
   {
+    public: using Type = NoData;
     /// \brief Components with no data are always equal to another instance of
     /// the same type.
     /// \param[in] _component Component to compare to
@@ -418,6 +448,25 @@ namespace components
     // Documentation inherited
     public: std::unique_ptr<BaseComponent> Clone() const override;
 
+    /// \brief Returns the unique ID for the component's type.
+    public: static constexpr ComponentTypeId typeIdStatic()
+    {
+      // This call triggers Argument Dependent Lookup (ADL) to find the
+      // `componentTypeId` function defined by the GZ_SIM_REGISTER_COMPONENT
+      // macro in the component's namespace.
+      //
+      // We pass a null pointer of the type `Component*` to trigger ADL. ADL
+      // searches the namespaces of the template arguments of `Component`,
+      // which includes the namespace where the tag type (and thus the
+      // component) is defined. This removes the constraint that all
+      // components must be defined in the `gz::sim::components` namespace.
+      //
+      // Passing `Component*` also ensures that components sharing the same
+      // tag type but having different data types generate unique function
+      // overloads, preventing ODR collisions.
+      return componentTypeId((Component*)nullptr);
+    }
+
     // Documentation inherited
     public: ComponentTypeId TypeId() const override;
 
@@ -429,7 +478,9 @@ namespace components
 
     /// \brief Unique ID for this component type. This is set through the
     /// Factory registration.
-    public: inline static ComponentTypeId typeId{0};
+    public:
+    // [[deprecated("Use TypeId() or TypeIdStatic() for constexpr context. Editing this variable at runtime is not supported.")]]
+    inline static ComponentTypeId typeId = typeIdStatic();
 
     /// \brief Unique name for this component type. This is set through the
     /// Factory registration.
@@ -514,7 +565,7 @@ namespace components
   template <typename DataType, typename Identifier, typename Serializer>
   ComponentTypeId Component<DataType, Identifier, Serializer>::TypeId() const
   {
-    return typeId;
+    return typeIdStatic();
   }
 
   //////////////////////////////////////////////////
@@ -545,7 +596,7 @@ namespace components
   template <typename Identifier, typename Serializer>
   ComponentTypeId Component<NoData, Identifier, Serializer>::TypeId() const
   {
-    return typeId;
+    return typeIdStatic();
   }
 
   //////////////////////////////////////////////////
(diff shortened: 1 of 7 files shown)


luca-della-vedova:
I'm not sure what is happening with the CI, it shows as tests all passed but somehow still fails?

caguero (commented):
Should we add some of the details to the migration guide? E.g.: typeId deprecation.

include/gz/sim/components/Component.hh
```diff
@@ -383,7 +412,10 @@ namespace components
 
     /// \brief Unique ID for this component type. This is set through the
     /// Factory registration.
-    public: inline static ComponentTypeId typeId{0};
+    public:
+    [[deprecated("Use TypeId() or TypeIdStatic()."\
```
caguero: Nit: Missing space after .

caguero:
> I'm not sure what is happening with the CI, it shows as tests all passed but somehow still fails?

I do see a deprecation warning. There's still an example to update:

```
examples/plugin/command_actor/CommandActor.cc:95
```

caguero (commented):
I leave here an alternative idea:

Replace the ADL hook with a traits specialization in `gz::sim::components`. The macro emits a full specialization; `TypeIdStatic()` reads from it.

```cpp
namespace gz::sim::components
{
  template <typename T, typename = void>
  struct ComponentTraits
  {
    static_assert(sizeof(T) == 0,
      "Component not registered. Did you forget GZ_SIM_REGISTER_COMPONENT?");
  };

  template <typename DataType, typename Identifier, typename Serializer>
  class Component : public BaseComponent
  {
    public: static constexpr ComponentTypeId TypeIdStatic()
    {
      return ComponentTraits<Component>::kTypeId;
    }
  };
}

#define GZ_SIM_REGISTER_COMPONENT(_compType, _classname)              \
  namespace gz::sim::components                                       \
  {                                                                   \
    template <> struct ComponentTraits<_classname>                    \
    {                                                                 \
      static constexpr ::gz::sim::ComponentTypeId kTypeId =           \
        ::gz::common::hash64(_compType);                              \
      static constexpr const char *kName = _compType;                 \
    };                                                                \
  }                                                                   \
  /* existing Factory registration struct — unchanged */
```

The `[[deprecated]] typeId`, the `TypeIdStatic()`-based initializer, and the call-site sweep all stay as-is. This only swaps the core mechanism.

### Advantages

- **Compile-time error on "forgot to register"** — the primary template's `static_assert` fires with a named diagnostic, instead of the ADL fallback silently returning `0`.
- **Robust to macro invocation site** — ADL requires the `componentTypeId(T*)` overload to live in a namespace ADL reaches from `Component*` (usually the tag's namespace). If the user invokes the macro from global scope or the wrong namespace, ADL misses it and `TypeIdStatic()` returns 0 while the old `typeId` field stays correct — silent bifurcation. Traits always land in `gz::sim::components`, independent of where the macro is typed.
- **Immune to name shadowing** — specializations don't participate in overload resolution, so an unrelated `componentTypeId` in the TU can't hijack lookup.
- **More idiomatic** — traits specialization is the standard extension pattern (`std::hash`, `std::iterator_traits`, `fmt::formatter`). The current ADL design needs a 12-line comment to explain the `static_cast<Component*>(nullptr)` trick.
- **Natural home for future per-component metadata** (name, serializer, category, …) in one place.

### Cons

- **Macro opens `gz::sim::components`.** The ADL design's one real win is keeping the macro expansion in the user's own namespace. Traits give that up — same pattern as specializing `std::hash`, so not unusual, but strictly more boilerplate in the macro body.
- **No user-namespace free function** (which some might view as a loss if they want to customize typeId per-type from their own namespace — not a use case I see here).

azeey:
https://github.com/gazebosim/gz-sim/pull/3447 actuallystarted out with this approach, but was changed to the ADL approach in https://github.com/luca-della-vedova/gz-sim/pull/1. The main reason is to avoid breaking existing code as much as possible and allow users to define component in any namespace of their choosing. With the template specialization approach, anyone that has defined a component outside of the `gz::sim::components` namespace would have to update their code. Even though this is targetting `main` and will likely not be backported, I think we should still try to minimize code breakage when possible.

Addressing the advantages mentioned:

* Compile-time error on "forgot to register": 
  >  the primary template's static_assert fires with a named diagnostic, instead of the ADL fallback silently returning 0.
  * Yes, it would be nice to have a compile time check for this, but I'm not sure if it's worth breaking existing code. We could have a runtime check when a component is created for the first time. We could have a `static_assert` in `template <typename T> constexpr ComponentTypeId componentTypeId(T*)` to prevent unregistered components completely (haven't verified).

* Robust to macro invocation site: 
  > ADL requires the componentTypeId(T*) overload to live in a namespace ADL reaches from Component* (usually the tag's namespace). If the user invokes the macro from global scope or the wrong namespace, ADL misses it and TypeIdStatic() returns 0 while the old typeId field stays correct — silent bifurcation. Traits always land in gz::sim::components, independent of where the macro is typed.
  * As shown in src/Component_TEST.cc, the macro can be called from the global scopre or any custom namespace, so I'm not sure if this is an issue. See https://github.com/gazebosim/gz-sim/pull/3488/changes#diff-54fdc1cb3fca7db581cffd2efcd05b3d6bb703069c8b2b2c5104887495900b10R516-R520

* Immune to name shadowing:
  > specializations don't participate in overload resolution, so an unrelated componentTypeId in the TU can't hijack lookup.
  * If we are concerned about this, we can rename the function to something more obsecure. Since user will not have to define this themselves, I don't think it needs to be short/egonomic.

* More idiomatic:
  > traits specialization is the standard extension pattern (std::hash, std::iterator_traits, fmt::formatter). The current ADL design needs a 12-line comment to explain the static_cast<Component*>(nullptr) trick.
  * This is true, but I don't think it's worth breaking existing code.

* Natural home for future per-component metadata (name, serializer, category, …) in one place.
  * This is true, but I don't think it's worth breaking existing code.

include/gz/sim/components/Component.hh
```diff
@@ -383,7 +412,10 @@ namespace components
...
+        "Editing this variable at runtime is not supported.")]]
+       inline static ComponentTypeId typeId = TypeIdStatic();
```
azeey: Other than some existing tests, I doubt anyone is actually modifying `typeId`. What do you think about changing `typeId` to a constexpr instead of deprecating it. It would dramatically reduce the diff in this PR.

include/gz/sim/components/Component.hh
```diff
@@ -200,6 +200,16 @@ namespace serializers
 
 namespace components
 {
+  /// \brief Fallback function for ADL component registration.
+  /// This returns 0 for components that are not registered using the macro.
+  /// It is a template so it has lower priority than the non-template
+  /// overloads defined by the GZ_SIM_REGISTER_COMPONENT macro.
+  template <typename T>
+  constexpr ComponentTypeId componentTypeId(T*)  // NOLINT(readability/casting)
+  {
```
azeey, line 210: Consider putting a `static_assert` here to prevent forgetting to register components

include/gz/sim/components/Factory.hh
```diff
@@ -404,41 +403,56 @@ namespace components
...
+/// discover the component's unique ID via Argument Dependent Lookup (ADL).
+/// This removes the constraint that all components must be defined inside the
+/// `gz::sim::components` namespace, enabling custom components to be defined
+/// in any namespace.
+///
+/// We take a pointer to `_classname` as the argument to avoid name collisions
+/// and support distinguishing components that share the same tag type but have
+/// different data types.
+#define GZ_SIM_REGISTER_COMPONENT(_compType, _classname) \
+inline constexpr ::gz::sim::ComponentTypeId componentTypeId(_classname* ptr) \
```
azeey: As @caguero pointed out, we might have a name collision with`componentTypeId`. Consider naming it to something more obscure, maybe `gzSimFactoryComponentTypeId`?

include/gz/sim/components/Component.hh
luca-della-vedova: Ah not having to tick/tock this makes it a lot easier, done in https://github.com/gazebosim/gz-sim/pull/3488/commits/2cc516e26f12e65894579a6304b7c0b553ac84c7 and as you said reduced the diff a lot. I removed the `TypeIdStatic()` function altogether since the access to the field has the same function.

include/gz/sim/components/Component.hh
luca-della-vedova: https://github.com/gazebosim/gz-sim/pull/3488/commits/f7d9c56cddb894031b693d3e072ba24fec5e9a1e, as somewhat expected this did indeed create some compile failures in cases where we were using components without registering them. Luckily this only happened in a unit test so I fixed it in the same commit.
I considered adding a test for the new assertion behavior but it is a bit more boiler-plate-y since we want to test a compile failure. This means we must create a new .cc file with a bunch of boilerplate, build it in a separate pipeline and make sure that it doesn't compile.
I thought the extra machinery and code wasn't worth it but let me know if you want me to do it.

include/gz/sim/components/Factory.hh
luca-della-vedova: Done https://github.com/gazebosim/gz-sim/pull/3488/commits/7d1c6c6a06e49fddbaead7f0ff327698f918d4c4, in the same commit I also added the migration guide as mentioned by @caguero

azeey (approved):

PR #3503, opened 2026-05-05
Add reset test helpers and migrate reset integration tests

XINJIANGMO:
## Summary
This PR adds shared reset test utilities and migrates the existing reset integration tests to use them.

The new helper centralizes:

- world reset requests through `/world/<name>/control`
- the two-step reset propagation flow
- stepping until simulation-side conditions become true
- waiting for asynchronous transport messages
- topic message capture through a reusable `MsgReceiver`

## Details

The reset flow intentionally keeps `ConsumeResetRequest` and `ApplyWorldReset` as separate public helpers. They currently share the same one-step implementation, but represent different observable reset phases in tests:

1. consume the reset request and set internal reset state
2. apply the reset through the simulation runner

This split is mainly for auditing the reset lifecycle itself. Most plugin reset tests can use `RequestAndApplyWorldReset` directly.

Migrated tests:
- `reset.cc`
- `reset_sensors.cc`
- `reset_detachable_joint.cc`

No system plugin behavior is changed.

## Checklist
- [x] Signed all commits for DCO
- [ ] Added a screen capture or video to the PR description that demonstrates the fix (as needed)
- [ ] Added tests
- [ ] Updated documentation (as needed)
- [ ] Updated migration guide (as needed)
- [ ] Consider updating Python bindings (if the library has them)
- [ ] `codecheck` passed (See [contributing](https://gazebosim.org/docs/all/contributing#contributing-code))
- [ ] All tests passed (See [test coverage](https://gazebosim.org/docs/all/contributing#test-coverage))
- [ ] Updated Bazel files (if adding new files). Created an issue otherwise.
- [ ] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Agazebosim+archived%3Afalse+) to support the maintainers
- [ ] Was GenAI used to generate this PR? If so, make sure to add "Generated-by" to your commits. (See [this policy](https://osralliance.org/wp-content/uploads/2025/05/OSRF-Policy-on-the-Use-of-Generative-Tools-Generative-AI-in-Contributions.pdf) for more info.)

**Note to maintainers**: Remember to use **Squash-Merge** and edit the commit message to match the pull request summary while retaining `Signed-off-by` and `Generated-by` messages.

**Backports:** If this is a backport, please use **Rebase and Merge** instead.

PR #3541, opened 2026-05-11
Fix Contact state contamination on world reset

XINJIANGMO:
# 🦟 Bug fix

Fixes #

## Summary

During the reset audit, the `Contact` system failed the reset behavior test. After a world reset, contact sensor data did not reliably recover for the next episode.

## Test logic

1. Loaded the existing `contact.sdf` world.
2. Subscribed to the contact sensor topic and waited for valid contact data.
3. Executed a direct world reset with `Server::ResetAll()` to avoid transport reset flakiness.
4. Re-subscribed after reset to avoid late pre-reset transport messages.
5. Checked that contact data is published again and that the contact count / contact positions are valid after reset.

## Root cause

`Contact` created its runtime sensor objects from `EachNew<ContactSensor>`. After a world reset, the sensor entities still exist in the ECM, so they are not reported as “new” again. If the runtime sensor map is cleared or becomes stale across reset, the system cannot rebuild contact sensors from `EachNew`, and contact publishing may not recover correctly.

## Fix logic

1. Add `ISystemReset` support to `Contact`.
2. Clear the runtime `entitySensorMap` on reset.
3. Re-scan existing `ContactSensor` components during reset using a full `Each<ContactSensor>` pass.
4. Keep the normal `PreUpdate` path using `EachNew<ContactSensor>` so regular runtime behavior remains unchanged.
5. Avoid recreating duplicate `ContactSensorData` components when rebuilding after reset.

## Checklist
- [x] Added tests

### DIFF
--- a/src/systems/contact/Contact.cc
+++ b/src/systems/contact/Contact.cc
@@ -85,7 +85,10 @@ class gz::sim::systems::ContactPrivate
 {
   /// \brief Create sensors that correspond to entities in the simulation
   /// \param[in] _ecm Mutable reference to ECM.
-  public: void CreateSensors(EntityComponentManager &_ecm);
+  /// \param[in] _includeExisting True to scan existing contact sensor
+  /// components as well as newly-created ones.
+  public: void CreateSensors(EntityComponentManager &_ecm,
+              bool _includeExisting = false);
 
   /// \brief Update and publish sensor data
   /// \param[in] _ecm Immutable reference to ECM.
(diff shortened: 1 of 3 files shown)


test/integration/contact_system.cc
```diff
@@ -272,6 +305,53 @@ TEST_F(ContactSystemTest,
...
+  {
+    Subscription<msgs::Contacts> initialContacts;
+    transport::Node node;
+    initialContacts.Subscribe(node, "/test_multiple_collisions", 1);
+
+    ASSERT_TRUE(waitForContacts(initialContacts));
+  }
+
+  server.ResetAll();
+  server.Run(true, 2, false);
```
arjo129: Is there a reason we run this for 2 iterations here?  Won't `waitForContacts` run this 2000 iterations?

If there is a reason to do this we should document it.

arjo129 (changes requested):
These are looking like theyre ready for merging I just have a question regarding the test itself.

test/integration/contact_system.cc
XINJIANGMO: My original intent was to give `ResetAll()` a small settle window. I just worried about the first frame might be unstable after a reset. It can be removed.

arjo129: Thats a fair point. TBH I recommend using the ecs to get the `ContactSensorData` component instead of transport.

https://github.com/gazebosim/gz-sim/blob/90e82be8e2a1978cd5d20f60a29cc868a78bc39b/src/systems/contact/Contact.cc#L223

XINJIANGMO: make sense! I will update!

arjo129:
~~This seems to introduce a regression in the physics integration test.~~

XINJIANGMO:
> This seems to introduce a regression in the physics integration test.

~Maybe due to CI flaky? I open PR drafts for adding reset test. They also failed. Local test is fine.~

See in [gazebosim/gz-sim#3552](https://github.com/gazebosim/gz-sim/issues/3552)

src/systems/contact/Contact.cc
```diff
@@ -154,61 +157,79 @@ void ContactSensor::Publish()
 }
 
 //////////////////////////////////////////////////
-void ContactPrivate::CreateSensors(EntityComponentManager &_ecm)
+void ContactPrivate::CreateSensors(EntityComponentManager &_ecm,
+    bool _includeExisting)
 {
   GZ_PROFILE("ContactPrivate::CreateSensors");
-  _ecm.EachNew<components::ContactSensor>(
```
arjo129, line 160: One thing Im noticing is there is a recurring theme of `EachNew` being problematic for reset. While we can cobble around the current API, perhaps we should think of `EachNew` how should work without needing this hack.

arjo129 (approved):

src/systems/contact/Contact.cc
XINJIANGMO: yes! It is worth thinking a unified way to reset as a better design.

arjo129:

scpeters:
all the backports fail to build since they use a test helper added in https://github.com/gazebosim/gz-sim/pull/3503 that hasn't been backported

arjo129:
Thanks for spotting this.

PR #3556, opened 2026-05-14
Fix EnvironmentalSensorSystem state contamination on world reset

XINJIANGMO:
# 🦟 Bug fix

Fixes #

## Summary

This adds reset coverage for `EnvironmentalSensorSystem` and fixes stale environmental sensor state after reset.
## Audit logic
1. Start `environmental_sensor.sdf`.
2. Subscribe to humidity, `/wind_speed2d`, and `/wind_speed3d`.
3. First check that humidity and both wind readings are in their initial state.
4. Keep running until humidity clearly changes, so the test proves it has left the initial episode.
5. Call `Server::ResetAll()` directly.
6. Avoid timestamp assertions. After reset, only check that humidity goes back to the early low value, and that `wind2d / wind3d` still publish the expected configured values.

## Fix logic

- Add `ISystemReset` to `EnvironmentalSensorSystem`.
- On reset, clear the old sensor map and rebuild the current custom sensors.
- Mark the environmental data table for refresh after reset.
- In the next safe `PostUpdate`, wait for the current `Environment` data, send it back to the sensors, and refresh their runtime sessions.

## Checklist
- [x] Added tests


nkoenig:
Can you link to the bug that this is resolving please?

src/systems/environmental_sensor_system/EnvironmentalSensorSystem.cc
```diff
@@ -367,6 +367,67 @@ class gz::sim::EnvironmentalSensorSystemPrivate {
...
+    }
+    else
+    {
+      gzerr << "No sensor data loaded\n";
+    }
+
+    enableComponent<components::WorldLinearVelocity>(_ecm, _entity);
+
+    // Keep track of this sensor
+    this->entitySensorMap.insert(std::make_pair(_entity,
```
nkoenig: I think this is mostly fine, but using `insert` could lead to  silent failure if/when the `AddSensorEntity` function is called during the PreUpdate step and the sensor is already in the map. Using `std::unordered_map::insert` does not overwrite, so the sensor would be dropped if it's already in the map. I think you want to use [insert_or_assign](https://en.cppreference.com/cpp/container/unordered_map/insert_or_assign)

XINJIANGMO:
> Can you link to the bug that this is resolving please?

It is a process of [gsoc](https://github.com/osrf/osrf_wiki/wiki/GSoC-2026#qa-of-reset-of-core-systems). Actually, the bug is the test I added in this PR before fix.

src/systems/environmental_sensor_system/EnvironmentalSensorSystem.cc
XINJIANGMO: make sense! I will update it!

PR #3580, opened 2026-05-19
Fix VisualizeContacts for newly spawned collisions

### LINKED ISSUE #3519: VisualizeContacts plugin doesn't detect collisions for newly created meshes
## Environment
* OS Version: Ubuntu 24.04
* Source or binary build? Source, Gazebo 9.5

## Description
As noted in https://github.com/intrinsic-dev/aic/pull/533#issuecomment-4395442670
* Expected behavior: Creating new meshes after a visualize contacts plugins was enabled will show their collisions.
* Actual behavior: It doesn't.

## Steps to reproduce

1. Create a world with the Visualize Contacts GUI plugin enabled and Show Contacts.
2. Spawn some meshes.
3. Collisions between the newly spawned meshes will not be reported.
4. Removing and readding the plugin fixes the issue.

I believe this is due to the GUI pluggin sending service calls to create collision sensors for each collision mesh it finds. The plugin only does this _once_ [in initialization](https://github.com/gazebosim/gz-sim/blob/a51bef76cf8017c49c9c58e22bc49f9fec351994/src/gui/plugins/visualize_contacts/VisualizeContacts.cc#L186-L187), which means that any mesh that was added afterwards will not have its collision enabled unless the plugin is restarted and reinitialized.

There are several ways to go about this but my recommendation would be to add a service + marker component or server configuration to _globally_ enable collisions.
Then the `VisualizeContacts` plugin can make a single service call to enable all collisions, and the server will know that for every collision that gets added a contact sensor should be instantiated as well.

puneetdixit200:
## Summary

- Keep the existing initial collision scan for VisualizeContacts.
- Add an `EachNew` collision pass so collisions spawned after startup also request contact data.
- Share the collision request logic between the initial and new-entity paths.

Fixes #3519.

## Testing

- `git diff --check`
- Local build/tests were not run because this machine does not have the Gazebo development dependency stack installed.

Suggested reviewer test:

1. Start Gazebo with the VisualizeContacts GUI plugin enabled.
2. Load or create a world that has an existing contact pair and confirm contacts are visualized as before.
3. Spawn a new model with a collision element after the world is already running.
4. Cause the newly spawned collision to contact another entity.
5. Confirm the new collision's contacts are visualized without restarting Gazebo or reloading the plugin.

### DIFF
--- a/src/gui/plugins/visualize_contacts/VisualizeContacts.cc
+++ b/src/gui/plugins/visualize_contacts/VisualizeContacts.cc
@@ -61,9 +61,21 @@ inline namespace GZ_SIM_VERSION_NAMESPACE
   class VisualizeContactsPrivate
   {
     /// \brief Creates ContactSensorData for Collision components without a
-    /// Contact Sensor by requesting the /enable_contact service
+    /// Contact Sensor by requesting the /enable_collision service
     /// \param[in] Reference to the GUI Entity Component Manager
     public: void CreateCollisionData(EntityComponentManager &_ecm);
+
+    /// \brief Creates ContactSensorData for new Collision components without
+    /// a Contact Sensor by requesting the /enable_collision service
+    /// \param[in] Reference to the GUI Entity Component Manager
+    public: void CreateNewCollisionData(EntityComponentManager &_ecm);
+
+    /// \brief Request ContactSensorData creation for one collision entity.
+    /// \param[in] _entity Collision entity.
+    /// \param[in] Reference to the GUI Entity Component Manager.
+    /// \return True to continue entity iteration.
+    public: bool RequestCollisionData(
+                const Entity &_entity, EntityComponentManager &_ecm);
 
     /// \brief Transport node
     public: transport::Node node;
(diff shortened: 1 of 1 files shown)


arjo129:
Thanks for your contribution, can you use the PR template?

Also follow these instructions to fix DCO https://github.com/gazebosim/gz-sim/pull/3580/checks?check_run_id=76681756349

The testing does not seem to reflect how I would test this. Try to avoid copy pasting from AI agents without understanding the problem. Testing instructions are for me to test and review the problem at hand.

arjo129:
@puneetdixit200 please review our AI guidelines here: https://osralliance.org/wp-content/uploads/2025/05/OSRF-Policy-on-the-Use-of-Generative-Tools-Generative-AI-in-Contributions.pdf

```
Local build/tests were not run because this machine does not have the Gazebo development dependency stack installed.
```
I have very mixed feelings about this. This feels like you are wasting my time.

You should be signing commits using the `Generated-by:` tag line - not doing so suggests a degree of dishonesty.

Also you should fill the default PR template: https://github.com/gazebosim/.github/blob/main/PULL_REQUEST_TEMPLATE.md

If you did so manually instead of using whatever agentic tool you are using, you would already have known about the sign-off requirement. 

As it stands, I strongly suspect undisclosed AI use in this PR. In addition the way you have copy pasted "Suggested reviewer test" suggests to me a lack of regard for the project. If you do not rectify these issues, I will proceed to close this pull request.

PR #3590, opened 2026-05-21
Mitigate #3589: Restrict reading of resources to paths that live only within approved resource list

### LINKED ISSUE #3589: WebsocketServer 'asset' operation allows arbitrary file read
This is a security vulnerability detected using LLM assisted tools, but it's a tested legitimate attack vector. There is also no SECURITY.md to publish these findings. The recommendation is to remove `common:exists(assetUri)` and to do a proper check. First canonicalize paths using `std::filesystem::canonical()` and then verify that they are coming from a specific base folder e.g. . The issue I had was determining which was the intended base folder that it should only come from

## Environment
* OS Version: Ubuntu 24.04 (Docker), also verified via source review on gz-sim 9 main branch
* Source or binary build
  - Binary: gz-sim 8.11.0 / Harmonic
  - Source: main branch (gz-sim 9), commit verified 2026-05-20

## Description
* Expected behavior: The WebsocketServer `asset` operation should only serve simulation assets (meshes, textures, SDF resources) from Gazebo resource directories and the Fuel cache and not arbitrary files from the host filesystem (e.g. /etc/shadow, /etc/passwd, ....)

* Actual behavior: The `asset` WebSocket frame handler (`OnAsset`) accepts any filesystem path and returns the file contents to the caller with no path validation. When no `<authorization_key>` is configured (the default), this is exploitable pre-authentication by any network client that can reach port 9002.

**Root cause:** `OnAsset` in `WebsocketServer.cc` reads `assetUri` directly from the WebSocket frame (line 1127) and short-circuits into `resolvedPath = assetUri` when `common::exists(assetUri)` returns true (line 1132). Absolute paths like `/etc/passwd` pass this check and are opened with `std::ifstream` (line 1153) with no canonicalization, base-directory confinement, or allowlist. Even if the `common::exists()` short-circuit is bypassed, the `ResourcePathsResolveService` (ServerPrivate.cc) also accepts absolute paths via `common::absPath()` and `file://` prefix stripping without confinement.

## Steps to reproduce

1. Run gz-sim with the WebsocketServer plugin (no `<authorization_key>` configured):
   ```xml
   <plugin filename="gz-sim-websocket-server-system" name="gz::sim::systems::WebsocketServer">
     <port>9002</port>
   </plugin>

2. From any machine that can reach port 9002.  Run something like:
 `printf 'asset,/etc/passwd,,' | websocat -n1 ws://target:9002`

3. The full contents of /etc/passwd (or any readable file) are returned in the WebSocket response.

arjo129:
# 🦟 Bug fix

Fixes #3589

## Summary
#3589 describes a vulnerability with our websocket server code. This is due to unprotected read access. The fix applied here is to (1) only allow assets that live within the approved resource directories.  

## Checklist
- [ ] Added tests

Generated-by: Gemini-CLI


src/systems/websocket_server/WebsocketServer.cc
```diff
@@ -1149,6 +1168,43 @@ void WebsocketServer::OnAsset(int _socketId,
 
   if (!resolvedPath.empty())
   {
+    // Verify the path is inside allowed resource paths
+    std::string canonicalResolved = common::absPath(resolvedPath);
```
j-rivero: The path-allowlist is bypassable. gz::common::absPath() is just std::filesystem::absolute(), which preserves .. segments verbatim and does not follow symlinks — so with GZ_SIM_RESOURCE_PATH=/opt/models, the URI  `/opt/models/../../etc/passwd` passes the rfind check and gets served.

j-rivero (changes requested):
We probably need to patch other use cases.

src/systems/websocket_server/WebsocketServer.cc
j-rivero: A possible patch suggested by Claude. Mostly uses `std::filesystem::weakly_canonical` for dealing with .. and symlinks correctly. I did not curated it but can serve as a base for working on:

```diff
diff --git a/src/systems/websocket_server/WebsocketServer.cc b/src/systems/websocket_server/WebsocketServer.cc
index d029c8d52..0dbe23486 100644
--- a/src/systems/websocket_server/WebsocketServer.cc
+++ b/src/systems/websocket_server/WebsocketServer.cc
@@ -24,6 +24,7 @@
 #include <cstddef>
 #include <cstdio>
 #include <cstring>
+#include <filesystem>
 #include <fstream>
 #include <functional>
 #include <limits>
@@ -1168,26 +1169,44 @@ void WebsocketServer::OnAsset(int _socketId,
 
   if (!resolvedPath.empty())
   {
-    // Verify the path is inside allowed resource paths
-    std::string canonicalResolved = common::absPath(resolvedPath);
+    // Verify the path is inside allowed resource paths.
+    // common::absPath() / std::filesystem::absolute() does NOT collapse ".."
+    // segments and does NOT follow symlinks, so a prefix check on its
+    // output is bypassable (issue #3589). Use weakly_canonical, which
+    // resolves both, before comparing.
+    std::error_code ec;
+    std::filesystem::path canonicalResolvedPath =
+        std::filesystem::weakly_canonical(resolvedPath, ec);
+    std::string canonicalResolved =
+        ec ? std::string() : canonicalResolvedPath.string();
+
     bool allowed = false;
-    for (const std::string &resPath : sim::resourcePaths())
+    if (!canonicalResolved.empty())
     {
-      std::string canonicalRes = common::absPath(resPath);
-      if (canonicalRes.empty())
-        continue;
-
-      // Ensure trailing separator
-      if (canonicalRes.back() != '/' && canonicalRes.back() != '\\')
+      for (const std::string &resPath : sim::resourcePaths())
       {
-        canonicalRes = common::separator(canonicalRes);
-      }
+        std::error_code resEc;
+        std::filesystem::path canonicalResPath =
+            std::filesystem::weakly_canonical(resPath, resEc);
+        if (resEc)
+          continue;
+        std::string canonicalRes = canonicalResPath.string();
+        if (canonicalRes.empty())
+          continue;
 
-      if (canonicalResolved == common::absPath(resPath) ||
-          canonicalResolved.rfind(canonicalRes, 0) == 0)
-      {
-        allowed = true;
-        break;
+        std::string canonicalResWithSep = canonicalRes;
+        if (canonicalResWithSep.back() != '/' &&
+            canonicalResWithSep.back() != '\\')
+        {
+          canonicalResWithSep = common::separator(canonicalResWithSep);
+        }
+
+        if (canonicalResolved == canonicalRes ||
+            canonicalResolved.rfind(canonicalResWithSep, 0) == 0)
+        {
+          allowed = true;
+          break;
+        }
       }
     }
```

src/systems/websocket_server/WebsocketServer.cc
```diff
@@ -502,10 +504,27 @@ void WebsocketServer::Configure(const Entity & /*_entity*/,
   // We will handle logging
   lws_set_log_level( 0, lwsl_emit_syslog);
 
+  std::string address = "127.0.0.1";
+  if (sdf::ElementConstPtr addressElem = _sdf->FindElement("address"))
+  {
+    address = addressElem->Get<std::string>();
```
j-rivero: Not doing any sanitize action on the ->Get for the input of garbage. Refuse to start on garbage.

j-rivero:
Don't have too much experience on what the websocket needs to resolve to work but, are we sure that  `sim::resourcePaths()`  is enough for websocket to handle all possible assets to load? My agent suggest that it resolves only `GZ_SIM_RESOURCE_PATH`. I'm thinking about the fuel cache in .gz/fuel and other env variables that we can use to load stuff: plugins, media, etc.

src/systems/websocket_server/WebsocketServer.cc
arjo129: Thanks for spotting this. Fixed in [e042be](https://github.com/gazebosim/gz-sim/pull/3590/commits/7e042be496f398a3eae5b3364652f15503298cea)

src/systems/websocket_server/WebsocketServer.cc
arjo129: 02f5787

src/systems/websocket_server/WebsocketServer.cc
```diff
@@ -1149,6 +1194,47 @@ void WebsocketServer::OnAsset(int _socketId,
...
+        canonicalRes = common::separator(canonicalRes);
+      }
+
+      if (canonicalResolved == canonicalResNoSep ||
+          canonicalResolved.rfind(canonicalRes, 0) == 0)
+      {
+        allowed = true;
+        break;
+      }
+    }
```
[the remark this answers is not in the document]
arjo129: [b1b416a](https://github.com/gazebosim/gz-sim/pull/3590/commits/b1b416a0a7755cecedc1ecc6b236124f3e668581)

src/systems/websocket_server/WebsocketServer.cc
```diff
@@ -1149,6 +1194,47 @@ void WebsocketServer::OnAsset(int _socketId,
...
+      if (canonicalRes.back() != '/' && canonicalRes.back() != '\\')
+      {
+        canonicalRes = common::separator(canonicalRes);
+      }
+
+      if (canonicalResolved == canonicalResNoSep ||
+          canonicalResolved.rfind(canonicalRes, 0) == 0)
+      {
+        allowed = true;
+        break;
```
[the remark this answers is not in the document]
arjo129: [e593cfb](https://github.com/gazebosim/gz-sim/pull/3590/commits/e593cfb3023f5694c3d5484d6667a247c80e4e45)

src/systems/websocket_server/WebsocketServer.cc
```diff
@@ -502,10 +507,52 @@ void WebsocketServer::Configure(const Entity & /*_entity*/,
...
+    address = addressElem->Get<std::string>();
+  }
+  else if (sdf::ElementConstPtr ifaceElem = _sdf->FindElement("iface"))
+  {
+    address = ifaceElem->Get<std::string>();
+  }
+
+  // Sanitize the address to prevent potential security/injection issues
+  bool validAddress = true;
+  if (address.length() > 255)
```
j-rivero:
```suggestion
  if (address.empty() || address.length() > 255) validAddress = false;
```

The current setup fails for an empty address defined in <address></address> since it does not fall into false and the loop never starts, returning a validAddress = true.

src/systems/websocket_server/WebsocketServer.cc
```diff
@@ -502,10 +507,52 @@ void WebsocketServer::Configure(const Entity & /*_entity*/,
...
+      {
+        validAddress = false;
+        break;
+      }
+    }
+  }
+
+  if (!validAddress)
+  {
+    gzwarn << "Invalid characters or excessive length in address/iface ["
```
j-rivero: I would prefer it we fail hard here after a problematic user setup instead of defaulting to localhost.

src/systems/websocket_server/WebsocketServer.cc
```diff
@@ -502,10 +507,52 @@ void WebsocketServer::Configure(const Entity & /*_entity*/,
...
   memset(&info, 0, sizeof info);
   info.port = port;
-  info.iface = nullptr;
+  if (address == "0.0.0.0" || address == "all" || address == "*")
+  {
+    info.iface = nullptr;
+  }
+  else
+  {
+    info.iface = address.c_str();
```
j-rivero: My agent discovered a problem here with the lifetime of the address:

"""
Address is a local std::string in this Configure() method. 

libwebsockets stores info.iface by pointer on the vhost (vh->iface = info->iface; in [lws_create_vhost](https://libwebsockets.org/lws-api-doc-v2.1-stable/html/group__context-and-vhost.html#ga0c54c667ccd9b8b3dddcd123ca72f87c)) and references it later (e.g. retry binds, lws_get_vhost_iface, vhost teardown). 

After Configure() returns the buffer is destroyed and the server thread (spawned at line 604) runs against a dangling iface.
"""

Not sure if we want to fix this here since it also affects to sslCertFile / sslPrivateKeyFile. We probably need to wider the scope of the variable, probably to be a member class that goes further than the  lws_context.

src/systems/websocket_server/WebsocketServer.cc
```diff
@@ -1149,6 +1196,65 @@ void WebsocketServer::OnAsset(int _socketId,
 
   if (!resolvedPath.empty())
   {
+    std::error_code ec;
+    std::string canonicalResolved =
+        std::filesystem::weakly_canonical(resolvedPath, ec).string();
```
j-rivero, line 1200: weakly_canonical is not exception-safe so if we don't try / catch here the propagation of the exception can drop the whole thread and we change a security problem by a DoS :)  Thay may throw std::bad_alloc or filesystem_error, probably a "try { ... } catch (const std::exception &e) { ... " is enough.

src/systems/websocket_server/WebsocketServer.cc
```diff
@@ -1149,6 +1196,65 @@ void WebsocketServer::OnAsset(int _socketId,
...
+    bool allowed = false;
+
+    if (!ec)
```
j-rivero, line 1202: Are we here computing ClientConfig and resourcePaths() per every request? Probably better to move them to Configure and run them once.

src/systems/websocket_server/WebsocketServer.cc
```diff
@@ -1149,6 +1196,65 @@ void WebsocketServer::OnAsset(int _socketId,
...
+      {
+        allowedPaths.push_back(fuelCachePath);
+      }
+
+      for (const std::string &resPath : allowedPaths)
+      {
+        std::error_code pathEc;
+        std::string canonicalRes =
+            std::filesystem::weakly_canonical(resPath, pathEc).string();
+        if (pathEc || canonicalRes.empty())
```
j-rivero: should we emit a warning on skipped paths so the user has a sense of what is being omitted?

j-rivero (changes requested):

src/systems/websocket_server/WebsocketServer.cc
arjo129: ResourcePaths can be modified at run-time using this service: https://github.com/gazebosim/gz-sim/blob/839e621c852df43b8092d103f304779b47b91b61/src/ServerPrivate.cc#L494

Hence the need to recompute it every time. It could be possible to cache these but Im not sure its really worth adding complexity.

src/systems/websocket_server/WebsocketServer.cc
arjo129: What would that look like? Throwing an exception and crashing everything or letting the sim run with a gz-err but disabling this plugin?

j-rivero: That is a good point, I think that traditionally Gazebo has used the approach of "keep it running at all cost" where I do usually prefer "fail to start" under errors and warnings that are clearly not what the user expect. In this case of a plugin, could make sense to disable the plugin, we probably don't have an example of a plugin crashing the whole simulator. I'll bring it to the PMC for feedback.

src/systems/websocket_server/WebsocketServer.cc
j-rivero: Thanks for clarifying.

src/systems/websocket_server/WebsocketServer.cc
```diff
@@ -1149,6 +1196,87 @@ void WebsocketServer::OnAsset(int _socketId,
...
+      std::string data = BUILD_MSG(this->operations[ASSET], assetUri,
+          msg.GetTypeName(), msg.SerializeAsString());
+
+      this->QueueMessage(this->connections[_socketId].get(),
+          data.c_str(), data.length());
+      return;
+    }
+
     // Read the file
     std::ifstream infile(resolvedPath, std::ios_base::binary);
```
j-rivero:
```suggestion
    std::ifstream infile(canonicalResolved, std::ios_base::binary);
```

I think it is now critical to use the sanitized path.

j-rivero (changes requested):
One critical change I think and we are good to go..

j-rivero:
> @arjo129
[Address](https://github.com/gazebosim/gz-sim/pull/3590/commits/01be00195778f8227abe4160d0cd81d4f1f7e638) @jrivero['s comment](https://github.com/gazebosim/gz-sim/pull/3590/commits/01be00195778f8227abe4160d0cd81d4f1f7e638)

Let me fix my suggestion that was broken and add another check.

src/systems/websocket_server/WebsocketServer.cc
```diff
@@ -1147,37 +1207,126 @@ void WebsocketServer::OnAsset(int _socketId,
...
+    gzwarn << "Resolved Path is empty" << "\n";
+    return;
+  }
 
-    // Store the file in a protobuf message
-    gz::msgs::Bytes bytes;
-    bytes.set_data(fileBuffer);
+  bool allowed = false;
+  std::string canonicalResolved;
+  try
```
azeey, line 1196: SInce we're using `std::error_code`, do we need to catch exceptions?

azeey (commented):
It looks like we need further discussion on error handling. I think it would be better to have a narrowly scoped PR to fix the resource path problem, which can be backported, and then create another PR for network address related changes.

src/systems/websocket_server/WebsocketServer.cc
j-rivero: Good point. The `std::error_code` overload removes the filesystem_error exceptoon, which was half the original reason for the try/catch. The error_code overloads aren't noexcept and can still throw `std::bad_alloc`, and the surrounding .string()/vector/string work allocates too. Since an exception will end up in killing the plugin, there is a space for a potential DoS

j-rivero:
Refactor looks good so far. A couple of things I'm investigating with my agents:

1. Testing:
I asked the agent to build a white-box test that covers all the input options that were problematic here sanitizing the entries, it is in https://github.com/gazebosim/gz-sim/commit/2562ec1e536b9d1699d6472424f4187d811f8701 . Not sure if we want to add it but worked as far as I can tell.

2. Regression on new blocked but valid paths:
Asked the agent to track if we are blocked previously valid paths that are needed for the simulation. Relevant code is in [Util.cc addResourcePaths](https://github.com/gazebosim/gz-sim/blob/main/src/Util.cc#L521-L588) and the problematic bit affecting here is that we do not have a getter for that information. We are covering here GZ_SIM_RESOURCE_PATH + Fuel but seems to be also SDF_PATH and GZ_FILE_PATH. Right solution is to refactor that Util code to provide a getter on the paths, not sure if we want to go that route in this PR.

arjo129:
Thanks @j-rivero on iterating on this with me.

1. I think such a test is valuable to prevent future regressions
2. Reading through the [docs](https://gazebosim.org/api/sim/8/resources.html), it seems we use the term "may" work with SDF_PATH and GZ_FILE_PATH, but when we implement https://github.com/gazebosim/gz-sim/blob/7d1f3dc3a966aabe4ea203f4ef78a654294de55b/src/Util.cc#L507 we only read from `GZ_RESOURCE_PATHS`. I dont see us using these paths within sim code base outside of tests. It'd be good to clarify what their role really is.

Also I'll be opening a ticket regarding the demo as the javascript code of Gz3d.js that we package and ship in our demo is not working on main (the clock is being transimited correctly, but the scene is not loading thanks to some javacript errors).

That being said we should probably merge this specific PR. 

We will also need to backport it  to jetty. We should also check if `gz-launch` is effected on ionic. harmonic and fortress.

azeey:
`SDF_PATH` and `GZ_FILE_PATH` are environment variables consumed by libsdformat and gz-common. Even though `resourcePaths` doesn't return them, the underlying libraries will use them when resolving resource (e.g. mesh files). I have actually never used `GZ_FILE_PATH`, but I'm familiar with `SDF_PATH`. In the docs, we recommend against using these variables and only rely on `GZ_SIM_RESOURCE_PATH`. However, the change in this PR will indeed break those workflows. I think we can discuss whether we want websocket system to only work with `GZ_SIM_RESOURCE_PATH` going forward, but the safest thing for now is to continue supporting `GZ_FILE_PATH` and `SDF_PATH`.

arjo129:
> 2\. GZ_FILE_PATH

Thanks for the clarification. Ive added support to `SDF_PATH` and `GZ_FILE_PATH`  it in d11158b ! I've also cherry-picked @j-rivero's test to prevent further regressions.

src/systems/websocket_server/WebsocketServer.cc
```diff
@@ -1125,9 +1131,39 @@ void WebsocketServer::OnAsset(int _socketId,
...
+    -> std::vector<std::string>
+  {
+    std::vector<std::string> paths;
+    const char *envVal = std::getenv(_envVar.c_str());
+    if (envVal)
+    {
+      std::string envStr(envVal);
+      std::stringstream ss(envStr);
+      std::string path;
+      while (std::getline(ss, path, ':'))
```
j-rivero: Maybe using common::SystemPaths::Delimiter() instead of hardcoding : ?

j-rivero (commented):
Mostly fine, a single comment con delimiter

src/systems/websocket_server/WebsocketServer.cc
arjo129: [594fa46](https://github.com/gazebosim/gz-sim/pull/3590/commits/594fa46e5ee5c155eeb381d8765af881a9a6f500)

j-rivero (approved):

azeey (approved):

arjo129:
@mergifyio backport gz-sim10

PR #3592, opened 2026-05-22
Update some tests to be independent of entity ordering

luca-della-vedova:
# 🎉 New feature

Span out of #3447 

## Summary
Update the most straightforward tests to make them independent of order of entity in Each.
Both tests check that a query contains all the entities that should be present (i.e. a world has all the required models, model all the required links, link all the required collisions / visuals). However, it expects them to be in a specific order that is more of an implementation detail.
This PR, span out of the entt refactor, updates the tests so they still validate that all the entities are present but don't test anymore that they are iterated in a specific order.

## Test it
Tested by CI, or run the affected tests manually

## Checklist
- [x] Signed all commits for DCO
- [ ] Added a screen capture or video to the PR description that demonstrates the feature
- [ ] Added tests
- [ ] Added example and/or tutorial
- [ ] Updated documentation (as needed)
- [ ] Updated migration guide (as needed)
- [ ] Consider updating Python bindings (if the library has them)
- [ ] `codecheck` passed (See [contributing](https://gazebosim.org/docs/all/contributing#contributing-code))
- [ ] All tests passed (See [test coverage](https://gazebosim.org/docs/all/contributing#test-coverage))
- [ ] Updated Bazel files (if adding new files). Created an issue otherwise.
- [ ] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Agazebosim+archived%3Afalse+) to support the maintainers
- [ ] Was GenAI used to generate this PR? If so, make sure to add "Generated-by" to your commits. (See [this policy](https://osralliance.org/wp-content/uploads/2025/05/OSRF-Policy-on-the-Use-of-Generative-Tools-Generative-AI-in-Contributions.pdf) for more info.)

**Note to maintainers**: Remember to use **Squash-Merge** and edit the commit message to match the pull request summary while retaining `Signed-off-by` and `Generated-by` messages.

PR #3624, opened 2026-05-28
Make systems work independently of entity order

luca-della-vedova:
# 🎉 New feature

Part of #3608 

Based on #3592 (will rebase once it is merged), part of #3447. Update _systems_ to be robust to unordered iterations. Contains https://github.com/gazebosim/gz-sim/commit/9a0f0e298076059322911cbf83c10323cbcb7f15 to test in CI (that makes all the view unordered) that should be reverted first. We can discuss afterwards if we want to have that behavioral change in a separate PR to test it on rotary for a while, or just have it as part of the Entt refactor later on.

## Summary

An important aspect is that this PR relies on the fact that entity IDs are strictly monotonic and not reused, something that is also enforced in the Entt rewrite but we should consider revisiting in the future.
Key list of changes:

* `RenderUtil` now sorts the entities before spawning them, this pretty much preserves the previous behavior that relied on iteration order that spawned parent models before their nested ones.
* `Physics` has a similar change, where _only for models_ we push all of them into a `std::set<Entity>` first, then iterate through the set to spawn them.
* `SdfGenerator` as well, here we push into a `vector` then sort it since we use `EntitiesByComponents` already. The sorting + no entity recycling keeps sdf generation deterministic and tests / diffs predictable.
* `ModelCommandAPI` just sort all the results of `EntitiesByComponents`, again for test and API determinism.
* `LevelManager` is a lot more interesting... It seems the log was generated in a previous version of Gazebo when there was no `Name` component attached to default levels, specifically https://github.com/gazebosim/gz-sim/commit/517a73adf9f85111c3d1d851c7e00a8af624ae98 from 2024 introduced the `Name` component, but the binary logfile in https://github.com/gazebosim/gz-sim/tree/main/test/media/rolling_shapes_log was generated 4 years ago, hence accessing that component was invalid but for some reason didn't cause any issue until now. I couldn't really figure out how it was working before and I could have updated the log file but I thought if there is anyone that is relying on this behavior and playing back old logs we should avoid breaking them.

## Test it

Testing in CI, but running any sort of world should still work

## Checklist
- [ ] Added tests

Generated-by: Gemini 3


caguero (commented):
Switching to `std::unordered_set` removes the ascending-entity-id iteration order, but SceneBroadcaster (not touched in this PR) relies on it when building the scene graph in EachNew:

```
  // SceneBroadcaster.cc:817
  newGraph.AddVertex(_nameComp->Data(), modelMsg, _entity);
  newGraph.AddEdge({_parentComp->Data(), _entity}, true);
```

 `gz::math::graph::AddEdge` returns `NullEdge` if any vertex isn't present yet . With `std::set`, a parent model (lower id) was always added before its nested child in the same batch. With `unordered_set`, the child can be processed first. The later merge into the persistent sceneGraph only iterates newGraph.Edges(), so the dropped edge is never recovered; the nested model stays permanently disconnected, and late-joining GUI clients never render it.

Suggest sorting the model EachNew results by entity id here too (same fix applied in RenderUtil), or otherwise guaranteeing parents are added before children.

src/rendering/RenderUtil.cc
```diff
@@ -1166,6 +1167,19 @@ void RenderUtil::Update()
...
+  };
+  std::sort(newModels.begin(), newModels.end(), sortById);
+  std::sort(newActors.begin(), newActors.end(), sortById);
+  std::sort(newLinks.begin(), newLinks.end(), sortById);
+  std::sort(newVisuals.begin(), newVisuals.end(), sortById);
+  std::sort(newLights.begin(), newLights.end(), sortById);
+  std::sort(newParticleEmitters.begin(), newParticleEmitters.end(), sortById);
+  std::sort(newProjectors.begin(), newProjectors.end(), sortById);
+  std::sort(newSensors.begin(), newSensors.end(), sortById);
+
```
caguero: What about `newParticleEmittersCmds`, `newInertiaLinks, newCOMLinks`, `newWireframeVisualLinks`, `newCollisionLinks`, `newJointModels`, `newTransparentVisualLinks`?

src/systems/physics/Physics.cc
```diff
@@ -1244,160 +1244,170 @@ void PhysicsPrivate::CreateWorldEntities(const EntityComponentManager &_ecm,
...
+
+  for (const auto &_entity : entities)
+  {
+    const auto *_name = _ecm.Component<components::Name>(_entity);
+    const auto *_pose = _ecm.Component<components::Pose>(_entity);
+    const auto *_parent = _ecm.Component<components::ParentEntity>(_entity);
 
-        // Check if model already exists
-        if (this->entityModelMap.HasEntity(_entity))
+    if (_ecm.EntityHasComponentType(_entity, components::Recreate::typeId))
```
caguero: Can this happen according to line 1256?

src/systems/physics/Physics.cc
```diff
@@ -1244,160 +1244,170 @@ void PhysicsPrivate::CreateWorldEntities(const EntityComponentManager &_ecm,
...
-        if (_ecm.EntityHasComponentType(_entity, components::Recreate::typeId))
-          return true;
+        if (!_ecm.EntityHasComponentType(_entity, components::Recreate::typeId))
+          entities.insert(_entity);
+        return true;
+      });
+
+  for (const auto &_entity : entities)
+  {
+    const auto *_name = _ecm.Component<components::Name>(_entity);
```
caguero: I looks we're not leveraging that this information was also available in `EachNew`. Perhaps we could store it in a `map` and read it here or something similar.

src/SdfGenerator.cc
```diff
@@ -327,112 +328,127 @@ namespace sdf_generator
...
+          }
+          return true;
+        });
+    std::sort(modelEntities.begin(), modelEntities.end());
 
-          auto modelDir =
-              common::parentPath(_modelSdf->Data().Element()->FilePath());
+    for (const auto &modelEntity : modelEntities)
+    {
+      const auto *modelSdf = _ecm.Component<components::ModelSdf>(modelEntity);
```
caguero: Same here, it looks like we're missing the opportunity to use `_modelSdf` from the `Each` lambda?

src/SdfGenerator.cc
```diff
@@ -327,112 +328,127 @@ namespace sdf_generator
...
               _modelEntity);
-          if (parentComp && parentComp->Data() != _entity)
-            return true;
+          if (parentComp && parentComp->Data() == _entity)
+          {
+            modelEntities.push_back(_modelEntity);
+          }
+          return true;
+        });
+    std::sort(modelEntities.begin(), modelEntities.end());
```
caguero: I wonder if it would be more efficient to use `_ecm.ChildrenByComponents()`. This way, we should go over the inmediate child models of the world, as opposed to iterate over all model entities in the world. Something like:

```
  // Immediate child models of the world, via the entity graph (O(direct children))
  auto modelEntities = _ecm.ChildrenByComponents(_entity, components::Model());
  std::sort(modelEntities.begin(), modelEntities.end());

  for (const auto &modelEntity : modelEntities)
  {
    const auto *modelSdf = _ecm.Component<components::ModelSdf>(modelEntity);
    if (!modelSdf)            // preserve the old "must have ModelSdf" filter
      continue;
    ...
  }

```

caguero (commented):
I just made some comments, the more important thing to tackle should be `SceneBroadcaster`.

src/SdfGenerator.cc
luca-della-vedova: `ChildrenByComponents` is a funny one, _generally_ it should only be used when users expect the entity to have very few children, since it iterates over the much non optimized graph list of children, plus it needs to do a logn set lookup for every entity, so overall is very slow.
I ran the benchmark on this branch to show it:

```
lucadv@noble:~/ws_jetty_upstream$ build/gz-sim/bin/BENCHMARK_each --benchmark_filter="EntityByComponentsFixture/(ChildrenByComponents|EntitiesByComponents)"
----------------------------------------------------------------------------------------------
Benchmark                                                    Time             CPU   Iterations
----------------------------------------------------------------------------------------------
EntityByComponentsFixture/EntitiesByComponents/10        0.014 ms        0.014 ms        49997
EntityByComponentsFixture/EntitiesByComponents/100       0.065 ms        0.065 ms        10760
EntityByComponentsFixture/EntitiesByComponents/1000      0.604 ms        0.604 ms         1189
EntityByComponentsFixture/ChildrenByComponents/10        0.073 ms        0.073 ms         9563
EntityByComponentsFixture/ChildrenByComponents/100       0.790 ms        0.789 ms          887
EntityByComponentsFixture/ChildrenByComponents/1000       12.7 ms         12.7 ms           53
```

So for non trivial entity counts it is at least an order of magnitude slower than `EntitiesByComponents`. Hence it can be a good idea to use it, for example, to get the list of links from a particular model (since there might be a lot of links overall in the world but not so many links in a specific model), but on the other hand it shouldn't be used with entities that have a lot of children, the `World` is one of them.
It could theoretically make things faster if you had a world where there are many deeply nested models, where each model has at least >10 nested models in it, but I doubt that really happens much.

src/systems/physics/Physics.cc
luca-della-vedova: Indeed not really! Removed https://github.com/gazebosim/gz-sim/pull/3624/commits/9ee033014f8f468bfad70c968594149b03135e58

src/systems/physics/Physics.cc
luca-della-vedova: Good call out, component lookups are somewhat expensive, https://github.com/gazebosim/gz-sim/pull/3624/commits/3f921e2eeece99d6927e67b8a615e1c1c24416a8 and https://github.com/gazebosim/gz-sim/pull/3624/commits/8ea1d36f2f3fdf3b9066f0f4ccdc7826a694d63b (both this and the comment for SdfGenerator)

src/rendering/RenderUtil.cc
luca-della-vedova: It's been long enough that I forgot why I did it that way, from what I understand what _really_ matters is nested model construction order, so just simplify it to only sort models https://github.com/gazebosim/gz-sim/pull/3624/commits/ce1a565ed94d2c98bd4851038d30846cfb87d4e1

luca-della-vedova:
> Switching to `std::unordered_set` removes the ascending-entity-id iteration order, but SceneBroadcaster (not touched in this PR) relies on it when building the scene graph in EachNew:
> 
> ```
>   // SceneBroadcaster.cc:817
>   newGraph.AddVertex(_nameComp->Data(), modelMsg, _entity);
>   newGraph.AddEdge({_parentComp->Data(), _entity}, true);
> ```
> 
> `gz::math::graph::AddEdge` returns `NullEdge` if any vertex isn't present yet . With `std::set`, a parent model (lower id) was always added before its nested child in the same batch. With `unordered_set`, the child can be processed first. The later merge into the persistent sceneGraph only iterates newGraph.Edges(), so the dropped edge is never recovered; the nested model stays permanently disconnected, and late-joining GUI clients never render it.
> 
> Suggest sorting the model EachNew results by entity id here too (same fix applied in RenderUtil), or otherwise guaranteeing parents are added before children.

Ah makes sense, I completely missed that plugin, did it in https://github.com/gazebosim/gz-sim/pull/3624/commits/6b49d34b98bc4dbefa2cfe581f3d950c59f1737c (plus changed it to cache to components in a map in https://github.com/gazebosim/gz-sim/pull/3624/commits/8ea1d36f2f3fdf3b9066f0f4ccdc7826a694d63b)

src/rendering/RenderUtil.cc
```diff
@@ -1166,6 +1167,7 @@ void RenderUtil::Update()
   }
   this->dataPtr->updateMutex.unlock();
 
+
```
caguero: Nit: Remove extra line.

src/SdfGenerator.cc
```diff
@@ -327,112 +329,125 @@ namespace sdf_generator
...
+          if (parentComp && parentComp->Data() == _entity)
+          {
+            modelEntities.insert({_modelEntity, _modelSdf});
+          }
+          return true;
+        });
 
-          auto modelDir =
-              common::parentPath(_modelSdf->Data().Element()->FilePath());
+    for (const auto [modelEntity, modelSdf] : modelEntities)
```
caguero: Should we use `const auto &`?

caguero (approved):
Just a reminder to remove e91d02d , but looks good to me!

src/SdfGenerator.cc
luca-della-vedova: I went back and forth between "are they trivial types, are they not trivial types" longer than I should have! Done in https://github.com/gazebosim/gz-sim/pull/3624/commits/f354564437f514c541e3cd19679982ae59b72cc5

luca-della-vedova:
I was a bit too triggered by the randomness in the EntityTree GUI plugin (even though it's not caught in CI) so I pushed https://github.com/gazebosim/gz-sim/pull/3624/commits/871c90e09ca96cb46ceddce07b48c9e18bc53761, now the order of elements in the GUI at startup will be sorted.

Before:

Screenshot From 2026-07-01 13-55-02

After (preserve current behavior):

image

PR #3638, opened 2026-05-31
Fix DopplerVelocityLogSystem state contamination on world reset

XINJIANGMO:
# 🦟 Bug fix

Fixes #<NUMBER>

## Summary
Make `DopplerVelocityLogSystem` survive a world reset, and add tests that cover it.

Before this PR, after `ResetAll()` the DVL just stopped publishing usable velocity messages — the sensor objects were still attached to the previous episode's entities and parent visuals, and the system kept a pile of stale per-step requests around. This fixes that and adds tests for both bottom tracking and water-mass tracking.

## Audit Logic

1. Load a DVL world (`dvl_reset.sdf` for bottom tracking, `underwater_currents.sdf` for water-mass tracking).
2. Subscribe to `/dvl/velocity` and wait until the sensor locks to its expected at-rest / steady-state output.
3. Push the vehicle around with some linear and angular velocity so the output clearly leaves that state — this proves there is real runtime state to corrupt.
4. Stop driving it and call `Server::ResetAll()`.
5. Re-subscribe to `/dvl/velocity` so we don't read leftover messages from the previous episode.
6. After reset, just check the 4 beams re-lock and the output is back where it started. No timestamp checks.

## Fix Logic

* Add `ISystemReset` to `DopplerVelocityLogSystem`.
* On reset, clear all the stale per-step requests, the known-sensor set, the pending flags and `nextUpdateTime`, and resync `simTime` from `UpdateInfo`.
* Set a `resetSensorRebindPending` flag so the next `PreUpdate` walks `_ecm.Each<CustomSensor, ParentEntity>` (not `EachNew`) and re-queues every existing DVL entity as a `ResetSensor` request.
* Handle `ResetSensor` by **rebinding** the existing sensor instead of recreating it: match it by entity id or sensor name, refresh entity / parent / next update time, and re-parent its rendering sensors onto the current parent visual. Fall back to `CreateSensor` if nothing matches.
* Null-check `sensorManager.Sensor(id)` and `scene` in the few places that used to dereference them blindly, and don't publish in `DoPostUpdate` until the world state actually contains every known sensor.
* Add `test/worlds/dvl_reset.sdf` — a flat seabed, at-rest world with an analytically known beam range. The existing worlds either have currents or moving vehicles, so they can't be used as the "back to initial state" reference after reset.

## Checklist
- [x] Added tests

Generated-by: codex 5.5


test/integration/dvl_system_bottom_tracking.cc
```diff
@@ -174,3 +240,63 @@ TEST(DVLTest, GZ_UTILS_TEST_DISABLED_ON_MAC(BottomTracking))
...
+      "worlds", "dvl_reset.sdf");
+  TestFixtureWithModel fixture(worldFile, "dvl_model");
+
+  constexpr double seaBedDepth{20.};
+  const double expectedBeamRange =
+      (seaBedDepth + sensorPositionInSFMFrame.Z()) /
+      std::cos(beamInclination);
+
+  Subscription<DVLVelocityTracking> velocitySubscription;
+  transport::Node node;
```
caguero, line 257: The node should outlive the Subscription. We should move this declaration before `velocitySubscription`.

test/integration/dvl_system_bottom_tracking.cc
```diff
@@ -174,3 +240,63 @@ TEST(DVLTest, GZ_UTILS_TEST_DISABLED_ON_MAC(BottomTracking))
...
+        velocitySubscription.Last().velocity().mean()).Length() > 0.1;
+  }));
+
+  // Stop injecting velocity commands so the reset episode can settle at rest.
+  fixture.Manipulator().SetLinearVelocity(math::Vector3d::Zero);
+  fixture.Manipulator().SetAngularVelocity(math::Vector3d::Zero);
+  fixture.Simulator()->ResetAll();
+
+  // Ignore delayed messages from the previous episode.
+  Subscription<DVLVelocityTracking> postResetVelocitySubscription;
```
caguero: I think we could replace this block of lines 290-292 with `velocitySubscription.Clear()`

test/integration/dvl_system_bottom_tracking.cc
```diff
@@ -174,3 +240,63 @@ TEST(DVLTest, GZ_UTILS_TEST_DISABLED_ON_MAC(BottomTracking))
...
+  transport::Node postResetNode;
+  postResetVelocitySubscription.Subscribe(postResetNode, "/dvl/velocity", 1);
+
+  ASSERT_TRUE(fixture.StepUntil(2000, [&]
+  {
+    return postResetVelocitySubscription.Count() > 0u &&
```
caguero: This could be:
```
return velocitySubscription.Count() > 0u &&
     bottomTrackingAtRest(velocitySubscription.Last(), expectedBeamRange);
```

test/integration/dvl_system_bottom_tracking.cc
```diff
@@ -174,3 +240,63 @@ TEST(DVLTest, GZ_UTILS_TEST_DISABLED_ON_MAC(BottomTracking))
...
+        bottomTrackingAtRest(
+            postResetVelocitySubscription.Last(), expectedBeamRange);
+  }));
+  expectBottomTrackingAtRest(
```
caguero: And this could be:
```
expectBottomTrackingAtRest(velocitySubscription.Last(), expectedBeamRange);
```

test/integration/dvl_system_water_mass_tracking.cc
```diff
@@ -195,3 +216,60 @@ TEST(DVLTest, GZ_UTILS_TEST_DISABLED_ON_MAC(WaterMassTracking))
...
+//////////////////////////////////////////////////
+TEST(DVLTest, GZ_UTILS_TEST_DISABLED_ON_MAC(ResetRestoresWaterMassTracking))
+{
+  const std::string worldFile = common::joinPaths(
+      std::string(PROJECT_SOURCE_PATH), "test",
+      "worlds", "underwater_currents.sdf");
+  TestFixtureWithModel fixture(worldFile, "tethys");
+
+  Subscription<DVLVelocityTracking> velocitySubscription;
+  transport::Node node;
```
caguero, line 228: Same here, I'd move node before `velocitySubscription`.

test/integration/dvl_system_water_mass_tracking.cc
```diff
@@ -195,3 +216,60 @@ TEST(DVLTest, GZ_UTILS_TEST_DISABLED_ON_MAC(WaterMassTracking))
...
+    const math::Vector3d velocity =
+        msgs::Convert(velocitySubscription.Last().velocity().mean());
+    return (velocity - initialVelocity).Length() > 0.3;
+  }));
+
+  fixture.Manipulator().SetLinearVelocity(math::Vector3d::Zero);
+  fixture.Simulator()->ResetAll();
+
+  // Ignore delayed messages from the previous episode.
+  Subscription<DVLVelocityTracking> postResetVelocitySubscription;
```
caguero: These three lines could be:
```
velocitySubscription.Clear();
```

test/integration/dvl_system_water_mass_tracking.cc
```diff
@@ -195,3 +216,60 @@ TEST(DVLTest, GZ_UTILS_TEST_DISABLED_ON_MAC(WaterMassTracking))
...
+  transport::Node postResetNode;
+  postResetVelocitySubscription.Subscribe(postResetNode, "/dvl/velocity", 1);
+
+  ASSERT_TRUE(fixture.StepUntil(2000, [&]
+  {
+    if (postResetVelocitySubscription.Count() == 0u ||
```
caguero: How about:
```
if (velocitySubscription.Count() == 0u ||
     !validWaterMassTracking(velocitySubscription.Last()))
```

test/integration/dvl_system_water_mass_tracking.cc
```diff
@@ -195,3 +216,60 @@ TEST(DVLTest, GZ_UTILS_TEST_DISABLED_ON_MAC(WaterMassTracking))
...
+        !validWaterMassTracking(postResetVelocitySubscription.Last()))
+    {
+      return false;
+    }
+
+    const math::Vector3d velocity =
+        msgs::Convert(postResetVelocitySubscription.Last().velocity().mean());
```
caguero:
```
msgs::Convert(velocitySubscription.Last().velocity().mean());
```

src/systems/dvl/DopplerVelocityLogSystem.cc
```diff
@@ -383,6 +478,14 @@ void DopplerVelocityLogSystem::Implementation::DoPostUpdate(
         return true;
       });;
 
+    for (const auto &entity : this->knownSensorEntities)
```
caguero: I'm not sure about this block. If I understand correctly, a single sensor missing will avoid the entire update. I'd reconsider removing this entire block from here and instead, use `OnRender() to skip sensor whose entity isn't present in the latest world state.

caguero (commented):
Thanks for the patch! I just added some comments.

XINJIANGMO:
> Thanks for the patch! I just added some comments.

Thanks, these suggestions make sense. apply them and  build and the two DVL reset tests pass locally.

src/systems/dvl/DopplerVelocityLogSystem.cc
```diff
@@ -253,9 +285,98 @@ void DopplerVelocityLogSystem::Implementation::DoConfigure(
...
+}
+
+//////////////////////////////////////////////////
+void DopplerVelocityLogSystem::Implementation::DoReset(
+  const gz::sim::UpdateInfo &_info,
+  gz::sim::EntityComponentManager &)
+{
+  this->perStepRequests.clear();
+  this->knownSensorEntities.clear();
+  this->needsUpdate = true;
```
caguero: Can this be removed? I think the rendering thread sets `needsUpdate = true` itself when it handles the queued `ResetSensor` and `SetWorldState` requests. Could you confirm?

XINJIANGMO: make sense! Confirm it [here](https://github.com/gazebosim/gz-sim/blob/dba8747ea3332db1ffe429e6204ae1f18becf59b/src/systems/dvl/DopplerVelocityLogSystem.cc#L540)

PR #3660, opened 2026-06-05
Only publish to clock and stats topics when there are subscribers

iche033:
# 🦟 Bug fix

## Summary

Minor optimization. We don't need to publish to stats or clock topics if there are no subscribers. GUI always subscribes to `/world/<world_name>/stats` but I think we don't have any subscribers for the `/stats`, `/clock`, `/world/<world_name>/clock` topics in a typical out-of-box  gazebo simulation with default plugins. We can save some cpu by not publishing msgs at 1000 Hz.

Side note, I noticed that the `/stats` topic is published at 1000Hz while the `/world/<world_name>/stats` topic is 10Hz. Based on the inline comment (`// Publish the stats message. The stats message is throttled.`), I think it's not intentional and the `/stats` topic should also be throttled at 10Hz. I did not make this change in this PR.

## Checklist
- [ ] Added tests

### DIFF
--- a/src/SimulationRunner.cc
+++ b/src/SimulationRunner.cc
@@ -519,9 +519,16 @@ void SimulationRunner::PublishStats()
 {
   GZ_PROFILE("SimulationRunner::PublishStats");
 
-  // Create the world statistics message.
-  msgs::WorldStatistics msg;
-  msg.set_real_time_factor(this->realTimeFactor);
+  const bool publishStats = this->statsPub.HasConnections() ||
+                            (this->rootStatsPub.Valid() &&
+                             this->rootStatsPub.HasConnections());
+
+  const bool publishClock = this->clockPub.HasConnections() ||
+                            (this->rootClockPub.Valid() &&
+                             this->rootClockPub.HasConnections());
+
+  if (!publishStats && !publishClock)
+    return;
 
   auto realTimeSecNsec =
     math::durationToSecNsec(this->currentInfo.realTime);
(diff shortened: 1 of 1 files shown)


src/SimulationRunner.cc
```diff
@@ -519,58 +519,75 @@ void SimulationRunner::PublishStats()
 {
   GZ_PROFILE("SimulationRunner::PublishStats");
 
-  // Create the world statistics message.
-  msgs::WorldStatistics msg;
-  msg.set_real_time_factor(this->realTimeFactor);
+  const bool publishStats = this->statsPub.HasConnections() ||
+                            (this->rootStatsPub.Valid() &&
```
ahcorde: `HasConnections` already check if the connection is `Valid()`. Probably redundant

ahcorde (approved):
Just a minor comment

azeey:
Seems reasonable. Although, I wonder how it would interact with the logic that determines if there's another instance of Gazebo running. https://github.com/gazebosim/gz-sim/blob/90e82be8e2a1978cd5d20f60a29cc868a78bc39b/src/SimulationRunner.cc#L843-L854

In theory two instances of gz-sim could simulate two different worlds since the topic names would have the `/world/{name}/` prefix, but I highly doubt that actually works since there are various other topics that would clash.

src/SimulationRunner.cc
iche033: removed `Valid` check. [1376e24](https://github.com/gazebosim/gz-sim/pull/3660/commits/1376e241dbfb3e4d5d296c2cbec82bd3edc0a35b)

iche033:
> Seems reasonable. Although, I wonder how it would interact with the logic that determines if there's another instance of Gazebo running.
> 
> https://github.com/gazebosim/gz-sim/blob/90e82be8e2a1978cd5d20f60a29cc868a78bc39b/src/SimulationRunner.cc#L843-L854
> 
> In theory two instances of gz-sim could simulate two different worlds since the topic names would have the `/world/{name}/` prefix, but I highly doubt that actually works since there are various other topics that would clash.

I think that was added for subt but I don't exactly remember the reason, but yeah I think there are already a few other conflicting topic names.

PR #3651, opened 2026-06-03
Make EntityComponentManager State() run serially

luca-della-vedova:
# 🎉 New feature

## Summary

Split out from #3447.
While working on the Entt refactor I noticed that there really wasn't a lot of performance to be gained by running the State() function in a multithreaded way, I commented it out there and left as a TODO to figure out whether it was needed or not.
The idea was that we might be able to make things a bit faster by parallelizing but at the same time we currently:

* Launch a number of threads that will pretty much be equal to the number of cores in the machine, even if we have very few entities.
* Need to wait for all of them to be finished.
* When they are all finished we need to iterate through all the maps they populated and merge them into one.

This is really fairly complex and requires additional book-keeping (i.e. through the `componentTypeIndexDirty` flag). Turns out, I _think_ it's actually counter productive?

I split this PR in two commits:

* https://github.com/gazebosim/gz-sim/commit/c456ac01dec1ccabb167a23a51b64d6d027d2ca8 adds a benchmark for the multithreaded function. Note that we have two `State` functions, one that gets an input parameter and is multithreaded and one that doesn't. Now the benchmark tests both.
* https://github.com/gazebosim/gz-sim/commit/f6831d08bc60369d9d1ad4f05503643472c6cb4c is the real commit that removes all the multithreaded behavior and makes it run serially.

## Test it

Checkout https://github.com/gazebosim/gz-sim/commit/c456ac01dec1ccabb167a23a51b64d6d027d2ca8, build and run the test:

```
$ colcon build --packages-select gz-sim --merge-install --mixin rel-with-deb-info
$ build/gz-sim/bin/BENCHMARK_ecm_serialize
```

Checkout the commit that removes the multithreaded logic https://github.com/gazebosim/gz-sim/commit/f6831d08bc60369d9d1ad4f05503643472c6cb4c and rebuild + rerun the benchmark.

On my machine (12 core Ryzen Threadripper 5945WX):

```
Before:
BM_Serialize1ComponentMap/10        0.369 ms        0.365 ms         1782 num_components=1 num_entities=10 serialized_size=370
BM_Serialize1ComponentMap/50        0.661 ms        0.656 ms         1013 num_components=1 num_entities=50 serialized_size=1.89k
BM_Serialize1ComponentMap/100       0.826 ms        0.818 ms          847 num_components=1 num_entities=100 serialized_size=3.79k
BM_Serialize1ComponentMap/500        1.20 ms         1.19 ms          538 num_components=1 num_entities=500 serialized_size=20.136k
[...]

PR #3661, opened 2026-06-05
Remove per-call thread spawning from EntityComponentManager::State

caguero:
# 🦟 Bug fix / 🌱 Improvement

Credit to @azeey for the idea. 

## Summary

`EntityComponentManager::State(SerializedStateMap&, ...)`  is used to serialize the full state and called from `SceneBroadcaster`. The current code uses multiple `std::thread`s to serialize, and then, merged each thread messages back into the output map under a `mutex`.

Microbenchmarking showed that this per-call thread spawning dominated the cost (serializing 10 entities took ~140 µs, almost entirely thread setup/teardown) and that the serial merge-copy negated most of the parallel gain even for large worlds. Replacing it with a direct serial serialization is **1.0×–47× faster** across entity counts (10–47× for the common small/medium worlds, still faster at 10k entities).

## Test it

A new `BM_SerializeStateMap{1,5}Component` benchmark in `test/benchmark/ecm_serialize.cc` exercises the path.

## Checklist

- [x] Added tests (benchmark)
- [x] `codecheck` passed
- [x] All tests passed

**Note to maintainers**: Remember to use **Squash-Merge** and edit the commit message to match the pull request summary while retaining the `Signed-off-by` lines.

### DIFF
--- a/src/EntityComponentManager.cc
+++ b/src/EntityComponentManager.cc
@@ -70,10 +70,6 @@ class gz::sim::EntityComponentManagerPrivate
   public: void EraseEntityRecursive(Entity _entity,
       std::unordered_set<Entity> &_set);
 
-  /// \brief Allots the work for multiple threads prior to running
-  /// `AddEntityToMessage`.
-  public: void CalculateStateThreadLoad();
-
   /// \brief Copies the contents of `_from` into this object.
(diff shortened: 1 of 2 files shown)


caguero:
Duplicated of #3651

PR #3666, opened 2026-06-08
Benchmark for scopedName function in Utils.cc

cheriehu:
# 🎉 New feature

#3656

## Summary

Adds a benchmark for the `scopedName` function which as described in #3656 is quite inefficient. This benchmark can be used to quantify if any optimization done on the function improves performance. 

Also, this flamegraph (generated with help from https://github.com/caguero/gz-profiling.git) shows that the scopedName function does take up a significant portion of CPU when contacts are enabled.

3k_shapes_static_bullet_contact

## Test it

1. `colcon build --packages-select gz-sim --merge-install`
2. `build/gz-sim/bin/BENCHMARK_utils_benchmark`
3. Result should look like 
```
--------------------------------------------------------------------
Benchmark                          Time             CPU   Iterations
--------------------------------------------------------------------
BM_ScopedName/sensor/1           420 ns          420 ns      1652865
BM_ScopedName/projector/1        419 ns          419 ns      1673972
BM_ScopedName/model/1            401 ns          400 ns      1805916
```

## Checklist
- [X] Signed all commits for DCO
- [X] Added a screen capture or video to the PR description that demonstrates the feature
- [ ] Added tests
- [ ] Added example and/or tutorial
- [ ] Updated documentation (as needed)
- [ ] Updated migration guide (as needed)
- [ ] Consider updating Python bindings (if the library has them)
- [ ] `codecheck` passed (See [contributing](https://gazebosim.org/docs/all/contributing#contributing-code))
- [ ] All tests passed (See [test coverage](https://gazebosim.org/docs/all/contributing#test-coverage))
- [ ] Updated Bazel files (if adding new files). Created an issue otherwise.
- [ ] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Agazebosim+archived%3Afalse+) to support the maintainers
- [ ] Was GenAI used to generate this PR? If so, make sure to add "Generated-by" to your commits. (See [this policy](https://osralliance.org/wp-content/uploads/2025/05/OSRF-Policy-on-the-Use-of-Generative-Tools-Generative-AI-in-Contributions.pdf) for more info.)

**Note to maintainers**: Remember to use **Squash-Merge** and edit the commit message to match the pull request summary while retaining `Signed-off-by` and `Generated-by` messages.

**Backports:** If this is a backport, please use **Rebase and Merge** instead.

Issue #3656, opened 2026-06-04
Utility function for entity to string conversion performance


luca-della-vedova:
We have utility functions [here](https://github.com/gazebosim/gz-sim/blob/63b0440eef47a6c605cde469b0204fd9747dde3c/src/Util.cc#L279-L385) for converting an entity to a string for visualization reasons. This function in turn is called from [scopedName](https://github.com/gazebosim/gz-sim/blob/63b0440eef47a6c605cde469b0204fd9747dde3c/src/Util.cc#L156) that is used widely in the codebase, including in hot loops (such as updating collisions in physics, which for every contact point in every mesh in every simulation step calls the function twice https://github.com/gazebosim/gz-sim/blob/63b0440eef47a6c605cde469b0204fd9747dde3c/src/systems/physics/Physics.cc#L4552-L4555).

The function itself performs fairly poorly, it's a long chain of if statements that try to fetch a component which requires itself a few hash map lookups.
A constant time approach could involve creating a component for the "entity type", two main ideas:

## Label + typeId components

We could have a `components::Label` component that wraps a `std::string` which, when queried, will just directly return the entity human readable label without having to do up to 11 (as of now) component lookups.
The same for a `components::ComponentTypeId` (subject to name bikeshedding) for its typeId.

## Entity type enum

We could have something like:

```
enum class EntityType {
    World,
    Model,
    [...]
}

// Register a components::EntityType that wraps it

std::string entityTypeStr(const EntityType _type) {
    switch _type {
        case EntityType::World: return "world"; [...]
   }
}
```

This would add a bit of branching but still a switch statement with N cases is a lot cheaper that N component lookups.
Note that the potential brittleness here is that changing the entity type at runtime might require us to also update this component, for example something that was a World but then becomes a Model because its World component was removed but a Model component was added.
To be honest though, I'm not sure how supported this is right now / how much we want to support it.

luca-della-vedova:
Before going too deep into optimization it would be good to have a set of benchmarks to have some ground truth.
For example, we could create a world with N of models with a fairly deep hierarchy and call ScopedName on entities (i.e. call it on a sensor, that is child of a link, that is child of a model, that is nested inside another model, that is a child of the world).

Then we can use this benchmark to optimize the function itself.

Pengkun-ZHU:
Hi @luca-della-vedova, as an attempt to study Gazebo infra, I set off did a benchmark and enum solution as you mentioned, and filed a PR for demonstration.
[...]

PR #3667, opened 2026-06-08
Optimize scopedName function

### LINKED ISSUE #3608: Deep dive into Gazebo performance issues

luca-della-vedova:
# 🎉 New feature

Based on #3666, for the benchmark, will rebase after it is merged.

## Summary
Mitigate #3656 by reducing allocations and moves.

There are two notable changes.

### Return type of `entityTypeStr` changed to `string_view`

Since the function returns a constant, by returning a `string_view` we can avoid an additional allocation / deallocation. I tested on godbolt what is the difference between returning a `string` vs a `string_view` and there seems to be a significant change in the generated assembly [string version](https://godbolt.org/z/M4fcbGKcM) with about 50 lines of assembly, [string_view version](https://godbolt.org/z/77hT49o8x) with 37.
In this case, returning `string_view` is safe since it is a constant literal so there are no lifetime issues.
Note that this is an API / ABI breakage, so it will make this PR non backportable.
Something to consider is a _potential_ lifetime issue if #3656 will be implemented, specifically:

* If we change the behavior of `entityTypeStr` to read a `Label` component then
* A user calls the function and stores the `string_view` in their code then
* The component is removed / reallocated / entity despawned then
* This will be undefined behavior

I'm tempted to say it's not a huge issue since users of a  `string_view` should generally be careful since they know it's non owning and there are no calls of this function in the whole codebase other than the part touched by this PR, but thought I'd bring it up for completeness.

### Update the concatenation to reduce memory allocations / moves

Previously, we were constantly [pushing to the front of the string](https://github.com/gazebosim/gz-sim/blob/63b0440eef47a6c605cde469b0204fd9747dde3c/src/Util.cc#L167-L179), but every front push to a vector requires moving the whole vector the the right, we were potentially shifting the whole string up to 4 times per hierarchy level, this is without counting all the dynamic reallocations of the string.
The new implementation just pushes all the string views (stack allocated simple types) to the back of a vector then in one go reserves the final string capacity, iterates in reverse order and creates the final topic, removing all the moves altogether.

## Test it

Run the benchmark introduced in #3666:

```
$ colcon build --packages-up-to gz-sim --merge-install
$ build/gz-sim/bin/BENCHMARK_utils_benchmark
```

On my machine, output before optimization:

```
BM_ScopedName/bullet_sensor/1000         0.426 ms        0.426 ms         1674
BM_ScopedName/bullet_projector/1000      0.433 ms        0.433 ms         1562
BM_ScopedName/bullet_model/1000          0.384 ms        0.384 ms         1811
```

After optimization:

```
BM_ScopedName/bullet_sensor/1000         0.359 ms        0.359 ms         1934
BM_ScopedName/bullet_projector/1000      0.380 ms        0.380 ms         1881
BM_ScopedName/bullet_model/1000          0.331 ms        0.331 ms         2138
```

So about 15% performance improvement. Not massive but for a function that takes 20% of the time in the tested it can be significant.

## Checklist
- [ ] Added tests

Generated-by: Gemini 3.

### DIFF
--- a/include/gz/sim/Util.hh
+++ b/include/gz/sim/Util.hh
@@ -152,7 +152,7 @@ namespace gz
     /// In case the entity has more than one type, only one of them will be
     /// returned. This is not standard usage.
     ///
-    /// Note that this is different from  component type names.
+    /// Note that this is different from component type names.
     ///
     /// \param[in] _entity Entity to get the type for.
     /// \param[in] _ecm Immutable reference to ECM.
--- a/include/gz/sim/Util.hh
+++ b/include/gz/sim/Util.hh
@@ -21,6 +21,7 @@
 
 #include <memory>
 #include <string>
+#include <string_view>
 #include <unordered_set>
 #include <vector>
 
@@ -136,7 +137,7 @@ namespace gz
     /// \param[in] _entity Entity to get the type for.
     /// \param[in] _ecm Immutable reference to ECM.
     /// \return ID of entity's type-defining components.
-    std::string GZ_SIM_VISIBLE entityTypeStr(const Entity &_entity,
+    std::string_view GZ_SIM_VISIBLE entityTypeStr(const Entity &_entity,
         const EntityComponentManager &_ecm);
 
     /// \brief Get the world to which the given entity belongs.
--- a/include/gz/sim/Util.hh
+++ b/include/gz/sim/Util.hh
@@ -156,7 +156,7 @@ namespace gz
     ///
     /// \param[in] _entity Entity to get the type for.
     /// \param[in] _ecm Immutable reference to ECM.
-    /// \return ID of entity's type-defining components.
+    /// \return Lowercase string of entity's type-defining components.
     /// \deprecated Use entityTypeStrView instead to reduce memory allocations.
     std::string GZ_SIM_VISIBLE GZ_DEPRECATED(11) entityTypeStr(
         const Entity &_entity,
--- a/include/gz/sim/Util.hh
+++ b/include/gz/sim/Util.hh
@@ -137,7 +137,29 @@ namespace gz
     /// \param[in] _entity Entity to get the type for.
     /// \param[in] _ecm Immutable reference to ECM.
     /// \return ID of entity's type-defining components.
-    std::string_view GZ_SIM_VISIBLE entityTypeStr(const Entity &_entity,
+    std::string_view GZ_SIM_VISIBLE entityTypeStrView(const Entity &_entity,
+        const EntityComponentManager &_ecm);
+
+    /// \brief Generally, each entity will be of some specific high-level type,
+    /// such as "world", "sensor", "collision", etc, and one type only.
+    ///
+    /// This function returns a lowercase string for each type. For example,
+    /// "light", "actor", etc.
+    ///
+    /// In case the entity isn't of any known type, this will return an empty
+    /// string.
+    ///
+    /// In case the entity has more than one type, only one of them will be
+    /// returned. This is not standard usage.
+    ///
+    /// Note that this is different from  component type names.
+    ///
+    /// \param[in] _entity Entity to get the type for.
+    /// \param[in] _ecm Immutable reference to ECM.
+    /// \return ID of entity's type-defining components.
+    /// \deprecated Use entityTypeStrView instead to reduce memory allocations.
+    std::string GZ_SIM_VISIBLE GZ_DEPRECATED(11) entityTypeStr(
+        const Entity &_entity,
         const EntityComponentManager &_ecm);
 
     /// \brief Get the world to which the given entity belongs.
--- a/src/Util.cc
+++ b/src/Util.cc
@@ -145,6 +145,7 @@ std::string scopedName(const Entity &_entity,
   // withou occupying significant memory (string view is just a pointer + size)
   parts.reserve(10);
   auto entity = _entity;
+  size_t totalSize = 0;
 
   while (true)
   {
@@ -167,9 +168,11 @@ std::string scopedName(const Entity &_entity,
     if (!prefix.empty())
     {
       parts.push_back(name);
+      totalSize += name.size();
       if (_includePrefix)
       {
         parts.push_back(prefix);
+        totalSize += prefix.size();
       }
     }
 
@@ -183,11 +186,6 @@ std::string scopedName(const Entity &_entity,
     return "";
 
   std::string result;
-  size_t totalSize = 0;
-  for (const auto &part : parts)
-  {
-    totalSize += part.size();
-  }
   totalSize += (parts.size() - 1) * _delim.size();
   result.reserve(totalSize);
 
--- a/src/Util.cc
+++ b/src/Util.cc
@@ -141,6 +141,9 @@ std::string scopedName(const Entity &_entity,
     bool _includePrefix)
 {
   std::vector<std::string_view> parts;
+  // Pre-allocate 10 entries, a sensible default that will prevent reallocation
+  // withou occupying significant memory (string view is just a pointer + size)
+  parts.reserve(10);
   auto entity = _entity;
 
   while (true)
--- a/src/Util.cc
+++ b/src/Util.cc
@@ -145,7 +145,6 @@ std::string scopedName(const Entity &_entity,
   // without occupying significant memory (string view is just a pointer + size)
   parts.reserve(10);
   auto entity = _entity;
-  size_t totalSize = 0;
 
   while (true)
   {
@@ -168,11 +167,9 @@ std::string scopedName(const Entity &_entity,
     if (!prefix.empty())
     {
       parts.push_back(name);
-      totalSize += name.size();
       if (_includePrefix)
       {
         parts.push_back(prefix);
-        totalSize += prefix.size();
       }
     }
 
@@ -186,6 +183,11 @@ std::string scopedName(const Entity &_entity,
     return "";
 
   std::string result;
+  size_t totalSize = 0;
+  for (const auto &part : parts)
+  {
+    totalSize += part.size();
+  }
   totalSize += (parts.size() - 1) * _delim.size();
   result.reserve(totalSize);
 
--- a/src/Util.cc
+++ b/src/Util.cc
@@ -140,8 +140,7 @@ std::string scopedName(const Entity &_entity,
     const EntityComponentManager &_ecm, const std::string &_delim,
     bool _includePrefix)
 {
-  std::string result;
-
+  std::vector<std::string_view> parts;
   auto entity = _entity;
 
   while (true)
@@ -150,10 +149,10 @@ std::string scopedName(const Entity &_entity,
     auto nameComp = _ecm.Component<components::Name>(entity);
     if (nullptr == nameComp)
       break;
-    auto name = nameComp->Data();
+    const auto &name = nameComp->Data();
 
     // Get entity type
-    std::string prefix = entityTypeStr(entity, _ecm);
+    std::string_view prefix = entityTypeStr(entity, _ecm);
     if (prefix.empty())
     {
       gzwarn << "Skipping entity [" << name
@@ -164,21 +163,38 @@ std::string scopedName(const Entity &_entity,
     auto parentComp = _ecm.Component<components::ParentEntity>(entity);
     if (!prefix.empty())
     {
-      result.insert(0, name);
+      parts.push_back(name);
       if (_includePrefix)
       {
-        result.insert(0, _delim);
-        result.insert(0, prefix);
+        parts.push_back(prefix);
       }
     }
 
     if (nullptr == parentComp)
       break;
 
-    if (!prefix.empty())
-      result.insert(0, _delim);
-
     entity = parentComp->Data();
+  }
+
+  if (parts.empty())
+    return "";
+
+  std::string result;
+  size_t totalSize = 0;
+  for (const auto &part : parts)
+  {
+    totalSize += part.size();
+  }
+  totalSize += (parts.size() - 1) * _delim.size();
+  result.reserve(totalSize);
+
+  for (auto it = parts.rbegin(); it != parts.rend(); ++it)
+  {
+    if (it != parts.rbegin())
+    {
+      result += _delim;
+    }
+    result += *it;
   }
 
   return result;
@@ -331,10 +347,10 @@ ComponentTypeId entityTypeId(const Entity &_entity,
 }
 
 //////////////////////////////////////////////////
-std::string entityTypeStr(const Entity &_entity,
+std::string_view entityTypeStr(const Entity &_entity,
     const EntityComponentManager &_ecm)
 {
-  std::string type;
+  std::string_view type;
 
   if (_ecm.Component<components::World>(_entity))
   {
--- a/src/Util.cc
+++ b/src/Util.cc
@@ -155,7 +155,7 @@ std::string scopedName(const Entity &_entity,
     const auto &name = nameComp->Data();
 
     // Get entity type
-    std::string_view prefix = entityTypeStr(entity, _ecm);
+    std::string_view prefix = entityTypeStrView(entity, _ecm);
     if (prefix.empty())
     {
       gzwarn << "Skipping entity [" << name
@@ -349,8 +349,7 @@ ComponentTypeId entityTypeId(const Entity &_entity,
   return type;
 }
 
-//////////////////////////////////////////////////
-std::string_view entityTypeStr(const Entity &_entity,
+std::string_view entityTypeStrView(const Entity &_entity,
     const EntityComponentManager &_ecm)
 {
   std::string_view type;
@@ -401,6 +400,13 @@ std::string_view entityTypeStr(const Entity &_entity,
   }
 
   return type;
+}
+
+//////////////////////////////////////////////////
+std::string entityTypeStr(const Entity &_entity,
+    const EntityComponentManager &_ecm)
+{
+  return std::string(entityTypeStrView(_entity, _ecm));
 }
 
 //////////////////////////////////////////////////
--- a/src/Util.cc
+++ b/src/Util.cc
@@ -142,7 +142,7 @@ std::string scopedName(const Entity &_entity,
 {
   std::vector<std::string_view> parts;
   // Pre-allocate 10 entries, a sensible default that will prevent reallocation
-  // withou occupying significant memory (string view is just a pointer + size)
+  // without occupying significant memory (string view is just a pointer + size)
   parts.reserve(10);
   auto entity = _entity;
   size_t totalSize = 0;
--- a/src/Util_TEST.cc
+++ b/src/Util_TEST.cc
@@ -379,51 +379,51 @@ TEST_F(UtilTest, EntityTypeStr)
   EntityComponentManager ecm;
 
   auto entity = ecm.CreateEntity();
-  EXPECT_TRUE(entityTypeStr(entity, ecm).empty());
+  EXPECT_TRUE(entityTypeStrView(entity, ecm).empty());
 
   entity = ecm.CreateEntity();
   ecm.CreateComponent(entity, components::World());
-  EXPECT_EQ("world", entityTypeStr(entity, ecm));
+  EXPECT_EQ("world", entityTypeStrView(entity, ecm));
 
   entity = ecm.CreateEntity();
   ecm.CreateComponent(entity, components::Model());
-  EXPECT_EQ("model", entityTypeStr(entity, ecm));
+  EXPECT_EQ("model", entityTypeStrView(entity, ecm));
 
   entity = ecm.CreateEntity();
   ecm.CreateComponent(entity, components::Light());
-  EXPECT_EQ("light", entityTypeStr(entity, ecm));
+  EXPECT_EQ("light", entityTypeStrView(entity, ecm));
 
   entity = ecm.CreateEntity();
   ecm.CreateComponent(entity, components::Link());
-  EXPECT_EQ("link", entityTypeStr(entity, ecm));
+  EXPECT_EQ("link", entityTypeStrView(entity, ecm));
 
   entity = ecm.CreateEntity();
   ecm.CreateComponent(entity, components::Visual());
-  EXPECT_EQ("visual", entityTypeStr(entity, ecm));
+  EXPECT_EQ("visual", entityTypeStrView(entity, ecm));
 
   entity = ecm.CreateEntity();
   ecm.CreateComponent(entity, components::Collision());
-  EXPECT_EQ("collision", entityTypeStr(entity, ecm));
+  EXPECT_EQ("collision", entityTypeStrView(entity, ecm));
 
   entity = ecm.CreateEntity();
   ecm.CreateComponent(entity, components::Joint());
-  EXPECT_EQ("joint", entityTypeStr(entity, ecm));
+  EXPECT_EQ("joint", entityTypeStrView(entity, ecm));
 
   entity = ecm.CreateEntity();
   ecm.CreateComponent(entity, components::Sensor());
-  EXPECT_EQ("sensor", entityTypeStr(entity, ecm));
+  EXPECT_EQ("sensor", entityTypeStrView(entity, ecm));
 
   entity = ecm.CreateEntity();
   ecm.CreateComponent(entity, components::Actor());
-  EXPECT_EQ("actor", entityTypeStr(entity, ecm));
+  EXPECT_EQ("actor", entityTypeStrView(entity, ecm));
 
   entity = ecm.CreateEntity();
   ecm.CreateComponent(entity, components::ParticleEmitter());
-  EXPECT_EQ("particle_emitter", entityTypeStr(entity, ecm));
+  EXPECT_EQ("particle_emitter", entityTypeStrView(entity, ecm));
 
   entity = ecm.CreateEntity();
   ecm.CreateComponent(entity, components::Projector());
-  EXPECT_EQ("projector", entityTypeStr(entity, ecm));
+  EXPECT_EQ("projector", entityTypeStrView(entity, ecm));
 }
 
 /////////////////////////////////////////////////
--- a/src/Util_TEST.cc
+++ b/src/Util_TEST.cc
@@ -27,6 +27,7 @@
 
 #include <gz/fuel_tools/ClientConfig.hh>
 #include <gz/utils/ExtraTestMacros.hh>
+#include <gz/utils/SuppressWarning.hh>
 
 #include "gz/sim/components/Actor.hh"
 #include "gz/sim/components/Collision.hh"
@@ -376,54 +377,71 @@ TEST_F(UtilTest, EntityTypeId)
 /////////////////////////////////////////////////
 TEST_F(UtilTest, EntityTypeStr)
 {
+  // EntityTypeStr is deprecated, suppress once for the whole test
+  GZ_UTILS_WARN_IGNORE__DEPRECATED_DECLARATION
+
   EntityComponentManager ecm;
 
   auto entity = ecm.CreateEntity();
   EXPECT_TRUE(entityTypeStrView(entity, ecm).empty());
+  EXPECT_TRUE(entityTypeStr(entity, ecm).empty());
 
   entity = ecm.CreateEntity();
   ecm.CreateComponent(entity, components::World());
   EXPECT_EQ("world", entityTypeStrView(entity, ecm));
+  EXPECT_EQ("world", entityTypeStr(entity, ecm));
 
   entity = ecm.CreateEntity();
   ecm.CreateComponent(entity, components::Model());
   EXPECT_EQ("model", entityTypeStrView(entity, ecm));
+  EXPECT_EQ("model", entityTypeStr(entity, ecm));
 
   entity = ecm.CreateEntity();
   ecm.CreateComponent(entity, components::Light());
   EXPECT_EQ("light", entityTypeStrView(entity, ecm));
+  EXPECT_EQ("light", entityTypeStr(entity, ecm));
 
   entity = ecm.CreateEntity();
   ecm.CreateComponent(entity, components::Link());
   EXPECT_EQ("link", entityTypeStrView(entity, ecm));
+  EXPECT_EQ("link", entityTypeStr(entity, ecm));
 
   entity = ecm.CreateEntity();
   ecm.CreateComponent(entity, components::Visual());
   EXPECT_EQ("visual", entityTypeStrView(entity, ecm));
+  EXPECT_EQ("visual", entityTypeStr(entity, ecm));
 
   entity = ecm.CreateEntity();
   ecm.CreateComponent(entity, components::Collision());
   EXPECT_EQ("collision", entityTypeStrView(entity, ecm));
+  EXPECT_EQ("collision", entityTypeStr(entity, ecm));
 
   entity = ecm.CreateEntity();
   ecm.CreateComponent(entity, components::Joint());
   EXPECT_EQ("joint", entityTypeStrView(entity, ecm));
+  EXPECT_EQ("joint", entityTypeStr(entity, ecm));
 
   entity = ecm.CreateEntity();
   ecm.CreateComponent(entity, components::Sensor());
   EXPECT_EQ("sensor", entityTypeStrView(entity, ecm));
+  EXPECT_EQ("sensor", entityTypeStr(entity, ecm));
 
   entity = ecm.CreateEntity();
   ecm.CreateComponent(entity, components::Actor());
   EXPECT_EQ("actor", entityTypeStrView(entity, ecm));
+  EXPECT_EQ("actor", entityTypeStr(entity, ecm));
 
   entity = ecm.CreateEntity();
   ecm.CreateComponent(entity, components::ParticleEmitter());
   EXPECT_EQ("particle_emitter", entityTypeStrView(entity, ecm));
+  EXPECT_EQ("particle_emitter", entityTypeStr(entity, ecm));
 
   entity = ecm.CreateEntity();
   ecm.CreateComponent(entity, components::Projector());
   EXPECT_EQ("projector", entityTypeStrView(entity, ecm));
+  EXPECT_EQ("projector", entityTypeStr(entity, ecm));
+
+  GZ_UTILS_WARN_RESUME__DEPRECATED_DECLARATION
 }
 
 /////////////////////////////////////////////////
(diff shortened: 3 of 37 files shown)


include/gz/sim/Util.hh
```diff
@@ -136,7 +137,7 @@ namespace gz
     /// \param[in] _entity Entity to get the type for.
     /// \param[in] _ecm Immutable reference to ECM.
     /// \return ID of entity's type-defining components.
-    std::string GZ_SIM_VISIBLE entityTypeStr(const Entity &_entity,
+    std::string_view GZ_SIM_VISIBLE entityTypeStr(const Entity &_entity,
```
caguero: Consider adding a note in `Migration.md` about the API update.

test/benchmark/utils_benchmark.cc
```diff
@@ -0,0 +1,116 @@
+/*
+ * Copyright (C) 2019 Open Source Robotics Foundation
```
iche033:
```suggestion
 * Copyright (C) 2026 Open Source Robotics Foundation
```

test/benchmark/utils_benchmark.cc
```diff
@@ -0,0 +1,116 @@
...
+
+BENCHMARK_CAPTURE(BM_ScopedName, bullet_projector, "projector")
+    ->Arg(1000)
+    ->Unit(benchmark::kMillisecond);
+
+BENCHMARK_CAPTURE(BM_ScopedName, bullet_model, "model")
+    ->Arg(1000)
+    ->Unit(benchmark::kMillisecond);
+
+// OSX needs the semicolon, Ubuntu complains that there's an extra ';'
```
iche033:
```suggestion
// macOS needs the semicolon, Ubuntu complains that there's an extra ';'
```

test/benchmark/utils_benchmark.cc
```diff
@@ -0,0 +1,116 @@
...
+  for (auto _ : _st) {
+    for (int ii = 0; ii < iterations; ++ii)
+    {
+      scopedName(entity, ecm);
+    }
+  }
+}
+
+// NOLINTNEXTLINE
+BENCHMARK_CAPTURE(BM_ScopedName, bullet_sensor, "sensor")
```
iche033: maybe just `sensor` instead of `bullet_sensor`? same for the other `BENCHMARK_CAPTURE` below

test/benchmark/utils_benchmark.cc
```diff
@@ -0,0 +1,116 @@
...
+#include <benchmark/benchmark.h>
+
+#include "gz/sim/components/Link.hh"
+#include "gz/sim/components/Model.hh"
+#include "gz/sim/components/Name.hh"
+#include "gz/sim/components/ParentEntity.hh"
+#include "gz/sim/components/Projector.hh"
+#include "gz/sim/components/Sensor.hh"
+#include "gz/sim/components/World.hh"
+#include "gz/sim/components/Link.hh"
```
iche033: duplicate header 

```suggestion
```

iche033 (commented):
Is it possible to avoid breaking ABI while keeping the optimizations by creating a separate helper function, e.g. `entityTypeStrView`, and the existing one just calls the new helper function?

```C++
std::string entityTypeStr(const Entity &_entity, const EntityComponentManager &_ecm)
{
  return std::string(entityTypeStrView(_entity, _ecm));
}
```

`scopedName` could just call the new helper function.

src/Util.cc
```diff
@@ -164,23 +163,40 @@ std::string scopedName(const Entity &_entity,
...
   }
 
+  if (parts.empty())
+    return "";
+
+  std::string result;
+  size_t totalSize = 0;
+  for (const auto &part : parts)
+  {
+    totalSize += part.size();
```
ntfshard, line 189: BTW, you can do this accumulation in a same place where you adding elements to `parts`.
Not sure about benefits for this change, therefore not sure about applying this change.

Also for `parts` you are using vector. And vector resize is not very costly, but still. Maybe other container suits better for this and/or maybe do some pre-allocation, like 10 elements should be enough in a most cases I guess.

AFAIR, gnu C++ runtime use powers of 2 as an allocation steps (0, 1, 2, 4, 8...)

ntfshard:
> Is it possible to avoid breaking ABI while keeping the optimizations by creating a separate helper function, e.g. `entityTypeStrView`, and the existing one just calls the new helper function?
> 
> ```c++
> std::string entityTypeStr(const Entity &_entity, const EntityComponentManager &_ecm)
> {
>   return std::string(entityTypeStrView(_entity, _ecm));
> }
> ```
> 
> `scopedName` could just call the new helper function.

and even more, project has a deprecated policy: 
https://gazebosim.org/docs/latest/releases/#deprecation-strategy

ntfshard (commented):
is any positive effect on RTF in 3k_shapes.sdf simulation?

test/benchmark/utils_benchmark.cc
luca-della-vedova: This PR is based on #3666 created by @cheriehu  which is the one that introduces the benchmark so feedback on the benchmarks should go in that PR. For this specific one (and a few others) I believe they are already solved there, I just rebased and force pushed to incorporate the latest changes.
I just couldn't target #3666 branch because that would have ended up opening a PR on the fork, not this repo.

src/Util.cc
luca-della-vedova: > BTW, you can do this accumulation in a same place where you adding elements to `parts`. Not sure about benefits for this change, therefore not sure about applying this change.
> 

Yea the main reason I didn't do it was that performance wise it didn't really make any difference and since we push to the vector in more than one place this makes the code _slightly_ more repetitive / brittle if someone adds a push somewhere and forgets to increase the totalSize counter. This is what the diff would look like https://github.com/gazebosim/gz-sim/pull/3667/commits/750bbf549bbb9059e930c26de853ed08e2b61f49 (happy to keep it or revert it).

> Also for `parts` you are using vector. And vector resize is not very costly, but still. Maybe other container suits better for this and/or maybe do some pre-allocation, like 10 elements should be enough in a most cases I guess.
> 
> AFAIR, gnu C++ runtime use powers of 2 as an allocation steps (0, 1, 2, 4, 8...)

Very good point, I somewhat naively thought that since string_view is such a small type it wouldn't matter so much but it does, I had an additional ~10% performance improvement on the benchmark by following your 10 preallocation strategy. https://github.com/gazebosim/gz-sim/pull/3667/commits/eeb9a2723f2e4ba44e29c3e85bc0b400b689651d

luca-della-vedova:
> is any positive effect on RTF in 3k_shapes.sdf simulation?

It does, although in a minor way (as can be expected from this somewhat minor improvement).
This branched out of the flame graph in #3666, basically as soon as we initialize contact visualization the RTF drops _a lot_, and up to 20% of the time in each simulation is spent on the `scopedName` function.
This happens specifically [here](https://github.com/gazebosim/gz-sim/blob/189dd1705af2741cee63678bd5d7ba17a31d73c9/src/systems/physics/Physics.cc#L4552-L4555), where in each simulation step, for each contact point (and there could be many contacts) we call the `scopedName` function twice.
This is as far as we can optimize before thinking about radical things, like caching the name altogether and removing all these lookups, but those come with their own sets of tradeoffs (how do we detect and rebuild a stale hierarchy cache?).
In my machine, this improved the RTF from 1.5% to 1.6%, so about 6-7%. Again not mindblowing but fairly reasonable. There are various more optimizations we can do for that scenario (i.e. using arenas for protobufs, since a lot of time is spent allocating and deallocating contacts) but one at a time!

src/Util.cc
ntfshard: Execution time less than _0.5ms_ gave a hint that this change could impact performance.

Maybe https://github.com/gazebosim/gz-sim/commit/750bbf549bbb9059e930c26de853ed08e2b61f49 not very important in this case, should be benchmarked. If no performance -- initial version was fine. Up to you.

For vector grow, (mine speculation) maybe it's combination of pipeline/code cache flush (prefetched code in cache is not needed, must go to runtime to allocate memory, to release memory), maybe data cache also affected. Maybe something more. `perf` can tell more.

ntfshard:
> > is any positive effect on RTF in 3k_shapes.sdf simulation?
> 
> It does, although in a minor way (as can be expected from this somewhat minor improvement). This branched out of the flame graph in #3666, basically as soon as we initialize contact visualization the RTF drops _a lot_, and up to 20% of the time in each simulation is spent on the `scopedName` function. This happens specifically [here](https://github.com/gazebosim/gz-sim/blob/189dd1705af2741cee63678bd5d7ba17a31d73c9/src/systems/physics/Physics.cc#L4552-L4555), where in each simulation step, for each contact point (and there could be many contacts) we call the `scopedName` function twice. This is as far as we can optimize before thinking about radical things, like caching the name altogether and removing all these lookups, but those come with their own sets of tradeoffs (how do we detect and rebuild a stale hierarchy cache?). In my machine, this improved the RTF from 1.5% to 1.6%, so about 6-7%. Again not mindblowing but fairly reasonable. There are various more optimizations we can do for that scenario (i.e. using arenas for protobufs, since a lot of time is spent allocating and deallocating contacts) but one at a time!

I asked due to it happened a lot, when you have a positive benchmark results, but in a real workload it doesn't have any impact. Change from 1.5 to 1.6 can be a fluke. How is RTF measured? Reading `/stats` topic?
And correct me if I wrong, this all changes is only impact when we enable contact visualization? So, for users w/o this it will work in a same pace?

luca-della-vedova:
> I asked due to it happened a lot, when you have a positive benchmark results, but in a real workload it doesn't have any impact. Change from 1.5 to 1.6 can be a fluke. How is RTF measured? Reading `/stats` topic? And correct me if I wrong, this all changes is only impact when we enable contact visualization? So, for users w/o this it will work in a same pace?

I generally don't like "measuring RTF". How I did it? The very subjective way you would expect a user to, run the world, look at the RTF value in the UI. I don't think this is a good approach generally which is why I would be personally happy if something like https://github.com/gazebosim/gz-sim/pull/3671 landed, were we have a binary that we can execute N times and it can tell us "the RTF for this scenario over X runs is Y".
That being said, I don't think this is a fluke. Again the [flamegraph](https://private-user-images.githubusercontent.com/94511514/604219676-798d48e0-452a-470b-9a37-0dbc6427c035.svg?jwt=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3ODE2NjE3OTAsIm5iZiI6MTc4MTY2MTQ5MCwicGF0aCI6Ii85NDUxMTUxNC82MDQyMTk2NzYtNzk4ZDQ4ZTAtNDUyYS00NzBiLTlhMzctMGRiYzY0MjdjMDM1LnN2Zz9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNjA2MTclMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjYwNjE3VDAxNTgxMFomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPWQxNjg2Nzg0MGY4OTY1YzNlMzRhYTFlYjQwNThmN2YyZWFjYzZhMjMzOTkyMzljYWIzZTQwNjc4NjQ2OTJhNWEmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JnJlc3BvbnNlLWNvbnRlbnQtdHlwZT1pbWFnZSUyRnN2ZyUyQnhtbCJ9.--kUSlz_gqiDH0juCWHpkrF4mlpuGaEK2bUc2zmBeTE) in #3666 clearly showed that in a world with a lot of collisions we spend more than 20% of the whole simulation time in this function (actually some other optimizations landed after that flame graph, so I expect the value now to be ~25%). The latest micro benchmark run that only tests this function shows that this function is now ~33% faster, so I expect indeed 0.33 * 0.25 ~= 8% higher RTF.

This affects users that either enable contact visualization or enable contact sensors in general. I personally encountered this when working on a headless simulation of a robotic arm that had contact sensors to detect contacts with objects, which I believe could be a fairly common usecase.

ntfshard:
> > I asked due to it happened a lot, when you have a positive benchmark results, but in a real workload it doesn't have any impact. Change from 1.5 to 1.6 can be a fluke. How is RTF measured? Reading `/stats` topic? And correct me if I wrong, this all changes is only impact when we enable contact visualization? So, for users w/o this it will work in a same pace?
> 
> I generally don't like "measuring RTF". How I did it? The very subjective way you would expect a user to, run the world, look at the RTF value in the UI. I don't think this is a good approach generally which is why I would be personally happy if something like #3671 landed, were we have a binary that we can execute N times and it can tell us "the RTF for this scenario over X runs is Y". That being said, I don't think this is a fluke. Again the flamegraph in #3666 clearly showed that in a world with a lot of collisions we spend more than 20% of the whole simulation time in this function (actually some other optimizations landed after that flame graph, so I expect the value now to be ~25%). The latest micro benchmark run that only tests this function shows that this function is now ~33% faster, so I expect indeed 0.33 * 0.25 ~= 8% higher RTF.
> 
> This affects users that either enable contact visualization or enable contact sensors in general. I personally encountered this when working on a headless simulation of a robotic arm that had contact sensors to detect contacts with objects, which I believe could be a fairly common usecase.

Hey. No offense.

My main concern is about real performance impact on a whole system. (almost) Any function can be speed-up, but do it really change whore system performance that's the question. And I'm trying to find the answer.

For example, you mention that strings involve an allocation and deallocation. And for quite long strings it's true, but for a short strings in C++ runtimes usually available [Small String Optimization](https://pvs-studio.com/en/blog/terms/6658/) and here is an example https://godbolt.org/z/6ed5sG8bq (where difference between str object address and data is just 16 bytes for SSO). So even `collision` can fit w/o memory allocation. So maybe something other is quite expensive.

`entityTypeStr` has quite a lot of calls like : `_ecm.Component<components::T>(_entity)` which is quite interesting thing. If I see correctly, each call involve several search operations inside ecm. And it happen in order which mentioned in code. I can bet `Light` is less frequent that `Link`.

And about flamegraph, for templates and -O2, it can be a lot of confusion in a debug_info, which used to attribute the traces (it became not very accurate in other words)

That's why I asking this questions. And instantaneous RTF and RTF deviation is quite fellable for user IMHO.

luca-della-vedova:
> Hey. No offense.
> 

Too late, the offense took place! Some of the feedback is valid (the preallocate vector, again thanks for spotting it!) but some of it feels a bit too much bikeshedding on a fairly trivial straightforward change. Apologies for being blunt but a lot of it also feels like it would be avoided if instead of discussing on "matters of principle" the changes were actually looked at more in depth or tested.

> My main concern is about real performance impact on a whole system. (almost) Any function can be speed-up, but do it really change whore system performance that's the question. And I'm trying to find the answer.
> 

Agreed, and this is the purpose on the flamegraph that it shows that _when enabling contacts_ this function takes > 20% of the whole simulation time step. When contacts are enabled the RTF drops by more than 90% which is, honestly, not acceptable.
This can be easily tested, run 3k_shapes on bullet featherstone, wait for the RTF to go up after the physics puts the objects to sleep, enable contacts from the UI, see the huge drop.
I do happen to rely on that feature for my core work so this is important to me, but regardless I think it's pretty straightforward that we can't have this kind of performance drop that seems to be almost exclusively caused by inefficent data copies / protobuf population?

> For example, you mention that strings involve an allocation and deallocation. And for quite long strings it's true, but for a short strings in C++ runtimes usually available [Small String Optimization](https://pvs-studio.com/en/blog/terms/6658/) and here is an example https://godbolt.org/z/6ed5sG8bq (where difference between str object address and data is just 16 bytes for SSO). So even `collision` can fit w/o memory allocation. So maybe something other is quite expensive.
> 

I attached two godbolt links in the PR description, the [string](https://godbolt.org/z/M4fcbGKcM) and the [string_view](https://godbolt.org/z/77hT49o8x) API. I'm not an expert but from glancing at the code the `string_view` API generates 37 lines of assembly while the string API generates 50. Also the `string` API has two calls to a destructor while the `string_view` only one. This all points to me in the direction of "string_view is probably faster".
But again, rule zero I have been using in my performance work is that ultimately all that matters is profiling, benchmarking and testing changes.
After your comment I went back and applied your suggested change, the patch is below. Remove the new API and just use `std::string`.

```patch
diff --git a/include/gz/sim/Util.hh b/include/gz/sim/Util.hh
index fe9a87c4d..773b8f6de 100644
--- a/include/gz/sim/Util.hh
+++ b/include/gz/sim/Util.hh
@@ -120,26 +120,6 @@ namespace gz
     ComponentTypeId GZ_SIM_VISIBLE entityTypeId(const Entity &_entity,
         const EntityComponentManager &_ecm);
 
-    /// \brief Generally, each entity will be of some specific high-level type,
-    /// such as "world", "sensor", "collision", etc, and one type only.
-    ///
-    /// This function returns a lowercase string for each type. For example,
-    /// "light", "actor", etc.
-    ///
-    /// In case the entity isn't of any known type, this will return an empty
-    /// string.
-    ///
-    /// In case the entity has more than one type, only one of them will be
-    /// returned. This is not standard usage.
-    ///
-    /// Note that this is different from  component type names.
-    ///
-    /// \param[in] _entity Entity to get the type for.
-    /// \param[in] _ecm Immutable reference to ECM.
-    /// \return ID of entity's type-defining components.
-    std::string_view GZ_SIM_VISIBLE entityTypeStrView(const Entity &_entity,
-        const EntityComponentManager &_ecm);
-
     /// \brief Generally, each entity will be of some specific high-level type,
     /// such as "world", "sensor", "collision", etc, and one type only.
     ///
diff --git a/src/Util.cc b/src/Util.cc
index 91b1a4583..b9e623139 100644
--- a/src/Util.cc
+++ b/src/Util.cc
@@ -140,7 +140,7 @@ std::string scopedName(const Entity &_entity,
     const EntityComponentManager &_ecm, const std::string &_delim,
     bool _includePrefix)
 {
-  std::vector<std::string_view> parts;
+  std::vector<std::string> parts;
   // Pre-allocate 10 entries, a sensible default that will prevent reallocation
   // without occupying significant memory (string view is just a pointer + size)
   parts.reserve(10);
@@ -155,7 +155,7 @@ std::string scopedName(const Entity &_entity,
     const auto &name = nameComp->Data();
 
     // Get entity type
-    std::string_view prefix = entityTypeStrView(entity, _ecm);
+    std::string prefix = entityTypeStr(entity, _ecm);
     if (prefix.empty())
     {
       gzwarn << "Skipping entity [" << name
@@ -349,10 +349,10 @@ ComponentTypeId entityTypeId(const Entity &_entity,
   return type;
 }
 
-std::string_view entityTypeStrView(const Entity &_entity,
+std::string entityTypeStr(const Entity &_entity,
     const EntityComponentManager &_ecm)
 {
-  std::string_view type;
+  std::string type;
 
   if (_ecm.Component<components::World>(_entity))
   {
@@ -402,13 +402,6 @@ std::string_view entityTypeStrView(const Entity &_entity,
   return type;
 }
 
-//////////////////////////////////////////////////
-std::string entityTypeStr(const Entity &_entity,
-    const EntityComponentManager &_ecm)
-{
-  return std::string(entityTypeStrView(_entity, _ecm));
-}
-
 Entity worldEntity(const Entity &_entity,
     const EntityComponentManager &_ecm)
diff --git a/src/Util_TEST.cc b/src/Util_TEST.cc
index 2546c2042..08ae2cd84 100644
--- a/src/Util_TEST.cc
+++ b/src/Util_TEST.cc
@@ -379,51 +379,51 @@ TEST_F(UtilTest, EntityTypeStr)
   EntityComponentManager ecm;
 
   auto entity = ecm.CreateEntity();
-  EXPECT_TRUE(entityTypeStrView(entity, ecm).empty());
+  EXPECT_TRUE(entityTypeStr(entity, ecm).empty());
 
   entity = ecm.CreateEntity();
   ecm.CreateComponent(entity, components::World());
-  EXPECT_EQ("world", entityTypeStrView(entity, ecm));
+  EXPECT_EQ("world", entityTypeStr(entity, ecm));
 
   entity = ecm.CreateEntity();
   ecm.CreateComponent(entity, components::Model());
-  EXPECT_EQ("model", entityTypeStrView(entity, ecm));
+  EXPECT_EQ("model", entityTypeStr(entity, ecm));
 
   entity = ecm.CreateEntity();
   ecm.CreateComponent(entity, components::Light());
-  EXPECT_EQ("light", entityTypeStrView(entity, ecm));
+  EXPECT_EQ("light", entityTypeStr(entity, ecm));
 
   entity = ecm.CreateEntity();
   ecm.CreateComponent(entity, components::Link());
-  EXPECT_EQ("link", entityTypeStrView(entity, ecm));
+  EXPECT_EQ("link", entityTypeStr(entity, ecm));
 
   entity = ecm.CreateEntity();
   ecm.CreateComponent(entity, components::Visual());
-  EXPECT_EQ("visual", entityTypeStrView(entity, ecm));
+  EXPECT_EQ("visual", entityTypeStr(entity, ecm));
 
   entity = ecm.CreateEntity();
   ecm.CreateComponent(entity, components::Collision());
-  EXPECT_EQ("collision", entityTypeStrView(entity, ecm));
+  EXPECT_EQ("collision", entityTypeStr(entity, ecm));
 
   entity = ecm.CreateEntity();
   ecm.CreateComponent(entity, components::Joint());
-  EXPECT_EQ("joint", entityTypeStrView(entity, ecm));
+  EXPECT_EQ("joint", entityTypeStr(entity, ecm));
 
   entity = ecm.CreateEntity();
   ecm.CreateComponent(entity, components::Sensor());
-  EXPECT_EQ("sensor", entityTypeStrView(entity, ecm));
+  EXPECT_EQ("sensor", entityTypeStr(entity, ecm));
 
   entity = ecm.CreateEntity();
   ecm.CreateComponent(entity, components::Actor());
-  EXPECT_EQ("actor", entityTypeStrView(entity, ecm));
+  EXPECT_EQ("actor", entityTypeStr(entity, ecm));
 
   entity = ecm.CreateEntity();
   ecm.CreateComponent(entity, components::ParticleEmitter());
-  EXPECT_EQ("particle_emitter", entityTypeStrView(entity, ecm));
+  EXPECT_EQ("particle_emitter", entityTypeStr(entity, ecm));
 
   entity = ecm.CreateEntity();
   ecm.CreateComponent(entity, components::Projector());
-  EXPECT_EQ("projector", entityTypeStrView(entity, ecm));
+  EXPECT_EQ("projector", entityTypeStr(entity, ecm));
 }
 
(END)
+  EXPECT_EQ("link", entityTypeStr(entity, ecm));
 
   entity = ecm.CreateEntity();
   ecm.CreateComponent(entity, components::Visual());
-  EXPECT_EQ("visual", entityTypeStrView(entity, ecm));
+  EXPECT_EQ("visual", entityTypeStr(entity, ecm));
 
   entity = ecm.CreateEntity();
   ecm.CreateComponent(entity, components::Collision());
-  EXPECT_EQ("collision", entityTypeStrView(entity, ecm));
+  EXPECT_EQ("collision", entityTypeStr(entity, ecm));
 
   entity = ecm.CreateEntity();
   ecm.CreateComponent(entity, components::Joint());
-  EXPECT_EQ("joint", entityTypeStrView(entity, ecm));
+  EXPECT_EQ("joint", entityTypeStr(entity, ecm));
 
   entity = ecm.CreateEntity();
   ecm.CreateComponent(entity, components::Sensor());
-  EXPECT_EQ("sensor", entityTypeStrView(entity, ecm));
+  EXPECT_EQ("sensor", entityTypeStr(entity, ecm));
 
   entity = ecm.CreateEntity();
   ecm.CreateComponent(entity, components::Actor());
-  EXPECT_EQ("actor", entityTypeStrView(entity, ecm));
+  EXPECT_EQ("actor", entityTypeStr(entity, ecm));
 
   entity = ecm.CreateEntity();
   ecm.CreateComponent(entity, components::ParticleEmitter());
-  EXPECT_EQ("particle_emitter", entityTypeStrView(entity, ecm));
+  EXPECT_EQ("particle_emitter", entityTypeStr(entity, ecm));
 
   entity = ecm.CreateEntity();
   ecm.CreateComponent(entity, components::Projector());
-  EXPECT_EQ("projector", entityTypeStrView(entity, ecm));
+  EXPECT_EQ("projector", entityTypeStr(entity, ecm));
 }
 
```

This PR:

```
BM_ScopedName/sensor/1000         0.437 ms        0.437 ms         1565
BM_ScopedName/projector/1000      0.457 ms        0.457 ms         1519
BM_ScopedName/model/1000          0.391 ms        0.391 ms         1806
```

Your proposed change of using `string` instead:

```
BM_ScopedName/sensor/1000         0.514 ms        0.514 ms         1345
BM_ScopedName/projector/1000      0.535 ms        0.535 ms         1301
BM_ScopedName/model/1000          0.460 ms        0.460 ms         1505
```

It is 20% slower. That's it, this is part of my first comment of me feeling there wasn't enough due diligence to the comment about strings being efficent because of SSO. Apart from the godbolt links, testing the suggestion shows that it is not the case. Feel free to try the code before / after and let me know if it is different for you.

> `entityTypeStr` has quite a lot of calls like : `_ecm.Component<components::T>(_entity)` which is quite interesting thing. If I see correctly, each call involve several search operations inside ecm.

Yes it does and... The PR descriptions refers to https://github.com/gazebosim/gz-sim/issues/3656 that talks exactly about this, about how all these lookups _are_ expensive and we could consider other strategies such as having a single component that contains the string, or even a string that contains the fully scoped name. But that comes at an increased risk since changing the type means we need to change the string, or if we cache the fully scoped name we need to be able to detect hierarchy changes and propagate to scoped name changes.
Again I was very clear that all these component lookups are a bottleneck, I opened an issue two weeks ago to describe exactly what you are saying, I discussed tradeoffs there and linked the issue in the PR description, so this comment to me feels like that wasn't seen and / or not read through.

> And it happen in order which mentioned in code. I can bet `Light` is less frequent that `Link`.
> 

Yes and... did you see the benchmark that is attached to the PR description? It tests three different categories, sensors projectors and models. Why is that? Because Model is on the [top](https://github.com/gazebosim/gz-sim/blob/09366f48e86d47efe2dbabbdfdfb308fc8d5af1d/src/Util.cc#L343-L346) of the list, Projector is in the [bottom](https://github.com/gazebosim/gz-sim/blob/09366f48e86d47efe2dbabbdfdfb308fc8d5af1d/src/Util.cc#L379-L382), so the benchmark written by @cheriehu was exactly testing this scenario, to see how expensive this chain of `if` statements and `Component` calls are, and you can see that in the varying numbers.
Again this is clear from the PR description + benchmark result, so again this feels like an unnecessary comment. 

> And about flamegraph, for templates and -O2, it can be a lot of confusion in a debug_info, which used to attribute the traces (it became not very accurate in other words)
> 

I see what you are coming from but while you can say that for low level functions, depending on optimization, templates, the [flamegraph](https://private-user-images.githubusercontent.com/94511514/604219676-798d48e0-452a-470b-9a37-0dbc6427c035.svg?jwt=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3ODE3NTE0MjMsIm5iZiI6MTc4MTc1MTEyMywicGF0aCI6Ii85NDUxMTUxNC82MDQyMTk2NzYtNzk4ZDQ4ZTAtNDUyYS00NzBiLTlhMzctMGRiYzY0MjdjMDM1LnN2Zz9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNjA2MTglMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjYwNjE4VDAyNTIwM1omWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTM4MmUyZTRiZmYxMGU3OTkzOWY2NWE1YzA5NTU5Zjk1ZTFmZDM5Njk2Yjk0MzM5MDZmMjg3MmE4MDMzNWQ1MmImWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JnJlc3BvbnNlLWNvbnRlbnQtdHlwZT1pbWFnZSUyRnN2ZyUyQnhtbCJ9.d7AUBDzzDYvd0kEDUgb-EzfQqTdXTqlN38dvqBGslIE) shows `scopedName` itself being a big bottleneck. I do think flamegraphs for high level function will be reliable enough though and again I have done a lot of local tests (i.e. comment out the function, make it return a constant instead of doing any work to see how much of a bottleneck it is) and they all showed that indeed it is taking a lot of time.

> That's why I asking this questions. And instantaneous RTF and RTF deviation is quite fellable for user IMHO.

Yes and I fully acknowledge that but my final question is... what's a better way to do it? I provided a flamegraph that shows the bottleneck, paired with godbolt links to show assembly code reduction, together with a microbenchmark that tests the single function and shows before / after results, together with a manual run to check the RTF and finally I proposed better way to do RTF measurements in the future. I feel I have done enough due diligence for a fairly trivial change. Again that is why I'm pushing for the benchmark in #3671, so we can have a single executable that runs a simulation for X number of iterations and reports how long it took, then we can automate RTF measurements, but we haven't received any feedback on that yet. If there is any better way of measuring RTF and benchmarking "high level" performance I'm happy to discuss it and report findings.

ntfshard:
> Too late, the offense took place!

Ok. Let me start from scratch. It took some time to gather whole picture together.

The first things which caught my eye in the beginning _was performance data._ 
Difference in 20 assembly instructions (even including main, actually functions is shorter) shouldn't be 0.1 ms (or 100000 ns) on current average HW. It just feels not right.
Yes, all instructions are not equal in execution time and CPU has quite a lot of internal states which can affect time, but for more or less regular instructions it should be less than 10 clocks. (it can be even 0 due to pipeline)
CPU usually operating on GHz range, 100 clocks is in worst case 100e-9 or 100 ns

That's why I was concerned about whole program performance in previous messages.

And I extracted this change from `std::string` to `std::string_view` in independent benchmark. No ECM, just 2 object files to avoid compilator overoptimization an google benchmark:
https://godbolt.org/z/rWrdG3ahE

Yes, godbolt is not very good for benchmarking such small values, but it's available for anyone (and downloadable)
One of launches(not the best, not the worst) gave me such results: 
```
-----------------------------------------------------
Benchmark           Time             CPU   Iterations
-----------------------------------------------------
BM_getStr        86.3 ns         86.2 ns      7963758
BM_getView       2.90 ns         1.78 ns    463294341
```
And yes, difference exist. And from one side it's 40 times faster, but from other side it's still nanosecond range (talking exactly about this rough results).

So, I took a look in a benchmark, and it running 1000 times which Google benchmark show as one result.
Usually ppl rely on Google benchmark in statistic aggregation.
https://github.com/gazebosim/gz-sim/pull/3666/changes#diff-0e3357eed1cf2aaeec5820905d981b53d443cc082654d3bb9876084975c78d98R86

So, if we will take initial time of 0.43ms and divide it by 1000 it will be 430 ns per call. And after change it's 380 ns.
Yes, above 10%, but time benefit -50 ns per call. (from `BM_ScopedName/bullet_projector/1000` line)

**This is was mine question, I had feeling that something not match in numbers.**

You a right about some buffering on caller side if you have a lot of invocations of this function during one physics step.
Or maybe it that code should be revisited and this call is not necessary completely. Sometimes better algorithm (replacing square complexity by linear for example) is much better optimization than fight with constant part of complexity.

Also before I mentioned reordering conditions in `entityTypeStr` function due to more possible cases can be higher due to check is linear and it also can bring some gain (but it depends on benchmark, it can try to find links/joints/collisions/visuals as a very frequent things or it can try to find particles_emitter which is very specific things)

src/Util_TEST.cc
```diff
@@ -379,51 +379,51 @@ TEST_F(UtilTest, EntityTypeStr)
   EntityComponentManager ecm;
 
   auto entity = ecm.CreateEntity();
-  EXPECT_TRUE(entityTypeStr(entity, ecm).empty());
+  EXPECT_TRUE(entityTypeStrView(entity, ecm).empty());
```
azeey, line 386: We should keep around the test for `entityTypeStr` even if it's deprecated so that we don't break it by accident. You can use warning suppression macros.

include/gz/sim/Util.hh
```diff
@@ -136,7 +137,29 @@ namespace gz
...
+    /// This function returns a lowercase string for each type. For example,
+    /// "light", "actor", etc.
+    ///
+    /// In case the entity isn't of any known type, this will return an empty
+    /// string.
+    ///
+    /// In case the entity has more than one type, only one of them will be
+    /// returned. This is not standard usage.
+    ///
+    /// Note that this is different from  component type names.
```
iche033:
```suggestion
    /// Note that this is different from component type names.
```

include/gz/sim/Util.hh
```diff
@@ -136,7 +137,29 @@ namespace gz
...
+    ///
+    /// \param[in] _entity Entity to get the type for.
+    /// \param[in] _ecm Immutable reference to ECM.
+    /// \return ID of entity's type-defining components.
```
iche033:
```suggestion
    /// \return Lowercase string of entity's type-defining components.
```

iche033 (approved):
looks good to me. Thanks all for being thorough with the C++ / string optimization analysis and comments. For the changes proposed in this PR, I think there are sufficient benchmarking data to illustrate the benefits for the targeted scenario (headless sim with contact sensors). Approving to get these changes in first. Feel free to create another PR (or ticket) if anyone finds that these changes cause a perf regression or issue.

I made a couple very minor comments, but looks like they already exist in the original function. There are also some conflicts to resolve

src/Util_TEST.cc
luca-della-vedova: https://github.com/gazebosim/gz-sim/pull/3667/commits/0ddf697f7cf04af71cf8c5ba7a931cd2aadc1a7f

I went for a simple warning suppression at test level + duplicated entityTypeStr calls instead of a larger refactor so it's more straightforward later to remove it.

luca-della-vedova:
Will backport but remove the deprecation warning in the cherry-picks so users don't get any new warnings

luca-della-vedova:
@mergifyio backport gz-sim10 gz-sim9 gz-sim8 ign-gazebo6

PR #3668, opened 2026-06-08
Fix issue #3656: refactor entity string conversion

### LINKED ISSUE #3656: Utility function for entity to string conversion performance

Pengkun-ZHU:
# 🎉 New feature

Closes #3656 

## Summary

Cache entity type with an `EntityType` enum and `EntityTypeTag`, set automatically in `CreateComponent`. 
Old approach is kept as a fallback when tag is unknown, e.g., manual ECM setup by plugins, spawners not updated in current phase, etc.
Added in Google benchmark set for metrics and quantization.

## Test it

Toggle code back and forth to run BENCHMARK_scoped_name and compare result before vs after.

## Checklist
- [ ] Added tests

### DIFF
--- a/include/gz/sim/Types.hh
+++ b/include/gz/sim/Types.hh
@@ -77,6 +77,23 @@ namespace gz
       OneTimeChange = 2
     };
 
+    /// \brief A high-level entity classification for fast type lookup.
+    enum class EntityType
+    {
+      Invalid = 0,
+      World = 1,
+      Model = 2,
+      Light = 3,
+      Link = 4,
+      Collision = 5,
+      Visual = 6,
+      Joint = 7,
+      Sensor = 8,
+      Actor = 9,
+      ParticleEmitter = 10,
+      Projector = 11,
+    };
+
     /// \brief A unique identifier for a component type. A component type
     /// must be derived from `components::BaseComponent` and can contain plain
     /// data or something more complex like `gz::math::Pose3d`.
--- a/include/gz/sim/Util.hh
+++ b/include/gz/sim/Util.hh
@@ -139,6 +139,16 @@ namespace gz
     std::string GZ_SIM_VISIBLE entityTypeStr(const Entity &_entity,
         const EntityComponentManager &_ecm);
 
+    /// \brief Set cached entity type on an entity.
+    void GZ_SIM_VISIBLE setEntityTypeTag(EntityComponentManager &_ecm,
+        Entity _entity, EntityType _type);
+
+    /// \brief Get string label for a known EntityType (no ECM lookup)
+    std::string GZ_SIM_VISIBLE entityTypeStr(EntityType _type);
+
+    /// \brief Get component type ID for a known EntityType (no ECM lookup)
+    ComponentTypeId GZ_SIM_VISIBLE entityTypeComponentTypeId(EntityType _type);
+
     /// \brief Get the world to which the given entity belongs.
     /// \param[in] _entity Entity to get the world for.
     /// \param[in] _ecm Immutable reference to ECM.
--- a/include/gz/sim/components/EntityTypeTag.hh
+++ b/include/gz/sim/components/EntityTypeTag.hh
@@ -1,0 +1,40 @@
+/*
+ * Copyright (C) 2026 Open Source Robotics Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+*/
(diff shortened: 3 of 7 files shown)


luca-della-vedova (changes requested):
Please disclose whether AI was used in this PR (I suspect it was?)

This doesn't build because the benchmark was not committed, also adding the component only in `CreateComponent` wouldn't work for code paths that depend on `CreateComponentImplementation` to populate the ECM

Pengkun-ZHU:
Auto completion in Cursor was used during coding (not sure which model it internally uses when it generates the inline suggestions).
The approach is not proposed by AI, usage of enum was to follow your suggestion and I went with template due to its advantage performance-wise plus the ComponentTypeT is itself already templated.

For the missed benchmark, it must be I carelessly omitted it, I don't recall any reason not to include it in commit. I'll later get it back.

And true I did not fully cover all related parts, as this pr was only for ensuring the general approach doesn't deviate from community's expectation, I'll follow up and add if this solution is approved.

azeey:
We're working to change the label to "Assisted By" instead of "Generated By", but for now, please go ahead include "Generated By" in the PR description as well as your commits. If you don't know the model (using Auto mode), I'd suggest looking at your usage history or simply listing all the models the Auto mode can choose from.

Pengkun-ZHU:
Edited and added AI tool related info.

luca-della-vedova (commented):
I think the tricky part of this solution (and somewhat the reason I opened it as an issue ticket, although perhaps shouldn't have been a good first issue since it's a bit complicated!) is book-keeping and consistency.
Two main ideas I can think of:

* If the entity type changes for any reason the enum will be set to the same type. Now this is probably not a large use case but I think we should at least discuss this.
* If the component is created through another path (i.e. `CreateComponentImplementation`, used when deserializing state) this component won't be populated, which will negate any performance improvements in, for example, the GUI.

But in general caching is an interesting point that we should discuss, we could even take this a step further and cache the scopedName altogether (if hierarchy changes are uncommon enough that they are not an issue), but again it all comes with tradeoffs!

Pengkun-ZHU:
> I think the tricky part of this solution (and somewhat the reason I opened it as an issue ticket, although perhaps shouldn't have been a good first issue since it's a bit complicated!) is book-keeping and consistency. Two main ideas I can think of:
> 
> * If the entity type changes for any reason the enum will be set to the same type. Now this is probably not a large use case but I think we should at least discuss this.
> * If the component is created through another path (i.e. `CreateComponentImplementation`, used when deserializing state) this component won't be populated, which will negate any performance improvements in, for example, the GUI.
> 
> But in general caching is an interesting point that we should discuss, we could even take this a step further and cache the scopedName altogether (if hierarchy changes are uncommon enough that they are not an issue), but again it all comes with tradeoffs!

Many thanks for teaching, these are important points.

**Regarding Runtime type changes:**
I had assumed type is immutable after creation, my current hook sets the tag only once and does not update precedence if another type-marker is added later. To support mutable type, we could refresh the tag when type-marker components are added or removed, recomputing the winning type with the same precedence as today’s if-chain (~11 lookups per mutation). That cost should be negligible if mutations are rare. Bit flags could be considered only if that path became a bottleneck.

**Other creation paths (e.g. deserialization):**
I may not know every path, but for full ECM deserialization I treat this largely as backward compatibility: EntityTypeTag is a normal component, so state saved with this implementation should include the tag in the stream and restore it after deserialize. For older streams without the tag, the existing fallback should preserve behavior, we would only miss the performance win, not correctness. 
GUI / live sync is a separate concern, where ECM may get rebuilt from scene or partial component subsets may have markers without a tag, those paths would also use fallback until we populate tags there (e.g. impl hook or post-load refresh).

My think on the tool is, scopedName is hot in physics simulation, optimizing the common case while accepting fallback on rarer paths (old data, partial ECM, type mutation refresh) still seems like a worthwhile tradeoff.

Kindly correct if anything I missed/misunderstood.

PR #3696, opened 2026-06-18
Don't mark components as removed if they were just added

### LINKED ISSUE #3608: Deep dive into Gazebo performance issues

luca-della-vedova:
# 🎉 New feature

Performance improvement, Refer to the Zulip message [#Gazebo General > Benchmarking and performance CI integration @ 💬](https://openrobotics.zulipchat.com/#narrow/channel/526040-Gazebo-General/topic/Benchmarking.20and.20performance.20CI.20integration/near/603964845).

## Summary

Since #2228, as soon as there is any sort of system that sets some of the commands that are removed (i.e. since that PR Linear / Joint / Angular VelocityCmd), the following happens:

* We go through the path of component addition / removal, a bit painful but will hopefully be a lot better if #3447 lands.
* We trigger change detection because of component removal, which means the SceneBroadcaster will be forced to publish a message at every iteration.

As an example, I ran the `joint_controller.sdf` example:

```
$ gz sim -v3 joint_controller.sdf --physics-engine libgz-physics-bullet-featherstone-plugin -r
```

And I echoed one of the messages being printed out.

```
stats {
  sim_time {
    sec: 86
    nsec: 646000000
  }
  real_time {
    sec: 8
    nsec: 310121685
  }
  iterations: 86646
  step_size {
    nsec: 1000000
  }
}
state {
  entities {
    key: 8
    value {
      id: 8
      components {
        key: -7527930132038368260
        value {
          type: 10918813941671183356
          component: "3.37816e-09 -2.06795e-25 0.1 2.72698e-09 -8.75229e-08 -0.0474766"
        }
      }
    }
  }
  entities {
    key: 12
    value {
      id: 12
      components {
        key: -8367970844975482128
        value {
          type: 10078773228734069488
          component: " "
          remove: true
        }
      }
    }
  }
}
```

Note how the message contains the joint state (OK) plus a single component marked as removed. This happens at every single iteration.
The flow of operations is the following:
* PreUpdate: Systems populate their *Cmd component.
* Update: Physics processes it and removes it.
* PostUpdate: SceneBroadcaster detects that the component was removed and triggers a state send, with the content above.

I would argue (and I did in Zulip, together with a conversation with @azeey), that the information "component X was removed" is useless if component X was not present in the first place the last time the `SceneBroadcaster` was ran. The user didn't even have a chance to know that it was added!

This is indeed what this PR addresses, if a component was added and removed in the same step (so if no `ClearRemovedComponents` was called), don't treat it as a change.
Previously, because of the change detection, we would always fall in [this branch](https://github.com/gazebosim/gz-sim/blob/09366f48e86d47efe2dbabbdfdfb308fc8d5af1d/src/systems/scene_broadcaster/SceneBroadcaster.cc#L418-L421), which would cause the state to be sent at every step.
By contrast, now regardless of the RTF, the SceneBroadcaster publishes at 60 Hz.
The impact of  this is very significant on worlds that contain any sort of robot controller (so really all useful worlds out there), and becomes more significant as the RTF scales up (publish rate is now fixed at 60 Hz, instead of being once per simulation iteration).

## Backport Policy
I _believe_ this is most useful from Harmonic onwards since the PR that added component removal was merged in gz-sim8, but I might be wrong.
    - [x] Jetty
    - [x] Ionic
    - [x] Harmonic
    - [ ] Fortress
- [ ] I am not sure

## Test it

Before and after this PR run a demo world that includes a robot controller, as fast as possible, i.e.:

```
$ gz sim -v3 joint_controller.sdf --physics-engine libgz-physics-bullet-featherstone-plugin -r -z 10000000
```

In another shell, print the frequency of the messages:

```
$ gz topic -f -t /world/default/state
```

Check the RTF before and after this PR. For me, the simulation before this PR runs at 1000% RTF and has this frequency of (mostly useless) messages:

```
$ gz topic -f -t /world/default/state
average rate: 13135.4
```

After the PR, it runs at 2000% RTF so *double the performance* (!) and the rate of state updates is fixed at 60 Hz.

```
$ gz topic -f -t /world/default/state
average rate: 59.5
```

## Checklist
- [x] Added tests


arjo129 (commented):
In general, I think this seems like a good idea (and an extremely interesting finding). I'm trying to think of potential behaviour changes that could occur. If a plugin developer were using `EachRemoved<>` to detect removals from a `PreUpdate`/`Update` step in a `PostUpdate` step I could see minor behaviour changes. AFAIK this should be relatively harmless as the worst case scenario is that we run the simulation for one more time step before the removal logic kicks in but Id let this cook a bit in rotary before back porting. The nice side effect of this change is that it would reduce races where system order could change behaviour within a time step.

Code style wise the changes generally LGTM.

luca-della-vedova:
> In general, I think this seems like a good idea (and an extremely interesting finding). I'm trying to think of potential behaviour changes that could occur. If a plugin developer were using `EachRemoved<>` to detect removals from a `PreUpdate`/`Update` step in a `PostUpdate` step I could see minor behaviour changes. AFAIK this should be relatively harmless as the worst case scenario is that we run the simulation for one more time step before the removal logic kicks in but Id let this cook a bit in rotary before back porting. The nice side effect of this change is that it would reduce races where system order could change behaviour within a time step.
> 
> Code style wise the changes generally LGTM.

To my understanding the `EachRemoved` API operates on removed _entities_, not components, [here](https://github.com/gazebosim/gz-sim/blob/6011926799d6c87572085e919ea68e61a0573902/include/gz/sim/detail/EntityComponentManager.hh#L489), this PR operates purely on removed component detection and the only API that will behave differently is the one the `SceneBroadcaster` was using, which is `HasRemovedComponents`. I didn't find any user of that API other than `SceneBroadcaster` in our codebase.
What this PR could break (like every PR that is a behavior change! I'll skip the usual Ctrl-key XKCD) is users that care about the use case mentioned in the PR description, so they want to know that in a simulation step, before their system was run, a component with a certain ID was added, then removed, even if they didn't ever have access to its value.

arjo129 (approved):
Thanks for this finding and sorry for the confusion earlier. This is a very interesting solution! LGTM.

src/EntityComponentManager.cc
```diff
@@ -921,7 +935,14 @@ bool EntityComponentManager::RemoveComponent(
...
     std::lock_guard<std::mutex> lock(this->dataPtr->removedComponentsMutex);
-    this->dataPtr->removedComponents[_entity].insert(_typeId);
+    const auto addedCompIt = this->dataPtr->addedComponents.find(_entity);
+    if (addedCompIt == this->dataPtr->addedComponents.end() ||
+        addedCompIt->second.find(_typeId) == addedCompIt->second.end())
+    {
+      // Only add the component to the removedComponents map if it was not
+      // added in this iteration to reduce false positives.
+      this->dataPtr->removedComponents[_entity].insert(_typeId);
+    }
```
azeey, line 960: Wouldn't it be safer to apply this logic in `SetRemovedComponentsMsgs` since it will only affect what's serialized? Doing this here will prevent `removedComponents` from being updated, which changes the behavior of `HasRemovedComponents`.

luca-della-vedova: The change in behavior of `HasRemovedComponents` is exactly the intent of this PR, it returns true at every simulation step right now that [forces the StateBroadcaster to publish](https://github.com/gazebosim/gz-sim/blob/f2fcfdfdf6a105337a3001869982102ad872e1b7/src/systems/scene_broadcaster/SceneBroadcaster.cc#L401-L403) the state.

arjo129 (changes requested):
Based on further discussion there is debate about the behaviour change in the API. One option was to introduce a new API so that a behaviour change is not introduced in `HasRemovedComponents`. For this reason we should hold of on merging this till we come to a consensus.

azeey:
I think this is a behavior change. My initial assessment was only considering what was serialized and sent over the wire to the GUI. In that case, since the GUI ECM never knew the component existed, we could say there would be no behavior change. However, I've since realized that the scope of this change is wider than just the GUI.  And since it's trivial to add a new API that does what we want, I don't think it's worth the rist of breaking the old behavior.

Here are some suggestions for the new function name from Gemini.

1. `HasRemovedExistingComponents` (highlights that the component existed before the current iteration began)
1. `HasNetRemovedComponents` (uses the concept of net change from the start to the end of the pass)
1. `HasLostComponents` (implies the entity possessed the component at the start of the process)
1. `HasRemovedPriorComponents` (explicitly focuses on components belonging to a prior iteration)

I'd lean toward (1) as the clearer choice. (4) would be my secondary choice.

luca-della-vedova:
Tried in https://github.com/gazebosim/gz-sim/pull/3696/commits/807eaabe6b777039670d10cc6565d5aa453a4928 so the behavior of the previous API is unchanged.
I'm very ambivalent about this, I would be OK with this being the way to backport this to other branches safely but I wonder if we could get rid of it in `main`, the main (pun intended) reasons:

* It's yet another map about removed components. We already have `removedComponents` and `componentsMarkedAsRemoved` that are somewhat confusing, now we need another `allRemovedComponents` map, three maps with different sorts of removed components!
* The only public API for these hash maps is actually the boolean function, then they are an implementation detail of the `State` call that changes how it behaves (specifically through the `SetRemovedComponentsMsg` private function). Either we:
  *  Add a new parameter / `State` function to specify which type of behavior we want from removed components, include all or only include the ones that were not added in this step.
  * Store the behavior we want for removed component serialization and have a public getter / setter.
  * Have only one behavior and change it to what this PR originally did, which would mean users wouldn't get any sort of information about which components were added and removed in this step, they can only access the boolean API.
  * Have only one behavior and keep it unchanged (so serialize all removed components), which has the opposite issue, users don't have any way to know which previously existing components has been removed and State has all removed components. They would also have potentially a bunch of serialization overhead for information that is not relevant.
* With the idea that "the only exposed behavior is a bool API" I also tried https://github.com/gazebosim/gz-sim/pull/3696/commits/f870eb5d2e34c4fa571923667a0dbc73d7856d1b to avoid having yet another hash map, but as noted in the test comment it adds, it has the problem above (the behavior of State is ambiguous, either we include all removed components or only include the "previously existing ones").

PR #3711, opened 2026-06-22
Use components::ParentEntity for parent lookup

### LINKED ISSUE #3608: Deep dive into Gazebo performance issues

luca-della-vedova:
Part of #3447, minor performance improvement in `EntityComponentManager::ParentEntity` that is somewhat in a hot path in physics, specifically in this snippet where we iterate over every single link and get its parent. https://github.com/gazebosim/gz-sim/blob/c017986be541e95f4398dfaa60ca718958d83002/src/systems/physics/Physics.cc#L3785-L3793

This is also a correctness improvement and future proofing, as explained below.

## Summary

With this change we always make sure the `components::ParentEntity` is kept in sync with the graph by changing the `SetParentEntity` API to also add the component.
There is a bit of duplication since now the function that updates the graph is private, but the graph won't exist anymore if #3447 lands so that function will be gone.
There are two main reasons for this change:

### Performance

A minor improvement since now getting the parent is a component lookup which is a fair bit more efficient than the gz-math call. Specifically this is a flamegraph for the jetty world

jetty_bullet_new_main (1)

Doing a lot of zooming into `UpdatePhysics` shows that quite a lot of time is taken in `ParentEntity` because it needs to do lookups in (ordered) maps + do a bunch of expensive API calls.
I don't have a flamegraph for this change but I can AI generate a micro benchmark if it's of interest.

### Correctness

Perhaps more important, previously we had two sources of truth for parent information. The `EntityComponentManager::ParentEntity(Entity)` API and reading the `components::ParentEntity` component via `EntityComponentManager::Component<components::parentEntity>(Entity)`.
Furthermore, setting the parent through `SetParentEntity` API required users to remember that they need to also create the component if they want the data to be consistent, which is prone to errors.

This is clear from the diff where there are a lot of places where users had to call `SetParentEntity` followed by `CreateComponent<components::ParentEntity>`. And this is even more obvious in the cases where this was not done and as such were not correct. For example, by grepping for `SetParentEntity` I found two cases in battery modeling where the user forgot to call `CreateComponent` so we had a discrepancy between the hierarchy saved in the components and the one saved in the graph:

https://github.com/gazebosim/gz-sim/blob/c017986be541e95f4398dfaa60ca718958d83002/src/systems/battery_plugin/LinearBatteryPlugin.cc#L297-L303

and

https://github.com/gazebosim/gz-sim/blob/c017986be541e95f4398dfaa60ca718958d83002/src/systems/thruster/Thruster.cc#L687-L692

### Future-proofing / Entt refactor

The entt refactor uses exactly this approach of being purely component based and deletes the graph altogether, so getting this in separately will reduce the diff / risk of regressions with the refactor.

## Backport Policy
I think this is safe since CI passes, performance is slightly improved and the code is slightly more correct (in the battery plugins I linked), but I might be unaware of some expectations of other parts of the code.
    - [ ] Jetty
    - [ ] Ionic
    - [ ] Harmonic
    - [ ] Fortress
- [x] I am not sure
## Test it

Automated through CI

## Checklist
- [ ] Added tests

### DIFF
--- a/src/EntityComponentManager.cc
+++ b/src/EntityComponentManager.cc
@@ -1264,11 +1264,7 @@ bool EntityComponentManager::CreateComponentImplementation(
       return updateData;
     }
     const auto *parent = static_cast<const components::ParentEntity *>(_data);
-    if (!this->dataPtr->SetParentEntityGraph(_entity, parent->Data()))
-    {
-      gzerr << "Failed setting parent for entity " << _entity << " to "
-            << parent->Data() << std::endl;
-    }
+    this->dataPtr->SetParentEntityGraph(_entity, parent->Data());
   }
 
   return updateData;
--- a/src/EntityComponentManager.cc
+++ b/src/EntityComponentManager.cc
@@ -924,6 +924,12 @@ bool EntityComponentManager::RemoveComponent(
     this->dataPtr->removedComponents[_entity].insert(_typeId);
   }
 
+  // If the component is a components::ParentEntity, leave the entity parentless
+  if (_typeId == components::ParentEntity::typeId)
+  {
+    this->dataPtr->SetParentEntityGraph(_entity, kNullEntity);
+  }
+
   return true;
 }
 
@@ -1090,17 +1096,21 @@ Entity EntityComponentManager::ParentEntity(const Entity _entity) const
 bool EntityComponentManager::SetParentEntity(const Entity _child,
     const Entity _parent)
 {
-  bool successful = this->dataPtr->SetParentEntityGraph(_child, _parent);
+  // Validate input, child must exist and parent must either be kNullEntity
+  // (in which case we remove the parent) or a valid entity to set.
+  if (!this->HasEntity(_child))
+    return false;
+  if (_parent != kNullEntity && !this->HasEntity(_parent))
+    return false;
+
+  // Component creation and deletion take care of updating the graph
+  this->RemoveComponent<components::ParentEntity>(_child);
   if (_parent == kNullEntity)
   {
-    this->RemoveComponent<components::ParentEntity>(_child);
     return true;
   }
-  else if (successful)
-  {
-    this->CreateComponent(_child, components::ParentEntity(_parent));
-  }
-  return successful;
+  this->CreateComponent(_child, components::ParentEntity(_parent));
+  return true;
 }
 
 /////////////////////////////////////////////////
--- a/src/EntityComponentManager.cc
+++ b/src/EntityComponentManager.cc
@@ -1257,8 +1257,18 @@ bool EntityComponentManager::CreateComponentImplementation(
   // update the entities graph.
   if (_componentTypeId == components::ParentEntity::typeId)
   {
+    if (!_data)
+    {
+      gzerr << "Internal error: Invalid parent component detected, "
+            << "this should not happen." << std::endl;
+      return updateData;
+    }
     const auto *parent = static_cast<const components::ParentEntity *>(_data);
-    this->dataPtr->SetParentEntityGraph(_entity, parent->Data());
+    if (!this->dataPtr->SetParentEntityGraph(_entity, parent->Data()))
+    {
+      gzerr << "Failed setting parent for entity " << _entity << " to "
+            << parent->Data() << std::endl;
+    }
   }
 
   return updateData;
--- a/src/EntityComponentManager.cc
+++ b/src/EntityComponentManager.cc
@@ -74,6 +74,15 @@ class gz::sim::EntityComponentManagerPrivate
   /// `AddEntityToMessage`.
   public: void CalculateStateThreadLoad();
 
+  /// \brief Helper function to only update the entity graph for a new
+  /// parent - child relation. This does not do any book-keeping on the
+  /// components::ParentEntity component, for which the EntityComponentManager
+  /// SetParentEntity function should be used.
+  /// \param[in] _parent the new parent of the entity, or kNullEntity if the
+  /// entity should be left parentless.
+  /// \param[in] _child the entity to update the parent for.
+  public: bool SetParentEntityGraph(const Entity _parent, const Entity _child);
+
   /// \brief Copies the contents of `_from` into this object.
   /// \note This is a member function instead of a copy constructor so that
   /// it can have additional parameters if the need arises in the future.
@@ -515,7 +524,6 @@ Entity EntityComponentManager::CloneImpl(Entity _entity, Entity _parent,
   if (_parent != kNullEntity)
   {
     this->SetParentEntity(clonedEntity, _parent);
-    this->CreateComponent(clonedEntity, components::ParentEntity(_parent));
   }
 
   // make sure that the cloned entity has a unique name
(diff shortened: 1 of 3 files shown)


src/EntityComponentManager.cc
```diff
@@ -74,6 +74,15 @@ class gz::sim::EntityComponentManagerPrivate
...
   public: void CalculateStateThreadLoad();
 
+  /// \brief Helper function to only update the entity graph for a new
+  /// parent - child relation. This does not do any book-keeping on the
+  /// components::ParentEntity component, for which the EntityComponentManager
+  /// SetParentEntity function should be used.
+  /// \param[in] _parent the new parent of the entity, or kNullEntity if the
+  /// entity should be left parentless.
+  /// \param[in] _child the entity to update the parent for.
+  public: bool SetParentEntityGraph(const Entity _parent, const Entity _child);
```
caguero: I think we should swap `_parent` and `_child` here (and reorder the doxygen) as the definition is using the reverse order.

src/EntityComponentManager.cc
```diff
@@ -1072,24 +1080,39 @@ bool EntityComponentManager::HasEntity(const Entity _entity) const
...
+{
+  bool successful = this->dataPtr->SetParentEntityGraph(_child, _parent);
+  if (_parent == kNullEntity)
+  {
+    this->RemoveComponent<components::ParentEntity>(_child);
+    return true;
+  }
+  else if (successful)
+  {
+    this->CreateComponent(_child, components::ParentEntity(_parent));
```
caguero: Doesn't this internally call `SetParentEntityGraph()` again? How about something like:

```
  if (_parent == kNullEntity) {
    this->dataPtr->SetParentEntityGraph(_child, kNullEntity);
    this->RemoveComponent<components::ParentEntity>(_child);
    return true;
  }
  return this->CreateComponent(_child, components::ParentEntity(_parent)) != nullptr;
```

src/SdfEntityCreator.cc
caguero: Should we now replace this call with:
```
this->dataPtr->ecm->SetParentEntity(childEntity, kNullEntity);
```

so the component and the entity graph stay in sync.

caguero (commented):
Added a few comments.

src/SdfEntityCreator.cc
iche033: +1, should we provide a `RemoveParentEntity` helper function to keep the graph and ECM in sync?

src/EntityComponentManager.cc
luca-della-vedova: Ahhhhhhhh I really wish the compiler caught this kind of issues, thanks! https://github.com/gazebosim/gz-sim/pull/3711/commits/3d0ff6d47baa6bb590b65bfe3aa0d098c518ad68

src/SdfEntityCreator.cc
luca-della-vedova: Ah interesting, very nice finding! I suppose this is another case where previously the component was correct but the graph broken?
FIxed in https://github.com/gazebosim/gz-sim/pull/3711/commits/3f0bf167e59b5e44537ee469b7ae62c6567f073c.
I'm OK either way for the API, it is available as Carlos suggested  , where setting the parent to kNullEntity will actually remove the parent, but if it's not clear enough we could add something like:

```
bool RemoveParentEntity(const Entity _entity)
{
    return this->SetParentEntity(_entity, kNullEntity);
}
```

We would end up with two ways to remove parents then though.
In general, I'm not a big fan of "magic values" that users need to be subtly aware of by reading the docs or the code. In an ideal (for me) world we would converge into an API where the type itself makes it clear what the function does, either:

```
// Option 1, single API
// Clear from looking at the definition that the second argument is optional, fairly intuitive that setting to std::nullopt might remove the parent
bool SetParentEntity(const Entity _child, const std::optional<Entity> _parent);

// Option 2, two APIs
// _parent MUST be a valid entity or this will fail
bool SetParentEntity(const Entity _child, const Entity _parent);
bool RemoveParentEntity(const Entity _entity);
```

src/EntityComponentManager.cc
luca-della-vedova: Yes! This sent me into an interesting rabbit hole that resulted in a bit of rearchitecting how parent-hood is done in https://github.com/gazebosim/gz-sim/pull/3711/commits/de655259e3ddd0f199cf65f33d0fded19410bcad.
Specifically I added some input validation since the change you proposed (and also the change I implemented) both ignore the return value of the graph updating function, which means we need to catch the failure scenarios (invalid entities) earlier.

I added a similar check into `RemoveComponent` to detect when a `components::ParentEntity` component is being removed which updates the graph to leave the entity parentless.
This is symmetrical with the change in `CreateComponent` that, when a `components::ParentEntity` is detected it sets the graph.
As such, now all the `SetParentEntity` function has to do is just delete + recreate the component. As a side effect this will also automatically fix any other instances such as the one you found in SdfEntityCreator, where `components::ParentEntity` was removed without explicitly updating the graph.
So now creating or deleting a component is pretty much the same as this API call and both should work in an equivalent way, the only difference of this API is that it returns a boolean and fails if the proposed relationship is invalid, something that component creation cannot do.

The only possible scenario that is not fixed by this and also won't be fixed by the Entt refactor but has been existing since the dawn of this implementation is if users do something like:

```
auto* comp = ecm.Component<components::ParentEntity>(_entity);
comp->Data() = 1234;
```

I... will leave this as something for another day. Entt can make it a bit more robust through component hooks but it's a bit out of scope.

src/EntityComponentManager.cc
```diff
@@ -1224,8 +1257,8 @@ bool EntityComponentManager::CreateComponentImplementation(
   // update the entities graph.
   if (_componentTypeId == components::ParentEntity::typeId)
   {
-    auto parentComp = this->Component<components::ParentEntity>(_entity);
-    this->SetParentEntity(_entity, parentComp->Data());
+    const auto *parent = static_cast<const components::ParentEntity *>(_data);
```
caguero, line 1266: Nit: Should we check `if (_data)` before the casting?

caguero (approved):

luca-della-vedova:
What a journey!
I added the nullptr check and I thought I would be extra careful and actually log if the setting of the parent entity fails in the same commit.
That made the `ModelCommandAPI` unit test fail (???) which surfaced another inconsistency that existed for a long time!
It seems that when setting components through the workflow used for both GUI / ModelCommandAPI of:

* Use `State()` function to get the whole serialized ECM state.
* Use `SetState(state)` to deserialize it back.

When we serialize we iterate over an `unordered_map`, which means that the entities will be received in a random order, which means that the function might fail (if we try to add a child before a parent) and the hierarchy might be incorrect.
The error only became evident because I added a print statement, otherwise that function was actually failing all the time when using that workflow, meaning that most like both the UI / ModelCommandAPI ECM had wrong hierarchy information all along.
This also explains why, when I tried to switch over EntitiesByComponents for ChildrenByComponents call, ModelCommandAPI was not working and I had to revert it to read the component and not the graph https://github.com/gazebosim/gz-sim/pull/3362#issuecomment-3995059151.

I'll open an issue ticket with a detailed explanation but would recommend this in a followup since it probably existed since the existence of this library.

luca-della-vedova:
This is the issue https://github.com/gazebosim/gz-sim/issues/3729

caguero:
About backporting, I don't see any potential ABI break, the changes look clean to me.

However, I see a few behavior changes:
* `SetParentEntity()` now creates or removes the `ParentEntity` component. Before we were just updating the internal graph. 
* `RemoveComponent<ParentEntity>` now detaches the entity graph as well. Before we weren't doing it.

Strictly speaking we shouldn't change behavior, but perhaps in these cases that we're shipping a more correct version it's OK if we document it in the `Migration.md`.

azeey:
I agree with @caguero. Going with the [principle of least surprise](https://en.wikipedia.org/wiki/Principle_of_least_astonishment), I think the least surprising thing for our users is the behavior in this PR, so I support backporting it.

luca-della-vedova:

PR #3481, opened 2026-04-15
Use protobuf Arena allocation in publisher systems

caguero:
This patch adopts [Protobuf  Arena](https://protobuf.dev/reference/cpp/arenas/) allocation in four publisher systems to reduce per-step message-construction cost. The change replaces stack/heap-allocated message trees with arena-allocated ones, which lets sub-messages be bump-pointer allocated and freed in a single block.

I measured the impact of this patch with our gz-common profiler taking 2000 samples. I used a custom world with one model and ~100 links and ~100 joints. Each test drops the first 200 as warmup, considering 1800 samples effectively. I computed the median sample per run and repeated 3 separate runs. I ran it on Ubuntu 24.04.

### Baseline (no arena code)

```
JointStatePublisher::PostUpdate    1800   p50=185521  p50=189841  p50=189863  → 189 µs
PosePublisher::PostUpdate          1800   p50=136361  p50=148169  p50=147045  → 144 µs
PosePublisher::PublishPoses        1800   p50=36283   p50=38209   p50=37746   →  37 µs
SceneBroadcast::PoseUpdate         1800   p50=77212   p50=78963   p50=79955   →  79 µs
```

### Arena (this PR)

```
JointStatePublisher::PostUpdate    1800   p50=180004  p50=175759  p50=176778  → 178 µs
PosePublisher::PostUpdate          1800   p50=117658  p50=126601  p50=125570  → 123 µs
PosePublisher::PublishPoses        1800   p50=15403   p50=15753   p50=15587   →  15 µs
SceneBroadcast::PoseUpdate         1800   p50=70760   p50=70902   p50=71813   →  71 µs
```

### Wall time per 2,000-iteration run

| | run 1 | run 2 | run 3 | avg |
|---|---|---|---|---|
| Baseline | 1738 ms | 1705 ms | 1698 ms | 1714 ms |
| Arena | 1608 ms | 1677 ms | 1626 ms | 1637 ms |

Wall time gain: **−4.5%** end-to-end. 

As a sanity check I also ran the tests with the same world but with 1 link and joint:

### 1-joint world results

```
                            Baseline p50    Arena p50    Δ
JSP::PostUpdate                   4.65 µs      4.46 µs   −4%
PP::PostUpdate                    4.32 µs      4.12 µs   −4.5%
PP::PublishPoses                  1.18 µs      0.83 µs   −29%
SB::PoseUpdate                    1.88 µs      1.87 µs   ±0% (noise)
```

## Test it

[...]

PR #3712, opened 2026-06-22
Add CI for Bazel - Clang, Fix layering failure

luca-della-vedova:
# 🦟 Bug fix

Fix compiles on Bazel / Clang + add a CI step to find future regressions.

## Summary

#3481 broke Bazel + Clang builds where the layering step reports errors such as:

```
bazel-out/haswell-opt/bin/external/gz-sim+/src/systems/pose_publisher/PosePublisher_static_plugin.cc:20:10: error: module gz-sim+//:gz-sim-pose-publisher-system-static does not depend on a module exporting 'google/protobuf/arena.h'
   20 | #include <google/protobuf/arena.h>
```

It seems this wasn't caught by CI because we don't test Bazel + Clang, so this PR also adds a step to the Bazel CI builds to make sure we test that combination.
The first commit only adds the CI, you can find a failure in its new Bazel + Clang step https://github.com/gazebosim/gz-sim/actions/runs/27944359735/job/82685279109.
The second commit adds the actual fix. Here is a run for it showing green Bazel builds https://github.com/gazebosim/gz-sim/actions/runs/27949882252/job/82703886503?pr=3712.

## Backport Policy
    - [x] Jetty
    - [x] Ionic
    - [x] Harmonic
    - [ ] Fortress
- [ ] I am not sure

## Checklist
- [ ] Added tests

Generated-by: Gemini 3

### DIFF
--- a/.github/workflows/bazel.yml
+++ b/.github/workflows/bazel.yml
@@ -23,19 +23,4 @@ jobs:
           {"folder": ".", "bzlmodEnabled": false},
         ]
       exclude_windows: true
-      bazel_test_command: "bazel test --test_output=errors //..."
-
-  test-clang:
-    uses: bazel-contrib/.github/.github/workflows/bazel.yaml@v7.2.2
-    with:
-      folders: |
-        [
-          ".",
-        ]
-      exclude: |
-        [
-          {"folder": ".", "bzlmodEnabled": false},
-          {"folder": ".", "os": "macos-latest"},
-        ]
-      exclude_windows: true
       bazel_test_command: "bazel test --repo_env=CC=clang --repo_env=CXX=clang++ --test_output=errors //..."
--- a/.github/workflows/bazel.yml
+++ b/.github/workflows/bazel.yml
@@ -24,3 +24,18 @@ jobs:
         ]
       exclude_windows: true
       bazel_test_command: "bazel test --test_output=errors //..."
+
+  test-clang:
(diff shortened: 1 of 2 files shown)


iche033:
looks good in general. Just a note that with this we'll have a lot of bazel github actions running per PR. So I wonder if we should just keep the clang bazel builds as they are usually stricter than the gcc builds?

 > This is safe to backport to the following versions:
 >    ...
 >    [x] Fortress

 btw Fortress does not have bazel support so we don't need to backport the changes there

luca-della-vedova:
> looks good in general. Just a note that with this we'll have a lot of bazel github actions running per PR. So I wonder if we should just keep the clang bazel builds as they are usually stricter than the gcc builds?
> 
> > This is safe to backport to the following versions:
> > ...
> > [x] Fortress
> 
> btw Fortress does not have bazel support so we don't need to backport the changes there

Removed fortress from the backport. I'm not familiar enough to know whether gcc / clang behave so differently that we should keep both, I was using gz-sim on Clang and encountered this hence the PR.
It is true though that at least for this type of issues Clang seems more strict so perhaps adopting it as a default should be OK? Did it in https://github.com/gazebosim/gz-sim/pull/3712/commits/6795e307bf21552c4bc6447d4a1f668596b35cb8 CC @shameekganguly in case you want to chime in on the bazel / gcc / clang CI!

iche033 (approved):

luca-della-vedova:

PR #3671, opened 2026-06-10
Add benchmark for Server.Run operation 

cheriehu:
# 🎉 New feature

## Summary

Add benchmark to run a few worlds and time how long 1 step takes. Timer starts after a number steps to only measure performance after the simulation has reached a "steady-state". This delay seems to matter especially for the bullet-featherstone physics engine which can take a while before it reaches optimal performance.

## Backport Policy

- [x] This is safe to backport to the following versions:
    - [x] Jetty
    - [x] Ionic
    - [x] Harmonic
    - [x] Fortress
- [ ] This **should not** be backported
- [ ] I am not sure
- [ ] Other (fill in yourself)

## Test it

1. `colcon build --merge-install --packages-up-to gz-sim`
2. `build/gz-sim/bin/BENCHMARK_server_run`, optionally add the `--benchmark_filter="(/sdf)"` argument to run the benchmark without the 3k_shapes.sdf world, which can take a few minutes.
3. These are the results I get: 

```
--------------------------------------------------------------------------------------------
Benchmark                                                  Time             CPU   Iterations
--------------------------------------------------------------------------------------------
BM_RuntimeWorld/sdf_shapes_bullet/10                   0.011 ms        0.011 ms        60259
BM_RuntimeWorld/sdf_shapes_dart/1                      0.132 ms        0.132 ms         4868
BM_RuntimeWorld/sdf_gpu_lidar_sensor_bullet/10         0.010 ms        0.010 ms        52192
BM_RuntimeWorld/sdf_gpu_lidar_sensor_dart/1            0.020 ms        0.020 ms        35199
BM_RuntimeWorld/lengthy_sdf_3k_shapes_bullet/3000      0.626 ms        0.626 ms         1106
BM_RuntimeWorld/lengthy_sdf_3k_shapes_dart/1            30.3 ms         30.3 ms           24
BM_LoadWorld/sdf_shapes_bullet                           108 ms         85.7 ms            6
BM_LoadWorld/sdf_shapes_dart                             112 ms         89.6 ms            8
BM_LoadWorld/sdf_gpu_lidar_sensor_bullet                 181 ms          141 ms            5
BM_LoadWorld/sdf_gpu_lidar_sensor_dart                   182 ms          142 ms            5
BM_LoadWorld/lengthy_sdf_3k_shapes_bullet               5831 ms         4653 ms            1
BM_LoadWorld/lengthy_sdf_3k_shapes_dart                 6394 ms         5042 ms            1
```

## Checklist
[...]

PR #3714, opened 2026-06-23
Performance: don't create new msg::Contacts when adding contacts to ContactSensorData component

### LINKED ISSUE #3608: Deep dive into Gazebo performance issues

cheriehu:
# 🦟 Bug fix

Performance improvement for world with contacts. 

## Summary

When contacts are enabled, a significant percentage of the time is spent destroying Contacts (`gz::msgs::Contacts::~Contacts`) in the UpdateCollisions function. This is because a new `msgs::Contacts contactsComp;` message is created in an _ecm.Each() call and gets destroyed at the end of the function. 

~This can be improved by using protobuf arenas which allows for faster-on-average message allocation and destruction compared to heap-allocated proto messages. This approach is similar to the one in #3481.~

EDIT: No longer using protobuf arenas. This change now adds contacts directly to the Data() object in the ContactSensorData component with `_contacts->Data().add_contact();` instead of adding contacts to a new `contactsComp` message and then calling `SetData()` to update the ContactSensorData component data. 

I also added a benchmark based off #3671 which measures the time taken for `Server.Run()` with contacts enabled.

## Test it
1. `colcon build --merge-install --packages-select gz-sim`
2. `./build/gz-sim/bin/BENCHMARK_server_run --benchmark_filter="(BM_RuntimeWorldContacts)"`
3. I added [this](https://github.com/gazebosim/gz-sim/blob/454a0fe470cd065e0402fbbc9f5ffb0508ba5184/test/worlds/contact_without_entity_names.sdf#L12) in 3k_shapes.sdf to disable scopedName (which takes a lot of time, see #3656) to see the difference more clearly since this optimization has nothing to do with scopedName. 

### My results
**On main + this benchmark**
```
----------------------------------------------------------------------------------------------------
Benchmark                                                          Time             CPU   Iterations
----------------------------------------------------------------------------------------------------
BM_RuntimeWorldContacts/bullet_shapes_sdf/10                   0.009 ms        0.009 ms        67655
BM_RuntimeWorldContacts/bullet_gpu_lidar_sensor_sdf/10         0.013 ms        0.013 ms        39812
BM_RuntimeWorldContacts/bullet_breadcrumbs_sdf/10              0.061 ms        0.060 ms        11256
BM_RuntimeWorldContacts/lengthy_bullet_3k_shapes_sdf/3000       7.43 ms         7.42 ms           87
```

**On this branch**
```shell
----------------------------------------------------------------------------------------------------
Benchmark                                                          Time             CPU   Iterations
----------------------------------------------------------------------------------------------------
BM_RuntimeWorldContacts/bullet_shapes_sdf/10                   0.008 ms        0.008 ms        86071
BM_RuntimeWorldContacts/bullet_gpu_lidar_sensor_sdf/10         0.010 ms        0.010 ms        56524
BM_RuntimeWorldContacts/bullet_breadcrumbs_sdf/10              0.051 ms        0.051 ms        13962
BM_RuntimeWorldContacts/lengthy_bullet_3k_shapes_sdf/3000       5.50 ms         5.49 ms          116
```

** The benchmarks took me about 2 minutes to run

## Backport Policy

    - [ ] Jetty
    - [ ] Ionic
    - [ ] Harmonic
    - [ ] Fortress
- [X] I am not sure

## Checklist
- [X] Added tests
- [X] Was GenAI used to generate this PR? If so, make sure to add "Generated-by" to your commits. (See [this policy](https://osralliance.org/wp-content/uploads/2025/05/OSRF-Policy-on-the-Use-of-Generative-Tools-Generative-AI-in-Contributions.pdf) for more info.)

Generated-by: Gemini 3


caguero (commented):
Should we also run the benchmark with a world where the number of contacts is small?

cheriehu:
> Should we also run the benchmark with a world where the number of contacts is small?

Yep I think you're right, edited the PR to show the results for the rest of the worlds

test/benchmark/server_run.cc
```diff
@@ -0,0 +1,201 @@
...
+        if (_ecm.EntityHasComponentType(_entity,
+              components::ContactSensorData::typeId))
+        {
+          return true;
+        }
+        // Enable collision by creating the ContactSensorData component.
+        _ecm.CreateComponent(_entity, components::ContactSensorData());
+        return true;
+      });
+  });
```
luca-della-vedova, line 97: We should have some way to short circuit this query to avoid it having to go through the whole world in every simulation step and potentially adding too much overhead to the benchmark.
On the top of my head we could either have an `EachNew` query that will only run for new entities, hence do nothing after initialization is finished, or some sort of boolean flag that can be used to short circuit the iteration once initialized.

cheriehu: Makes sense! I went with the boolean flag because I think none of the entities are considered new since we're adding the system after stabilizing? When I replaced `Each` with `EachNew` contacts didn't get enabled.

luca-della-vedova: Ah I think this might be a system order issue :| EachNew on PreUpdate is tricky to get right. Boolean works as well!

src/systems/physics/Physics.cc
```diff
@@ -4613,11 +4618,19 @@ void PhysicsPrivate::UpdateCollisions(EntityComponentManager &_ecm)
...
+          google::protobuf::Arena::CreateMessage<msgs::Contacts>(
+            &this->contactsArena)
+        #endif
+        };
+
         if (entityContactMap.find(_collEntity1) == entityContactMap.end())
         {
           // Clear the last contact data
-          auto state = _contacts->SetData(contactsComp,
+          auto state = _contacts->SetData(*contactsComp,
```
azeey: Can we just clear the existing protobuf data in the component rather than assigning it? If we do that, we can do the allocation of `contactsComp` below this `if` block.

azeey (commented):
I realized while looking into this PR that there should already be a `msgs::Contacts` object in the component since each `Component<T>` default constructs the type it holds.

https://github.com/gazebosim/gz-sim/blob/3ae82babe8096b7cb56e4f10322cdb7759680aac/include/gz/sim/components/ContactSensorData.hh#L36-L38 

https://github.com/gazebosim/gz-sim/blob/3ae82babe8096b7cb56e4f10322cdb7759680aac/include/gz/sim/components/Component.hh#L395

So could we not simly reuse the data that's already in the component rather than allocating new ones (whether using arena or not)?

src/systems/physics/Physics.cc
cheriehu: I made the changes in 0e79c36824819238e95c971fe6f76f26d3e6e320 and 53cb608383d0fe2370f2d497861acaa24436d157

0e79c36824819238e95c971fe6f76f26d3e6e320 has the change to clear the msgs::Contacts in ContactSensorData directly instead of setting it to an empty Contacts message
53cb608383d0fe2370f2d497861acaa24436d157 removes the arena and the creation of the `contactsComp` message and instead uses the Data() object in ContactSensorData to modify the contacts. 

edit: reverted 53cb608383d0fe2370f2d497861acaa24436d157

These are the benchmark results compared to main. Seems like the changes all seem to be better than the situation on main. Happy to hear anyone's thoughts/if I've made a mistake in making the changes

```
main (~14%)
BM_RuntimeWorldContacts/lengthy_bullet_3k_shapes_sdf/3000       6.88 ms         6.86 ms         1000
BM_RuntimeWorldContacts/lengthy_bullet_3k_shapes_sdf/3000       6.83 ms         6.81 ms         1000
BM_RuntimeWorldContacts/lengthy_bullet_3k_shapes_sdf/3000       7.17 ms         7.15 ms         1000
BM_RuntimeWorldContacts/lengthy_bullet_3k_shapes_sdf/3000       6.70 ms         6.69 ms         1000

"no arena, set contacts directly from _contacts->Data().add_contact(); (53cb608)" ~17%
BM_RuntimeWorldContacts/lengthy_sdf_3k_shapes_bullet/3000       5.80 ms         5.78 ms         1000
BM_RuntimeWorldContacts/lengthy_sdf_3k_shapes_bullet/3000       5.66 ms         5.64 ms         1000
BM_RuntimeWorldContacts/lengthy_sdf_3k_shapes_bullet/3000       6.25 ms         6.20 ms         1000
BM_RuntimeWorldContacts/lengthy_sdf_3k_shapes_bullet/3000       6.03 ms         5.98 ms         1000

"Use msgs::Contacts object in ContactSensorData component to clear contacts" (0e79c36)
BM_RuntimeWorldContacts/lengthy_sdf_3k_shapes_bullet/3000       5.43 ms         5.41 ms         1000
BM_RuntimeWorldContacts/lengthy_sdf_3k_shapes_bullet/3000       5.52 ms         5.50 ms         1000
BM_RuntimeWorldContacts/lengthy_sdf_3k_shapes_bullet/3000       5.80 ms         5.78 ms         1000

"update with changes to main" (3c0a0bd7)
BM_RuntimeWorldContacts/lengthy_sdf_3k_shapes_bullet/3000       5.31 ms         5.30 ms         1000
BM_RuntimeWorldContacts/lengthy_sdf_3k_shapes_bullet/3000       6.27 ms         6.24 ms         1000
BM_RuntimeWorldContacts/lengthy_sdf_3k_shapes_bullet/3000       5.66 ms         5.65 ms         1000
```

src/systems/physics/Physics.cc
```diff
@@ -4613,24 +4618,29 @@ void PhysicsPrivate::UpdateCollisions(EntityComponentManager &_ecm)
...
           // Clear the last contact data
-          auto state = _contacts->SetData(contactsComp,
-            this->contactsEql) ?
+          bool changed = _contacts->Data().contact_size() > 0;
+          if (changed)
+          {
+            _contacts->Data().Clear();
+          }
+
+          auto state = changed ?
```
luca-della-vedova, line 4662: I believe this is a behavior change, and I wonder if this is part of the reason this implementation had an improved performance.
Specifically, the previous implementation used `SetData` that takes as an input a lambda that is be used to compare the previous and new component data and return true if they are different, false otherwise.
The comparison is [pretty expensive](https://github.com/gazebosim/gz-sim/blob/528d19de5744072b5b3bd6bd3046544b8c4c9256/src/systems/physics/Physics.cc#L453-L486) since it has to compare, one by one, all the collisions.
So the previous implementation will return an "unchanged" collision if the collision was the same between update steps. Since this implementation instead only checks if there are collisions, two consecutive steps with the same collisions will report a change and as such incur in serialization.
Note that I'm not advocating for this behavior (the overhead from doing a lot of comparisons in every step is not trivial, and might be greater than the saving from not serializing unchanged collisions), but just pointing out that this is a behavior change that we should be careful about (and perhaps would make it trickier to backport?).

src/systems/physics/Physics.cc
```diff
@@ -4676,15 +4686,16 @@ void PhysicsPrivate::UpdateCollisions(EntityComponentManager &_ecm)
           }
         }
 
-        auto state = _contacts->SetData(contactsComp,
-          this->contactsEql) ?
+        bool changed = _contacts->Data().contact_size() > 0;
+        auto state = changed ?
```
luca-della-vedova: This is also a behavior change and potentially incorrect, going from having contacts to having no contact _should_ be considered a change, but this logic would set it as unchanged skipping its serialization

cheriehu: oh yes, that makes sense

src/systems/physics/Physics.cc
cheriehu: I think I felt this change was okay since in the previous version we were always creating an empty contacts message, `contactsComp`, and setting the contact data to that new empty message. So the `.contact_size()` will always be 0 for the new `contactsComp` message and so simply checking that the contact size is greater than 0 would be enough to replicate the previous behavior? Also since the first check in the old comparison function was also to check the length

(I also then copied this logic below which I realise now is incorrect)

luca-della-vedova: Ah sorry didn't realize what branch this was in. Indeed since this is in the branch of "if there are no collisions right now" this should be OK, so only the other one is problematic

luca-della-vedova (commented):
I have an idea to somewhat have the best of both worlds, but it's not very pretty.

So the root of the problem is that we need to have two different data structures, one that stores the current data and one that has the new data, so we can do change detection.
**Assuming we still want to do change detection**, we could have a "manual change detection", which would be pretty much a reimplementation of the lambda calling that `component->SetData` does behind the scenes, with the difference that instead of taking two protobufs as input it takes the existing component and the maps used to create the new component. Then, only if different, we clear and assign the data as recommended by @azeey and set the component as `PeriodicChange`.
With this approach we could get rid of the arena altogether.

From what I understand from the flamegraphs, the biggest issue in the current implementation + the issue in the arena implementation is that assigning the data inside the component destructs the whole data contained in the component, which calls a lot of deallocations. Then it immediately has to build a new heap allocated component, which will cause a lot of allocations.
By contrast, with @azeey suggestion if we only clear the component we will not deallocate the memory making, hopefully, its clearing a lot more efficient and its subsequent allocations a lot faster.

I think my proposal (again only if benchmarks / profiling show that it make sense! would be):

* This PR (backward compatible) Try this manual change detection + no additional protobufs approach.
* A future PR (but profile first to verify whether this would help), consider changing the component from a `gz::msgs::Contacts` to an arena allocated `gz::msgs::Contacts*`. This comes with breakages and tricky lifetime implications though, so we should only consider it if the performance gains are very significant.

cheriehu:
> **Assuming we still want to do change detection**, we could have a "manual change detection", which would be pretty much a reimplementation of the lambda calling that `component->SetData` does behind the scenes, with the difference that instead of taking two protobufs as input it takes the existing component and the maps used to create the new component. Then, only if different, we clear and assign the data as recommended by @azeey and set the component as `PeriodicChange`.

The performance does seem to improve with this new change! These are the new results (again with no scopedname)

```
BM_RuntimeWorldContacts/lengthy_sdf_3k_shapes_bullet/3000/iterations:1000       3.59 ms         3.59 ms         1000
BM_RuntimeWorldContacts/lengthy_sdf_3k_shapes_bullet/3000/iterations:1000       3.65 ms         3.65 ms         1000
BM_RuntimeWorldContacts/lengthy_sdf_3k_shapes_bullet/3000/iterations:1000       4.17 ms         4.16 ms         1000
```

Also checked that contacts still work at least on this diff drive world

https://github.com/user-attachments/assets/f8d557e8-5969-4c42-aa6f-a009470e935b

luca-della-vedova (commented):
This is excellent results! Am I understanding it right that performance almost doubled for the 3k shapes benchmark?
I wonder if in the future we should reconsider a similar change for other systems that implemented the Arena change, just in case it can make even those faster.
I'll try to get a review for the code itself in the coming days, just a small nit for now, would it be possible to get rid of the manual increment of `i` / `j` variables and just use a normal for loop instead of doing it range based?

cheriehu:
> This is excellent results! Am I understanding it right that performance almost doubled for the 3k shapes benchmark?

Yep seems like it almost doubles performance for me on 3k shapes. Could be because it skips updating any new contacts if the contact information stays the same
--> update: tried commenting out the `if (!changed) return true;` block and still seems to be faster than all the arena stuff?
```
commented out if (!changed) return true; 
BM_RuntimeWorldContacts/lengthy_sdf_3k_shapes_bullet/3000/iterations:1000       4.45 ms         4.44 ms         1000
BM_RuntimeWorldContacts/lengthy_sdf_3k_shapes_bullet/3000/iterations:1000       4.64 ms         4.63 ms         1000
BM_RuntimeWorldContacts/lengthy_sdf_3k_shapes_bullet/3000/iterations:1000       4.85 ms         4.84 ms         1000
```

> would it be possible to get rid of the manual increment of `i` / `j` variables and just use a normal for loop instead of doing it range based?

done in 07da40b - though it also adds an iterator variable for the unordered_map since the map can't be accessed by index?

src/systems/physics/Physics.cc
```diff
@@ -4616,12 +4649,16 @@ void PhysicsPrivate::UpdateCollisions(EntityComponentManager &_ecm)
       [&](const Entity &_collEntity1, components::Collision *,
           components::ContactSensorData *_contacts) -> bool
       {
-        msgs::Contacts contactsComp;
         if (entityContactMap.find(_collEntity1) == entityContactMap.end())
```
luca-della-vedova: Nit (not touched by this PR), we do two lookups here, we can make this slightly more efficient by only doing one lookup and caching the iterator, so instead of:

```
if (entityContactMap.find(_collEntity1) == entityContactMap.end()) {
...
}

// afterwards
const auto &contactMap = entityContactMap[_collEntity1];
```

We could do

```
const auto contactMapIt = entityContactMap.find(_collEntity1);
if (contactMapIt == entityContactMap.end()) {
...
}
// afterwards
const auto& contactMap = contactMapIt->second;
```

BUILD.bazel
```diff
@@ -1500,6 +1500,7 @@ gz_sim_system_libraries(
     visibility = ["//visibility:public"],
     deps = [
         ":gz-sim",
+        "@com_google_protobuf//:protobuf_lite",
```
luca-della-vedova: I think we can take this out since there is no more arena being added?

luca-della-vedova (approved):
Generally LGTM, I ran it and I can reproduce more than doubling the RTF when contacts are enabled.
A minor cleanup and a minor caching that could remove some further inefficiency

src/systems/physics/Physics.cc
```diff
@@ -610,6 +575,90 @@ class gz::sim::systems::PhysicsPrivate
...
+              !math::equal(body2force1.y(), body2force2.y(), 1e-6) ||
+              !math::equal(body2force1.z(), body2force2.z(), 1e-6))
+          {
+            return false;
+          }
+        }
+      }
+    }
+    return true;
+  }
```
luca-della-vedova, line 660: Generally LGTM! Since CI is also green + it's validated with a video I'm leaning towards merging it, this is the main behavior change if anyone wants to have a look / veto it. The equality function is much more comprehensive now (it also checks for normals and wrenches) and it gates the component update (i.e. if the component is not changed it will not be updated, previously it was always updated).
If either this logic is invalid / comprehensive enough or we rely on the behavior of always updating the component even if it's almost equal we could end up with unintended behavior.

luca-della-vedova:
Thank you!

I think this is safe to backport but I noticed that cherry-picking fails because we didn't backport the benchmakr PR, so will do that first.

cheriehu:
Thanks! 🙏

luca-della-vedova:
@mergifyio backport gz-sim8 gz-sim9 gz-sim10

PR #3737, opened 2026-06-26
Use marker array visualize contacts

### LINKED ISSUE #3730: Improve Performance of VisualizeContacts by using MarkerArray message
## Desired behavior
The [VisualizeContacts](https://github.com/gazebosim/gz-sim/blob/3ae82babe8096b7cb56e4f10322cdb7759680aac/src/gui/plugins/visualize_contacts/VisualizeContacts.cc#L1) GUI plugin currently uses individual `Marker` service requests for each contact. Using MarkerArray messages should be more efficient.

https://github.com/gazebosim/gz-sim/blob/3ae82babe8096b7cb56e4f10322cdb7759680aac/src/gui/plugins/visualize_contacts/VisualizeContacts.cc#L229-L249

## Alternatives considered
n/a

## Implementation suggestion
Use marker arrays as shown in https://github.com/gazebosim/gz-sim/blob/3ae82babe8096b7cb56e4f10322cdb7759680aac/examples/standalone/marker/marker.cc#L308

MobiusRoamer:
# 🦟 Bug fix

Fixes #3730.

## Summary
This updates `VisualizeContacts` to batch contact position markers into a single `Marker_V` request, then send it through `/marker_array` once per update. Previously, one `/marker` request was sent per contact position.
The request remains non-blocking, matching the existing behavior.

Also adds behavioral coverage for the `/marker_array` path in `test/integration/markers.cc`: the test sends a `Marker_V` with multiple markers, updates `MarkerManager`, and verifies that each marker is received.

## Backport Policy
    - [x] Jetty
    - [x] Ionic
    - [x] Harmonic
    - [x] Fortress
- [ ] I am not sure

## Checklist
- [x] Added tests

Generated by: GPT5.5, used for local test commands and finding build tools.

## Notes 
Relying on CI for an initial check so keeping this as draft

### DIFF
--- a/test/integration/markers.cc
+++ b/test/integration/markers.cc
@@ -16,7 +16,9 @@
  */
 
 #include <gtest/gtest.h>
+#include <gz/msgs/boolean.pb.h>
 #include <gz/msgs/marker.pb.h>
+#include <gz/msgs/marker_v.pb.h>
 
 #include <gz/common/Console.hh>
 #include <gz/common/Util.hh>
@@ -50,6 +52,11 @@ void markerCb(const msgs::Marker &_msg)
   mutex.lock();
   markerMsgs.push_back(_msg);
   mutex.unlock();
+}
+
+/////////////////////////////////////////////////
+void markerArrayCb(const gz::msgs::Boolean &, const bool)
+{
 }
 
 /////////////////////////////////////////////////
@@ -112,4 +119,75 @@ TEST_F(MarkersTest, MarkerPublisher)
   node.Request("/marker", markerMsg);
   markerManager.Update();
   wait(2, markerMsg);
+}
+
+/////////////////////////////////////////////////
+TEST_F(MarkersTest, MarkerArrayPublisher)
+{
+  std::map<std::string, std::string> params;
+  auto engine = gz::rendering::engine("ogre2", params);
+  auto scene = engine->CreateScene("testscene_marker_array");
+
+  // Clear messages received by previous tests.
+  mutex.lock();
+  markerMsgs.clear();
+  mutex.unlock();
+
+  MarkerManager markerManager;
+  markerManager.Init(scene);
+
+  // Function that Waits for messages to be received
+  auto wait = [&](std::size_t _size) {
+    for (int sleep = 0; sleep < 30; ++sleep)
+    {
+      // The service request is asynchronous, so keep processing messages.
+      markerManager.Update();
+      std::this_thread::sleep_for(std::chrono::milliseconds(100));
+
+      mutex.lock();
+      bool received = markerMsgs.size() == _size;
+      mutex.unlock();
+
+      if (received)
+        break;
+    }
+
+    mutex.lock();
+    EXPECT_EQ(markerMsgs.size(), _size);
+    mutex.unlock();
+  };
+
+  // subscribe to marker topic
+  transport::Node node;
+  node.Subscribe("/marker", &markerCb);
+
+  gz::msgs::Marker_V markerArrayMsg;
+
+  auto markerMsg0 = markerArrayMsg.add_marker();
+  markerMsg0->set_ns("array");
+  markerMsg0->set_id(0);
+  markerMsg0->set_action(gz::msgs::Marker::ADD_MODIFY);
+  markerMsg0->set_type(gz::msgs::Marker::SPHERE);
+  markerMsg0->set_visibility(gz::msgs::Marker::GUI);
+
+  auto markerMsg1 = markerArrayMsg.add_marker();
+  markerMsg1->set_ns("array");
+  markerMsg1->set_id(1);
+  markerMsg1->set_action(gz::msgs::Marker::ADD_MODIFY);
+  markerMsg1->set_type(gz::msgs::Marker::BOX);
+  markerMsg1->set_visibility(gz::msgs::Marker::GUI);
+
+  // Send an asynchronous marker array request.
+  EXPECT_TRUE(node.Request(
+      "/marker_array", markerArrayMsg, &markerArrayCb));
+
+  wait(2);
+
+  mutex.lock();
+  auto receivedMsgs = markerMsgs;
+  mutex.unlock();
+
+  ASSERT_EQ(receivedMsgs.size(), 2u);
+  EXPECT_EQ(markerMsg0->DebugString(), receivedMsgs[0].DebugString());
+  EXPECT_EQ(markerMsg1->DebugString(), receivedMsgs[1].DebugString());
 }
--- a/test/integration/markers.cc
+++ b/test/integration/markers.cc
@@ -128,51 +128,23 @@ TEST_F(MarkersTest, MarkerArrayPublisher)
   auto engine = gz::rendering::engine("ogre2", params);
   auto scene = engine->CreateScene("testscene_marker_array");
 
-  // Clear messages received by previous tests.
-  mutex.lock();
-  markerMsgs.clear();
-  mutex.unlock();
-
   MarkerManager markerManager;
   markerManager.Init(scene);
 
-  // Function that Waits for messages to be received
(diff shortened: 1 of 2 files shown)


arjo129:
I changed the base branch as I think you are working of gz-sim10 and your diff was absolutely huge when targetting main.

MobiusRoamer:
> I changed the base branch as I think you are working of gz-sim10 and your diff was absolutely huge when targetting main.

Thanks for fixing it. Yes, this change was based on `gz-sim10`; I missed that when opening the draft PR.

MobiusRoamer:
There isn’t an existing `VisualizeContacts` test, but is already marker behavior coverage in `test/integration/markers.cc`. I’ll add coverage there for the `/marker_array` path: send a `Marker_V` with multiple markers, update `MarkerManager`, and verify that the markers are processed. This should cover the batching behavior without adding heavy GUI plugin tests and flaky dependencies, etc. 

Have added behavioral test coverage for the `/marker_array` path in `test/integration/markers.cc`: the test sends a `Marker_V` containing 2 markers, and verifies that the individual markers are received.

azeey:
Thanks for the contribution @MobiusRoamer! I think this PR could benefit from a screen capture
Also, have you seen any performance difference between this and the old version?

MobiusRoamer:
> Thanks for the contribution @MobiusRoamer! I think this PR could benefit from a screen capture Also, have you seen any performance difference between this and the old version?


Yeah of course, thx for the suggestion.

### Behavioral Check & Performance
For a local reproduction of the behavioral test:

Pls use environment:

- Gazebo Sim 10.4.0
- Ubuntu 24.04
- `examples/worlds/visualize_contacts.sdf`
- Marker radius: `0.10 m`
- Update period: `200 ms`

I used a temporary copy of the example world and changed the cylinder pose from:

```
<pose>-0.3 0 1.15 0 0 0</pose>
```
to:
```
<pose>-0.3 0 3.15 0 0 0</pose>
```
The equivalent setup commands are 
```
cp examples/worlds/visualize_contacts.sdf \
  /tmp/visualize_contacts_drop.sdf

sed -i \
  's|<pose>-0.3 0 1.15 0 0 0</pose>|<pose>-0.3 0 3.15 0 0 0</pose>|' \
  /tmp/visualize_contacts_drop.sdf
```

then run the simulation using two terminals:

Terminal1: start the sim server
```
unset LD_LIBRARY_PATH
unset GZ_GUI_PLUGIN_PATH

gz sim -s -v 3 /tmp/visualize_contacts_drop.sdf
```
Terminal2: start the local built GUI and plugin
```
cd /path/to/gz-sim
export GZ_GUI_PLUGIN_PATH="$PWD/build/gz-sim/lib:${GZ_GUI_PLUGIN_PATH}"

./build/gz-sim/bin/gz-sim-gui-client -v 3
```
(replace /path/to/gz-sim with your local path, and others according to ur machine if needed)


During the check I collected request count stats using temporary tracker in `src/gui/plugins/visualize_contacts/VisualizeContacts.cc`: (line 261 was changed to the following block)
```
if (markerMsgs.marker_size() > 0)
  {
    static std::size_t markerCount = 0;
    static std::size_t batchCount = 0;

    markerCount += markerMsgs.marker_size();
    ++batchCount;

    if (batchCount == 10)
    {
      const double reduction =
          100.0 * (markerCount - batchCount) / markerCount;

      gzmsg << "[VisualizeContacts stats] markers=" << markerCount
            << ", old_requests=" << markerCount
            << ", batched_requests=" << batchCount << std::endl;

      markerCount = 0;
      batchCount = 0;
    }

    this->dataPtr->node.Request(
        "/marker_array", markerMsgs, &OnMarkerArrayResponse);
  }
```
The above has been deleted so not part of the commit of course. Then the raw results were
```text
markers=172, old_requests=172, batched_requests=10
markers=224, old_requests=224, batched_requests=10
markers=232, old_requests=232, batched_requests=10
```
So e.g. across 10 updates in first line, old implementation would have sent 172 individual requests, new implementation sends 10. To view the case at play, see screen capture at the end.


### Side note on latest commit
On my latest commit: When testing the GUI, previous attempt did not display the markers. So I changed the `/marker_array` from previous no-response overload to current response-callback overload. `OnMarkerArrayResponse` is intentionally empty as a result. Here I assume rsp data is not needed, the function only provides a callback to force a rqst/reply without blocking GUI updates. 

Looked further into why no contact markers where displayed when `/marker_array` used the two-argument request overload, that is,
```cpp 
`node.Request("/marker_array", markerMsgs);`
```
hides the markers.

MarkerManager defines the `/marker_array` handler with this signature:

```cpp
bool OnMarkerMsgArray(
    const msgs::Marker_V &_req,
    msgs::Boolean &_res);
```
(src/rendering/MarkerManager.cc, line79-80)
Gazebo Transport passes both var's type names Marker_V and Boolean into firstHandler (gazebosim/gz-transport/blob/gz-transport14/include/gz/transport/detail/Node.hh, line 502-510). 

Yet the two-arg overload hardcodes rsp type as `gz::msgs::Empty` (same file Node.hh, line 718-731), hence the call
```cpp
node.Request("/marker_array", markerMsgs);
```
maps Marver_V to Empty while its meant to be Boolean. So I do think 
```cpp
void OnMarkerArrayResponse(const gz::msgs::Boolean &,const bool)
{
}
```
solves it despite empty. Open to changes on this one.



Have also updated the regression test accordingly. 
The markers now display correctly. 
### Screen Capture
https://github.com/user-attachments/assets/e5898206-6e01-4a3f-9276-a58b2dda4222

MobiusRoamer:
Catching this up with latest branch, will make it ready once checks r done, let me know if changes are needed

arjo129:
You don't need to wait for tests to pass (other than ones this may affect). Feel free to mark it as ready for review.

test/integration/markers.cc
```diff
@@ -113,3 +120,74 @@ TEST_F(MarkersTest, MarkerPublisher)
...
+    }
+
+    mutex.lock();
+    EXPECT_EQ(markerMsgs.size(), _size);
+    mutex.unlock();
+  };
+
+  // subscribe to marker topic
+  transport::Node node;
+  node.Subscribe("/marker", &markerCb);
```
arjo129: Why subscribe to the marker topic? This feels redundant for the test.

test/integration/markers.cc
arjo129, line 1: Thanks for adding this test for the MarkerManager! Im assuming you are aware that this test tests a different component (the MarkerManager) and not `VisualizeContacts`. For future reference this could live in a different PR - I'll let it slide for this time.

arjo129 (changes requested):
I tested this myself and it works:

Image

However, I think some of the code in the test is redundant. As for performance improvements, I can see how this could be beneficial. We are however lacking reproducible benchmarks in this department.

Once the redundant code is removed it looks good to me.

test/integration/markers.cc
MobiusRoamer: Thanks for the review! Totally agree, no indirection via pub/sub needed. 

I removed the subscription. Now chekcing this directly: after sending the `/marker_array` reqyest, the test calls `MarkerManager::Update()` to process each queued marker through `ProcessMarkerMsg()`, then creates visual ID. Then we retrieve the visuals with `scene->VisualByName(...)` and uses assertions to check they are non-null in the scene.

test/integration/markers.cc
MobiusRoamer: Yes that's a fair point. I will keep it in mind.

arjo129 (approved):
Thanks for iterating on this PR. LGTM.

arjo129:
@mergifyio backport gz-sim10 gz-sim9 gz-sim8 ign-gazebo6

azeey:
@arjo129 this one already targetted `gz-sim10`. I tneeds to be ported to `main`.

arjo129:
@Mergifyio backport main

PR #3745, opened 2026-06-29
Fix race between asset download and entity creation

caguero:
# 🦟 Bug fix

This patch makes `ServerPrivate::DownloadAssets` wait for the download thread by joining it instead of waiting on a condition variable. A join cannot miss the completion signal or wake up spuriously, and it also guarantees that the download thread has finished updating every runner before the entities are created.

## Summary

The condition variable wait had no predicate, so a spurious or early notification could run `CreateEntities()` before the download thread finished. In addition, the error path with `SdfErrorBehavior::EXIT_IMMEDIATELY` notified the waiting thread and kept updating the runners, so a server that had already exited with errors ended up with entities created and system plugins loaded.

This patch waits by joining the download thread instead, returns early from the thread on the error path, and skips entity creation when the server exited with errors. The mutex, condition variable, and predicate members are no longer needed and were removed.

The strengthened `ServerFixture.SdfErrorExit` test asserts that a server that exited with errors has zero entities and zero system plugins. It fails 20 out of 20 runs without this patch and passes 20 out of 20 with it.

## Backport Policy

    - [x] Jetty
    - [ ] Ionic
    - [ ] Harmonic
    - [ ] Fortress
- [ ] I am not sure

## Checklist
- [x] Added tests

Assisted-by: Claude Fable 5


Old-Ding (commented):
I think the predicate can still become true too early on the `localErrors` path. This block sets `downloadAssetsComplete` and notifies before the thread reaches the later loop that updates each runner's world SDF and calls `SetCreateEntities()`.

That means the waiting thread can wake and call `CreateEntities()` while the download thread is still mutating runner state, which seems to preserve part of the ordering race that the predicate is meant to remove. Since the unconditional `WaitForAssets()` block at the end also sets the predicate and notifies, would it be safer to set `downloadAssetsComplete` only at that final completion point, or return immediately after the error path if the later runner updates should not run?

caguero:
> I think the predicate can still become true too early on the `localErrors` path. This block sets `downloadAssetsComplete` and notifies before the thread reaches the later loop that updates each runner's world SDF and calls `SetCreateEntities()`.
> 
> That means the waiting thread can wake and call `CreateEntities()` while the download thread is still mutating runner state, which seems to preserve part of the ordering race that the predicate is meant to remove. Since the unconditional `WaitForAssets()` block at the end also sets the predicate and notifies, would it be safer to set `downloadAssetsComplete` only at that final completion point, or return immediately after the error path if the later runner updates should not run?

Thanks, @Old-Ding! I actually changed my mind and removed the entire condition variable in favor or the download thread + `join`. I think it's cleaner. [bbb956f](https://github.com/gazebosim/gz-sim/pull/3745/commits/bbb956fcf5a2b438275953a237808fdaa1ce4002)

PR #3746, opened 2026-06-29
Use SetComponentData in Link velocity setters

caguero:
# Clean :broom: 

This patch replaces the custom get-or-create logic in Link's linear/angular velocity setters with the existing `EntityComponentManager::SetComponentData` helper.

## Backport Policy

    - [x] Jetty
    - [x] Ionic
    - [x] Harmonic
    - [x] Fortress
- [ ] I am not sure

## Checklist
- [ ] Added tests

Generated-by: Claude Opus 4.8


luca-della-vedova (commented):
There is a subtle difference, which is that `SetComponentData` defaults to `SetData` which will, behind the scenes, call `operator==` for Vector3, which is wasted work since, for this API, the user doesn't care whether data changed or not.
I can't quite tell if the compiler will be able to optimize it away though

caguero:
> There is a subtle difference, which is that `SetComponentData` defaults to `SetData` which will, behind the scenes, call `operator==` for Vector3, which is wasted work since, for this API, the user doesn't care whether data changed or not. I can't quite tell if the compiler will be able to optimize it away though

We could modify the `SetComponent()` API with:

```
public: template<typename ComponentTypeT>
             bool SetComponentData(const Entity _entity,
             const typename ComponentTypeT::Type &_data,
             bool _compare = true);
```

so that we can skip the comparison if needed.

caguero:
> > There is a subtle difference, which is that `SetComponentData` defaults to `SetData` which will, behind the scenes, call `operator==` for Vector3, which is wasted work since, for this API, the user doesn't care whether data changed or not. I can't quite tell if the compiler will be able to optimize it away though
> 
> We could modify the `SetComponent()` API with:
> 
> ```
> public: template<typename ComponentTypeT>
>              bool SetComponentData(const Entity _entity,
>              const typename ComponentTypeT::Type &_data,
>              bool _compare = true);
> ```
> 
> so that we can skip the comparison if needed.

I'm actually more inclined to not extend the `SetComponentData()` API with the compare flag. 

I think the extra comparison is negligible in practice. Physics removes the velocity cmd components at the end of each update step, so the common path here is the create branch, which does no do the comparison at all. The compare only runs if the setter is called more than once within a single step, and it's just three double equality checks on an inline template, so it should be crazy fast compared with the component lookup itself.

@luca-della-vedova , what do you think?

ahcorde (approved):

ahcorde:
https://github.com/Mergifyio backport gz-sim10 gz-sim9 gz-sim8

caguero:
@mergifyio backport ign-gazebo6

PR #3750, opened 2026-06-30
Fix hierarchy inconsistency when serializing + deserializing state

### LINKED ISSUE #3729: Hierarchy inconsistency when serializing into state + deserializing from state
# Bug report

## Environment
* OS Version:  Ubuntu 24.04
* Source or binary build? Source, on main but probably existed since forever.

## Description
* Expected behavior: Setting the graph for the ECM from the ModelCommandAPI should succeed.
* Actual behavior:  It fails!

## Steps to reproduce

Apply this patch to latest main, to print out when setting the hierarchy fails

```
diff --git a/src/EntityComponentManager.cc b/src/EntityComponentManager.cc
index 5db48a674..a48edc095 100644
--- a/src/EntityComponentManager.cc
+++ b/src/EntityComponentManager.cc
@@ -1225,7 +1225,11 @@ bool EntityComponentManager::CreateComponentImplementation(
   if (_componentTypeId == components::ParentEntity::typeId)
   {
     auto parentComp = this->Component<components::ParentEntity>(_entity);
-    this->SetParentEntity(_entity, parentComp->Data());
+    if(!this->SetParentEntity(_entity, parentComp->Data()))
+    {
+      gzerr << "Failed setting parent for entity " << _entity << " to "
+            << parentComp->Data() << std::endl;
+    }
   }
 
   return updateData;
```

Then build and run the ModelCommandAPI unit test:

```
build/gz-sim/bin/UNIT_ModelCommandAPI_TEST
```

The test looks at stdout output and it will fail because a lot of the calls actually fail and errors are printed!

## Output

```
[redacted]
+(2026-06-25 21:59:45.684) [error] [EntityComponentManager.cc:1230] Failed setting parent for entity 16 to 13
+(2026-06-25 21:59:45.684) [error] [EntityComponentManager.cc:1230] Failed setting parent for entity 15 to 13
+(2026-06-25 21:59:45.684) [error] [EntityComponentManager.cc:1230] Failed setting parent for entity 11 to 9
+(2026-06-25 21:59:45.684) [error] [EntityComponentManager.cc:1230] Failed setting parent for entity 6 to 5
+(2026-06-25 21:59:45.747) [error] [EntityComponentManager.cc:1230] Failed setting parent for entity 12 to 1
+(2026-06-25 21:59:45.747) [error] [EntityComponentManager.cc:1230] Failed setting parent for entity 17 to 1
+(2026-06-25 21:59:45.750) [error] [EntityComponentManager.cc:1230] Failed setting parent for entity 2 to 1
+(2026-06-25 21:59:45.750) [error] [EntityComponentManager.cc:1230] Failed setting parent for entity 7 to 5
+(2026-06-25 21:59:45.750) [error] [EntityComponentManager.cc:1230] Failed setting parent for entity 8 to 1
+(2026-06-25 21:59:45.759) [error] [EntityComponentManager.cc:1230] Failed setting parent for entity 3 to 1
+(2026-06-25 21:59:45.759) [error] [EntityComponentManager.cc:1230] Failed setting parent for entity 4 to 1
 - Sensor [16]
   - Name: rgbd_camera

[  FAILED  ] ModelCommandAPI.RgbdCameraSensor (962 ms)
```

## Root cause + fix information

As noted in https://github.com/gazebosim/gz-sim/pull/3711#issuecomment-4799963763, what I _think_ is happening is that when serializing into `State` functions, we iterate over the `unordered_map` of components, which means the `State` message will not be ordered.
Afterwards, we use this message to populate a new ECM, but if a child is created before its parent the call fails and the hierarchy gets inconsistent.
After this is fixed the ModelCommandAPI should also be updated again to use ChildrenByComponents as mentioned in https://github.com/gazebosim/gz-sim/pull/3362#issuecomment-3995059151.
Ideally there would also be a unit test for the `EntityComponentManager` where we:

* Create a hierarachy
* Serialize it into a State message
* Deserialize it into a new ECM
* Check that the new ECM has the exact same hierarchy

luca-della-vedova:
# 🦟 Bug fix

Fixes #3729 

## Summary

Add a pre-spawning step to the `SetState` functions so we are sure that entities are present when, afterwards, we try to add components to them.
Previously, since we spawned entities and their components in one go, we ended up in cases where an entity would be spawned before its parent, and updating the graph failed as a consequence.
I'm splitting this PR in three:

* The first commit https://github.com/gazebosim/gz-sim/pull/3750/commits/b27af72963eab05a0291aec830f12cef0a394ce8 adds a failing unit test, together with logging for failures in setting the hierarchy that will make the ModelCommandAPI unit test fail.
* The second commit https://github.com/gazebosim/gz-sim/pull/3750/commits/1de67aeaedc3cc4feb2e015f325c949c01e5dab7 will hopefully make CI green again by fixing the behavior.
* The third commit https://github.com/gazebosim/gz-sim/pull/3750/commits/ace4f90e0b4335ff989522c24d02324eb48c56b1 migrates the ModelCommandAPI to use ChildrenByComponents, which previously didn't work because of this bug as mentioned in #3729 

## Backport Policy
This is a fix and should be backported, BUT since it is based on the `ParentEntity` change the diff might become a bit gnarly if we decide to only backport this but not the ParentEntity change.
    - [x] Jetty
    - [x] Ionic
    - [x] Harmonic
    - [x] Fortress
- [ ] I am not sure

## Checklist
- [x] Added tests

Generated-by: Gemini 3

### DIFF
--- a/src/EntityComponentManager.cc
+++ b/src/EntityComponentManager.cc
@@ -1264,7 +1264,11 @@ bool EntityComponentManager::CreateComponentImplementation(
       return updateData;
     }
     const auto *parent = static_cast<const components::ParentEntity *>(_data);
-    this->dataPtr->SetParentEntityGraph(_entity, parent->Data());
+    if (!this->dataPtr->SetParentEntityGraph(_entity, parent->Data()))
+    {
+      gzerr << "Failed setting parent for entity " << _entity << " to "
+            << parent->Data() << std::endl;
+    }
   }
 
   return updateData;
--- a/src/EntityComponentManager.cc
+++ b/src/EntityComponentManager.cc
@@ -1898,8 +1898,20 @@ void EntityComponentManager::SetState(
     {
       this->dataPtr->CreateEntityImplementation(entity);
     }
+  }
 
-    // Create / remove / update components
+  // Create / remove / update components
+  for (int e = 0; e < _stateMsg.entities_size(); ++e)
+  {
+    const auto &entityMsg = _stateMsg.entities(e);
+
+    Entity entity{entityMsg.id()};
+
+    if (entityMsg.remove())
+    {
+      continue;
+    }
+
     for (int c = 0; c < entityMsg.components_size(); ++c)
     {
       const auto &compMsg = entityMsg.components(c);
@@ -1996,8 +2008,20 @@ void EntityComponentManager::SetState(
     {
       this->dataPtr->CreateEntityImplementation(entity);
     }
+  }
 
-    // Create / remove / update components
+  // Create / remove / update components
+  for (const auto &iter : _stateMsg.entities())
+  {
+    const auto &entityMsg = iter.second;
+
+    Entity entity{entityMsg.id()};
+
+    if (entityMsg.remove())
+    {
+      continue;
+    }
+
     for (const auto &compIter : iter.second.components())
     {
       const auto &compMsg = compIter.second;
--- a/src/EntityComponentManager_TEST.cc
+++ b/src/EntityComponentManager_TEST.cc
@@ -3222,6 +3222,76 @@ TEST_P(EntityComponentManagerFixture,
 }
 
 //////////////////////////////////////////////////
+/// \brief Test that after serializing and deserializing state the hierarchy
+/// is preserved.
+TEST_P(EntityComponentManagerFixture,
+       GZ_UTILS_TEST_DISABLED_ON_WIN32(StateMsgHierarchy))
+{
+  static constexpr std::size_t NUM_ENTITIES = 100;
+  EntityComponentManager originalECMStateMap;
+  EntityComponentManager otherECMStateMap;
+
+  // A series of entities where half have no parent and half have the previous
+  // entity as a parent.
+  std::vector<Entity> entities;
+  entities.reserve(NUM_ENTITIES);
+  auto parentEntity = kNullEntity;
+  for (std::size_t i = 0; i < NUM_ENTITIES; ++i)
+  {
+    auto entity = originalECMStateMap.CreateEntity();
+    originalECMStateMap.CreateComponent(entity,
+        components::Name("entity_" + std::to_string(i)));
+    if (i % 2)
+    {
+      EXPECT_TRUE(originalECMStateMap.SetParentEntity(entity, parentEntity));
+    }
+    parentEntity = entity;
+    entities.push_back(entity);
+  }
+
+  // update the other ECM to have the new entity and component
+  msgs::SerializedStateMap stateMapMsg;
+  originalECMStateMap.State(stateMapMsg);
+  otherECMStateMap.SetState(stateMapMsg);
+
+  ASSERT_EQ(otherECMStateMap.EntityCount(), originalECMStateMap.EntityCount());
+  for (std::size_t i = 0; i < NUM_ENTITIES; ++i)
+  {
+    auto desc = otherECMStateMap.Descendants(entities[i]);
+    // Descendants includes self, remove it to simplify asserts
+    desc.erase(entities[i]);
+    if (i % 2 == 0)
+    {
+      // Entity with no descendents
+      ASSERT_EQ(desc.size(), 1);
+      EXPECT_EQ(*desc.begin(), entities[i + 1]);
+    } else {
+      EXPECT_EQ(desc.size(), 0);
+    }
+  }
+
+  EntityComponentManager otherECMState;
+  auto stateMsg = originalECMStateMap.State();
+  otherECMState.SetState(stateMsg);
+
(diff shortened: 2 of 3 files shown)


caguero (commented):
Not sure if we should handle this here but my understanding is that this patch fixes hierarchy on component creation. However, an update to an existing `ParentEntity` only hits the "Deserialize the existing component" path, which never calls `SetParentEntityGraph`. So a re-parent sent over state sync (e.g. detachable-joint  attach/detach) would leave a stale graph edge and `Descendants/ChildrenByComponents` would report the old hierarchy.

src/EntityComponentManager.cc
```diff
@@ -1264,7 +1264,11 @@ bool EntityComponentManager::CreateComponentImplementation(
       return updateData;
     }
     const auto *parent = static_cast<const components::ParentEntity *>(_data);
-    this->dataPtr->SetParentEntityGraph(_entity, parent->Data());
+    if (!this->dataPtr->SetParentEntityGraph(_entity, parent->Data()))
+    {
+      gzerr << "Failed setting parent for entity " << _entity << " to "
```
caguero, line 1269: Will this show up even if the parent appears later and nothing actually breaks?

src/EntityComponentManager.cc
caguero, line 1988: Just worth checking if we can extract a helper function to be used in both `SetState()` functions.

src/EntityComponentManager.cc
luca-della-vedova: It will, but the parent appearing later _will_ be a breakage since this call fails and this entity will still be without a parent, unless somehow we find a way to detect this case and try again.

luca-della-vedova:
> Not sure if we should handle this here but my understanding is that this patch fixes hierarchy on component creation. However, an update to an existing `ParentEntity` only hits the "Deserialize the existing component" path, which never calls `SetParentEntityGraph`. So a re-parent sent over state sync (e.g. detachable-joint attach/detach) would leave a stale graph edge and `Descendants/ChildrenByComponents` would report the old hierarchy.

Yes this is in general a problem with any type of edits to the `ParentEntity` component, before it even reaches this whole serialize / deserialize state (that as you mentioned is broken), if any user at any point was to do something like:

```
auto* parent = ecm.Component<components::ParentEntity>(entity);
*parent->Data() = newParent;
```

This was not (and still is not) caught in any way, so from this moment forward this ECM will have a broken graph, then when this change gets propagated the client will also have a broken graph.
To my understanding this case of changing the value of a parent is probably never done? You mentioned detachable joints, I am not very familiar with them, I had a look it seems that adding a detachable joint doesn't actually change the hierarchy. It saves the information about who is the parent and who is the child in its component:

https://github.com/gazebosim/gz-sim/blob/7579cc47533d85a65df5b8aa6383bcbaffacb838/src/systems/detachable_joint/DetachableJoint.cc#L351-L356

Then this data is used to create a fixed joint in the physics engine without actually saying that the parent has the joint as a child and the child has the joint as a parent:

https://github.com/gazebosim/gz-sim/blob/7579cc47533d85a65df5b8aa6383bcbaffacb838/src/systems/physics/Physics.cc#L2077-L2078

But yea I think in general the cases of dynamic hierarchy changes during the simulation are not / have not really been supported.

src/EntityComponentManager.cc
luca-della-vedova: The types are just so slightly different with different APIs that they make it a bit tricky.
What I would _really_ like is getting rid of one of them, but it seems not super straightforward.

src/EntityComponentManager_TEST.cc
```diff
@@ -3221,6 +3221,76 @@ TEST_P(EntityComponentManagerFixture,
...
+
+  ASSERT_EQ(otherECMStateMap.EntityCount(), originalECMStateMap.EntityCount());
+  for (std::size_t i = 0; i < NUM_ENTITIES; ++i)
+  {
+    auto desc = otherECMStateMap.Descendants(entities[i]);
+    // Descendants includes self, remove it to simplify asserts
+    desc.erase(entities[i]);
+    if (i % 2 == 0)
+    {
+      // Entity with no descendents
```
caguero, line 3265: // Parent of exactly one child.

src/EntityComponentManager_TEST.cc
```diff
@@ -3221,6 +3221,76 @@ TEST_P(EntityComponentManagerFixture,
...
+      ASSERT_EQ(desc.size(), 1);
+      EXPECT_EQ(*desc.begin(), entities[i + 1]);
+    } else {
+      EXPECT_EQ(desc.size(), 0);
```
caguero, line 3269: // A leaf, so no descendants

(same in the next block)

caguero (approved):
Looks good then, besides a minor comment fix.

arjo129:
@osrf-jenkins run tests

arjo129:

PR #3759, opened 2026-07-04
test: build display-dependent unit tests

### LINKED ISSUE #3496: Tests skipped due to missing display should still be compiled
## Desired behavior
As noted in https://github.com/gazebo-tooling/release-tools/issues/1490, there are tests in gz-sim that are being skipped on macOS due to a missing display. However, we should still build these tests to catch potential compile errors. For example, I ran into a build error on `gz-sim8` on macOS that's not caught by CI since `reset_sensors.cc` is never compiled on that platform. There's a fix for it in https://github.com/gazebosim/gz-sim/pull/3000, but it was not backported to gz-sim8.

## Alternatives considered
None

## Implementation suggestion
Instead of appending tests that need display to `gtest_source`, we can accumulate a separate list and call `gz_build_tests` on them. We can then obtain the list of test targets from `gz_build_tests` and call [`set_test_properties`](https://cmake.org/cmake/help/v3.10/command/set_tests_properties.html) and set the [`DISABLED`](https://cmake.org/cmake/help/v3.10/prop_test/DISABLED.html) property on them.

Old-Ding:
## Summary
- Build display-dependent unit tests as separate `gz_build_tests` invocations instead of removing them from the source lists when no valid display is available.
- Disable their CTest entries when `VALID_DISPLAY` or `VALID_DRI_DISPLAY` is false, including the generated `check_` tests that would otherwise fail without XML output.
- Keep the existing target-specific setup for `UNIT_Server_Rendering_TEST` and `UNIT_ModelCommandAPI_TEST` intact.

Fixes #3496

## Testing
- `git diff --check origin/main..HEAD`
- Installed the Python-distributed CMake binary locally and ran `cmake -S . -B build\verify-display-tests -DBUILD_TESTING=ON -DENABLE_GUI=OFF`; configuration stopped at the expected missing local `gz-cmake` dependency before project CMake logic could run.
- Ran a temporary CMake/CTest harness that mocks `gz_build_tests(TEST_LIST ...)`; `ctest -C Debug --show-only=json-v1` confirmed these tests are registered and `DISABLED: true` when display detection is false:
  - `UNIT_Server_Rendering_TEST`
  - `check_UNIT_Server_Rendering_TEST`
  - `UNIT_ModelCommandAPI_TEST`
  - `check_UNIT_ModelCommandAPI_TEST`

Not run: full gz-sim configure/build/tests, because this Windows workspace does not have the Gazebo development dependencies such as `gz-cmake` installed.

## CI notes
- The current Jenkins build URL still reports `gz_sim-ci-pr_any-noble-amd64` failure, but `consoleText` returns HTTP 404 and `testReport/api/json` returns HTTP 503 as of 2026-07-10 14:42 CST.
- The last accessible failure evidence pointed to `UNIT_PeerTracker_TEST`, while this PR only changes display-dependent test registration for `UNIT_Server_Rendering_TEST` and `UNIT_ModelCommandAPI_TEST`; `network/PeerTracker_TEST.cc` is in the regular unit-test list, not `tests_needing_display`.

## Checklist
- [ ] Added tests (not applicable; this changes test registration)
- [ ] Updated documentation (not needed)
- [ ] Updated migration guide (not needed)
- [ ] `codecheck` passed (not run; local Gazebo dependencies are unavailable)
- [ ] All tests passed (not run; local Gazebo dependencies are unavailable)
- [x] While waiting for a review on this PR, I have helped review another open pull request: #3745
- [x] GenAI was used; the commit includes `Generated-by: OpenAI Codex`

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -143,12 +143,10 @@ set(tests_needing_display
 
 # Add systems that need a valid display here.
 # \todo(anyone) Find a way to run these tests with a virtual display such Xvfb
-# or Xdummy instead of skipping them
-if(VALID_DISPLAY AND VALID_DRI_DISPLAY)
-  list(APPEND gtest_sources ${tests_needing_display})
-else()
+# or Xdummy instead of disabling them at test time
+if(NOT VALID_DISPLAY OR NOT VALID_DRI_DISPLAY)
   message(STATUS
-    "Skipping these UNIT tests because a valid display was not found:")
+    "Disabling these UNIT tests because a valid display was not found:")
   foreach(test ${tests_needing_display})
     message(STATUS " ${test}")
   endforeach(test)
@@ -160,7 +158,7 @@ if (MSVC)
   # members that don't get exported, so they trigger this warning. However, the
   # warning is not important since those members do not need to be interfaced
   # with.
-  set_source_files_properties(${sources} ${gtest_sources} COMPILE_FLAGS "/wd4251 /wd4146")
+  set_source_files_properties(${sources} ${gtest_sources} ${tests_needing_display} COMPILE_FLAGS "/wd4251 /wd4146")
 endif()
 
 # Create the library target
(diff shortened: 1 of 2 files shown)


Old-Ding:
The noble amd64 failure is from `PeerTracker.PeerTrackerStale`, not from the display-test CMake changes in this PR:

- `PeerTracker.PeerTrackerStale` expected `stalePeers == 1` but observed `2`.
- Windows and Homebrew checks passed for this PR.

I opened #3765 as a separate, narrower fix for that existing flaky timing assertion so this PR can stay scoped to display-dependent tests.

PR #3770, opened 2026-07-04
doc: clarify ServerConfig SDF string parameter

Old-Ding:
## Summary
- clarify that `SetSdfString` takes SDF string contents, not an SDF file path
- fix a nearby typo in the `SetSdfFile` parameter docs

## Testing
- `git diff --check`


ahcorde (approved):

azeey:

PR #3771, opened 2026-07-04
fix: correct actuator_number error message grammar

Old-Ding:
## Summary
- correct the missing actuator_number error message grammar in multicopter and spacecraft actuator systems

## Testing
- `git diff --check`


ahcorde (approved):

azeey:

PR #3772, opened 2026-07-04
docs: consolidate wording corrections

Old-Ding:
# Bug fix

## Summary
- Consolidate the small, non-behavior-changing documentation and diagnostic wording corrections previously split across #3762, #3768, #3769, #3770, #3771, #3772, and #3774.
- Correct malformed or inaccurate text in the HPC tutorial, script examples, conversion documentation, server configuration documentation, and actuator parameter documentation.
- Improve grammar in existing light and actuator diagnostic messages without changing control flow or runtime behavior.

Related issue: #3371

## Backport Policy
- [x] This is safe to backport to the following versions:
    - [x] Jetty
    - [ ] Ionic
    - [ ] Harmonic
    - [ ] Fortress
- [ ] This **should not** be backported
- [ ] I am not sure
- [ ] Other (fill in yourself)

## Testing
- `git diff --check`
- Targeted old/new text assertions for all seven source changes
- DCO, package validation, Bazel, Ubuntu Noble, Ubuntu Noble GUI-disabled, Ubuntu Resolute, and Windows `gz_sim-pr-cnlwin` passed on the consolidated head

## CI notes
- `Ubuntu Resolute CI (GUI disabled)` failed in the existing `TrackedVehicleTest.PublishCmd` at `test/integration/tracked_vehicle_system.cc:366`: `poses.back().Pos().Z()` was `0.599049`, just below the `> 0.6` threshold. The consolidated diff does not touch that test, tracked-vehicle behavior, physics code, or its world files.
- Windows `gz_sim-pr-cnlwin` build 1358 passed for the consolidated head.

## Checklist
- [x] Signed all commits for DCO
- [ ] Added a screen capture or video to the PR description (not applicable: text-only changes)
- [ ] Added tests (not applicable: documentation and diagnostic text only)
- [x] Updated documentation
- [ ] Updated migration guide (not applicable)
- [ ] Consider updating Python bindings (not applicable)
- [ ] `codecheck` passed (no separate codecheck context was reported)
- [ ] All tests passed (one unrelated integration failure remains)
- [ ] Updated Bazel files (not applicable: no files added)
- [ ] While waiting for a review on this PR, please help review another open pull request
- [x] Was GenAI used to generate this PR? The required commit attribution is included below.

Assisted-by: OpenAI Codex

PR #3774, opened 2026-07-05
doc: fix example README typos

Old-Ding:
## Summary
- fix wording in the websocket server example README
- close a missing inline-code backtick in the log video recorder README

## Testing
- `git diff --check`


ahcorde (approved):

azeey:

PR #3775, opened 2026-07-05
Fix AdvancedLiftDrag pitch moment sideslip term

### LINKED ISSUE #3695: AdvancedLiftDrag: Cm alpha contribution appears to be overwritten by beta contribution
## Environment

- Gazebo Sim version: 8.11.0
- System/plugin: `gz::sim::systems::AdvancedLiftDrag`
- Branch/source checked: `gz-sim8`
- Source file: `src/systems/advanced_lift_drag/AdvancedLiftDrag.cc`

## Description

I noticed a possible issue in `AdvancedLiftDrag.cc` when computing the pitching moment coefficient `Cem`.

The code first computes `Cem` from the angle of attack using `Cem0`, `Cema`, `CemaStall`, and `alphaStall`. This suggests the expected pre-stall behavior is:

```cpp
Cem = Cem0 + Cema * alpha;
```
with post-stall handling using `CemaStall`.
However, immediately after that, the code appears to overwrite Cem with the sideslip contribution:
```cpp
Cem = this->Cemb * this->beta;
Cem += Cem_ctrl_tot;
```
Because this uses assignment instead of accumulation, the previously computed alpha-based pitching moment is discarded. As a result, Cem0, Cema, and Cema_stall may have no effect on the final pitching moment coefficient.

## Expected behavior
The sideslip contribution should probably be added to the existing alpha-based pitching moment, similar to a coefficient build-up method.
A possible change would be:
```cpp
Cem += this->Cemb * this->beta;
Cem += Cem_ctrl_tot;
```
instead of:
```cpp
Cem = this->Cemb * this->beta;
Cem += Cem_ctrl_tot;
```
## Why this seems unintended
There is a nearby comment saying:
```cpp
// Add sideslip effect, if any
```
This sounds like the beta contribution should be additive rather than replacing the alpha-based pitching moment.
Also, for a longitudinal case with beta = 0 and no control surfaces, the current implementation makes the final static Cem become zero, even if Cem0 and Cema are configured in the SDF.

## Question
Could you confirm whether this is intended behavior, or whether this should be changed to:
```cpp
Cem += this->Cemb * this->beta;
```

Old-Ding:
## Summary
- Preserve the alpha / stall-based `Cem` contribution when adding the `Cemb * beta` sideslip term.
- Add an integration regression test that loads `AdvancedLiftDrag` and checks beta = 0 still produces the configured pitch moment.

Fixes #3695

## Testing
- `git diff --check HEAD~1..HEAD`
- `git ls-files --eol src\systems\advanced_lift_drag\AdvancedLiftDrag.cc test\integration\lift_drag_system.cc`

Not run: `INTEGRATION_lift_drag_system` because this Windows machine does not have `cmake` or `ninja` installed.

### DIFF
--- a/src/systems/advanced_lift_drag/AdvancedLiftDrag.cc
+++ b/src/systems/advanced_lift_drag/AdvancedLiftDrag.cc
@@ -738,7 +738,7 @@ void AdvancedLiftDragPrivate::Update(EntityComponentManager &_ecm)
     Cem = this->Cem0 + this->Cema * this->alpha;
   }
   // Add sideslip effect, if any
-  Cem = this->Cemb * this->beta;
+  Cem += this->Cemb * this->beta;
 
   Cem += Cem_ctrl_tot;
 
--- a/test/integration/lift_drag_system.cc
+++ b/test/integration/lift_drag_system.cc
@@ -15,6 +15,8 @@
  *
  */
 
+#include <chrono>
+
 #include <gtest/gtest.h>
 
 #include <gz/msgs/double.pb.h>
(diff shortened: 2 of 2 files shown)


arjo129:
@Old-Ding I've merged #3817, which solves this exact same problem. For changes involving complex physics, it is critical that the contributions are backed and verified by humans. While the code diff is trivial, altering fundamental math can break behavior for a lot of users. Evaluating that risk and deciding on an acceptable rollout strategy is a human problem, not one we can leave to an AI. In this case, the author of #3817 is a specialist in a community that heavily uses this plugin, which gave us the necessary confidence to merge their change.

PR #3800, opened 2026-07-08
ci: update Bazel 7 version to 7.7.0

Old-Ding:
﻿## Summary
- update `.bazelversion` from `7.4.0` to `7.7.0`

## Root cause
The Bazel CI reusable workflow reads this file for the Bazel 7 matrix entry. Current dependency resolution pulls `assimp@6.0.3`, which declares `bazel_compatibility: [>=7.7.0]`, so the Bazel 7.4.0 job fails during module resolution before tests run.

## Impact
This keeps the existing Bazel 7 / Bazel 8 coverage while using the minimum Bazel 7 release compatible with the resolved dependency graph.

## Validation
- `git diff --check origin/main...HEAD`
- confirmed the `bazelbuild/bazel` `7.7.0` release tag exists through the GitHub API

Not run locally: Bazel tests, because this Windows environment does not have `bazel` or `bazelisk` installed.


azeey:
@shameekganguly there are similar failures elsewhere as well (e.g. https://github.com/gazebosim/gz-rendering/pull/1309). Can you take a look?

shameekganguly:
> @shameekganguly there are similar failures elsewhere as well (e.g. https://github.com/gazebosim/gz-rendering/pull/1309). Can you take a look?

Ah, sorry about this. Will forward fix with the bump to bazel 9.

shameekganguly:
Superceded  by https://github.com/gazebosim/gz-sim/pull/3802

PR #3650, opened 2026-06-03
Skip Recreate component query if no component exists

luca-della-vedova:
# 🎉 New feature

Skip `EachNoCache` if no component was ever created.

## Summary

`EachNoCache` is a very expensive operation that goes through all the entities to look for an entity with a specific component.
We have only one instance of usage in the `ProcessRecreateEntitiesRemove`, which looks for entities with the `Recreate` component.
However, this component is only ever instantiated in the Component Inspector Editor GUI plugin which is rarely used by users, and the cost of traversing the whole graph in an uncached way can be very expensive.
The trivial solution is to just bypass the call if the component was never created.
Based on the findings by @caguero in his [Gazebo profiling report](https://caguero.github.io/gz-profiling/2026-04-21/findings_report.pdf), first optimization recommendation.

Note that in the current ECM implementation we never clear the `createdCompTypes` map so as soon as the component is created once the short circuit will stop being effective for the rest of the run, however I feel this is not a big issue for two reasons:

1) Again the component is only created in a GUI plugin that most people might not be familiar with.
2) The Entt rewrite in #3447 makes `Each` and `EachNoCache` perfectly equivalent. The `EachNoCache` will just return immediately and do nothing if no entities with the component exist.

## Test it

With and without this PR run the 3k_shapes demo:

```bash
$ gz sim -v3 3k_shapes.sdf -r --physics-engine gz-physics-bullet-featherstone-plugin
```

On my machine, before the PR I get about 20%, after the PR about 50% so a RTF of more than 2x, but happy to hear if other people can reproduce or not.

## Checklist
- [x] Signed all commits for DCO
- [ ] Added a screen capture or video to the PR description that demonstrates the feature
- [ ] Added tests
- [ ] Added example and/or tutorial
- [ ] Updated documentation (as needed)
- [ ] Updated migration guide (as needed)
- [ ] Consider updating Python bindings (if the library has them)
- [ ] `codecheck` passed (See [contributing](https://gazebosim.org/docs/all/contributing#contributing-code))
- [ ] All tests passed (See [test coverage](https://gazebosim.org/docs/all/contributing#test-coverage))
- [ ] Updated Bazel files (if adding new files). Created an issue otherwise.
- [ ] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Agazebosim+archived%3Afalse+) to support the maintainers
[...]

PR #3811, opened 2026-07-09
Fix: Make SetPose UI not editable for planes

### LINKED ISSUE #3483: Pose UI silently fails to apply changes to planes
Greetings!

## Environment
* OS Version: Alt Linux Education P11
* Source or binary build?
  Source, release 10.1.1, last commit is 570d53e5fd7f4852e6f32567f2f4ac3b73b80250, tagged `gz-sim10_10.1.1`
* Not a rendering bug.

## Description
* Expected behavior: Pose UI for planes is either not editable or applies edits
* Actual behavior: Pose UI for planes is editable but silently fails to apply edits

## Steps to reproduce
1. Open empty built-in world
2. Choose ground plane
3. Try to adjust it's position

## Output
Image
Image
Image

## Comment
Seems like the edit is rejected to avoid engine crash (Physics.cc 2551). Is this an intended limitation of the engine?
I'm guessing either
1. Engine is not supposed to crash in response to this request
or
2. Relevant fields are not supposed to be editable 
or
3. The error message is supposed to be visible on the UI

Originally reported here: https://bugzilla.altlinux.org/57553

silverducks143:
# 🦟 Bug fix

Fixes #3483 

## Summary
Fix makes Pose UI fields read-only when selection has plane geometry.
To achieve this, Component Inspector was extended with an extra property - modelContainsPlane - and that property was added to readOnly block in Pose3d.qml. Screenshot with PR applied:
image

### Steps to reproduce the bug:
1. Open empty built-in world
2. Choose ground plane
3. Try to adjust it's position

Expected behavior: Pose UI for planes is either not editable or applies edits
Actual behavior: Pose UI for planes is editable but silently fails to apply edits

## Backport Policy
    - [ ] Jetty
    - [ ] Ionic
    - [ ] Harmonic
    - [ ] Fortress
- [X] I am not sure

## Checklist
- [X] Added a screen capture or video to the PR description that demonstrates the fix (as needed)
- [ ] Added tests


src/gui/plugins/component_inspector/ComponentInspector.cc
```diff
@@ -537,6 +541,30 @@ void ComponentInspector::Update(const UpdateInfo &,
       }
       this->NestedModelChanged();
 
+      // Check if any collision in this model has plane geometry.
+      bool containsPlane = false;
+      _ecm.EachNoCache<components::Collision,
```
luca-della-vedova: ⚠️  that this API performs horribly, in fact removing it from our SimulationRunner alone doubled our RTF https://github.com/gazebosim/gz-sim/pull/3650.
Did you check how this change performs in a world with many entities (i.e. 3k_shapes)? Does it need to be the `NoCache` variant of `Each`?
If it affects performance significantly I think we should not backport it, if #3447 gets in `EachNoCache` will have no significant performance hit so merging it forward should be OK.

silverducks143: I wasn't really paying attention to performance. You're right - even on a mid-size scene impact is noticeable, let alone on `3k_shapes`.
Yeah, `NoCache` variant is indeed unnecessary here.
It's also iterating over everything, which is also unnecessary.

Thanks for the catch.
I've modified the check to fix both problems.

silverducks143:
Does not work on latest patch, investigating.

silverducks143:
Found the issue. 
This fix requires merging of this backport:
https://github.com/gazebosim/gz-sim/pull/3839
Merges cleanly into gz-sim10.5.0 on my end.
Once merged, no further changes to the fix are necessary.

cheriehu:
Testing the behavior on jetty (merged the latest gz-sim10 branch locally) but it doesn’t seem to work for me. The SetPose UI is still editable for planes. When I try to edit it, there is an error on the terminal that informs me it’s not editable. I’m on macOS Tahoe.

But if I cherry-pick [e09d448](https://github.com/gazebosim/gz-sim/pull/3811/commits/e09d4489472f6c8b383a595b8c6d9c5203aedb3b) onto gz-sim on main (rotary), the fix works and the UI is not editable.

on jetty:
image

on rotary (main):
image

silverducks143:
Have you applied fix from backport #3839 to jetty before applying this fix? This behavior matches my testing before I merged said backport.

To clarify, the fix worked on gz-sim v10.1.0 and broke on v10.2.0+. With the backport it works on v10.2.0+ again. PR from which the backport was generated is already merged on main (rotary), so it works there with no additional actions required.

The error in terminal is pre-existing and not created by this fix.

cheriehu:
Ahh ok my bad I missed that, yep applying that fix works! 
image

PR #3816, opened 2026-07-10
Get clock cpu load

jmachowinski:
# 🦟 Bug fix

## Summary

    fix: Don't wake render thread without reason
    
    The sensor thread was awakened for no reason ever ms.
    This lead to a high cpu load on clock_gettime.

    fix: Fixed broken system timestamp
    
    The timestamp was computed from multiple calls to
    std::chrono::system_clock::now(). This was a inefficient
    and b broken in case the second flipped between calls.

## Backport Policy

    - [x] Jetty
    - [ ] Ionic
    - [ ] Harmonic
    - [ ] Fortress
- [ ] I am not sure

## Checklist
- [x] Added a screen capture or video to the PR description that demonstrates the fix (as needed)
- [ ] Added tests

### DIFF
--- a/src/SimulationRunner.cc
+++ b/src/SimulationRunner.cc
@@ -599,9 +599,16 @@ void SimulationRunner::PublishStats()
     clockMsg.mutable_real()->set_nsec(realTimeSecNsec.second);
     clockMsg.mutable_sim()->set_sec(simTimeSecNsec.first);
     clockMsg.mutable_sim()->set_nsec(simTimeSecNsec.second);
-    clockMsg.mutable_system()->set_sec(GZ_SYSTEM_TIME_S());
+
+    auto curTime = std::chrono::system_clock::now();
+    auto seconds = std::chrono::duration_cast<std::chrono::seconds>(
+      curTime.time_since_epoch()).count();
+    auto ns = std::chrono::duration_cast<std::chrono::nanoseconds>(
+      curTime.time_since_epoch()).count();
+    clockMsg.mutable_system()->set_sec(seconds);
     clockMsg.mutable_system()->set_nsec(
-        GZ_SYSTEM_TIME_NS() - GZ_SYSTEM_TIME_S() * GZ_SEC_TO_NANO);
+      ns - seconds * GZ_SEC_TO_NANO);
+
     this->clockPub.Publish(clockMsg);
 
     // Only publish to root topic if no others are.
--- a/src/SimulationRunner.cc
+++ b/src/SimulationRunner.cc
@@ -600,7 +600,7 @@ void SimulationRunner::PublishStats()
     clockMsg.mutable_sim()->set_sec(simTimeSecNsec.first);
     clockMsg.mutable_sim()->set_nsec(simTimeSecNsec.second);
 
-    auto curTime = std::chrono::system_clock::now();
+    auto curTime = GZ_SYSTEM_TIME();
     auto seconds = std::chrono::duration_cast<std::chrono::seconds>(
       curTime.time_since_epoch()).count();
     auto ns = std::chrono::duration_cast<std::chrono::nanoseconds>(
--- a/src/systems/sensors/Sensors.cc
+++ b/src/systems/sensors/Sensors.cc
@@ -396,7 +396,7 @@ void SensorsPrivate::RunOnce()
 {
   {
     std::unique_lock<std::mutex> cvLock(this->renderMutex);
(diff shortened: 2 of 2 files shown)


jmachowinski:
Before fix : 
image

After fix:
image

src/SimulationRunner.cc
```diff
@@ -599,10 +599,15 @@ void SimulationRunner::PublishStats()
...
+    auto curTime = std::chrono::system_clock::now();
+    auto seconds = std::chrono::duration_cast<std::chrono::seconds>(
+      curTime.time_since_epoch()).count();
+    auto ns = std::chrono::duration_cast<std::chrono::nanoseconds>(
+      curTime.time_since_epoch()).count();
+
+    clockMsg.mutable_system()->set_sec(seconds);
     clockMsg.mutable_system()->set_nsec(
-        GZ_SYSTEM_TIME_NS() - GZ_SYSTEM_TIME_S() * GZ_SEC_TO_NANO);
-    this->clockPub.Publish(clockMsg);
```
luca-della-vedova, line 605: CI failure is real, it seems you removed the line that actually publishes the clock so it won't be published anymore?

luca-della-vedova (commented):
I can't comment on the sensor CV change since I'm not too familiar with that, there is an error in the clock logic that makes the logic a noop / the test segfault.
Apart from that, can you share a way for maintainers to reproduce the performance gains? perhaps if it improves on any of our benchmarks or a custom world to try?

src/SimulationRunner.cc
jmachowinski: Fixed, that broke on the cherry pick from jetty...

jmachowinski:
Any world with a few render sensors will do (e.g. cameras). In my test we had 2 cameras at 30 fps in the scene. And a 5 ms update loop in the physics.

To see the change in performance, just use perf.

Also note there is a second big source of high CPU usage, that we patched out in our local version with this:
      constexpr auto kSpinThreshold = 00us;

src/SimulationRunner.cc
```diff
@@ -599,9 +599,16 @@ void SimulationRunner::PublishStats()
     clockMsg.mutable_real()->set_nsec(realTimeSecNsec.second);
     clockMsg.mutable_sim()->set_sec(simTimeSecNsec.first);
     clockMsg.mutable_sim()->set_nsec(simTimeSecNsec.second);
-    clockMsg.mutable_system()->set_sec(GZ_SYSTEM_TIME_S());
+
+    auto curTime = std::chrono::system_clock::now();
```
luca-della-vedova: Nit, it seems there is a macro in gz-common, `GZ_SYSTEM_TIME()` can we use that since that seemed to be the convention?

luca-della-vedova (approved):
Thanks for your patience! Overall LGTM, I had a look at the `RunOnce` code path and it seems it's called in a continuous loop, so indeed having a shorter timeout will just make it return earlier, have the boolean flags not set and then be triggered to run immediately afterwards.

https://github.com/gazebosim/gz-sim/blob/2d3c7c777cc0a959666e3de497a38dca71b92409/src/systems/sensors/Sensors.cc#L537-L540

Furthermore, we notify the condition variable [here](https://github.com/gazebosim/gz-sim/blob/2d3c7c777cc0a959666e3de497a38dca71b92409/src/systems/sensors/Sensors.cc#L576) when setting `running` to false (that would stop the above loop) hence so this change shouldn't even introduce a delay in stopping.

The clock change also sounds reasonable to me, indeed calling `clock::now()` several times is risky if the second flips, regardless of performance implications it's more correct.

Thanks for the contribution! Just a minor nit to use a macro instead of re-implementing it, but generally LGTM

src/SimulationRunner.cc
jmachowinski: I don't like the macro for two reasons:
- It is a marco
- Looking at the code with the macro I would need to lookup the macro to understand the code. E.g. in the end it makes it harder to read and does not offer any advantage.

I don't care either way though. 
I rebased the patch and created a second branch get_clock_cpu_load_macro just pick the version you prefer

luca-della-vedova:
Thanks for iterating! I noticed you mentioned this can only be backported to Jetty, any particular reason for that vs backporting to all released versions?

jmachowinski:
I only tested it on jetty not sure about the rest therefore I did not check the boxes.

luca-della-vedova:

PR #3847, opened 2026-07-23
Add Entt vendored headers

luca-della-vedova:
# 🎉 New feature

Part of #3447, only add the vendored headers + changes to BUILD.bazel.
The headers are copied from the v3.16.0 branch upstream, with one additional change I had to do (add missing headers, separate commit for simplicity) to avoid CI failures when building with Bazel + Clang.

## Summary

I chose v3.16.0 since it's the latest released version. We could point to the latest `main` if we wanted but I haven't tried that out, I think among other things it uses C++ concepts.

## Backport Policy
    - [ ] Jetty
    - [ ] Ionic
    - [ ] Harmonic
    - [ ] Fortress
- [x] This **should not** be backported
- [ ] I am not sure
## Test it

Tested through CI, as part of #3447 

## Checklist
- [ ] Added tests


azeey:
Would it be possible to put this inside `vendor/entt` similar to `vendor/backward-cpp` and add supporting files such as `LICENSE`, a README containing the hash used, a patch file containing the changes you made, etc? cc @j-rivero

caguero (commented):
Should we exclude the entire `entt` tree from `gz_install_all_headers()`? At least for now.

caguero:
Should we try to upstream your missing-include patch?

azeey:
> Should we exclude the entire entt tree from gz_install_all_headers()? At least for now.

I think they'll need to be installed because this is a template heavy library. Our ECM templates instantiate EnTT templates, so they need to be accessible by downstream users.

luca-della-vedova:
> Should we try to upstream your missing-include patch?

I went into a bit of a rabbit hole and I couldn't find a straightforward way to reproduce it in CI upstream and make a case for it but I'll keep digging.
Anyway I can try to do that for `main` (which contains a newer version than this, since it needs C++-20) but realistically I doubt it will be backported to 3.16.
Specifically, the [upstream repo](https://github.com/skypjack/entt) only has a 3.13.x branch, it seems 3.16.0 was just a tag on main:

```
lucadv@lucadv02:~/entt$ git branch --contains v3.16.0
* main
```

So strictly speaking there isn't really a branch to backport the fix to, unless of course the maintainer branches off the 3.16.0 tag to create a 3.16.x branch, merges our fix to main, backports it and makes a new patch release, but I feel that is a bit of an ask!

luca-della-vedova:
I tried a different approach of using CMakeLists FetchContent API to get the library at configuration time, together with adding a dependency to the Bazel build.
I'm also discussing with the maintainer upstream in https://github.com/skypjack/entt/pull/1346 whether they would be interested in uploading their crate to BCR and / or have some fixes backported to a stable branch so we can remove the patch and cleanup the bazel dependency.

From a high level, what would be better between this approach and just copying in the code? I suppose this approach commits a lot less code but is a bit more complex to navigate

azeey:
> From a high level, what would be better between this approach and just copying in the code? I suppose this approach commits a lot less code but is a bit more complex to navigate

I seem to recall that having internet access during a build is not encouraged by our infra team. None of our vendored packages do that currently, except mujoco, but that's a temporary workaround. @j-rivero might have a more accurate opinion.

j-rivero:
> I seem to recall that having internet access during a build is not encouraged by our infra team. None of our vendored packages do that currently, except mujoco, but that's a temporary workaround. @j-rivero might have a more accurate opinion.

Adding the requirement of a working internet connection to the build package, particularly at a configuration time, it is a call to problems that I would personally avoid if possible and other projects (like Debian, Ubuntu) directly forbid.

luca-della-vedova:
OK I copied the headers back in and into a gz/sim/detail folder, I thought having them in a `vendor` folder would have created some slightly confusing behavior since we would need some CMake machinery to include a folder under `vendor`, and it might not be super immediate to users where the header is coming from.
I also added a README with information on what commit we are based off, what was copied and what was not, as well as a patch file with the diff.
I'll get to upstreaming the patches next week or so, please let me know if it looks OK from a high level!

azeey (approved):

luca-della-vedova:
It seems CI on this PR will be eternally red because we have a test that builds a library that includes an auto-generated header that will include recursively all generated headers, including Entt.
Because of how Entt is designed this will result in that test having a compile warning for switch with no default statement, specifically this is what I get:

```
[...]
~/ws_jetty_upstream/src/gz-sim/include/gz/sim/detail/vendor/entt/core/any.hpp:73:9: warning: switch missing default case [-Wswitch-default]
   73 |         switch(const auto *elem = static_cast<const Type *>(value.data()); req) {
```

I merged this change inside #3447, but unless we add some CMake machinery to avoid including the Entt headers / suppress warnings (and revert it later) I doubt we will be able to have this merged.

luca-della-vedova:
This is an alternative to make CI green and this merge-able https://github.com/gazebosim/gz-sim/pull/3847/commits/f25d3acf56a537e17d6ee47f6a696b20a9632349, the header that seems to cause the issue in CI is actually unused so I removed it together with a bunch of other headers that also included it.
This should make our patch files and code slightly smaller, at the expense of a bit more of a diff towards the main repo.
@azeey what do you think?

luca-della-vedova:
@osrf-jenkins run tests

PR #3849, opened 2026-07-23
Build display-dependent tests without running them

### LINKED ISSUE #3496: Tests skipped due to missing display should still be compiled

tylergibbs1:
# 🎉 New feature

Closes #3496

## Summary

Display-dependent unit, command, integration, and performance tests are now
always registered and compiled. When CMake cannot find a valid display or DRI
display, CTest disables both each generated test and its `check_` companion.
This lets headless and macOS CI catch compile regressions without attempting to
run rendering tests or validate result files that were never produced.

The existing rendering links and MSVC compile flags now also apply when these
tests are built on a machine without a display.

Some portions of this pull request were generated using OpenAI Codex (GPT-5).

## Backport Policy

    - [ ] Jetty
    - [ ] Ionic
    - [ ] Harmonic
    - [ ] Fortress
- [x] I am not sure

## Test it

Configure on a machine without a display:

```sh
unset DISPLAY WAYLAND_DISPLAY
cmake -S . -B build -DBUILD_TESTING=ON -DSKIP_PYBIND11=ON
cmake --build build
ctest --test-dir build -N
```

The display-dependent executables should be built, while their runtime tests
and `check_` companion tests should be listed as disabled.

Locally validated on macOS against the Rotary dependency stack:

- all 28 display-dependent executables compiled and linked
- all 56 runtime and result-check CTest entries were disabled
- `codecheck` completed successfully

## Checklist

- [ ] Added a screen capture or video to the PR description that demonstrates the feature (not applicable: build-system-only change)
- [x] Added tests (existing display-dependent tests now cover headless compilation)
- [ ] Added example and/or tutorial (not applicable)
- [ ] Updated documentation (not applicable)
- [ ] Updated migration guide (not applicable)
- [ ] Consider updating Python bindings (not applicable)
- [ ] All tests passed (See [test coverage](https://gazebosim.org/docs/all/contributing#test-coverage)) (targeted build and CTest validation passed; full matrix is delegated to CI)
- [x] Updated Bazel files (if adding new files). Created an issue otherwise. (not applicable: no files added)

Assisted-by: OpenAI Codex (GPT-5)
Generated-by: OpenAI Codex (GPT-5)

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -141,26 +141,15 @@ set(tests_needing_display
   Server_Rendering_TEST.cc
 )
 
-# Add systems that need a valid display here.
-# \todo(anyone) Find a way to run these tests with a virtual display such Xvfb
-# or Xdummy instead of skipping them
-if(VALID_DISPLAY AND VALID_DRI_DISPLAY)
-  list(APPEND gtest_sources ${tests_needing_display})
-else()
-  message(STATUS
-    "Skipping these UNIT tests because a valid display was not found:")
-  foreach(test ${tests_needing_display})
-    message(STATUS " ${test}")
-  endforeach(test)
-endif()
-
 if (MSVC)
   # Warning #4251 is the "dll-interface" warning that tells you when types used
   # by a class are not being exported. These generated source files have private
   # members that don't get exported, so they trigger this warning. However, the
   # warning is not important since those members do not need to be interfaced
   # with.
-  set_source_files_properties(${sources} ${gtest_sources} COMPILE_FLAGS "/wd4251 /wd4146")
+  set_source_files_properties(
+    ${sources} ${gtest_sources} ${tests_needing_display}
+    COMPILE_FLAGS "/wd4251 /wd4146")
 endif()
 
 # Create the library target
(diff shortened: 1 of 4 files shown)


azeey:
Thanks for the contribution @tylergibbs1! https://github.com/gazebosim/gz-sim/pull/3759 might also be fixing the same issue as linked in https://github.com/gazebosim/gz-sim/issues/3496. Do you mind reviewing that and see if the two can be combined?

tylergibbs1:
thanks for pointing me to #3759. i compared the two. it covers the same unit and command test changes, while #3849 also covers the integration and performance test groups. that extra integration coverage matters for #3496 because `reset_sensors.cc`, the example in that issue, lives there.

i checked the current ci failures too. the resolute job hit the same unrelated `PeerTrackerStale` failure. windows built and ran the tests, but the newly compiled integration targets surfaced 19 existing `MSVC` conversion warnings, so jenkins marked the build unstable even though the tests passed.

since #3759 came first, my preference is to use it as the base, add the integration and performance pieces from this pr, and close #3849. if you'd rather keep this pr, i can fold #3759 into this branch and preserve the original authorship. which route would you and @Old-Ding prefer?

tylergibbs1:
Thank you for the pointer. I read #3759. The two pull requests overlap.

Both use the same method. Each one registers the display-dependent tests as a separate `gz_build_tests` call. Each one then sets the CTest entries to `DISABLED` when there is no valid display. Neither one removes the sources.

#3759 changes the UNIT tests in `src/` and `src/cmd/`. This pull request changes those tests, and also the INTEGRATION and PERFORMANCE tests. It adds the MSVC `/wd4251` flags for those two directories. It also makes the `gz-rendering` links unconditional. This last change is necessary, because the targets now exist on a machine without a display.

#3759 came first. The clearest path is to merge #3759 first. Then I rebase this pull request to the INTEGRATION and PERFORMANCE part only. I can also put all the work in one pull request, with credit to @Old-Ding. Please tell me which one you prefer.

One small difference: #3759 guards the `check_` entries with `Python3_Interpreter_FOUND`. This pull request uses `if(TEST check_${test_target})`. Both work, because `gz_build_tests` is a macro. The macro calls `find_package(Python3)` and sets the variable in the scope of the caller. A test for the entry itself depends less on that order.

PR #3685, opened 2026-06-16
Cache physics step output

luca-della-vedova:
# 🎉 New feature

Part of #3608 

Cache forward step output in Physics system.

Marking as a draft while I test CI and whether this has any implications I'm not aware of.

## Summary

When benchmarking a large world (3k_shapes), I noticed that a large amount of time was spent in reserving vectors to push the [world poses](https://github.com/gazebosim/gz-physics/blob/8a48173082a38b591ce1d5ca2829227d87ae8505/bullet-featherstone/src/SimulationFeatures.cc#L355) and the [changed poses](https://github.com/gazebosim/gz-physics/blob/8a48173082a38b591ce1d5ca2829227d87ae8505/bullet-featherstone/src/SimulationFeatures.cc#L371).
This was puzzling, since we call `clear()` just before the reserve but `clear()` does not change the capacity of the underlying memory, it only resets the size of the vector.
With some printouts and backtracing I realized we are creating a brand new `ForwardStep::Output` object at every simulation iteration, which means allocating + deallocating two possibly large vectors at every simulation step.
I just cached the step output in the private data member.

## Backport Policy

- [x] This is safe to backport to the following versions:
    - [x] Jetty
    - [x] Ionic
    - [x] Harmonic
    - [x] Fortress
- [ ] This **should not** be backported
- [ ] I am not sure
- [ ] Other (fill in yourself)
## Test it

Run the 3k_shapes world with bullet featherstone:

```
$ gz sim -v3 3k_shapes.sdf --physics-engine libgz-physics-bullet-featherstone-plugin -z 1000000 -r
$ gz sim -g
```

In my (fast) machine this brings up the RTF of the 3k_shapes.sdf with bullet featherstone from ~110% to ~160%, so about a 50% increase.

## Checklist
- [x] Signed all commits for DCO
- [ ] Added a screen capture or video to the PR description that demonstrates the feature
- [ ] Added tests
- [ ] Added example and/or tutorial
- [ ] Updated documentation (as needed)
- [ ] Updated migration guide (as needed)
- [ ] Consider updating Python bindings (if the library has them)
- [ ] `codecheck` passed (See [contributing](https://gazebosim.org/docs/all/contributing#contributing-code))
- [ ] All tests passed (See [test coverage](https://gazebosim.org/docs/all/contributing#test-coverage))
- [ ] Updated Bazel files (if adding new files). Created an issue otherwise.
[...]

PR #3850, opened 2026-07-24
Avoid copying cached physics step output

hugogo1998:
# 🦟 Bug fix

Follow-up to #3685.

## Summary

#3685 caches `ForwardStep::Output` in `PhysicsPrivate`, but `Step()` still returns it by value. This copies the cached `CompositeData` before `ChangedLinks()` consumes it.

This PR returns the cached output by const reference instead. Paused updates use a static empty output, preserving the existing behavior. This is an internal implementation change and does not affect the public API or ABI.

## Backport Policy

  - [ ] Jetty
  - [ ] Ionic
  - [ ] Harmonic
  - [ ] Fortress
- [x] I am not sure

### Performance

I compared the parent commit (`66bfce0`) with this PR (`1bad4e0`). Both were built with `RelWithDebInfo` and pinned to CPU 8 on an AMD EPYC 9554.

The existing `BM_RuntimeWorld/lengthy_sdf_3k_shapes_bullet/3000` benchmark showed no meaningful difference: 1.09 ms/step on the parent and 1.08 ms/step on this PR. After its 3,000-step warm-up, most objects are asleep and there is little changed-link output to copy.

To exercise the affected path, I used a temporary copy of the same 3,003-shape world with the dynamic shapes moved to `z=100`. Each run measured 500 falling steps after one warm-up step. I ran 30 alternating parent / PR pairs on the same CPU:

| | Parent | This PR |
|---|---:|---:|
| Median step time | 12.744 ms | 12.238 ms |
| Estimated RTF | 0.0785 | 0.0817 |

This is a 4.13% increase in throughput / RTF, or a 3.97% reduction in step time. The PR was faster in 23 of 30 pairs. A paired bootstrap placed the throughput improvement between 2.45% and 5.88% at the 95% confidence level.

The machine had CPU frequency scaling enabled, so the runs were alternated and the median is reported to limit the effect of system noise.

The benchmark command was:

```sh
podman run --rm --cpuset-cpus 8 \
  --env GZ_CONFIG_PATH=/workspace/build/test/conf/RelWithDebInfo \
  --env GZ_IP=127.0.0.1 \
  --env GZ_SIM_SYSTEM_PLUGIN_PATH=/workspace/build/lib \
  --volume "$source_dir:/workspace" \
  --workdir /workspace \
  --entrypoint /workspace/build/bin/BENCHMARK_server_run \
  localhost/gz-ci-noble-rotary-deps:latest \
  --benchmark_filter='BM_RuntimeWorld/active_sdf_3k_shapes_bullet/1/iterations:500$'
```

The temporary benchmark registration was:

```cpp
BENCHMARK_CAPTURE(BM_RuntimeWorld, active_sdf_3k_shapes_bullet,
                  "gz-physics-bullet-featherstone-plugin",
                  "3k_shapes_falling.sdf")
    ->Arg(1)
    ->Iterations(500)
    ->Unit(benchmark::kMillisecond);
```

### Profiling

I rebuilt both versions with `-fno-omit-frame-pointer` and recorded the same workload with:

```sh
perf record -e cycles:u -F 499 -m 8192 -g --call-graph fp \
  -o /tmp/pr3850.data -- \
  podman run --rm --cpuset-cpus 8 \
  --env GZ_CONFIG_PATH=/workspace/build/test/conf/RelWithDebInfo \
  --env GZ_IP=127.0.0.1 \
  --env GZ_SIM_SYSTEM_PLUGIN_PATH=/workspace/build/lib \
  --volume "$source_dir:/workspace" \
  --workdir /workspace \
  --entrypoint /workspace/build/bin/BENCHMARK_server_run \
  localhost/gz-ci-noble-rotary-deps:latest \
  --benchmark_filter='BM_RuntimeWorld/active_sdf_3k_shapes_bullet/1/iterations:500$'
```

In the parent profile, `PhysicsPrivate::Step` accounted for 29.90% of samples. The `SpecifyData::SpecifyData` copy directly below it accounted for 3.22%.

With this PR, that copy stack is absent and `PhysicsPrivate::Step` accounts for 26.80% of samples.

**Before: cached output returned by value**

Before flamegraph

**After: cached output returned by const reference**

After flamegraph

Full-resolution flamegraphs are also available in [this gist](https://gist.github.com/hugogo1998/1c761ab5712518bbc3ecd59fae5fbe41).

### Correctness

A clean `RelWithDebInfo` build, `codecheck`, and `INTEGRATION_play_pause` passed.

`INTEGRATION_physics_system` had the same five failures on the parent and PR builds, with no new failures introduced by this change.

## Checklist

- [x] Screen capture is not applicable because there is no visible behavior change
- [x] Existing physics and play / pause integration tests cover the changed paths
- [x] Documentation is not affected
- [x] Migration guide is not affected
- [x] Python bindings are not affected
- [x] `codecheck` passed
- [ ] All tests passed — see the parent comparison above
- [x] Bazel files are not affected
- [ ] While waiting for a review on this PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Agazebosim+archived%3Afalse+) to support the maintainers
- [x] Was GenAI used to generate this PR? If so, make sure to add "Assisted-by" to your commits.

Assisted-by: OpenAI Codex


hugogo1998:
Hi @luca-della-vedova , could you please take a look when you have time? I'd appreciate any feedback, and pls let me know if I overlook anything. Thanks

luca-della-vedova:
Since this is a performance improvement I would like to see some results that show that this is indeed faster.
We have some benchmarks that run high level tests [here](https://github.com/gazebosim/gz-sim/blob/main/test/benchmark/server_run.cc).
If the improvement is not captured in the benchmark, you could also capture a flamegraph through `perf` and attach a before / after flamegraph showing a reduced amount of time spent in the function you are optimizing, together with the exact command you ran and an estimation of the change in simulation RTF that you observed.
You can refer to any of the recent performance PRs to see how to test / report findings (i.e. https://github.com/gazebosim/gz-physics/pull/1005). The testing section in your PR doesn't really say anything about the performance improvements and instead seems like an AI generated "passed CI" testing result.

hugogo1998:
@luca-della-vedova Thank you very much for the comment and pointers. Let me profile the code before and after patch on the benchmarks and show you some proofs later.

hugogo1998:
Thanks for the guidance. I updated the PR description with the benchmark setup, exact commands, RTF estimate, and before / after flamegraphs.

The existing 3k-shape benchmark is effectively post-sleep after its 3,000-step warm-up and showed no meaningful difference (1.09 vs 1.08 ms/step). To exercise `ForwardStep::Output`, I moved the same 3,003 dynamic shapes to `z=100` in a temporary world and measured 500 falling steps after one warm-up step.

Across 30 alternating parent / PR pairs, median step time went from 12.744 ms to 12.238 ms. Estimated RTF went from 0.0785 to 0.0817, a 4.13% throughput improvement. The PR was faster in 23 of 30 pairs, and the paired bootstrap 95% interval was +2.45% to +5.88%.

The profile also shows the targeted change directly: the `SpecifyData::SpecifyData` copy below `PhysicsPrivate::Step` accounted for 3.22% of the parent profile and is absent after the change. `PhysicsPrivate::Step` went from 29.90% to 26.80% of samples.

The host had frequency scaling enabled, so I used alternating runs and report the median rather than claiming every individual run is faster.

hugogo1998:
To me though the improvement does not look "huge", I feel it is in the hot path so it should still be meaningful. But if you feel it's not an significant improvement I'm fine with closing it.

luca-della-vedova (approved):
Thanks for the flamegraphs! This sounds reasonable to me, the gain is not huge but it's significant enough and it doesn't really introduce any additional complexity other than a static const variable.

LGTM with green CI

iche033 (approved):

src/systems/physics/Physics.cc
```diff
@@ -1125,12 +1126,9 @@ void Physics::Update(const UpdateInfo &_info, EntityComponentManager &_ecm)
   {
     this->dataPtr->CreatePhysicsEntities(_ecm);
     this->dataPtr->UpdatePhysics(_ecm);
-    gz::physics::ForwardStep::Output stepOutput;
-    // Only step if not paused.
-    if (!_info.paused)
-    {
-      stepOutput = this->dataPtr->Step(_info.dt);
-    }
+    static const gz::physics::ForwardStep::Output emptyStepOutput;
```
azeey: Objects with static storage duration are forbidden in our [style guide](https://google.github.io/styleguide/cppguide.html#Static_and_Global_Variables) due to [static initialization order fiasco](https://en.cppreference.com/cpp/language/siof). Can we maybe update `ChangedLinks` to take a `const std::optional<Output> &` instead?

azeey (changes requested):

src/systems/physics/Physics.cc
hugogo1998: Thanks for catching this. Updated in 3b8ff91cf: `ChangedLinks` now takes a `const std::optional<ForwardStep::Output> &`, and paused updates pass an empty automatic optional instead of a static object. I also made the cached output itself an in-place optional and return it by const reference, so the unpaused path still avoids copying `ForwardStep::Output`.

azeey (approved):

src/systems/physics/Physics.cc
azeey: Thanks for the quick fix!

luca-della-vedova:

PR #3851, opened 2026-07-24
Reconstruct legacy ComponentStorage vector-to-deque fix as reference

hugogo1998:
### Disclosure

This PR is being submitted by me, not by an automated agent.

An automated profiling tool I use originally identified the optimization described here. Before submitting anything, I checked the current `main` branch myself and found that the affected code no longer exists—the underlying issue had already been eliminated by a later architectural refactor. Rather than discarding the finding, I reconstructed the historical implementation, verified the original behavior, and wrote up the analysis below myself. This PR is therefore documentation of a historical performance issue, **not** a change to the current implementation.

### Summary

This PR does **not** change the behavior or performance of current `gz-sim`.

The optimization opportunity originally found by the profiling tool applied to the historical `ComponentStorage<T>` implementation. That class no longer exists in the current codebase, and the performance issue it contained has already been eliminated by a later redesign. This PR preserves a reconstructed reference implementation and benchmark demonstrating the historical issue, since it seemed worth documenting rather than silently discarding.

### What actually happened here

The historical `ComponentStorage<T>` (`include/ignition/gazebo/detail/ComponentStorageBase.hh`) stored each component type in a `std::vector` reserved to 100 elements and expanded in blocks of 100. Every expansion returned an `expanded` flag, and `EntityComponentManager::CreateComponentImplementation` responded by calling `RebuildViews()`, an operation proportional to the number of views, entities, and component types.

As a result, populating a world of *E* entities triggered a full rebuild roughly every 100 insertions of a component type. Since each rebuild was itself O(E), overall world population became O(E²).

Current `gz-sim` no longer uses this storage design. Component storage is now

```cpp
std::unordered_map<Entity,
    std::vector<std::unique_ptr<components::BaseComponent>>>
```

where each component is individually heap allocated. Growing the per-entity vector moves only owning pointers, not component objects, and ordinary component creation now performs incremental updates through `view->MarkEntityToAdd()`. `RebuildViews()` remains only for the full entity-reset path.

In other words, the architectural redesign independently eliminated the same bottleneck that the profiling tool identified in the historical implementation.

### What's in this PR

`src/legacy_component_storage/` (not referenced by any build file and not compiled):

- `ComponentStorageBase_reference.hh` — reconstructed historical `ComponentStorage<T>` from `ignition-gazebo` / `gz-sim` commit `34ac465`, trimmed to only the storage logic relevant to the performance issue.
- `ComponentStorageBase_ours.hh` — the same implementation using `std::deque` instead of `std::vector`.
- `ComponentStorageBase_TEST.cc` — standalone verification code used during reconstruction.

### Results (historical implementation)

| entities | vector (old) | deque | speedup |
|---|---:|---:|---:|
| 1,000 | 1133 us | 339 us | 3.3x |
| 2,000 | 4207 us | 770 us | 5.5x |
| 4,000 | 18256 us | 1805 us | 10.1x |
| 8,000 | 73051 us | 3224 us | 22.7x |

The improvement grows with scale because the historical implementation is genuinely O(E²), while the `std::deque` version avoids the repeated reallocations that trigger global rebuilds. These measurements describe the historical code path only—they do **not** represent current `gz-sim`, where the bottleneck no longer exists.

### Verification

I verified the reconstruction by running `ComponentStorageBase_TEST.cc`, creating 8,000 simulated components through both implementations.

Both versions produced identical component IDs and stored values throughout the run. The original vector implementation raised its `expanded` flag 79 times, matching the expected ~E/100 expansion behavior that previously triggered full view rebuilds. The `std::deque` version never raised the flag.

### Why submit this?

Normally I would simply drop a finding if the target code had disappeared. This case felt different because the profiling tool uncovered a genuine algorithmic issue, but by the time I investigated it had already been solved independently by a larger architectural change.

Rather than letting that analysis disappear entirely, I reconstructed and documented it as a historical reference. If that isn't useful to keep in the repository, I'm completely happy for the PR to be closed.

### DIFF
--- a/src/legacy_component_storage/ComponentStorageBase_TEST.cc
+++ b/src/legacy_component_storage/ComponentStorageBase_TEST.cc
@@ -1,0 +1,80 @@
+/*
+ * Standalone differential check for the legacy_component_storage reference
+ * reconstruction: not wired into any build file, not run by CI. Compares
+ * ComponentStorageRef (vector, the historical behavior) against
+ * ComponentStorageOurs (deque) for (a) identical stored values and ids, and
+ * (b) the "expanded" (reallocation-triggered-rebuild) flag: never true for
+ * ours, true ~E/100 times for the reference, confirming the mechanism the
+ * optimization removes.
+ *
+ * Build and run manually, e.g.:
+ *   c++ -O2 -std=c++17 ComponentStorageBase_TEST.cc -o storage_test && ./storage_test
+ */
+
+#include <cassert>
+#include <cstdio>
+#include "ComponentStorageBase_reference.hh"
+#include "ComponentStorageBase_ours.hh"
+
+using legacy_component_storage::ComponentStorageRef;
+using legacy_component_storage::ComponentStorageOurs;
+
+struct Pose
+{
+  double x, y, z;
+  bool operator==(const Pose &_o) const
+  {
+    return x == _o.x && y == _o.y && z == _o.z;
+  }
+};
+
+int main()
+{
+  const int kNumEntities = 8000;
+
+  ComponentStorageRef<Pose> ref;
+  ComponentStorageOurs<Pose> ours;
+
+  int refExpandedCount = 0;
+  int oursExpandedCount = 0;
+
+  for (int i = 0; i < kNumEntities; ++i)
+  {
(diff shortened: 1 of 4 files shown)


hugogo1998:
Sorry, I reflect on this PR and realize it might not be good to be merged. I'm closing it for now.

PR #3852, opened 2026-07-24
Physics: cache per-parent pose inverse, reuse ChangedLinks scratch map

hugogo1998:
### Description

This PR contains two small reductions in repeated work performed during the physics system's per-step link pose synchronization.

1. `PhysicsPrivate::UpdateSim` computes each non-canonical link's pose relative to its parent model by first taking the inverse of the parent model's world pose (`Pose3d::Inverse()`). Links that share the same parent all recompute the same inverse independently. This caches the inverse once per parent entity and reuses it for all sibling links during that simulation step.

2. `PhysicsPrivate::ChangedLinks()` builds a `std::map<Entity, FrameData3d>` containing every link whose pose changed during the current step, returns it by value, and reconstructs the map again on the next physics update. This changes the map into a persistent member, clears it at the start of each call, and returns it by reference so the same storage is reused across simulation steps.

Neither change alters the algorithm or the output—they simply avoid recomputing or reallocating values that are identical either way.

### Testing

I don't currently have a working `gz-sim` build with `gz-physics`/DART available, so I wasn't able to benchmark these changes directly. Rather than guessing at a speedup, I'm leaving this PR without performance numbers.

I also looked into a third optimization while reviewing this code: avoiding the physics engine's velocity/acceleration computation for links that only require world pose updates. That would require a pose-only query in the `gz-physics` `LinkFrameSemantics` feature. After checking every `physics::Link` call site in this file, I couldn't find such an API, so I left that optimization out rather than proposing a change that depends on functionality that doesn't exist today.

### DIFF
--- a/src/systems/physics/Physics.cc
+++ b/src/systems/physics/Physics.cc
@@ -281,7 +281,10 @@ class gz::sim::systems::PhysicsPrivate
   /// std::map is used because canonical links must be in topological order
   /// to ensure that nested models with multiple canonical links are updated
   /// properly (models must be updated in topological order).
-  public: std::map<Entity, physics::FrameData3d> ChangedLinks(
+  /// Returns a reference to the persistent changedLinkFrameData member
+  /// (cleared, not reallocated, at the top of each call) rather than a
+  /// fresh map by value.
+  public: std::map<Entity, physics::FrameData3d> &ChangedLinks(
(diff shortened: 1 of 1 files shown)


hugogo1998:
Sorry, I need to think more about this PR. I'm closing it for now.

PR #3853, opened 2026-07-24
JointStatePublisher: build joint message once, update fields in place

hugogo1998:
## Summary

Reuse the `JointStatePublisher` protobuf message between simulation updates instead of rebuilding its joint tree on every call to `PostUpdate()`.

The set and order of joints are fixed during configuration, so the message structure and invariant fields only need to be initialized once. Subsequent updates overwrite the timestamp, model pose, and per-joint pose, position, velocity, and force fields before publishing.

This also removes the protobuf arena, since the message and its submessages now persist for the lifetime of the system.

A standalone benchmark measured the following time per update:

| Joints | Current | This change | Speedup |
| ---: | ---: | ---: | ---: |
| 4 | 1.290 µs | 0.472 µs | 2.73x |
| 20 | 5.802 µs | 2.053 µs | 2.80x |
| 40 | 11.478 µs | 4.165 µs | 2.76x |
| 80 | 22.893 µs | 8.189 µs | 2.80x |

Serialized output matched the current implementation across five updates of a 25-joint model.

## Backport Policy

    - [ ] Jetty
    - [ ] Ionic
    - [ ] Harmonic
    - [ ] Fortress
- [x] I am not sure

## Test it

The change was tested with a standalone harness that compares the serialized output and measures message update time.

It has not yet been built or tested in a complete `gz-sim` workspace.

## Checklist

- [ ] Added tests
- [ ] `codecheck` passed (See [[contributing](https://gazebosim.org/docs/all/contributing#contributing-code)](https://gazebosim.org/docs/all/contributing#contributing-code))
- [ ] All tests passed (See [[test coverage](https://gazebosim.org/docs/all/contributing#test-coverage)](https://gazebosim.org/docs/all/contributing#test-coverage))
- [ ] While waiting for a review on your PR, please help review [[another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Agazebosim+archived%3Afalse+)](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Agazebosim+archived%3Afalse+) to support the maintainers
- [x] Was GenAI used to generate this PR? If so, make sure to add `Assisted-by` to your commits. (See [[this policy](https://osralliance.org/wp-content/uploads/2025/05/OSRF-Policy-on-the-Use-of-Generative-Tools-Generative-AI-in-Contributions.pdf)](https://osralliance.org/wp-content/uploads/2025/05/OSRF-Policy-on-the-Use-of-Generative-Tools-Generative-AI-in-Contributions.pdf) for more info.)

### DIFF
--- a/src/systems/joint_state_publisher/JointStatePublisher.cc
+++ b/src/systems/joint_state_publisher/JointStatePublisher.cc
@@ -115,6 +115,14 @@ void JointStatePublisher::Configure(
       std::chrono::duration_cast<std::chrono::steady_clock::duration>(period);
   }
 
+  this->jointStateMsg.set_name(this->model.Name(_ecm));
+  this->jointStateMsg.set_id(this->model.Entity());
+  for (const Entity &joint : this->joints)
+  {
+    msgs::Joint *jointMsg = this->jointStateMsg.add_joint();
+    jointMsg->set_name(_ecm.Component<components::Name>(joint)->Data());
+    jointMsg->set_id(joint);
+  }
 }
 
 //////////////////////////////////////////////////
@@ -208,26 +216,6 @@ void JointStatePublisher::PostUpdate(const UpdateInfo &_info,
   this->lastUpdateTime = _info.simTime;
 
   msgs::Model &msg = this->jointStateMsg;
-
-  // Build the static message shape once (name/id + one Joint per joint,
-  // with name/id). The joint set is fixed after Configure(), so the tree
-  // shape never changes; only the numeric fields below change each step.
-  if (!this->jointStateMsgBuilt)
-  {
-    msg.set_name(this->model.Name(_ecm));
-    msg.set_id(this->model.Entity());
-    this->jointMsgHandles.clear();
-    for (const Entity &joint : this->joints)
-    {
-      msgs::Joint *jointMsg = msg.add_joint();
-      jointMsg->set_name(_ecm.Component<components::Name>(joint)->Data());
-      jointMsg->set_id(joint);
-      this->jointMsgHandles.emplace_back(joint, jointMsg);
-    }
-    this->jointStateMsgBuilt = true;
-  }
-
-  // Per-step: overwrite only the changing numeric fields (no allocation).
   msg.mutable_header()->mutable_stamp()->CopyFrom(
       convert<msgs::Time>(_info.simTime));
 
@@ -240,10 +228,10 @@ void JointStatePublisher::PostUpdate(const UpdateInfo &_info,
   static bool hasWarned {false};
 
   // Process each joint
-  for (auto &handle : this->jointMsgHandles)
+  int jointIndex = 0;
+  for (const Entity &joint : this->joints)
   {
-    const Entity joint = handle.first;
-    msgs::Joint *jointMsg = handle.second;
+    msgs::Joint *jointMsg = msg.mutable_joint(jointIndex++);
 
     // Set the joint pose
     pose = _ecm.Component<components::Pose>(joint);
--- a/src/systems/joint_state_publisher/JointStatePublisher.cc
+++ b/src/systems/joint_state_publisher/JointStatePublisher.cc
@@ -17,8 +17,6 @@
 
 #include "JointStatePublisher.hh"
 
-#include <google/protobuf/arena.h>
-
 #include <gz/msgs/model.pb.h>
 
 #include <chrono>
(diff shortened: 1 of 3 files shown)


hugogo1998:
Hi @arjo129, could you please take a look when you have a chance? I’d especially appreciate your feedback on whether caching the protobuf message and joint handles is a reasonable approach here. Thanks!

arjo129:
@luca-della-vedova is taking the lead on performance optimization. If he has some time I'll ask him to take a look.

hugogo1998:
Thanks, I'd appreciate it if @luca-della-vedova could take a look. Please let me know if I overlook something, and any feedback is appreciated, thanks

luca-della-vedova:
Similar to my comment in https://github.com/gazebosim/gz-sim/pull/3850#issuecomment-5311343282, there should be a way for maintainers to reproduce the performance gains (i.e. we have some joint state publisher sample worlds, can you try to run them and report what is the performance gain for you together with sharing the procedure you used?).
This PR also adds a bit of complexity + some new cached variables and binary flags, so I would like to see significant improvements in an existing world before adding this in.

hugogo1998:
Thanks @luca-della-vedova! I've pushed two follow-up commits: one simplifies the change to address the added-complexity concern, and one adds a reproducible benchmark on the existing worlds that already load `JointStatePublisher`.

### Simplification (d1f935c)
Dropped the `jointStateMsgBuilt` flag and the `jointMsgHandles` vector. Since the joint set is fixed after `Configure()`, the persistent message is now built there and `PostUpdate()` just updates fields in place via `msg.mutable_joint(i)`. The only new member is a single `msgs::Model`:

```diff
-    private: google::protobuf::Arena arena;
+    /// \brief Persistent joint-state message, built once then updated in place.
+    private: msgs::Model jointStateMsg;
```

### Reproducing the benchmark
The two existing example worlds that instantiate `JointStatePublisher` are `examples/worlds/lift_drag.sdf` (3 joints) and `examples/worlds/advanced_lift_drag_system.sdf` (7 joints). I added cases for both to `test/benchmark/server_run.cc` using a dedicated `BM_RuntimeWorldNoThrottle` that disables real-time throttling, so the timed value is the per-step **compute** cost (tens of µs) rather than the wall-clock sleep that matches the target RTF. The existing `BM_RuntimeWorld` cases are untouched, and the publisher runs every step.

Build with benchmarks enabled, then A/B within this branch (this keeps the benchmark identical and only swaps the system under test):

```bash
# "after" (this PR)
taskset -c 0 ./bin/BENCHMARK_server_run \
  --benchmark_filter='existing_.*joint_state_publisher' \
  --benchmark_repetitions=10 --benchmark_report_aggregates_only=true

# "before": revert just the system to main, rebuild, rerun the same command
git checkout main -- src/systems/joint_state_publisher/
#   <rebuild BENCHMARK_server_run>
git checkout HEAD -- src/systems/joint_state_publisher/   # restore afterwards
```

My box is shared, so to be robust against contention I ran *before* and *after* back-to-back for 15 rounds and report the median of the per-round differences (paired), pinned to one core.

### Results — median time per simulation step

| World | Joints | before (main) | after (this PR) | Speedup |
| --- | ---: | ---: | ---: | ---: |
| `lift_drag.sdf` | 3 | ~45.2 µs | ~43.8 µs | **~3%** (−1.5 µs; faster in 14/15 rounds) |
| `advanced_lift_drag_system.sdf` | 7 | ~67.0 µs | ~63.5 µs | **~5%** (−3.5 µs; faster in 15/15 rounds) |

`after` was faster in 29 of 30 paired rounds. The saving scales with the number of published joints (~0.5 µs/joint), matching the per-update micro-benchmark in the description — so a robot publishing many joints benefits more, while a small 3-joint world still sees a measurable but smaller improvement. Happy to add a higher-joint sample world if you'd like to see the trend at 20–40 joints.

Environment: gz-sim 11 / gz-physics 10, bullet-featherstone, Ubuntu Noble.

hugogo1998:
The improvement is 3%-5%, not sure if that's a meaningful improvement for gz-sim. Is the change too large and verbose to be considered? @luca-della-vedova

src/systems/joint_state_publisher/JointStatePublisher.cc
```diff
@@ -209,35 +215,23 @@ void JointStatePublisher::PostUpdate(const UpdateInfo &_info,
...
       this->model.Entity());
   if (pose)
-    msgs::Set(msg->mutable_pose(), pose->Data());
+    msgs::Set(msg.mutable_pose(), pose->Data());
 
   static bool hasWarned {false};
 
   // Process each joint
+  int jointIndex = 0;
   for (const Entity &joint : this->joints)
```
luca-della-vedova: Let's just change this to be an index based for loop instead of adding another variable

luca-della-vedova (commented):
I can reproduce the performance improvements and imho they are small but the code itself is actually slightly simpler (no more arena dependency, manual resets / ifdefs). If we don't count the test it's actually a negative line diff which is nice.
A minor comment for on iteration / manual indices, then I would lean towards actually reverting the test before merging since we would have to change it have a new test world for this but it's minor enough that we probably don't need a benchmark specifically for this.

On a completely separate note, if you have spare cycles feel free to look at `PosePublisher` as well, it has a similar potential bottleneck and protobuf arenas were introduced to optimize it, but I wonder if we can have a similar change to simplify the code + make it faster.

hugogo1998:
Addressed in 1b295a0: switched to an index-based loop and removed the benchmark additions. Thanks!

hugogo1998:
Hi @luca-della-vedova, could you take another look?

luca-della-vedova (approved):

luca-della-vedova:
Bazel CI failure is unrelated (gz-transport)

luca-della-vedova:

Issue #3829, opened 2026-07-16
:farmer: Unit_server_test failing by timeout after 240 secs in gz_sim


miguelgonrod:
## Environment
* OS Version: Ubuntu noble
* Source

## Description
* Expected behavior: To pass without test regressions
* Actual behavior: Failing with test regressions

## Steps to reproduce

1. Go to https://build.osrfoundation.org/job/gz_sim-ci-main-noble-amd64
2. Build a job
3. See the test failing flaky

### Reference Build
- https://build.osrfoundation.org/job/gz_sim-ci-main-noble-amd64/278

### Failing tests
- UNIT_Server_TEST.test_ran

### Log Output
```
39: (2026-07-15 20:12:59.648) [debug] [AirPressureSensor.cc:123] Air pressure for [air_pressure_model::link::air_pressure_sensor] advertised on [world/air_pressure_sensor/model/air_pressure_model/link/link/sensor/air_pressure_sensor/air_pressure]
39: (2026-07-15 20:12:59.649) [info] [SceneBroadcaster.cc:655] Serving scene information on [/world/air_pressure_sensor/scene/info]
39: (2026-07-15 20:12:59.649) [info] [SceneBroadcaster.cc:664] Serving graph information on [/world/air_pressure_sensor/scene/graph]
39: (2026-07-15 20:12:59.649) [info] [SceneBroadcaster.cc:675] Serving full state on [/world/air_pressure_sensor/state]
39: (2026-07-15 20:12:59.649) [info] [SceneBroadcaster.cc:684] Serving full state (async) on [/world/air_pressure_sensor/state_async]
39: (2026-07-15 20:12:59.649) [info] [SceneBroadcaster.cc:692] Publishing scene information on [/world/air_pressure_sensor/scene/info]
39: (2026-07-15 20:12:59.649) [info] [SceneBroadcaster.cc:701] Publishing entity deletions on [/world/air_pressure_sensor/scene/deletion]
39: (2026-07-15 20:12:59.649) [info] [SceneBroadcaster.cc:710] Publishing state changes on [/world/air_pressure_sensor/state]
39: (2026-07-15 20:12:59.649) [info] [SceneBroadcaster.cc:721] Publishing pose messages on [/world/air_pressure_sensor/pose/info]
39: (2026-07-15 20:12:59.649) [info] [SceneBroadcaster.cc:732] Publishing dynamic pose messages on [/world/air_pressure_sensor/dynamic_pose/info]
[...]

PR #3855, opened 2026-07-27
[flaky test] Fix lost stop request when Server teardown races SimulationRunner startup

j-rivero:
# 🦟 Bug fix

Part of gazebosim/gz-sim#3829

## Summary

A stop request emitted while SimulationRunner::Run was still starting up (e.g. Server destroyed right after a non-blocking Run) was overwritten by the unconditional 'running = true' assignment, so the run loop never terminated and ServerPrivate's destructor blocked forever joining the run thread. This is the cause of the long-standing UNIT_Server_TEST timeout flakiness (ServerConfigSensorPlugin and siblings): the test's service request is answered by the transport thread, so the test can reach teardown while the run thread is still in startup.

More info in https://github.com/gazebosim/gz-sim/issues/3829#issuecomment-5072087361

## Fix

Honor the pending stop by making the startup transition atomic: a new runStateMutex guards every write to the running/stopReceived pair, and Run() checks 'stopReceived' and sets 'running' under that lock. Either OnStop() wins and Run() returns without stepping, or Run() wins and the stop's 'running = false' necessarily lands afterwards and sticks. Reads stay lock-free — both flags remain atomics and the run loop conditions are untouched — so there is no per-iteration cost.

Run() now returns false instead of true when it does not step, so an aborted run is reported to the caller rather than looking like a successful zero-iteration run. Document that 'stopReceived' is monotonic and that stopping is therefore permanent: a Server cannot be run again once stopped, which was already relied upon by the SdfErrorBehavior EXIT_IMMEDIATELY path but was never stated in the public API docs.

Add a deterministic regression test (StopBeforeRun) that stops the runner before Run() and asserts it returns failure without stepping.

## Backport Policy

    - [x] Jetty
    - [x] Ionic
    - [x] Harmonic
    - [x] Fortress
- [ ] I am not sure

## Checklist
- [ ] Added tests

Assisted-By: Claude Fable 5, Opus 5

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -1601,6 +1601,7 @@ gz_sim_system_libraries(
         "@com_google_protobuf//:protobuf_lite",
         "@gz-common//profiler",
         "@gz-math",
+        "@gz-msgs",
         "@gz-msgs//:gzmsgs_cc_proto",
         "@gz-plugin//:register",
         "@gz-transport",
--- a/examples/worlds/sensors.sdf
+++ b/examples/worlds/sensors.sdf
@@ -24,6 +24,10 @@
       name="gz::sim::systems::AirPressure">
     </plugin>
     <plugin
+      filename="gz-sim-air-speed-system"
+      name="gz::sim::systems::AirSpeed">
+    </plugin>
+    <plugin
       filename="gz-sim-altimeter-system"
       name="gz::sim::systems::Altimeter">
     </plugin>
@@ -157,7 +161,7 @@
           <always_on>1</always_on>
           <update_rate>30</update_rate>
           <visualize>true</visualize>
-          <topic>air_pressure</topic>
+          <topic>air_speed</topic>
           <enable_metrics>true</enable_metrics>
           <air_speed>
             <pressure>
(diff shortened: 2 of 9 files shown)


j-rivero:
Test failure is probably unrelated to this PR, it is failing in previous build of the CI job.

caguero (commented):
I made some suggestions for you to consider in caguero/fix_3829_combined that can address this PR and #3858 directly.

j-rivero:
> I made some suggestions for you to consider in caguero/fix_3829_combined that can address this PR and #3858 directly.

I have some doubts that the branch can cover all the cases reported in both of my PRs, particularly the #3858 .  Please check the test `RunNonBlockingShortMultiple` pinned to a single core.

caguero:
> > I made some suggestions for you to consider in caguero/fix_3829_combined that can address this PR and #3858 directly.
> 
> I have some doubts that the branch can cover all the cases reported in both of my PRs, particularly the #3858 . Please check the test `RunNonBlockingShortMultiple` pinned to a single core.

What's your concern? As a sanity check I ran `RunNonBlockingShortMultiple` pinned to one CPU as a sanity check and I got:

```
base    pass (100 iterations)
    #3855   hang
    #3858   pass
    #3931   pass, and 5 repetitions (500 iterations) also pass
```

j-rivero:
> What's your concern?

No more concerns at this point, go ahead and reopen #3931 so we can comment there

PR #3863, opened 2026-08-03
fix: skip light visual lookup when Sensors path has no mapping (#3862)

### LINKED ISSUE #3862: Spurious "Could not find visual for entity: 0" error when a light is reconfigured while a rendering sensor is active
### Environment

- **OS**: Ubuntu 24.04 (noble)
- **Gazebo Sim**: 8.11.0 (`gz-sim-vendor` for ROS 2 Jazzy); the same code is in 8.14.0 and unchanged on `main`
- **Render engine**: ogre2, headless server (`gz sim -s`)
- **Install**: binaries

### Description

Whenever a light is reconfigured through `components::LightCmd` — via the `/world/<world>/light_config` service, or by any system plugin that sets the component — the server prints, once per light each time its `LightCmd` is processed:

```
[Err] [SceneManager.cc:615] Could not find visual for entity: 0
```

Functionally nothing is wrong: the light is recolored correctly and sensor data is unaffected. But it is logged at `Err` level, and a plugin that animates lights at e.g. 10 Hz makes the server log unusable.

What makes this confusing to diagnose is that the messages **only start appearing once something subscribes to a rendering sensor topic**, so it reads like a sensor/mesh problem. It isn't: the `Sensors` system only notifies its render thread while a rendering sensor `HasConnections()`, and `RenderUtil::Update()` — where the offending code lives — therefore only runs then. Any `entityLights` entries queued before that are flushed the moment the first subscriber connects.

### Cause

`RenderUtilPrivate::UpdateLights()` resolves the light's *visual* with `std::map::operator[]`:

https://github.com/gazebosim/gz-sim/blob/32ac5467fa2bbb49c95eac468c7c61ae7abfed16/src/rendering/RenderUtil.cc#L3012-L3016

```cpp
rendering::VisualPtr lightVisual =
    this->sceneManager.VisualById(
      this->matchLightWithVisuals[light.first]);
```

`matchLightWithVisuals` is only ever populated when `enableSensors == false`:

https://github.com/gazebosim/gz-sim/blob/32ac5467fa2bbb49c95eac468c7c61ae7abfed16/src/rendering/RenderUtil.cc#L1271-L1280

The `Sensors` system calls `SetEnableSensors(true, ...)`:

https://github.com/gazebosim/gz-sim/blob/32ac5467fa2bbb49c95eac468c7c61ae7abfed16/src/systems/sensors/Sensors.cc#L808

so in the server-side `RenderUtil` that map is always empty. `operator[]` then default-inserts `kNullEntity` (`0`), and `SceneManager::VisualById(0)` logs the error and returns `nullptr`. The following `if (lightVisual)` check means the rest of the function behaves correctly — only the error log is spurious.

### Steps to reproduce

Stock world, no extra plugins.

**Terminal 1** — headless server:

```bash
gz sim -v 4 -s -r gpu_lidar_sensor.sdf
```

**Terminal 2** — reconfigure the light while nothing subscribes to the lidar. No error is printed:

```bash
gz service -s /world/gpu_lidar_sensor/light_config \
  --reqtype gz.msgs.Light --reptype gz.msgs.Boolean --timeout 2000 \
  --req 'name: "sun", type: DIRECTIONAL, diffuse: {r: 1, g: 0, b: 0, a: 1}'
```

**Terminal 3** — subscribe to the lidar:

```bash
gz topic -e -t /lidar
```

Terminal 1 immediately prints `Could not find visual for entity: 0`, and every further `light_config` call adds exactly one more.

Measured here: 5 `light_config` calls with no subscriber → 0 errors; subscribing to `/lidar` → 1 error (the queued update being flushed); 5 more calls while subscribed → exactly 5 more errors.

### Suggested fix

Skip the visual update when no light visual is known, instead of looking up `kNullEntity`:

```cpp
auto visualIt = this->matchLightWithVisuals.find(light.first);
if (visualIt != this->matchLightWithVisuals.end())
{
  rendering::VisualPtr lightVisual =
      this->sceneManager.VisualById(visualIt->second);
  if (lightVisual)
    lightVisual->SetVisible(light.second.visualize_visual());
}
```

Note that `RemoveEntities` already uses `matchLightWithVisuals[_entity]` the same way at [RenderUtil.cc#L2484](https://github.com/gazebosim/gz-sim/blob/32ac5467fa2bbb49c95eac468c7c61ae7abfed16/src/rendering/RenderUtil.cc#L2484), which would schedule entity `0` for removal in the sensors-only render util — probably worth guarding too.

Happy to open a PR if this direction looks right.

trippyogi:
# Bug fix

Fixes #3862

## Summary
Sensors-enabled `RenderUtil` never populates `matchLightWithVisuals`, but `UpdateLights` used `operator[]`, which inserted entity `0` and made `SceneManager::VisualById(0)` log a spurious Err. Light updates still worked.

Use `find()` in `UpdateLights` and `RemoveRenderingEntities` so missing mappings are skipped.

## Backport Policy
- [x] I am not sure

## Checklist
- [ ] Added tests
- [ ] `codecheck` / full suite (not run locally on Windows)
- [x] Was GenAI used to generate this PR? If so, make sure to add Assisted-by to your commits.

Assisted-by: Cursor (Composer)

**Note to maintainers**: Squash-Merge and retain `Signed-off-by` / `Assisted-by`.

### DIFF
--- a/src/rendering/RenderUtil.cc
+++ b/src/rendering/RenderUtil.cc
@@ -2481,9 +2481,12 @@ void RenderUtilPrivate::RemoveRenderingEntities(
       [&](const Entity &_entity, const components::Light *)->bool
       {
         this->removeEntities[_entity] = _info.iterations;
-        this->removeEntities[matchLightWithVisuals[_entity]] =
-          _info.iterations;
-        matchLightWithVisuals.erase(_entity);
+        auto visualIt = this->matchLightWithVisuals.find(_entity);
+        if (visualIt != this->matchLightWithVisuals.end())
+        {
+          this->removeEntities[visualIt->second] = _info.iterations;
+          this->matchLightWithVisuals.erase(visualIt);
+        }
         return true;
       });
 
@@ -3009,11 +3012,15 @@ void RenderUtilPrivate::UpdateLights(
     auto l = std::dynamic_pointer_cast<rendering::Light>(node);
     if (l)
     {
-      rendering::VisualPtr lightVisual =
-          this->sceneManager.VisualById(
-            this->matchLightWithVisuals[light.first]);
-      if (lightVisual)
-        lightVisual->SetVisible(light.second.visualize_visual());
+      // Light visuals exist only when enableSensors is false (see Create path).
+      auto visualIt = this->matchLightWithVisuals.find(light.first);
+      if (visualIt != this->matchLightWithVisuals.end())
+      {
+        rendering::VisualPtr lightVisual =
+            this->sceneManager.VisualById(visualIt->second);
+        if (lightVisual)
+          lightVisual->SetVisible(light.second.visualize_visual());
+      }
 
       if (!light.second.is_light_off())
       {


iche033 (approved):
thanks for your contribution. Changes look good

iche033:

PR #3871, opened 2026-08-08
Support entity_namespace in user commands entity creation

C88-YQ:
# 🎉 New feature

## Summary

This PR adds new services for creating entities with a namespace:
  - `/world/<world>/create_with_ns`
  - `/world/<world>/create_with_ns_multiple`

The new services use `EntityFactoryWithNs` messages. When a namespace is provided, it is applied to the created model. When no namespace is provided, the model keeps the namespace from the SDF if one exists.

This also updates entity cloning so cloned entities can either:
  - use a namespace passed by the caller
  - keep the original entity namespace when no new namespace is passed

Tests were updated to cover entity creation with and without namespaces, namespace overrides, and namespace handling during cloning.

## Backport Policy

    - [x] Jetty
    - [x] Ionic
    - [x] Harmonic
    - [x] Fortress
- [ ] I am not sure

## Test it
 Build the project, then run the related tests:

```bash
ctest -R "EntityComponentManager|user_commands"
 ```

## Checklist
- [x] Added tests

### DIFF
--- a/include/gz/sim/EntityComponentManager.hh
+++ b/include/gz/sim/EntityComponentManager.hh
@@ -31,6 +31,7 @@
 #include <unordered_map>
 #include <utility>
 #include <vector>
+#include <optional>
 
 #include <gz/common/Console.hh>
 #include <gz/math/graph/Graph.hh>
@@ -111,6 +112,7 @@ namespace gz
       /// \param[in] _name The name that should be given to the cloned entity.
       /// Set this to an empty string if the cloned entity name should be
       /// auto-generated to something unique.
+      /// \param[in] _ns The namespace that should be given to the cloned entity.
       /// \param[in] _allowRename True if _name can be modified to be a unique
       /// name if it isn't already a unique name. False if _name cannot be
       /// modified to be a unique name. If _allowRename is set to False, and
@@ -123,7 +125,9 @@ namespace gz
       /// cloned.
       /// \sa Clone
       public: Entity Clone(Entity _entity, Entity _parent,
-                  const std::string &_name, bool _allowRename);
+                  const std::string &_name,
+                  const std::optional<std::string> &_ns,
+                  bool _allowRename);
 
       /// \brief Get the number of entities on the server.
       /// \return Entity count.
@@ -377,13 +381,16 @@ namespace gz
       /// \param[in] _entity The entity to clone.
       /// \param[in] _parent The parent of the cloned entity.
       /// \param[in] _name The name that should be given to the cloned entity.
+      /// \param[in] _ns The namespace that should be given to the cloned entity.
       /// \param[in] _allowRename True if _name can be modified to be a unique
       /// name if it isn't already a unique name. False if _name cannot be
       /// modified to be a unique name.
       /// \return The cloned entity. kNullEntity is returned if cloning failed.
       /// \sa Clone
       private: Entity CloneImpl(Entity _entity, Entity _parent,
-                  const std::string &_name, bool _allowRename);
+                  const std::string &_name,
+                  const std::optional<std::string> &_ns,
+                  bool _allowRename);
 
       /// \brief A version of Each() that doesn't use a cache. The cached
       /// version, Each(), is preferred.
--- a/include/gz/sim/EntityComponentManager.hh
+++ b/include/gz/sim/EntityComponentManager.hh
@@ -31,7 +31,6 @@
 #include <unordered_map>
 #include <utility>
 #include <vector>
-#include <optional>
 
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Comment by C88-YQ** [comment]
`EntityFactory` already added `entity_namespace` in https://github.com/gazebosim/gz-msgs/pull/612, but that commit cannot be backported to older branches.

gazebosim/gz-msgs#595 will be backported to older branches, and `EntityFactoryWithNs` is planned to be deprecated in gazebosim/gz-msgs#616

To make this PR easier to backport, this PR does not handle `entity_namespace` in `EntityFactory` yet. For now, namespace support is only available through `EntityFactoryWithNs`.

I will open a follow-up PR to add `entity_namespace` support for `EntityFactory`, and also add the deprecation warning for `EntityFactoryWithNs`. That follow-up PR will not be backported.

File: include/gz/sim/EntityComponentManager.hh [hunk 1]
```diff
@@ -123,7 +125,8 @@ namespace gz
       /// cloned.
       /// \sa Clone
       public: Entity Clone(Entity _entity, Entity _parent,
-                  const std::string &_name, bool _allowRename);
+                  const std::string &_name, const std::string &_ns,
+                  bool _allowRename);
```
luca-della-vedova: I believe this would be an immediate API breakage for downstream users that are using the `Clone` function.
We try to "tick-tock" APIs, or at least not break them completely, I can think of two ways to offer a cleaner migration path:

* Add the namespace as the last parameter with a default value, code that doesn't pass it as a parameter would still compile and use the default value (perhaps no namespace?).
* Create a new `Clone` function with the namespace with a mandatory namespace parameter and mark the old one as deprecated. The old one would call the new `Clone` with a reasonable default for the namespace.

I'm fairly neutral on which way to go

src/EntityComponentManager.cc
```diff
@@ -544,13 +549,25 @@ Entity EntityComponentManager::CloneImpl(Entity _entity, Entity _parent,
   }
   this->CreateComponent(clonedEntity, components::Name(clonedName));
 
+  auto originalNsComp = this->Component<components::Namespace>(_entity);
+  if (!_ns.empty())
```
luca-della-vedova, line 551: What if, instead of using an empty string as a "no namespace was requested" we have namespace be a `std::optional<std::string>` so if users specify a `std::nullopt` value we know that they don't want a namespace to be applied?

src/EntityComponentManager.cc
```diff
@@ -544,13 +549,25 @@ Entity EntityComponentManager::CloneImpl(Entity _entity, Entity _parent,
...
+  {
+    this->CreateComponent(clonedEntity, components::Namespace(_ns));
+  }
+  else if (nullptr != originalNsComp && !originalNsComp->Data().empty())
```
luca-della-vedova: A bit conflicted here, from what I saw in the previous PR if the namespace is empty we don't add a component in the first place. It's probably an extreme case but I wonder if the behavior here should be of copying the original namespace component regardless of its value

src/EntityComponentManager.cc
C88-YQ: @luca-della-vedova Thanks for the suggestion! We actually discussed this quite a few times, and we implemented both approaches at different points.

In the end, we decided not to distinguish between “not set” and “set to an empty string”. The main reason is that we want the behavior to be consistent between `gz-sim` and `ros_gz`.

On the `ros_gz` side, entity_namespace in [simulation_interfaces/srv/SpawnEntity.srv](https://github.com/ros-simulation/simulation_interfaces/blob/main/srv/SpawnEntity.srv) is a string, so we cannot tell whether the user did not set it or explicitly set it to an empty string. Changing this interface would be a breaking change and could affect many downstream users, so we decided to keep the current behavior.

Also, for `UserCommands`, if a user wants to clear an existing namespace when spawning an entity, they can use "/" instead of an empty string. Since `SpawnEntity` can only create entities directly under the world, the final behavior is the same.

For other cases, “not set” and “set to empty” currently have the same behavior, so we decided not to distinguish between them.

include/gz/sim/EntityComponentManager.hh
C88-YQ: Thanks a lot for the suggestion! That makes sense to me. I went with the first approach and added the namespace as the last parameter with a default value.

src/EntityComponentManager.cc
C88-YQ: That makes sense. The check here was indeed a bit redundant, so I’ve removed it. Thanks for pointing this out!

PR #3873, opened 2026-08-10
Report protobuf serialization errors

Daniiiil1:
## Summary

- propagate protobuf serialization and deserialization failures through the stream state
- parse generic protobuf messages into a temporary object before updating component data
- cover message, vector, converted-component, semantic-description, and semantic-tags serializers

## Root cause

The protobuf APIs return `false` on failure, but the serializers discarded those results. Callers therefore could not observe malformed input or failed output, and the generic message serializer could leave component data partially updated after a failed parse.

## Testing

- `git diff --check`
- `cppcheck --enable=warning,performance,portability --std=c++17 --language=c++ --suppress=missingIncludeSystem --suppress=missingInclude -Iinclude ...`
- `bazel test //:Component_TEST` was attempted locally, but the macOS Bazel toolchain failed in the unchanged external `tinyxml2` dependency because it could not locate `<cctype>`, before this target compiled

Fixes #3385

### DIFF
--- a/include/gz/sim/components/SemanticDescription.hh
+++ b/include/gz/sim/components/SemanticDescription.hh
@@ -44,8 +44,10 @@ class SemanticDescriptionSerializer
   {
     msgs::StringMsg msg;
     msg.set_data(_desc);
-    auto result = msg.SerializeToOstream(&_out);
-    (void)result;
+    if (!msg.SerializeToOstream(&_out))
+    {
+      _out.setstate(std::ios_base::failbit);
+    }
     return _out;
   }
 
@@ -53,10 +55,13 @@ class SemanticDescriptionSerializer
   static std::istream &Deserialize(std::istream &_in, std::string &_desc)
   {
     msgs::StringMsg msg;
-    if(msg.ParsePartialFromIstream(&_in))
+    if (!msg.ParsePartialFromIstream(&_in))
     {
-      _desc = msg.data();
+      _in.setstate(std::ios_base::failbit);
+      return _in;
     }
+
+    _desc = msg.data();
     return _in;
   }
 };
--- a/include/gz/sim/components/Serialization.hh
+++ b/include/gz/sim/components/Serialization.hh
@@ -22,7 +22,7 @@
 #pragma warning(disable: 4251)
 #endif
 
-#include <google/protobuf/message_lite.h>
+#include <google/protobuf/message.h>
 
 #ifdef _MSC_VER
 #pragma warning(pop)
@@ -30,6 +30,8 @@
 
 #include <gz/msgs/double_v.pb.h>
 
+#include <ios>
+#include <memory>
 #include <string>
 #include <vector>
 #include <sdf/Sensor.hh>
@@ -125,8 +127,10 @@ namespace serializers
         msg = gz::msgs::Convert(_data);
       }
 
-      auto result = msg.SerializeToOstream(&_out);
-      (void)result;
+      if (!msg.SerializeToOstream(&_out))
+      {
+        _out.setstate(std::ios_base::failbit);
+      }
       return _out;
     }
 
@@ -140,6 +144,7 @@ namespace serializers
       MsgType msg;
       if (!msg.ParseFromIstream(&_in))
       {
+        _in.setstate(std::ios_base::failbit);
         return _in;
       }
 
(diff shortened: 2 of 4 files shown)


include/gz/sim/components/SemanticDescription.hh
```diff
@@ -44,19 +44,24 @@ class SemanticDescriptionSerializer
   {
     msgs::StringMsg msg;
     msg.set_data(_desc);
-    auto result = msg.SerializeToOstream(&_out);
-    (void)result;
+    if (!msg.SerializeToOstream(&_out))
```
caguero, line 47: It looks like this component is using the `StringSerializer`, not the custom `SemanticDescriptionSerializer`. I think the right call is to remove the entire `SemanticDescriptionSerializer` class.

caguero (commented):
Thanks for the patch! I left a few comments.

include/gz/sim/components/Serialization.hh
```diff
@@ -140,6 +144,7 @@ namespace serializers
       MsgType msg;
       if (!msg.ParseFromIstream(&_in))
       {
+        _in.setstate(std::ios_base::failbit);
```
caguero, line 147: This new failbit is never checked by the callers. In `EntityComponentManager` the stream is a local variable and nothing looks at it after `Deserialize`, so the error would go unnoticed. It would be good to check the stream there and `gzerr` an error.

include/gz/sim/components/Serialization.hh
```diff
@@ -203,8 +211,10 @@ namespace serializers
     public: static std::ostream &Serialize(std::ostream &_out,
         const google::protobuf::Message &_msg)
     {
-      auto result = _msg.SerializeToOstream(&_out);
-      (void)result;
+      if (!_msg.SerializeToOstream(&_out))
```
caguero, line 214: Same problem on the write side. The callers in `EntityComponentManager` do not check the stream and still publish the bytes from `ostr.str()`, so a broken payload still goes out. Could we check the stream before using the result.

include/gz/sim/components/Serialization.hh
```diff
@@ -215,8 +225,16 @@ namespace serializers
     public: static std::istream &Deserialize(std::istream &_in,
         google::protobuf::Message &_msg)
     {
-      auto result = _msg.ParseFromIstream(&_in);
-      (void)result;
+      // Parse into a temporary message so malformed data can't leave the
+      // component partially updated.
+      std::unique_ptr<google::protobuf::Message> parsed(_msg.New());
```
caguero, line 230: This adds a heap allocation and a full message copy on every call. This code runs a lot during state sync, so large worlds will pay for it. Could we keep the old parse in place here?

include/gz/sim/components/Serialization.hh
```diff
@@ -203,8 +211,10 @@ namespace serializers
     public: static std::ostream &Serialize(std::ostream &_out,
         const google::protobuf::Message &_msg)
     {
-      auto result = _msg.SerializeToOstream(&_out);
-      (void)result;
+      if (!_msg.SerializeToOstream(&_out))
+      {
+        _out.setstate(std::ios_base::failbit);
```
caguero, line 216: I think this could throw an exception now.

include/gz/sim/components/Serialization.hh
```diff
@@ -125,8 +127,10 @@ namespace serializers
         msg = gz::msgs::Convert(_data);
       }
 
-      auto result = msg.SerializeToOstream(&_out);
-      (void)result;
+      if (!msg.SerializeToOstream(&_out))
+      {
+        _out.setstate(std::ios_base::failbit);
```
caguero, line 132: I believe failbit is sticky. After one failed write, all later writes to the same stream are dropped until someone calls `clear()`. Before this change, later writes still worked.

src/Component_TEST.cc
```diff
@@ -532,6 +552,93 @@ TEST_F(ComponentTest, IStream)
...
+  }
+
+  // Serialization failures should be observable through the output stream.
+  auto expectSerializationFailure = [](const auto &_component)
+  {
+    FailingOutputBuffer buffer;
+    std::ostream out(&buffer);
+    EXPECT_TRUE(out.good());
+    _component.Serialize(out);
+    EXPECT_TRUE(out.fail());
```
caguero, line 617: These checks also pass without the fix. The failing buffer already sets badbit, so `out.fail()` returns true on main too. Checking the failbit directly would make the test more robust: 
```
EXPECT_TRUE(out.rdstate() & std::ios_base::failbit)
```

PR #3883, opened 2026-08-15
TrackedVehicle: read track_height value from sdf

ntfshard:
# 🦟 Bug fix

Fixes N/A and #2309

## Summary
This change should fix missing read from sdf for `track_height` value. Also in examples and tests was plural, also fixed (expecting doxygen should have higher priority in this ambiguous case). Also fixed documentation for body_link value

## Backport Policy

- [x] This is safe to backport to the following versions:
    - [x] Jetty
    - [x] Ionic
    - [ ] Harmonic
    - [x] Fortress
- [ ] This **should not** be backported
- [ ] I am not sure
- [ ] Other (fill in yourself)

## Checklist
- [x] Signed all commits for DCO
- [ ] Added a screen capture or video to the PR description that demonstrates the fix (as needed)
- [ ] Added tests
- [ ] Updated documentation (as needed)
- [ ] Updated migration guide (as needed)
- [ ] Consider updating Python bindings (if the library has them)
- [ ] `codecheck` passed (See [contributing](https://gazebosim.org/docs/all/contributing#contributing-code))
- [ ] All tests passed (See [test coverage](https://gazebosim.org/docs/all/contributing#test-coverage))
- [ ] Updated Bazel files (if adding new files). Created an issue otherwise.
- [ ] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Agazebosim+archived%3Afalse+) to support the maintainers
- [ ] Was GenAI used to generate this PR? If so, make sure to add "Assisted-by" to your commits. (See [this policy](https://osralliance.org/wp-content/uploads/2025/05/OSRF-Policy-on-the-Use-of-Generative-Tools-Generative-AI-in-Contributions.pdf) for more info.)

**Note to maintainers**: Remember to use **Squash-Merge** and edit the commit message to match the pull request summary while retaining `Signed-off-by` and `Generated-by` messages.

**Backports:** If this is a backport, please use **Rebase and Merge** instead.

Issue #2309, opened 2024-02-08
TrackedVehicle system, the body_link element is not in the api documenation


rodericktaylor:
## Environment
* OS Version: All
* Source or binary build?
  source build of branch gz-sim8.

## Description

The body_link element is not documented in the api documentation for src/systems/tracked_vehicle/TrackedVehicle.hh.

azeey:
Thanks for reporting the issue. Would you be able to open a PR adding the documentation?

PR #3888, opened 2026-08-17
TrackedVehicle: read track_height from sdf

ntfshard:
# 🦟 Bug fix

Fixes N/A and #2309

## Summary
(re-target to main of https://github.com/gazebosim/gz-sim/pull/3883)

This change should fix missing read from sdf for `track_heigh`t value. Also fixed in examples and tests field was plural (expecting doxygen should have higher priority in this ambiguous case). Also fixed documentation for `body_link` value

## Backport Policy

    - [x] Jetty
    - [x] Ionic
    - [x] Harmonic
    - [x] Fortress
- [ ] I am not sure

## Checklist
- [ ] Added tests

### DIFF
--- a/src/systems/tracked_vehicle/TrackedVehicle.cc
+++ b/src/systems/tracked_vehicle/TrackedVehicle.cc
@@ -275,6 +275,8 @@ void TrackedVehicle::Configure(const Entity &_entity,
       this->dataPtr->tracksSeparation).first;
   this->dataPtr->steeringEfficiency = _sdf->Get<double>("steering_efficiency",
       this->dataPtr->steeringEfficiency).first;
+  this->dataPtr->trackHeight = _sdf->Get<double>("track_height",
+      this->dataPtr->trackHeight).first;
 
   // Instantiate the speed limiters.
   this->dataPtr->limiterLin = std::make_unique<math::SpeedLimiter>();
--- a/examples/worlds/tracked_vehicle_simple.sdf
+++ b/examples/worlds/tracked_vehicle_simple.sdf
@@ -1075,7 +1075,7 @@
                 <right_track><link>front_right_flipper</link></right_track>
                 <right_track><link>rear_right_flipper</link></right_track>
                 <tracks_separation>0.4</tracks_separation>
-                <tracks_height>0.18094</tracks_height>
+                <track_height>0.18094</track_height>
                 <steering_efficiency>0.5</steering_efficiency>
                 <!--debug>1</debug-->
             </plugin>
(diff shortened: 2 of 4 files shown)


src/systems/tracked_vehicle/TrackedVehicle.cc
```diff
@@ -275,6 +275,8 @@ void TrackedVehicle::Configure(const Entity &_entity,
       this->dataPtr->tracksSeparation).first;
   this->dataPtr->steeringEfficiency = _sdf->Get<double>("steering_efficiency",
       this->dataPtr->steeringEfficiency).first;
+  this->dataPtr->trackHeight = _sdf->Get<double>("track_height",
+      this->dataPtr->trackHeight).first;
```
luca-della-vedova, line 279: Minor (but not so minor), should we instead use `tracks_height` so we don't need to change the world file? If we want to be fully consistent we can change the name of the data pointer variable from `trackHeight` to `tracksHeight`.

Two main reasons for this suggestion:

1) If there are existing SDFs out there that were based on the example they will just see the fix immediately, without having to change their SDF.
2) It is also consistent with the other SDF parameter `tracks_separation`.

This PR as is makes the two parameters inconsistent at the SDF level (one is singular, one is plural) and existing worlds out there will still have the issue.

ntfshard: As I mentioned in a description, I assume doxygen is a documentation and it has higher priority than examples. Link to documentation (other versions are also available): https://gazebosim.org/api/gazebo/6/classignition_1_1gazebo_1_1systems_1_1TrackedVehicle.html

About fix to other, there is a thing: if anyone tried this plugin and realized its not working well, why a ticket was not submitted. Personally I found this when tried figure why and how this parameter is used.
And if nobody use it, they do not need this fix. And other thing, if they use and do not know that it's a bug inside, this fix will surprise them.

Also I'm not sure why track(s) should be plural in context of height.
For distance between a pair of tracks it's understandable. But here we have height of a track, not stack of them.

arjo129: Traditionally, to avoid breaking existing SDF worlds that copied the original example while keeping Doxygen and semantic naming intact (track_height is grammatically correct for a single track's profile), we'd support both.

We could parse track_height as the primary parameter, check for tracks_height as a fallback with a deprecation warning (ignwarn/gzmsg), and update the example SDF and Doxygen to reference track_height uniformly.

That being said this is a lot of churn for relatively low value. If @ntfshard is against it we can just go with what he has done and support only `track_height`. In any case this will be a behaviour change if people had configured a `track_height`/`track_heights`.

j-rivero:
> This change should fix missing read from sdf for `track_heigh`t value. Also fixed in examples and tests field was plural (expecting doxygen should have higher priority in this ambiguous case). Also fixed documentation for `body_link` value

Could you please demonstrate with output data how a change in `track_height` affects the Odometry generated for a given case?

PR #3878, opened 2026-08-11
Add namespace support to DiffDrive

C88-YQ:
# 🎉 New feature

## Summary
This PR adds namespace support to the DiffDrive plugin.

**Changes include:**
  - Update DiffDrive to resolve `cmd_vel`, `enable`, `odometry`, and `tf` topics using the model namespace when available.

**Expected behavior:**
  - If a model has a namespace, DiffDrive uses the namespace as the topic prefix instead of `/model/{model_name}`.
  - Relative custom topics such as `cmd_vel` are resolved under the namespace.
  - Absolute custom topics such as `/cmd_vel` are kept unchanged.
  - Existing behavior without a namespace falls back to the model-scoped default topics.

## Backport Policy
- [ ] This is safe to backport to the following versions:
    - [x] Jetty
    - [x] Ionic
    - [x] Harmonic
    - [ ] Fortress
- [ ] This **should not** be backported
- [ ] I am not sure
- [ ] Other (fill in yourself)

## Test it
1. Build
2. Launch the following SDF file with:

   <details>
     <summary>world.sdf</summary>

     ``` xml
     <?xml version="1.0" ?>
   
     <sdf version="1.12">
       <world name="test_world">
   
         <plugin
           filename="gz-sim-imu-system"
           name="gz::sim::systems::Imu">
         </plugin>
         <plugin
           filename="gz-sim-sensors-system"
           name="gz::sim::systems::Sensors">
           <render_engine>ogre2</render_engine>
         </plugin>
         <plugin
           filename="gz-sim-scene-broadcaster-system"
           name="gz::sim::systems::SceneBroadcaster">
         </plugin>
   
         <light name="sun" type="directional">
           <cast_shadows>true</cast_shadows>
           <pose>0 0 10 0 0 0</pose>
           <diffuse>0.8 0.8 0.8 1</diffuse>
           <specular>0.2 0.2 0.2 1</specular>
           <attenuation>
             <range>1000</range>
             <constant>0.9</constant>
             <linear>0.01</linear>
             <quadratic>0.001</quadratic>
           </attenuation>
           <direction>-0.5 0.1 -0.9</direction>
         </light>
   
         <model name="ground_plane">
           <static>true</static>
           <link name="link">
             <collision name="collision">
               <geometry>
                 <plane>
                   <normal>0 0 1</normal>
                   <size>100 100</size>
                 </plane>
               </geometry>
             </collision>
             <visual name="visual">
               <geometry>
                 <plane>
[...]

PR #3897, opened 2026-08-19
Add helper functions for namespaced topic resolution and unit test setup

C88-YQ:
# 🎉 New feature

Related to #3878

## Summary

This PR mainly adds helper functions for topic namespace support and for related tests.

- `resolvedTopicName`
  - Resolves topic names from SDF plugin elements.
  - Uses a non-empty custom topic from SDF when available.
  - Prefixes relative custom topics with the provided namespace.
  - Leaves absolute custom topics unchanged.
  - Falls back to the default topic when the SDF element is missing or empty.
  - Adds tests coverage for relative, absolute, empty, and missing topic elements.

- `UnitTestUtil`
  - Adds reusable helpers for setting up unit test contexts.
  - Supports loading a world from an SDF string and creating its entity.
  - Supports loading a model from an SDF string and creating its entity.
  - Optionally extracts the first plugin SDF for test use.

## Backport Policy
    - [x] Jetty
    - [x] Ionic
    - [x] Harmonic
    - [ ] Fortress
- [ ] I am not sure

## Test it
```
cd build/gz-sim
ctest
```

## Checklist
- [x] Added tests
- [x] Updated Bazel files (if adding new files). Created an issue otherwise.


luca-della-vedova (commented):
The Bazel CI failure looks real:

```
test/helpers/UnitTestUtil.hh:24:10: error: module //test:Helpers does not depend on a module exporting 'sdf/Root.hh'
   24 | #include <sdf/Root.hh>
      |          ^
test/helpers/UnitTestUtil.hh:25:10: error: module //test:Helpers does not depend on a module exporting 'sdf/World.hh'
   25 | #include <sdf/World.hh>
      |          ^
test/helpers/UnitTestUtil.hh:26:10: error: module //test:Helpers does not depend on a module exporting 'sdf/Model.hh'
   26 | #include <sdf/Model.hh>
      |          ^
test/helpers/UnitTestUtil.hh:27:10: error: module //test:Helpers does not depend on a module exporting 'sdf/Plugin.hh'
   27 | #include <sdf/Plugin.hh>
      |          ^
```

I think you might need to add a dependency to sdformat for the bazel test targets

C88-YQ:
@luca-della-vedova 

Sorry for the late reply.

The original Bazel layering error is fixed by adding `@sdformat` to `//test:Helpers`.

The current CI failure is different and comes from `gz-transport` / `zenoh-cpp`, where `std::optional<zenoh::CancellationToken>` fails to compile with clang/libstdc++. So it appears unrelated to this PR.

include/gz/sim/Util.hh
```diff
@@ -78,6 +78,23 @@ namespace gz
...
+    /// topic is available, the default topic is used.
+    /// \param[in] _sdf SDF to read the topic name from.
+    /// \param[in] _sdfElement Name of the SDF child element containing the
+    /// topic name.
+    /// \param[in] _ns Namespace to prepend to relative topic name.
+    /// \param[in] _defaultTopic Topic to use when no custom topic is specified.
+    /// \return A valid Gazebo Transport topic name, or an empty string if no
+    /// valid topic could be generated.
+    std::string GZ_SIM_VISIBLE resolvedTopicName(
+        const std::shared_ptr<const sdf::Element> &_sdf,
```
azeey, line 94: Can you forward declare `sdf::Element` since `sdf/Element.hh` is not directly included here.

src/Util.cc
```diff
@@ -282,6 +282,32 @@ std::string scopedNamespace(const EntityComponentManager &_ecm,
   return result;
 }
 
+std::string resolvedTopicName(const std::shared_ptr<const sdf::Element> &_sdf,
```
azeey, line 285: Include `sdf/Element.hh`.

include/gz/sim/Util.hh
```diff
@@ -78,6 +78,23 @@ namespace gz
...
+        const std::string &_sdfElement, const std::string &_ns,
```
azeey, line 95:
```suggestion
        const std::string &_sdfElementName, const std::string &_namespace,
```

src/Util.cc
```diff
@@ -282,6 +282,32 @@ std::string scopedNamespace(const EntityComponentManager &_ecm,
...
+  {
+    std::string customTopic = _sdf->Get<std::string>(_sdfElement);
+
+    if (!customTopic.empty())
+    {
+      // Only prepend namespace to relative topic names.
+      // Absolute topic names (starting with '/') are left unchanged.
+      if (customTopic.front() != '/')
+      {
+        customTopic = _ns + "/" + customTopic;
```
azeey, line 301: Is it a problem if `_ns` is `/`?

src/Util_TEST.cc
```diff
@@ -308,6 +308,40 @@ TEST_F(UtilTest, ScopedNamespace)
...
+  EXPECT_EQ("/ns/test_relative_topic",
+    resolvedTopicName(sdf, "relative_topic", "/ns", "/default_relative_topic"));
+  EXPECT_EQ("/test_absolute_topic",
+    resolvedTopicName(sdf, "absolute_topic", "", "/default_absolute_topic"));
+  EXPECT_EQ("/test_absolute_topic",
+    resolvedTopicName(sdf, "absolute_topic", "/ns", "/default_absolute_topic"));
+  EXPECT_EQ("/default_empty_topic",
+    resolvedTopicName(sdf, "empty_topic", "", "/default_empty_topic"));
+  EXPECT_EQ("/default_empty_topic",
+    resolvedTopicName(sdf, "empty_topic", "/ns", "/default_empty_topic"));
```
azeey, line 338: Can you add a test for a default empty topic that is relative to ensure that the namespace is used in that case?

test/helpers/UnitTestUtil.hh
```diff
@@ -0,0 +1,107 @@
...
+/// \param[in] _eventMgr Event manager used while creating entities.
+/// \param[out] _entity The entity corresponding to the world
+/// \param[out] _pluginSdf Optional copy of the world's first plugin SDF.
+/// If nullptr, no plugin will be stored
+inline void LoadWorldContext(
+  const std::string &_sdfString,
+  EntityComponentManager &_ecm,
+  EventManager &_eventMgr,
+  Entity &_entity,
+  sdf::Plugin *_pluginSdf = nullptr)
```
azeey, line 48: I'm not sure passing a pointer here is correct. If you want the caller's pointer to be updated, then we'll need to pass a double pointer. If you want the caller's plugin object to be updated (sdf::Plugin value copy), then it'd be better to pass by reference. Better yet, just return the value (see https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#f20-for-out-output-values-prefer-return-values-to-output-parameters)

include/gz/sim/Util.hh
```diff
@@ -78,6 +78,23 @@ namespace gz
...
+    /// namespace, while absolute topic names are left unchanged. If no custom
+    /// topic is available, the default topic is used.
+    /// \param[in] _sdf SDF to read the topic name from.
+    /// \param[in] _sdfElement Name of the SDF child element containing the
+    /// topic name.
+    /// \param[in] _ns Namespace to prepend to relative topic name.
+    /// \param[in] _defaultTopic Topic to use when no custom topic is specified.
+    /// \return A valid Gazebo Transport topic name, or an empty string if no
+    /// valid topic could be generated.
+    std::string GZ_SIM_VISIBLE resolvedTopicName(
```
azeey, line 93: We have three consecutive string parameters here which can make it easy to accidentally swap the arguments at the call site. Consider creating a struct with `string_view` fields.

include/gz/sim/Util.hh
C88-YQ: Thanks for the suggestion! Done :smile:

src/Util.cc
C88-YQ: Thanks for the suggestion! Done :smile:

src/Util.cc
C88-YQ: Thanks for the reminder! I updated the prefix handling so `/` is preserved without appending another slash.

src/Util_TEST.cc
C88-YQ: @azeey My idea is that `default_topic` should already be fully resolved before calling this function, so `_ns` is not added to `default_topic` here.

The main reason is to keep the old behavior. Before namespace support, `/model/<model_name>` was only added for default topics, not for user-defined topics. If we use `/model/<model_name>` as `_ns` when no namespace is set, it would also affect user-defined relative topics and change the old behavior.

I see two possible approaches:

1. Resolve the full `default_topic` before calling this function, which is what I currently do.
2. Pass two namespace/prefix values: one for the explicit namespace and one for the default topic prefix.

I chose the first approach, so I didn't add a test for a relative `default_topic` here. Do you think the second approach would be better?

include/gz/sim/Util.hh
C88-YQ: Thanks for the suggestion! Done :smile:

test/helpers/UnitTestUtil.hh
C88-YQ: Thanks for the suggestion! I changed `_pluginSdf` to be passed by reference since the caller's plugin object is updated here. I also added overloads for the cases where the plugin SDF isn't needed.

include/gz/sim/Util.hh
C88-YQ: Thanks for the suggestion! That makes sense.

I ended up using `std::string` for the fields because the SDF APIs used here (`HasElement` and `Get`) take `std::string`, so using `std::string_view` would require additional conversions inside the function.

Is there any case here where `std::string_view` would still be preferable that I might be overlooking?

src/Util_TEST.cc
C88-YQ: I also saw your related comment in #3878 about handling this convention inside `resolvedTopicName` instead of in each system. I think that makes sense.

I updated `resolvedTopicName` to also take a `defaultTopicPrefix` and handle the prefix selection internally. It now uses the namespace when provided, otherwise it falls back to the original default prefix.

I also added test coverage for a relative default topic.

include/gz/sim/Util.hh
azeey: I was hoping to avoid copies when creating the struct. But since all the SDF APIs need `std::string`, what you have seems good.

PR #3446, opened 2026-03-30
Server: add PeekEcm and PokeEcm for direct ECM access

arjo129:
🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸

# 🎉 New feature

Closes #<NUMBER>

## Summary
This PR introduces a more direct and efficient way to interact with the simulation state through the gz::sim::Server class, particularly targeted at improving the testing experience and API parity.

  Key Changes:

   1. ECM Introspection (PeekEcm and PokeEcm):
       * Added PeekEcm (const) and PokeEcm (non-const) to the Server class. These methods allow users to pass a lambda that receives a reference to the EntityComponentManager.
       * These methods are thread-safe and intended for use when the server is stopped (e.g., before or after a Run call).
       * Impact: Simplifies unit and integration tests by allowing state verification without requiring a TestFixture or a simulation step.

   2. Server Metrics API:
       * Exposed SimTime, IterationCount, EntityCount, and SystemCount directly in the Server class.
       * Provides a standard, top-level way to query the simulation status.

   3. Python Bindings:
       * Added full Python bindings for all the new server-level methods.
       * Handled the std::function wrapping for poke_ecm to ensure Python callbacks can safely mutate the ECM reference.

   4. Test Improvements:
       * Refactored test/integration/entity_semantics_system.cc to use the new PeekEcm API, demonstrating a significant reduction in boilerplate and a more direct testing pattern.
       * Added python/test/server_TEST.py to verify the new Python bindings.

## TODO
Audit all tests to see opportunities where this can be used.

## Checklist
- [ ] Signed all commits for DCO
- [ ] Added a screen capture or video to the PR description that demonstrates the feature
- [ ] Added tests
- [ ] Added example and/or tutorial
- [ ] Updated documentation (as needed)
- [ ] Updated migration guide (as needed)
- [ ] Consider updating Python bindings (if the library has them)
- [ ] `codecheck` passed (See [contributing](https://gazebosim.org/docs/all/contributing#contributing-code))
- [ ] All tests passed (See [test coverage](https://gazebosim.org/docs/all/contributing#test-coverage))
- [ ] Updated Bazel files (if adding new files). Created an issue otherwise.
[...]

PR #3936, opened 2026-08-30
Add EcmGuard for thread-safe RAII access to the EntityComponentManager

azeey:
# 🎉 New feature

Alternative to https://github.com/gazebosim/gz-sim/pull/3446/

## Summary
This PR introduces `Server::Ecm()`, which returns an `EcmGuard` to provide safe, synchronous read and write access to the EntityComponentManager.

In C++, `EcmGuard` uses the RAII pattern to acquire the server's `runMutex`. This naturally blocks the simulation from starting while the ECM is being inspected or mutated. Conversely, if the simulation is actively stepping, it safely rejects access to prevent race conditions.

In Python, this exposes `server.ecm()` as a native context manager. This enables clean, linear scripting without the need for asynchronous callbacks—making it especially powerful for driving synchronous Reinforcement Learning (Gym) environments.

### Usage Examples

**C++ Usage:**
```cpp
// Safely acquire the ECM lock
if (auto ecmGuard = server.Ecm()) 
{
    // Mutate or read from the ECM safely
    Entity newEntity = ecmGuard->CreateEntity();
    
    // The lock is automatically released when 'ecmGuard' goes out of scope,
    // allowing the simulation to resume safely.
}
```

**Python Usage (RL Gym Loop Example):**
```python
# The context manager provides flat, synchronous access
for _ in range(5):
    with server.ecm() as ecm:
        # Inject forces or manipulate entities
        chassis.add_world_force(ecm, action_force)
    
    # Step the simulation forward
    server.run(True, 1, False)

# Safely extract the final observation state
with server.ecm() as ecm:
    obs = chassis.world_pose(ecm)
```

## Backport Policy

    - [x] Jetty
    - [x] Ionic
    - [ ] Harmonic
    - [ ] Fortress
- [ ] I am not sure

## Test it
1. Build the branch
2. Run the C++ server tests: `ctest -R UNIT_Server_TEST`
3. Run the Python server tests: `ctest -R server_TEST`
4. Use the new API in a Python script via `with server.ecm() as ecm:` and verify it behaves correctly.

## Checklist
- [x] Added tests

Assisted-by: Gemini 3.1 Pro

### DIFF
--- a/python/src/gz/sim/Server.cc
+++ b/python/src/gz/sim/Server.cc
@@ -90,6 +90,7 @@ void defineSimServer(pybind11::object module)
     "Resets a specific simulation runner under this server.")
   .def("ecm",
     pybind11::overload_cast<const std::size_t>(&gz::sim::Server::Ecm),
+    pybind11::call_guard<pybind11::gil_scoped_release>(),
     pybind11::arg("runner_id") = 0,
     "Acquire a thread-safe lock on the ECM as a context manager.")
   .def("iteration_count", &gz::sim::Server::IterationCount,
--- a/python/src/gz/sim/Server.cc
+++ b/python/src/gz/sim/Server.cc
@@ -15,10 +15,14 @@
  */
 
 
+#include <pybind11/chrono.h>
 #include <pybind11/pybind11.h>
+#include <pybind11/stl.h>
 
+#include <gz/sim/EntityComponentManager.hh>
 #include <gz/sim/Server.hh>
 #include <gz/sim/ServerConfig.hh>
+#include <gz/sim/Types.hh>
 
 #include "Server.hh"
 
@@ -30,6 +34,36 @@ namespace python
 {
 void defineSimServer(pybind11::object module)
 {
+  // EcmGuard
+  pybind11::class_<gz::sim::Server::EcmGuard>(module, "EcmGuard",
+    "RAII guard for exclusive thread-safe access to the "
+      "EntityComponentManager.")
+    .def("__enter__", [](gz::sim::Server::EcmGuard &self)
+        -> gz::sim::EntityComponentManager & {
+      if (!self) {
+        throw pybind11::value_error(
+          "Cannot enter invalid EcmGuard (server is running or "
+          "runner_id is invalid)");
+      }
+      return *self;
+    }, pybind11::return_value_policy::reference_internal)
+    .def("__exit__", [](gz::sim::Server::EcmGuard &self,
+                        pybind11::object /*exc_type*/,
+                        pybind11::object /*exc_val*/,
+                        pybind11::object /*exc_tb*/) -> pybind11::bool_ {
+      self.Reset();
+      return pybind11::bool_(false);
+    })
+    .def("valid", &gz::sim::Server::EcmGuard::Valid,
+      "Check whether the guard holds an active lock.")
+    .def("__bool__", &gz::sim::Server::EcmGuard::operator bool)
+    .def("ecm", pybind11::overload_cast<>(&gz::sim::Server::EcmGuard::Ecm),
+      pybind11::return_value_policy::reference_internal,
+      "Get reference to the underlying EntityComponentManager.")
+    .def("reset", &gz::sim::Server::EcmGuard::Reset,
+      "Explicitly release the lock.");
+
+  // Server
   pybind11::class_<gz::sim::Server,
     std::shared_ptr<gz::sim::Server>>(module, "Server",
     "The main simuulation server class. This class manages the simulation "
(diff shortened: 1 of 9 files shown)


azeey:
cc @arjo129

src/Server_TEST.cc
```diff
@@ -1384,3 +1386,154 @@ TEST_P(ServerFixture, WorldControlIgnoredOnExit)
...
+    auto g3 = server.Ecm();
+    EXPECT_TRUE(g3.Valid());
+  }
+
+  // 4. Invalid States
+  {
+    EXPECT_FALSE(server.Ecm(999).Valid());  // Out of bounds runner
+
+    server.Run(false, 0, false);
+    while (!server.Running()) GZ_SLEEP_MS(10);
```
arjo129: Knit: Busy  waiting. 
This could run endlessly (or for the test timeout duration).

python/src/gz/sim/Server.cc
```diff
@@ -53,7 +87,23 @@ void defineSimServer(pybind11::object module)
   .def("reset_all", &gz::sim::Server::ResetAll,
     "Resets all simulation runners under this server.")
   .def("reset", &gz::sim::Server::Reset,
-    "Resets a specific simulation runner under this server.");
+    "Resets a specific simulation runner under this server.")
+  .def("ecm",
```
arjo129, line 91: According to my AI reviewer


Server.run uses
`pybind11::call_guard<pybind11::gil_scoped_release>()`, but Server.ecm does not.
 -  If Python thread A calls server.ecm() while another
  thread is holding runMutex, thread A waits on runMutex
  while still holding the Python GIL.
 -  If the thread holding runMutex executes Python callbacks or systems requiring the GIL, this causes an immediate, unrecoverable AB/BA deadlock.

src/Server.cc
```diff
@@ -39,6 +39,93 @@
...
+//////////////////////////////////////////////////
+EntityComponentManager *Server::EcmGuard::operator->()
+{
+  return this->dataPtr ? this->dataPtr->ecm : nullptr;
+}
+
+//////////////////////////////////////////////////
+const EntityComponentManager *Server::EcmGuard::operator->() const
+{
+  return this->dataPtr ? this->dataPtr->ecm : nullptr;
```
arjo129, line 89: Should we throw an exception or at least log an error rather than returning a NULL ptr?

src/Server.cc
```diff
@@ -39,6 +39,93 @@
...
+}
+
+//////////////////////////////////////////////////
+EntityComponentManager &Server::EcmGuard::operator*()
```
arjo129, line 93: These methods could potentially segfault if the guard is invalid.

arjo129 (commented):
This is a much nicer API than my originally proposed API. One concern I have is the number of things `this->dataPtr->runMutex` is doing. It might be worth opening a ticket to see if we could refactor it.

Other than that I just have some minor comments. I'll re-align my demo with this new API.

src/Server.cc
```diff
@@ -433,6 +520,45 @@ bool Server::Reset(const std::size_t _runnerId)
   return true;
 }
 
+//////////////////////////////////////////////////
+Server::EcmGuard Server::Ecm(const std::size_t _runnerId)
+{
+  EcmGuard guard;
+  std::unique_lock<std::mutex> lock(this->dataPtr->runMutex);
```
arjo129, line 527: `dataPtr->runMutex` is doing a lot of heavy work.  If we hold on to the `runMutex` (as we do in the current implementation) its worth noting that `ServerPrivate::Stop()` will deadlock if called from another thread.

src/Server.cc
azeey: We don't use exceptions in the Gazebo codebase, so I'd hesitate to do that. I believe the only time `dataPtr` would be null is if the EcmGuard has been moved from. In C++,  if a user tries to use a moved-from object without reinitializing it, the behavior is undefined. This is like using a `std::unique_ptr` after you've moved from it. Gemini had an `assert` here right before the `return`, but I removed it because typically don't use that either in the codebase, but I could add that back in. It would only be useful in debug builds though.

src/Server.cc
azeey: Yes, but see https://github.com/gazebosim/gz-sim/pull/3936#discussion_r3895984841

src/Server.cc
azeey: Would it deadlock? Wouldn't the other thread just wait till `runMutex` is available? IMO having two mutexes is much more likely to result in deadlocks due to priority inversion. By using just `runMutex`, the only risk is that the user would forget they have an `EcmGuard` in the scope and try to call `Server::Run` or `Server::Stop`. I think having a noncallback based access to the ECM is worth the risk though.

src/Server_TEST.cc
azeey: Yup, using `waitUntil` in https://github.com/gazebosim/gz-sim/pull/3936/commits/68d490362a5d50cc431c2b7b1c02a3e458ff4168

python/src/gz/sim/Server.cc
azeey: Nice catch. This made me think more about the design of the Server API. We have calls like `EntityCount` which don't lock any mutex, and others APIs that do. Calling `EntityCount` while a simulation is running in a separate thread is a data race. It seems like we have an inconsistent API. We really need to answer how `Server` is meant to be used:
1.  in lock step mode requiring no thread synchorinization
2. with the users own/external thread synchronization.
3. Completely thread safe.
  
If our use case is (1) or (2), we can really simplify a lot of things. Even this `EcmGuard` class wouldn't be necessary. OTOH, if we want (3) to be able to run simulation async and monitor it's progress from a separate thread, we need all the public APIs to be thread synchronized, which can be costly. For now, we have a hybrid and so I have gone ahead and added `gil_scoped_release` as you suggested.

I've also learned that the non blocking `Server::Run` is actually never used in our codebase except in tests. I'll continue investigating this, but maybe we should just remove that non-blocking API and have a clear API where we expect the user to either do lock step or provide their own thread synchronization.

arjo129 (approved):
LGTM

azeey:
Let's hold off on backporting until other python changes have landed and have been tested.

PR #3957, opened 2026-09-03
Expose ECS components and ECM operations to Python

azeey:
# 🎉 New feature

Split out from https://github.com/gazebosim/gz-sim/pull/3494

## Summary

Expose Gazebo ECS components to Python and extend EntityComponentManager bindings with component lifecycle and data manipulation methods.

Key additions:
- ComponentPybindRegistry: A thread-safe, type-erased registry for Python getter and setter trampolines mapped by component type ID, supporting runtime registration from plugins with Pimpl ABI protection.
- Python ECM bindings: Expose entity hierarchy operations (parent_entity, set_parent_entity), component queries (entity_has_component_type, component), creation (create_component), modification (set_component_data), change notification (set_changed), and deletion (remove_component).
- Component coverage: Python bindings for standard components and internal structs (DetachableJointInfo, TemperatureRangeInfo, RayInfo, RaycastResultInfo, RaycastDataInfo).
- Unit tests: Comprehensive Python test suite verifying component lifecycle, type validation, in-place data mutation, entity hierarchy, and 100% parity with C++ ComponentFactory.

## Backport Policy

    - [x] Jetty
    - [x] Ionic
    - [x] Harmonic
    - [x] Fortress
- [ ] I am not sure

## Test it

Run the Python component test suite:
```bash
pytest python/test/components_TEST.py
# or via ctest:
ctest -R components_TEST --output-on-failure
```

Reviewers can also interactively inspect components in Python:
```python
from gz.sim import EntityComponentManager, components
from gz.math import Pose3d

ecm = EntityComponentManager()
e = ecm.create_entity()

# Tag component
ecm.create_component(e, components.Model)
assert ecm.entity_has_component_type(e, components.Model)

# Data component with reference mutation
ecm.create_component(e, components.Pose, Pose3d(1, 2, 3, 0, 0, 0))
pose = ecm.component(e, components.Pose)
pose.set_x(10.0)
assert ecm.component(e, components.Pose).x() == 10.0
```

## Checklist
- [x] Added tests

Assisted-by: Gemini 3.8


python/src/gz/sim/Components.cc
```diff
@@ -0,0 +1,195 @@
...
+{
+namespace sim
+{
+namespace python
+{
+
+/////////////////////////////////////////////////
+void defineComponentStructs(pybind11::module &m)
+{
+  namespace comps = gz::sim::components;
```
arjo129, line 52: I presume for the skipped components we would have to add them here?

python/src/gz/sim/EntityComponentManager.cc
```diff
@@ -36,9 +44,109 @@ void defineSimEntityComponentManager(pybind11::object module)
...
+           auto setter =
+               ComponentPybindRegistry::Instance()->Setter(_comp.typeId);
+           if (!setter)
+           {
+             throw pybind11::type_error(
+                 "Component type is not registered for Python manipulation");
+           }
+           if (!setter(self, _entity, _data, false))
+           {
+             throw pybind11::type_error(
```
arjo129: Knit: Should this be a type error or ValueError/KeyError since the entity is not found.

include/gz/sim/CMakeLists.txt
```diff
@@ -1,9 +1,19 @@
...
+set(excluded_dirs python)
+if(NOT ENABLE_GUI)
+  list(APPEND excluded_dirs "gui")
 endif()
 
+gz_install_all_headers(EXCLUDE_DIRS ${excluded_dirs})
+
+install(
+  DIRECTORY python
+  DESTINATION "${GZ_INCLUDE_INSTALL_DIR_FULL}/${PROJECT_INCLUDE_DIR}"
```
arjo129, line 14: knit: Consider adding `FILES_MATCHING PATTERN "*.hh"`

src/ComponentPybindRegistry.cc
```diff
@@ -0,0 +1,157 @@
...
+    return _name.substr(prefix.length());
+  }
+  return _name;
+}
+
+/////////////////////////////////////////////////
+void ComponentPybindRegistry::Register(ComponentTypeId _typeId, uintptr_t _id,
+                                       GetterFn _getter, SetterFn _setter)
+{
+  std::string name = CleanName(
```
arjo129: In C++, static initialization order across translation units is undefined. If a python component static initializer executes before the component's core `Factory` initializer, `Factory::Instance()->Name(_typeId)` returns "", permanently caching an empty name in the `PybindDescriptor`. Perhaps lazily looking up component names in the proxy may be a better idea?

arjo129 (commented):
This is in general a really nice API. Ive got a few minor knits, but in general it feels very ergonomic. I will be adding a few more component classes along with a new RL example using these new APIs shortly.

python/src/gz/sim/Components.cc
azeey: Yes, for internal structs. For protobuf based ones, I'm hoping to create a followup PR that just uses serialization/deserialization.

include/gz/sim/CMakeLists.txt
azeey: Done in c89a53419a4f

python/src/gz/sim/EntityComponentManager.cc
azeey: Changed to `KeyError` in c89a53419a4f

src/ComponentPybindRegistry.cc
azeey: Great catch! Also, the fact that there was a lookup in Getter to get the name was bothering me. So now, the name is passed in from the registration macro (since it was already available at compile time) and it is captured in a lambda instead. 949a2625c073

azeey:
I did an audit of the registered components and found some discrepancies. Some components were incorrectly registered when they were structs or protobuf types which currently don't have pybind11 casters. To make this more robust, I've added a test that checks that is possible to instantiate all components known to the C++ registery except those marked to be skipped. Now, if anyone adds a new component without updating `PyComponents.cc`, the test will fail. Users will either need to register the component to python or add it to `KNOWN_SKIPPED_COMPONENTS` in `components_TEST.py`. This is not ideal, but we hardly add new components, so I didn't think finding a more automated/niceer way to do it was worth the effort.

include/gz/sim/components/SlipComplianceCmd.hh
```diff
@@ -42,7 +42,7 @@ namespace components
   using SlipComplianceCmd =
     Component<std::vector<double>, class SlipComplianceCmdTag,
               serializers::VectorDoubleSerializer>;
-  GZ_SIM_REGISTER_COMPONENT("gz_sim_components.SlipComplianceCmd ",
+  GZ_SIM_REGISTER_COMPONENT("gz_sim_components.SlipComplianceCmd",
```
arjo129: Does this change affect our ability to backport?

azeey:
I initially had the component getter use `pybind11::return_value_policy::reference` to match the semantics of C++, but I found out that some types, such as std::string and all primitives use the built-in type-caster and have value semantics. Trying to mutate components holding those types would not actually work, so  in https://github.com/gazebosim/gz-sim/pull/3957/commits/3380227854219cfa3f563e8a5aaf44eb64957786, I've changed the getter to return by value. This way, only `set_component_data` can be used to write a component, which is less confusing.

include/gz/sim/python/ComponentPybindRegistry.hh
```diff
@@ -0,0 +1,317 @@
...
+        {
+          return _ecm.SetComponentData<T>(_entity, data);
+        }
+
+        T *comp = _ecm.Component<T>(_entity);
+        if (nullptr == comp)
+        {
+          return _ecm.CreateComponent(_entity, T(data)) != nullptr;
+        }
+        comp->Data() = data;
```
luca-della-vedova, line 263: Not necessarily for this PR but I wonder if we should slowly start to move away from the non const `comp->Data()`, in favor of `comp->SetData([...])`, two main reasons:

We have a deprecation notice that promises deprecating it in Citadel (!) https://github.com/gazebosim/gz-sim/blob/7a6d8d4b45f06f46aaf6ff494ecbfceda84682d3/include/gz/sim/components/Component.hh#L377-L382.

Going through an explicit setting data API rather than returning a mutable reference would make it easier to "intercept" changes and, for example, run change detection in entt (to hold invariants, such as parent changes that are intercepted to update children), or even our own change detection (one time / periodic changes)

include/gz/sim/python/ComponentPybindRegistry.hh
```diff
@@ -0,0 +1,317 @@
...
+  /// component name.
+  /// \param[in] _name Name of the component.
+  /// \return The getter function returning the component data cast to a python
+  /// object (or ComponentProxy for NoData).
+  static ComponentPybindRegistry::GetterFn CreateGetter(const char *_name)
+  {
+    return [_name](const gz::sim::EntityComponentManager &_ecm,
+                   const gz::sim::Entity &_entity) -> pybind11::object
+    {
+      (void)_name;
```
luca-della-vedova, line 168: Nit, this can be scoped down a bit since (if I understand corrctly) it's only meant to suppress unused variable warning in one branch of the `if` statement?

python/test/components_TEST.py
```diff
@@ -0,0 +1,346 @@
...
+
+        # Mutating the returned value must NOT reach the ECM.
+        pose.set_x(42.0)
+        self.assertAlmostEqual(1.0, ecm.component(e, components.Pose).x())
+
+        # Successive reads return independent snapshot objects.
+        a = ecm.component(e, components.Pose)
+        b = ecm.component(e, components.Pose)
+        a.set_x(7.0)
+        self.assertAlmostEqual(1.0, b.x())
```
luca-della-vedova, line 112: This is quite a footgun imho. Two alternative proposals:

* Should we make the `component` function return a `const`? I haven't done Python in some time but I suppose users would have to call a python copy function explicitly to edit it the way this test does and it would be extremely clear that they are working on a copy and the original will be unchanged, it is a bit more overhead in its current incantation though, which could be potentially mitigated by...
* If we do the above how about revisiting the "return a copy" and return a const reference instead? If objects are const there is less risk of confusing behavior with return by value vs reference since users can't actually edit the object. Apart from the aliasing risk (to be acknowledged I suppose, we also say in C++ not to hold component pointers for extended periods of time) the reference vs copy would be just a performance difference and not a behavior difference.

python/test/components_TEST.py
```diff
@@ -0,0 +1,346 @@
...
+            ecm.component(e, "invalid_type")
+
+        with self.assertRaises(TypeError):
+            ecm.entity_has_component_type(e, 12345)
+
+        with self.assertRaises(TypeError):
+            ecm.remove_component(e, 12345)
+
+        with self.assertRaises(TypeError):
+            ecm.set_changed(e, components.Pose.type_id)
```
luca-della-vedova, line 252: Let's add a comment into why this is expected to throw (I suppose because it should be `components.Pose` not `components.Pose.type_id`?)

python/test/components_TEST.py
```diff
@@ -0,0 +1,346 @@
...
+        """Test ECM entity creation, existence, and hierarchy."""
+        ecm = EntityComponentManager()
+        parent = ecm.create_entity()
+        child = ecm.create_entity()
+
+        self.assertTrue(ecm.has_entity(parent))
+        self.assertTrue(ecm.has_entity(child))
+        self.assertFalse(ecm.has_entity(999999))
+
+        self.assertEqual(0, ecm.parent_entity(child))
```
luca-della-vedova, line 264: In "our" ECM 0 is not really an ID, rather it's `kNullEntity`.
I wonder if we should map `kNullEntity` to `None` in Python so it's clear from this API that this entity has no parent rather than having a parent of 0?
If this is not a good idea we could consider exporting a `null entity` constant in Python instead?

python/test/components_TEST.py
```diff
@@ -0,0 +1,346 @@
...
+
+            # For data components with non-None default data, roundtrip through setter
+            if val is not None and not isinstance(val, components.ComponentProxy):
+                ecm.set_component_data(e, comp, val)
+
+            ecm.remove_component(e, comp)
+
+        self.assertEqual(
+            len(all_factory_comps) - len(KNOWN_SKIPPED_COMPONENTS),
+            tested_count)
```
luca-della-vedova, line 342: This looks duplicated with the test above `test_component_registration_parity`, should we only keep it there and have this test only make sure that round trips work?

python/test/components_TEST.py
azeey: Unfortunately, it's virtually impossible to do `const` for classes bound using pybind11. Instead of `component`, what if we only exposed `component_data` or `component_data_copy`. This would accomplish two things. (1) `component` in C++ gives a wrapper around the data, but here, we're returning the data directly, which is a clear mismatch. `component_data` would be explicit that the data is returned. (2) `component_data` is a copy in C++ already since it returns a `std::optional`. However, if a user doesn't know the C++ already, it's still not super clear whether it's a copy or reference. We could offer `component_data_copy`, but that not the most ergonomic name and wouldn't match the C++ API.

This also has implications on what we return from `ecm.each`. The most performant thing is to return a copy of the data rather than a wrapper, but that's not what C++ users expect. I propose exposing `ecm.each_data` instead.

There's a little wrinkle with `NoData` components. The C++ ComponentData does not compile if you give it a `NoData` component. In python, we'd have to return something since there's no other way to get that information on an entity. But I think that tradeoff is okay.

include/gz/sim/components/SlipComplianceCmd.hh
azeey, line 45: I've split this out into https://github.com/gazebosim/gz-sim/pull/4000 since we can't backport it

include/gz/sim/detail/EntityComponentManager.hh
```diff
@@ -58,12 +58,12 @@ namespace traits
 /// equality operator.
 /// If `DataType` doesn't have an equality operator defined, it will return
 /// false.
-/// For doubles, `gz::math::equal` will be used.
+/// For floating-point types, `gz::math::equal` will be used.
```
azeey, line 61: Split out into https://github.com/gazebosim/gz-sim/pull/4001 since it's unrelated to this PR

python/test/components_TEST.py
luca-della-vedova: Ah I see, I guess the `component_data` approach with an optional return sounds the most reasonable for mirroring the C++ API and not binding us (sorry couldn't resist) to maintain a slightly mismatched API in the long term.
Maybe in the future we can figure out a strategy to make sure we can do reference for all types, or maybe just have static_asserts to only expose a `component` API for types that support references rather than values, although with Python not being a compiled language that is also a recipe for runtime exceptions.

Same for `each`, a new method `each_data` sounds reasonable to me.

include/gz/sim/components/SlipComplianceCmd.hh
luca-della-vedova: Then we should revert it here right? Same for the other change

python/test/components_TEST.py
azeey: By "this", do you mean the single `assertEqual` or the whole test. If it's the latter, not sure what's duplicated exactly. I guess you can argue whether we need the `has_python_bindings` checks at all since they would implicitly be covered by `test_all_registered_components_roundtrip`

azeey:
I've made some changes per our discussion. There are a few more other changes as well:
- Renamed the `AddPybindGetterSetter` struct to `ComponentOps` since we're there are now more component operations stored there.
- Add a "Creator" ComponentOp which is used for creating new components instead of relying on Setter. This mirrors the C++ API better and calls CreateComponent which handles change notification.
- Change `component` and to `component_data` as discussed in https://github.com/gazebosim/gz-sim/pull/3957#discussion_r4039691451 to indicate that the the function returns the data, not a wrapper around it.
- Remove the `compare` flag from Setter/set_component_data since that's not available in C++. If we decide to add that later, we can do it in C++ as well.

@luca-della-vedova PTAL

include/gz/sim/components/SlipComplianceCmd.hh
azeey: Just needed an update from main

python/test/components_TEST.py
azeey: We already have `gz.sim.K_NULL_ENTITY`, so I used that in https://github.com/gazebosim/gz-sim/pull/3957/commits/b63aa3fcfd76c9a53fc0a170b6e20a69f1c88b58

python/test/components_TEST.py
azeey: Added a comment in https://github.com/gazebosim/gz-sim/pull/3957/commits/b63aa3fcfd76c9a53fc0a170b6e20a69f1c88b58

include/gz/sim/python/ComponentPybindRegistry.hh
azeey: Done in https://github.com/gazebosim/gz-sim/pull/3957/commits/b63aa3fcfd76c9a53fc0a170b6e20a69f1c88b58

include/gz/sim/python/ComponentPybindRegistry.hh
azeey: Now using SetComponentData, so this is no longer relevant.

azeey: But, I generally agree. The only issue with SetData is that you need to provide a comparator function, which is inconvenient. We should set the default parameter to `CompareData<T>` so it gets figured out automatically. I also wonder if there are use cases where we don't want to compare. In the Physics system, for example, there's code that executes only on links that the physics engine marked as having changed poses. Do we really need to do an extra comparison there? No way to tell without measurements of course.

Issue #3960, opened 2026-09-07
Windows: gz sim -v4 shapes.sdf never opens a window; exception in RenderEngineManager plugin lookup


Foadsf:
### Environment

- **OS**: Windows 11 Enterprise 10.0.26200 (x64)
- **Install**: binary, exactly as described in [Binary Installation on Windows 11](https://gazebosim.org/docs/latest/install_windows/) — pixi 0.79.0, `gz-sim = "10.*"` from conda-forge, with the two documented Qt activation variables
- **Versions**: `gz sim --versions` → `10.5.0`; `libgz-sim 10.5.0`, `libgz-rendering 10.0.2`, `libgz-msgs 12.0.2`
- **GPU**: NVIDIA RTX A2000 Laptop GPU (driver 32.0.15.9595) + Intel UHD Graphics (32.0.101.7085). Physical console session, not RDP.

### Description

Following the Windows binary install tutorial to the letter, `gz sim -v4 shapes.sdf` never opens a window. The tutorial presents this command as the working path (the note about `gz sim` with no arguments points at #3859, which is a different symptom).

`gz-sim-main` and `gz-sim-gui-client` both start and stay alive, and `Process.Responding` stays `True`, but no window is ever mapped and `MainWindowTitle` stays empty. I held the run for 60 s and 75 s across three separate attempts.

Headless works fine on the same install:

```
gz sim -s -v4 -r --iterations 200 shapes.sdf
```

runs DART, serves the scene topics, and shuts down cleanly.

### Steps to reproduce

1. Install per the Windows binary tutorial (`pixi.toml` with `gz-sim = "10.*"` plus the documented `QT_QPA_PLATFORM_PLUGIN_PATH` / `QML2_IMPORT_PATH`).
2. `pixi run gz sim -v4 shapes.sdf`
3. No window appears. Both processes remain resident.

### Output

QML loads and the world loads, then a C++ exception is thrown out of the render-engine lookup and caught by Qt:

```
Stack trace (most recent call last):
#29  BaseThreadInitThunk
...
#23  gz::sim::v10::gui::runGui
#22  QCoreApplication::exec
...
#13  QQuickItem::event
#10  GzPluginHook
#9   GzPluginHook
#8   GzPluginHook
#7   gz::rendering::v10::RenderEngineManager::Engine
#6   gz::rendering::v10::RenderEngineManager::Instance
#5   gz::rendering::v10::RenderEngineManager::Instance
[...]

PR #3961, opened 2026-09-07
Add a Windows build job to CI

Foadsf:
🎉 New feature

Contributes to #168 (Windows support).

## Summary

gz-sim's GitHub Actions CI is Ubuntu-only. Windows is exercised on the buildfarm and in the conda-forge feedstock, both of which run after the fact, so a change that breaks the Windows build is not visible on the pull request that introduces it.

This adds a `Windows CI` job that configures and builds the source tree with MSVC against the released Gazebo libraries from conda-forge.

## Design notes

- **Dependencies come from the released conda-forge packages.** Depending on `gz-sim = "10.*"` pulls in every upstream Gazebo library at the version the collection actually ships with; the gz-sim binaries that arrive with it are unused, since the build under test overrides them. This keeps the job to a single compile instead of a 16-repository source build, which would not fit a pull-request gate.
- **Tests are off** (`-DBUILD_TESTING=OFF`), matching the [documented Windows source build](https://gazebosim.org/docs/latest/install_windows_src/). This job is a build-regression gate. Turning tests on for Windows is worth doing but is a separate piece of work — for a datapoint in the other direction, gz-math's suite passes 130/130 on Windows today.
- **`.github/ci/windows/pixi.lock` is committed** so `setup-pixi` can compute a cache key; without it the action fails outright.
- **`windows-2022` is pinned** rather than `windows-latest`, so a runner-image rollover is a deliberate change rather than a surprise failure.

## Verified before proposing

The workflow was run on a fork rather than submitted untested. It is green end to end — dependency install, configure and a full MSVC build of the source tree all succeed on `windows-2022`, in **20 minutes** wall clock on a cold cache. That is a comfortable fit for a per-PR gate.

## Why now

Building on Windows is not hypothetical for this repository — the conda-forge feedstock ships `gz-sim 10.5.0` for `win-64` today, and the project publishes Windows binary and source install tutorials. A per-PR build gate is the cheapest way to keep that from silently rotting.

Separately, and not addressed by this PR: on a clean install following the Windows binary tutorial, `gz sim -v4 shapes.sdf` does not open a window on my machine — reported in #3960.

Happy to fold this into `ci.yml` as an extra job rather than a separate workflow, or to adjust the runner image and pinning, if you'd prefer.

### DIFF
--- a/.github/ci/windows/pixi.lock
+++ b/.github/ci/windows/pixi.lock
@@ -1,0 +1,3741 @@
+version: 7
+platforms:
+- name: win-64
+  virtual-packages:
+  - __win=10.0
+  - __archspec=0=x86_64
+environments:
+  default:
+    channels:
+    - url: https://conda.anaconda.org/conda-forge/
+    packages:
+      win-64:
+      - conda: https://conda.anaconda.org/conda-forge/noarch/ca-certificates-2026.7.22-h4c7d964_0.conda
+      - conda: https://conda.anaconda.org/conda-forge/noarch/font-ttf-dejavu-sans-mono-2.37-hab24e00_0.tar.bz2
+      - conda: https://conda.anaconda.org/conda-forge/noarch/font-ttf-inconsolata-3.000-h77eed37_0.tar.bz2
+      - conda: https://conda.anaconda.org/conda-forge/noarch/font-ttf-source-code-pro-2.038-h77eed37_0.tar.bz2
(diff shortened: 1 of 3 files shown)


azeey:
We do have PR jobs for windows running on Jenkins. For example, for this PR, the job is https://build.osrfoundation.org/job/gz_sim-pr-cnlwin/1617/. @j-rivero may correct me, but we don't have any plan or interest in adding Windows Github Action jobs since the Jenkins job is sufficient.

PR #3520, opened 2026-05-07
Fix Buoyancy state contamination on world reset

XINJIANGMO:
# 🦟 Bug fix

Depends on #3503 

## Summary
During the reset audit, the `Buoyancy` plugin failed the tests. After an environment reset, the model didn’t consistently return to its early-episode trajectory.

## Test logic
1. Recorded an early trajectory window of a model in `graded_buoyancy.sdf` before resetting.
2. Let the simulation run for a while, ensuring the system moves explicitly away from its initial state.
3. Executed a world reset.
4. Recorded the post-reset trajectory for the exact same early-episode time window.
5. Compared both trajectories to ensure they match identically, confirming whether the reset successfully brought the system back to a clean initial state.

## Root cause
The issue stems from how `Buoyancy` tracks the running state of links. It originally relied on the `EachNew` entity scan to fetch and cache components like `Volume` and `CenterOfVolume`. When the world resets, the entities are restored but they aren't strictly considered "new" again by the `ECM`. This led to broken buoyancy states where required components couldn't be reliably rebuilt on reset.

## Fix logic
1. Add ISystemReset for the `Buoyancy`
2. Clear the cached states: `centerOfVolumes`,`volumes` and `buoyancyForces`
3. Introduce a `rescanEntities` flag. Once a reset occurs, this flag triggers a fresh, full entity component scan in the very next `PreUpdate` step, gracefully recovering the required runtime components.

## Checklist
- [x] Signed all commits for DCO
- [ ] Added a screen capture or video to the PR description that demonstrates the fix (as needed)
- [x] Added tests
- [ ] Updated documentation (as needed)
- [ ] Updated migration guide (as needed)
- [ ] Consider updating Python bindings (if the library has them)
- [ ] `codecheck` passed (See [contributing](https://gazebosim.org/docs/all/contributing#contributing-code))
- [ ] All tests passed (See [test coverage](https://gazebosim.org/docs/all/contributing#test-coverage))
- [ ] Updated Bazel files (if adding new files). Created an issue otherwise.
- [ ] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Agazebosim+archived%3Afalse+) to support the maintainers
[...]

PR #3978, opened 2026-09-11
Buoyancy: add runtime enable/disable services and a BuoyancyEnable model system

ahcorde:
# 🎉 New feature

## Summary

Bring gz_buoyancy's registration which means a `<enable_by_default>` option and `/world/<world>/buoyancy/enable` and `/disable services`, with re-scanning of existing links and a denied set that survives world resets, plus a new `buoyancy-enable` model system that registers a model's own displacement links under the name it was spawned with. 

Added integration tests for the services, self-registration and renamed spawns. 

Requires https://github.com/gazebosim/gz-math/pull/847

## Backport Policy

    - [x] Jetty
    - [x] Ionic
    - [x] Harmonic
    - [x] Fortress
- [ ] I am not sure
## Checklist
- [x] Added tests

Assisted-by: Claude Opus 4.8


caguero (commented):
Thanks for this patch, having the ability to spawn vehicles with buoyancy enabled is great!

I've been actually thinking about the same issue and I'd like to suggest an alternative that doesn't need additional plugins or transport:

https://github.com/gazebosim/gz-sim/tree/caguero/buoyancy_marked_collisions

In this variant a link marks the collisions that displace water with an attribute, `gz:buoyancy="true"`. A link with marked collisions floats by those alone, whatever name the model was spawned under and whether or not the `<enable>` list names it; its other collisions stay contact geometry. The world keeps the same `<enable_by_default>` flag as this PR, so it can describe the fluid and float nothing on its own. Everything else is unchanged: worlds with no list still float everything, worlds with a list still float only what it names, and the reset rescan from #3520 covers the new path too.

Why I think it is worth considering:

* No second system and no transport. The mark is data in the model, read from the collision element the system already has. Registration over services needs a model side plugin that keeps asking every step until the world answers.
* The mark also says which geometry is buoyancy geometry, so a vehicle can keep a contact hull and a displacement box on one link without a dummy link to separate them. With a zero collide bitmask the displacement box stops touching things.
* It is a small patch in the plugin, most changes are adding tests.
* It keeps every current behaviour.

The branch carries some integration tests: restricted world, a link floated by its mark next to a same sized contact box that must not count, a spawn under a different name, precedence over the list in a default world, a nested model, and a reset.

What do you think?

PR #3982, opened 2026-09-14
MulticopterMotorModel: publish rotor speed on motorSpeedPubTopic

MDEAGEWT:
# 🦟 Bug fix

## Summary

`examples/worlds/multicopter_velocity_control.sdf` sets `<motorSpeedPubTopic>motor_speed/N</motorSpeedPubTopic>` on every rotor of both vehicles, but `MulticopterMotorModel` never read the tag or published anything on it — a silently dead parameter since the plugin was ported from `rotors_simulator` in 2019, where connecting the publisher was left as a TODO ([archived review](https://osrf-migration.github.io/ignition-gh-pages/#!/ignitionrobotics/ign-gazebo/pull-requests/322)).

To reproduce: run `gz sim -r multicopter_velocity_control.sdf`, then `gz topic -l | grep motor_speed` — nothing is listed, even though the SDF asks for 11 such topics.

This PR implements the tag:

* `Configure` advertises a `gz.msgs.Double` publisher on `<robotNamespace>/<motorSpeedPubTopic>`, but only when the tag is present, so worlds that don't set it are unaffected.
* `PreUpdate` publishes the rotor's angular velocity in rad/s when the topic has subscribers. The value is `turningDirection * realMotorVelocity`, so the nominal spin direction reads positive for both cw and ccw rotors and a consumer doesn't need to know `turningDirection` or `rotorVelocitySlowdownSim`. (Upstream RotorS publishes the raw joint velocity instead — happy to match that if you prefer.)

No new system interfaces and no publish-rate parameter.

**Verify it**: the new `MulticopterTest.MotorSpeedPub` commands a positive motor speed in `quadcopter.sdf` (which mixes cw and ccw rotors) and checks all 4 `/X3/motor_speed/N` topics converge to it with a positive sign. Manually:
```
gz sim -r multicopter_velocity_control.sdf
gz topic -e -t /X4/motor_speed/0
```

## Backport Policy
    - [x] Jetty
    - [x] Ionic
    - [x] Harmonic
    - [ ] Fortress
- [ ] I am not sure

No API/ABI change and no behavior change unless the tag is set. I built the plugin and ran `INTEGRATION_multicopter` with this commit applied on gz-sim8, gz-sim9 and gz-sim10: clean build, all tests pass. Fortress not checked.

## Checklist
- [ ] Added a screen capture or video to the PR description that demonstrates the fix
- [x] Added tests

Assisted-by: Claude Fable 5.1


test/integration/multicopter.cc
```diff
@@ -140,6 +143,72 @@ TEST_F(MulticopterTest, GZ_UTILS_TEST_DISABLED_ON_WIN32(CommandedMotorSpeed))
...
+  const std::size_t nIters{500};
+  const double cmdSpeed{100};
+  testSystem.OnPostUpdate(
+      [&](const UpdateInfo &_info, const EntityComponentManager &)
+      {
+        if (_info.iterations == iterTestStart)
+        {
+          msgs::Actuators msg;
+#if GOOGLE_PROTOBUF_VERSION >= 7035000
+          msg.mutable_velocity()->resize(4, cmdSpeed);
```
arjo129: Is there a reason we need to run this during the phase postUpdate?  FWIW this entire test file is probably one of the first places where we should consider using the new `EcmGuard` API.

arjo129 (approved):
FWIW it looks good to me, the test file probably should be updated so it's easier to read, rather than the current set up.

test/integration/multicopter.cc
MDEAGEWT: No particular reason, I was just following the existing Relay/PostUpdate pattern in this file. Since the new test doesn't touch the ECM at all, I removed the callback in 41eaffc9e and made it a plain sequence: step, publish, step, assert. I left EcmGuard out because nothing else in this file uses it yet. If you'd like me to try it here, let me know.

PR #3984, opened 2026-09-14
log: use parent directory when playback path points at state.tlog

### LINKED ISSUE #846: Non-intuitive error when specifying path to state.tlog for playback
## Environment
* OS Version: Bionic
* Source or binary build? Binary Dome

## Description
* Expected behavior: If I mistakenly specify the whole path to `state.tlog` and not just to its directory when passing a path to `--playback`, Ignition Gazebo either recognizes this common error and fixes it for me, or at least shows an informative error message.
* Actual behavior: 

```
[Err] [Zip.cc:126] Error opening zip archive: '/media/data/bags/state.tlog'
[Err] [LogPlayback.cc:474] Failed to extract recording to [/media/data/bags/state_extracted]
[Err] [LogPlayback.cc:207] Cannot play back files.
```

harsh839:
# 🦟 Bug fix

Fixes #846

## Summary
When a user passes the full path to `state.tlog` (rather than the recording **directory**) to playback, `LogPlayback::Configure` previously failed. gz-sim now recognizes this common mistake, logs *"Path points to a .tlog file, using parent directory [...] instead"*, and automatically uses the parent directory so playback works as expected. Non-`.zip`, non-`.tlog` paths now get an informative error explaining that a recording directory or a zip of one is expected.

Reproduce before: `gz sim --playback path/to/logs/state.tlog` fails to configure playback.
After: the notice above is logged and playback loads `state.tlog` from the parent directory.

## Backport Policy
    - [ ] Jetty
    - [ ] Ionic
    - [ ] Harmonic
    - [ ] Fortress
- [x] I am not sure

## Checklist
- [ ] Added tests

Assisted-by: harsh839 <harshbhargav440@gmail.com>


kvnloo:
One edge case looks worth tightening before this lands: the new branch treats any existing ".tlog" file as if it were "state.tlog".

For example, if a user passes "/logs/other.tlog" and "/logs/state.tlog" also exists, "Configure()" rewrites "logPath" to "/logs", and "Start()" then opens "/logs/state.tlog". That means playback can silently use a different file than the one the user supplied.

Since #846 is specifically about the common mistake of passing the full path to "state.tlog", would it make sense to gate this recovery path on "common::basename(logPath) == "state.tlog"" and keep other ".tlog" files on the unsupported-file path?

It also looks like "test/integration/log_system.cc" already has the right fixture in "RecordAndPlayback" to cover this cheaply: one case where the exact "state.tlog" path succeeds, and one arbitrary ".tlog" filename that is rejected. That would pin the intended behavior without broadening the change.

Assisted-by: OpenAI ChatGPT GPT-5.6 Sol

harsh839:
@[kvnloo](https://github.com/kvnloo): Good catch, fixed in 4fc8cc0. The recovery is now gated on common::basename(logPath) == "state.tlog", so an arbitrary .tlog file (e.g. /logs/other.tlog sitting next to a state.tlog) goes down the unsupported-file path instead of silently playing a sibling recording. RecordAndPlayback now covers both sides: the exact state.tlog path plays back (recovered to its parent directory), and a separate myLog.tlog copy is rejected (no state messages received).

src/systems/log/LogPlayback.cc
```diff
@@ -173,14 +173,29 @@ void LogPlayback::Configure(const Entity &,
...
-    if (extension != "zip")
+    if (extension == "tlog" && common::basename(this->dataPtr->logPath) ==
+                                   std::string("state.tlog"))
     {
-      gzerr << "Please specify a zip file.\n";
-      return;
+      // The user likely passed the path to the state.tlog file inside a
+      // recording, rather than the recording directory itself.
+      // Use the parent directory so playback finds state.tlog.
+      gzmsg << "Path points to a .tlog file, using parent directory ["
```
arjo129: Personally I'd go with a `gzwarn`. Also this error is only triggered if the file is `state.tlog` this could be confusing.

